PR #124, opened 2015-01-24
gigasecond:  use full time test cases

soniakeys:
- cases_test.go regenerated with new cases from x-common.
- flexibility added in gigasecond_test.go to parse full times.
- testVersion bumped to 2 to reflect new test cases.

This was a small change to the test program just to handle the new data.  The example needed no changes other than bumping testVersion.  Existing solutions should similarly continue to work.


kytrinyx:
Excellent, thank you!

PR #125, opened 2015-01-24
clock: add test program comments, improve failure message

soniakeys:
Proposed alternative to PR https://github.com/exercism/xgo/pull/121.  TestVersion is unchanged because this change is in comments and error messages only.  It cannot affect whether any existing solution passes or fails.  (Update:  Mm, well, not strictly true.  The logic was rewritten a little, but it should pass and fail exactly the same.)

### DIFF
--- a/clock/clock_test.go
+++ b/clock/clock_test.go
@@ -1,6 +1,21 @@
 package clock
 
-import "testing"
+import (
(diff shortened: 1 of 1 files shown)


kytrinyx:
This looks good, thanks!

PR #128, opened 2015-02-01
custom-set test program rewrite.

soniakeys:
addresses issue 116, slice based solutions.
slice based example added.  also now uses json test cases from x-common.
it's kind of a monster.


kytrinyx:
This is interesting. I like that it has two different example solutions.

PR #129, opened 2015-02-01
code gen function usable across exercises

soniakeys:
When I wrote the template-based data converter for gigasecond, I realized that much of it was not specific to gigasecond and could be reused.  Custom-set turned out to be able to use most of this code without change.  This PR collects the common code into a package that can be used across exercises.

The other exercise with json test data currently is clock.  It was the first generator I wrote and doesn't use templates.  If this PR is accepted, a good test of it will be to replace the clock generation code with a template and a call to this new package.

None of this affects what the solver sees btw.  It's all for the xgo developer.

### DIFF
--- a/config.json
+++ b/config.json
@@ -72,7 +72,8 @@
     "point-mutations"
   ],
   "ignored": [
-    "img"
+    "img",
+    "gen"
   ],
   "foregone": [
     "space-age",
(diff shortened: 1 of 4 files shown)


kytrinyx:
This looks great!

PR #136, opened 2015-03-04
Add connect exercise

pminten:
Would appreciate some critique, I've been out of Go for a few years.

One thing I'm particularly unsure about is what's more idiomatic, the current

```go
coords := make([]coord, b.height)
for i := 0; i < b.height; i++ {
	coords[i] = coord{x: 0, y: i}
}
```

or

```go
coords := make([]coord, 0, b.height)
for i := 0; i < b.height; i++ {
	coords = append(coords, coord{x: 0, y: i})
}
```

### DIFF
--- a/config.json
+++ b/config.json
@@ -64,7 +64,8 @@
     "simple-cipher",
     "paasio",
     "minesweeper",
-    "robot-simulator"
+    "robot-simulator",
+    "connect"
   ],
   "deprecated": [
     "accumulate",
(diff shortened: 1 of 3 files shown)


soniakeys:
Hello Peter,
For that specific question, I like the first version better.  It's simpler and I think simple is idiomatic and more readable.

kytrinyx:
Yes, I also like the first version better.

Would you add "connect" to the config.json, please?

pminten:
@kytrinyx Done.

kytrinyx:
Thank you!

PR #137, opened 2015-03-04
Add tournament exercise

pminten:


kytrinyx:
@soniakeys Would you review this?

@pminten -- Would you mind rebasing this onto master? The config.json files are conflicting.

pminten:
Looks like it should now apply.

soniakeys:
I'll try to have a look later today.  Been busy...

soniakeys:
Here I am.  Okay,

*  I would like to see new exercises use TestVersion.  It's an easy thing to put in.  See any recently touched exercise for an example.  You just define a value `testVersion` in the the test program and require the solver define a matching `TestVersion` in their solution.  The test program validates they are equal.  Any xgo developer that changes the test program or test data increments testVersion.  The benefit of all this is that nipickers can see which test version a posted solution was written for and be spared confusion over why an old posted solution might not pass current tests.  (Hmm, maybe I should add this to the xgo readme.)

*  I've been trying something new with common test data in x-common.  See recent commits that mention common test data for examples.  Unlike testVersion though, it's a little complex so if you don't want to mess with it, that's fine.

*  In the benchmark, we've been abandoning the complexity of stopping and starting the timer.  The purpose of the benchmarks are just so people can experiment with changes to their implementations and see if things get better or worse.  So benchmarks can just use` i < b.N` for the outer loop, run over all the test cases in an inner loop, and benchmark the combined time for running all test cases.

*  The function `prepare` appears to be dead code.  Maybe it was left over from something and can be removed now?

*  I really don't like ignoring bad data, as in the second test case.  It's kind of Go culture that errors are not ignored.  In this case a typo by a data entry clerk might go undetected until incorrect scores were posted.  The readme does not specify to ignore errors so I really don't think we should.  I think test cases with bad data are good but that the expected result should be an error value.

*  The intention of "incomplete competition" isn't clear to me.  Is the point that there is a tie for second place?

*  Speaking of the tie for second place, the example doesn't seem to guarantee output order, sorting only on points with the non-stable sort.Sort.

*  One thing you might consider is requiring a specific function signature for `Tally`.  I've debated this in other exercises and usually decided against it, but you might want it here.  As it is, I would expect to see solutions where people defined `Tally(*string.Reader, *bytes.Buffer) error` and then got nitpicked for not using io.Reader, io.Writer.  We as the exercise developers usually lean in favor of allowing more flexibility in solutions and letting discussion follow in nitpicking, but here it seems fine to just require the Go idiom up front.  You can put a line like

        var _ func(io.Reader, io.Writer) error = Tally

in the test program and the compiler will enforce the API.

pminten:
Thanks for the review @soniakeys. I've pushed a commit that fixes these issues. Thought it may be better to do it as a separate commit since that's easier to review, but could squash if it that's desired.

With regard to prepare, that was a stupid copy/paste bug on my side. Though it's kind of weird that Go didn't complain about it. C compilers do complain about the C equivalent (static functions that aren't being referenced) and it's easy enough to check such bugs (you just need a call graph, which a compiler has anyway).

kytrinyx:
@soniakeys I've submitted a PR that explains testVersion in the README: https://github.com/exercism/xgo/pull/140

soniakeys:
@kytrinyx Awesome.  That looks good.  I hadn't decided yet if the mechanism needed to work just like that, but it seems to have worked well so far.  I'm happy to have it in the readme.

soniakeys:
@pminten Also awesome.  I have one more request -- with a bit of hidden motive -- I'd rather not support comment lines in the data.  My motive is that I'd like to see solutions using the standard library package `encoding/csv`.  Go's csv reader can handle semicolons as separators, and blank lines, but not comment lines.  Again for support I'll cite that the x-common readme doesn't mention comment lines.  Now, the readmes have to be interpreted pretty liberally, I know.  We pretty freely add new requirements in specific language implementations.  Here though, adding this requirement complicates a potential standard library solution.

soniakeys:
Oh, here's a problem I forgot to mention before, the way the benchmark runner works, the `b.N` loop needs to be the outer loop, like

```
func BenchmarkTally(b *testing.B) {
	for i := 0; i < b.N; i++ {
		for _, tt := range happyTestCases {
			var buffer bytes.Buffer
			Tally(strings.NewReader(tt.input), &buffer)
		}
		for _, s := range errorTestCases {
			var buffer bytes.Buffer
			Tally(strings.NewReader(s), &buffer)
		}
	}
}
```

pminten:
Thanks, I've amended the last commit to implement your suggestion.

pminten:
Just noticed I didn't receive emails for all the comments on this thread. Just noticed the CSV comment, fixed that and changed the code to use the csv module since that looks slightly better.

soniakeys:
LGTM.  I think it's ready to go, but I think @kytrinyx would like changes rebased and squashed.

kytrinyx:
I do prefer them rebased and squashed :)

pminten:
Here you go.

kytrinyx:
Thank you

PR #141, opened 2015-03-13
Ledger refactoring exercise

pminten:
This is my first attempt at a refactoring exercise, as we discussed on the exercism mailing list.

Have kept this unsquashed in order to make it easier to understand how I've gone from the example to the exercise. Will squash after review.

I'm not sure where in the line of exercises this should go, added to the end of config.json just to make travis work.

Because this exercise has two versions (ledger.go and example.go) that should (and do) both pass the tests I had to use a build flag. To test example.go and not ledger.go you need to run `go test -example`. I've changed .travis.yml to do this automatically.

Note that it's not really desirable to test ledger.go automatically as it contains a race condition which leads to tests failing if it's triggered. The race is very race (I haven't seen it actually trigger) but it could result in spurious test failures.

### DIFF
--- a/ledger/ledger_test.go
+++ b/ledger/ledger_test.go
@@ -1,0 +1,308 @@
+package ledger
+
+import (
+	"reflect"
+	"strings"
+	"testing"
+)
+
+const testVersion = 1
+
+var successTestCases = []struct {
+	name     string
+	currency string
+	locale   string
+	entries  []Entry
+	expected string
+}{
+	{
+		name:     "empty ledger",
+		currency: "USD",
+		locale:   "en-US",
+		entries:  nil,
+		expected: `
+Date       | Description               | Change
+`,
+	},
+	{
+		name:     "one entry",
+		currency: "USD",
+		locale:   "en-US",
+		entries: []Entry{
+			{
+				Date:        "2015-01-01",
+				Description: "Buy present",
+				Change:      -1000,
+			},
+		},
+		expected: `
+Date       | Description               | Change
+01/01/2015 | Buy present               |      ($10.00)
+`,
+	},
+	{
+		name:     "credit and debit",
+		currency: "USD",
+		locale:   "en-US",
+		entries: []Entry{
+			{
+				Date:        "2015-01-02",
+				Description: "Get present",
+				Change:      1000,
+			},
+			{
+				Date:        "2015-01-01",
+				Description: "Buy present",
+				Change:      -1000,
+			},
+		},
+		expected: `
+Date       | Description               | Change
+01/01/2015 | Buy present               |      ($10.00)
+01/02/2015 | Get present               |       $10.00 
+`,
+	},
+	{
+		name:     "multiple entries on same date ordered by description",
+		currency: "USD",
+		locale:   "en-US",
+		entries: []Entry{
+			{
+				Date:        "2015-01-01",
+				Description: "Buy present",
+				Change:      -1000,
+			},
+			{
+				Date:        "2015-01-01",
+				Description: "Get present",
+				Change:      1000,
+			},
+		},
+		expected: `
+Date       | Description               | Change
+01/01/2015 | Buy present               |      ($10.00)
+01/01/2015 | Get present               |       $10.00 
+`,
+	},
+	{
+		name:     "final order tie breaker is change",
+		currency: "USD",
+		locale:   "en-US",
+		entries: []Entry{
+			{
+				Date:        "2015-01-01",
+				Description: "Something",
+				Change:      0,
+			},
+			{
+				Date:        "2015-01-01",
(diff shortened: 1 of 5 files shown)


ledger/ledger_test.go
```diff
@@ -0,0 +1,308 @@
...
+			},
+		},
+		expected: `
+Date       | Description               | Change
+01/01/2015 | Buy present               |      ($10.00)
+01/01/2015 | Get present               |       $10.00 
+`,
+	},
+	{
+		name:     "final order tie breaker is change",
```
kytrinyx: Does this mean that the value of the change is sorted in increasing order?

ledger/ledger_test.go
```diff
@@ -0,0 +1,308 @@
...
+		}
+		expected := tt.expected[1:] // Strip initial newline
+		if actual != expected {
+			t.Fatalf("FormatLedger for input named %q was expected to return...\n%s\n...but returned...\n%s",
+				tt.name, strings.Replace(expected, " ", "_", -1), strings.Replace(actual, " ", "_", -1))
+		}
+	}
+}
+
+func TestFormatLedgerFail(t *testing.T) {
```
kytrinyx: This is super nitpicky to the point where maybe it's irrelevant, but I like it when there's a symmetry between the names of the tests: _success_ vs _failure_, or _succeed_ vs _fail_.

If I'm being too pedantic please say so, and I won't take it personally.

kytrinyx:
I think this looks really good, and I can't wait to try it. I'm really looking forward to seeing what a refactoring exercise will look like on exercism.

@soniakeys do you want to (and have time to) take a look at it?

ledger/ledger_test.go
pminten: Nah, it's a good point. Have changed that.

ledger/ledger_test.go
pminten: Yes, the ordering is ORDER BY date ASC, description ASC, change ASC.

soniakeys:
I have do time. :)  I read the mailing list thread and took a quick look at the code here so far.  I could probably look for some more ways to defactor the code.  First though, I'm not sure about the Entry definition.  Hm, discovering ISO 4217 now.  So `Entry.cents` is technically a "minor currency unit."  Is there a point using cents?  I would kind of expect this field to be a decimal amount in the major currency unit represented in a string type, that is `"10.00"` rather than `1000`.

soniakeys:
(Also I just sent a PR to run CI with -race,  inspired by the race here in ledger.go.)

soniakeys:
Nah, I like it.  We could add problems to ledger.go forever, but it has more than enough stuff to fix as it is.

soniakeys:
Hmm, well, my suggestion of supplying amounts as strings would enable solutions that didn't use numeric amounts at all.  Maybe that's okay.  Or, one way to force people to pick a numeric representation would be to add a balance column.

pminten:
I don't really like strings in Entry because I think of Entry as some kind of internal data structure of a larger application. Keeping currency amounts in strings internally seems like a headache.

The problem with ints is that different currencies behave differently.

A euro is typically divided up in at most cents (at least in a ledger) while a bitcoin is divided up much further and it makes little sense to divide a single Iranian rial.

Perhaps some abstraction is needed. From the point of view of the ledger implementation the exact type doesn't matter, what matters is that it can be converted to a string with an appropriate number of decimal places. I'll try an experiment with this approach.

soniakeys:
I actually think ints make a fine internal representation for money although I know lots of people argue the point.  I was thinking we shouldn't prescribe the internal format, but I see now that the exercise is one of working on a prescribed internal format.  Avoiding this would expand the exercise somewhat.  It would involve a Ledger type, with constructor, AddEntry and Format methods.  Okay, maybe this exercise doesn't need to go there.

pminten:
Maybe this will work, I've added a Currency interface which just has a method to return an amount as an integer with, effectively, the decimal dot moved a desired number of spaces to the right.

Btw, I would argue against ints as a currency type, at least on a 32-bit platform. ;)

pminten:
Ledger types with constructors, update methods, etc does seem to be overkill for this exercise. However I can imagine a second refactoring exercise where we have the user implement a ledger that reads a CSV file and has some methods to return a view with a specific time range and the total change and stuff. Of course such an exercise would have less of the annoying little details like different currency types.

soniakeys:
Yeah, I agree, this one is plenty big enough as it is.

soniakeys:
The Currency interface is okay, but I don't think it's needed.  I think integer cents is fine for this.

pminten:
I've removed the currency interface commit again and squashed and rebased the other ones. So this one should be ready for merge.

soniakeys:
Thank you.

PR #142, opened 2015-03-14
travis.yml: 1.4.2, also try -race.

soniakeys:
also reduce paasio concurrency a bit to pass with -race.

-race makes sense with a couple of exercises now using concurrency.  (This doesn't have to go ahead of ledger, it could wait its turn.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,8 +16,11 @@ go:
 # 32 bit target is as easy as specifying GOARCH.  We test both to accomodate
 # users that might be running 32 bit environments and make sure our code
 # works for them.
+#
+# Race detection is valuable for our problems that involve concurrency.
+# The race detector isn't supported on 386 though.
 env:
-  - TESTARCH=amd64
+  - TESTARCH=amd64 RACE=-race
   - TESTARCH=386
 
 # Travis doc for the install step says it is for any dependencies of the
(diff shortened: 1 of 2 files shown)


pminten:
Wouldn't this give intermittent failures when running the tests? As I read the example code there's a big fat race in Write() since that calls Write() on the io.Writer outside of the mutex. the TestWriteTotal test runs multiple goroutines that call Write() and the implementation of bytes.Buffer doesn't seem to be thread-safe.

soniakeys:
Paasio may have some issues.  They should probably be filed as a separate issue.

kytrinyx:
What's the status of this? Did Paasio ever get updated?

soniakeys:
Issues of potentially unsynchronized writes in Paasio are independent of this PR.  Running the race detector can't cause any problems that aren't already there, it can only find problems.  I'll rebase this after the current travis PR goes in.

kytrinyx:
OK, cool (and merged).

soniakeys:
Katrina, take a look if you like.  The change to Paasio lowers the number of goroutines from 10000 to 8000 because the race detector has a limit of 8192 goroutines.

kytrinyx:
That makes sense. Thanks!

PR #144, opened 2015-03-16
 Make sorting in ledger example somewhat clearer

pminten:
Just saw this style in another go package, it's much clearer than the nested
if-then-else's.


kytrinyx:
Very nice, thank you.

PR #146, opened 2015-03-17
Use typed constants whenever applicable

pminten:
Typed constants are harder to use in wrong places.


kytrinyx:
Yes, this is better, thanks!

PR #147, opened 2015-03-17
Fill in a simple SETUP.md

### LINKED ISSUE #145: Add helpful information to the SETUP.md
The contents of the SETUP.md file gets included in
the README.md that gets delivered when a user runs the `exercism fetch`
command from their terminal.

At the very minimum, it should contain a link to the relevant
language-specific documentation on
[help.exercism.io](http://help.exercism.io).

It would also be useful to explain in a generic way how to run the tests.
Remember that this file will be included with _all_ the problems, so it gets
confusing if we refer to specific problems or files.

Some languages have very particular needs in terms of the solution: nested
directories, specific files, etc. If this is the case here, then it would be
useful to explain what is expected.

------

Thanks, @tejasbubane for suggesting that we add this documentation everywhere.
See [exercism.io#2198](https://github.com/exercism/exercism.io/issues/2198).

pminten:
closes #145

### DIFF
--- a/SETUP.md
+++ b/SETUP.md
@@ -1,0 +1,4 @@
+To run the tests simply run the command `go test` in the exercise directory.
+
+For more detailed info about the Go track see the [help
+page](http://help.exercism.io/getting-started-with-go.html).


soniakeys:
LGTM.  I'm happy with this being short, just a jog of what to do and a link to the detailed information online.

kytrinyx:
Yeah, I like this. Thanks!

PR #149, opened 2015-03-18
Use the versatile switch statement in tournament

pminten:


kytrinyx:
This looks more idiomatic, thank you.

PR #150, opened 2015-03-19
Add tree-building refactoring test

pminten:
This is an implementation of exercism/x-common#60.

The interesting part here is that converting a list of id/parent pairs to a tree when using a language with mutable arrays, can be quite efficient: most of the algorithm is linear, only the sorting of children isn't but even that has good complexity. There is also a much more simplistic way which is more than quadratic.

To illustrate this, these are the benchmark results from the example (fast algorithm):

```
BenchmarkTwoTree              50          23799606 ns/op
BenchmarkTenTree            1000           2653044 ns/op
BenchmarkShallowTree        1000           2955220 ns/op
```

and these are from the exercise file (slow algorithm):

```
BenchmarkTwoTree               1        10763446466 ns/op
BenchmarkTenTree              10         249729689 ns/op
BenchmarkShallowTree           5         428928462 ns/op
```

Hopefully the user will learn that refactoring isn't just about coding style, sometimes you need to take a step back and look at the algorithm as well.

### DIFF
--- a/tree-building/tree.go
+++ b/tree-building/tree.go
@@ -1,0 +1,111 @@
+// +build !example
+
+package tree
+
+import (
+	"errors"
+	"fmt"
+)
+
+const TestVersion = 1
+
+type Record struct {
+	Id, Parent int
+}
+
+type Node struct {
+	Id       int
+	Children []*Node
+}
+
+type Mismatch struct{}
+
+func (m Mismatch) Error() string {
+	return "c"
+}
+
+func Build(records []Record) (*Node, error) {
+	if len(records) == 0 {
+		return nil, nil
+	}
+	root := &Node{}
+	todo := []*Node{root}
+	n := 1
+	for {
+		if len(todo) == 0 {
+			break
+		}
+		newTodo := []*Node(nil)
+		for _, c := range todo {
+			for _, r := range records {
+				if r.Parent == c.Id {
+					if r.Id < c.Id {
+						return nil, errors.New("a")
+					} else if r.Id == c.Id {
+						if r.Id != 0 {
+							return nil, fmt.Errorf("b")
+						}
+					} else {
+						n++
+						switch len(c.Children) {
+						case 0:
+							nn := &Node{Id: r.Id}
+							c.Children = []*Node{nn}
+							newTodo = append(newTodo, nn)
+						case 1:
+							nn := &Node{Id: r.Id}
+							if c.Children[0].Id < r.Id {
+								c.Children = []*Node{c.Children[0], nn}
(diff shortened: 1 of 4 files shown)


tree-building/tree.go
```diff
@@ -0,0 +1,111 @@
...
+		todo = newTodo
+	}
+	if n != len(records) {
+		return nil, Mismatch{}
+	}
+	if err := chk(root, len(records)); err != nil {
+		return nil, err
+	}
+	return root, nil
+}
```
kytrinyx: This function is fantastic in terms of refactoring opportunity!

kytrinyx:
I really like this exercise. Again I'm going to defer to @soniakeys, since she's far more experienced with Go, and has really been doing great things with the Go track here on exercism.

tree-building/tree_test.go
```diff
@@ -0,0 +1,305 @@
...
+		t.Fatalf("Found TestVersion = %v, want %v", TestVersion, testVersion)
+	}
+	for _, tt := range successTestCases {
+		actual, err := Build(tt.input)
+		if err != nil {
+			t.Fatalf("Build for test case %q returned error %q.  Error not expected.",
+				tt.name, err)
+		}
+		if !reflect.DeepEqual(actual, tt.expected) {
+			t.Fatalf("Build for test case %q was expected to return\n%s\nbut returned:\n%s",
```
soniakeys: By [Code Review Comments](https://github.com/golang/go/wiki/CodeReviewComments#useful-test-failures) this message could be improved to use got-want order.

tree-building/tree_test.go
```diff
@@ -0,0 +1,305 @@
...
+				tt.name, tt.expected, actual)
+		}
+	}
+}
+
+func TestMakeTreeFailure(t *testing.T) {
+	for _, tt := range failureTestCases {
+		actual, err := Build(tt.input)
+		if err == nil {
+			t.Fatalf("Build for test case %q was expected to fail but returned: %s.",
```
soniakeys: Similarly here.  (I'm not even sure how consistent other exercises are with this, but I think the order of "Build returned %s, expected error" or "Build = %s, want error" would be better for consistency.  And, I think consistency in wording or identifier naming is less important than consistency in the order of information.)

pminten: Agreed in the general case of a relatively short result message. For the special case of multiline return messages such as in ledger I think "Build was expected to return error but returned:\n%s" is better since it avoids having to add a line after the actual message.

pminten:
Amended the commit with fixes for the test report order.

soniakeys:
Nice, thank you!

PR #152, opened 2015-03-23
Change clock.Add prototype in comment

pminten:
This confused me when making the exercise as it seemed the tests didn't expect a
a return value, which would force me to use a pointer receiver. Making the
comment more in line with what the test code actually expects should fix this.


soniakeys:
It should help anyway.  For being second in config.json this exercise has raise lots of questions.

PR #153, opened 2015-03-26
Two minor fixes to the testcases of the Go exercises

JensChrG:
One changes a format verb, and the other an error text


soniakeys:
Thank you!

PR #154, opened 2015-03-28
Add poker exercise

pminten:
See exercism/x-common#62.


soniakeys:
Gosh, that will take people a while.  There should be lots of variation in solutions as well!

PR #155, opened 2015-03-29
Use standard char functions in crypto-square

pminten:
This is slightly less efficient according to the benchmarks but more extensible.

### DIFF
--- a/crypto-square/example.go
+++ b/crypto-square/example.go
@@ -3,18 +3,17 @@ package cryptosquare
 import (
 	"math"
 	"strings"
+	"unicode"
 )
 
 const TestVersion = 1
 
 func norm(r rune) rune {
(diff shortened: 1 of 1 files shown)


soniakeys:
In general I don't think changes are worthwhile to improve the quality of the examples.  The contributing guide currently notes "The example solution will not be delivered as part of the problem, and is simply used as a sanity check to make sure that the test suite makes sense."  I'd like to see PRs on examples justified on bases of correctness -- and if an incorrect example was passing the test program, then changes to the test program should be considered as well.

pminten:
Fair enough. I sometimes use the examples to see if there are any improvements possible over my own solution (since later exercises in a track generally don't attract nits), but I understand your point.

PR #158, opened 2015-04-01
Change int -> int64 in largest-series-product

### LINKED ISSUE #157: Int type used in largest-series-product may be too small
The largest-series-product test has this in its test cases:

```go
var tests = []struct {
	digits  string
	span    int
	product int
	ok      bool
}{
...
	{pe1k, 13, 23514624000, true}, // new PE problem
}
```

The problem here is that the number 23514624000 doesn't fit in an int32. In go the type int may be implemented as int32 or int64, most likely based on the machine word size. That means this will work for a 64-bit platform but will likely break at compile time on a 32-bit platform (since you're not allowed to have overflowing constants).

I'm not quite sure how to best fix this though. Changing the result size to int64 would make the most sense but would break user code.

pminten:
This prevents errors on 32-bit systems with a number that doesn't fit in int32.

Fixes #157

### DIFF
--- a/largest-series-product/largest_series_product_test.go
+++ b/largest-series-product/largest_series_product_test.go
@@ -2,10 +2,20 @@ package lsproduct
 
 import "testing"
 
+// Define a function LargestSeriesProduct(string, int) (int64, error).
+//
+// Also define an exported TestVersion with a value that matches
+// the internal testVersion here.
+
+const testVersion = 1
+
+// Retired testVersions
+// (none) 1c5d360b98fc3a9f59c1e5e2f3a668ff8438419d
+
 var tests = []struct {
 	digits  string
 	span    int
-	product int
+	product int64
 	ok      bool
 }{
 	{"0123456789",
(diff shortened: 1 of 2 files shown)


largest-series-product/largest_series_product_test.go
```diff
@@ -56,6 +66,9 @@ const pe1k = "73167176531330624919225119674426574742355349194934" +
 	"71636269561882670428252483600823257530420752963450"
 
 func TestLargestSeriesProduct(t *testing.T) {
+	if TestVersion != testVersion {
+		t.Fatalf("Found TestVersion = %v, want %v", TestVersion, testVersion)
+	}
```
soniakeys: I was suggesting that in line 71/84 below, the test could read `case int64(p) != test.product:`.  That would allow existing incorrect solutions to continue to compile.  They would pass tests on 64 bit architectures but fail on 32.  Without the conversion, we're forcing incorrect solutions to fail at compile time.  With the cast, lots of people will write incorrect code.  Others could nitpick these solutions then.  The advantage is that people will discuss and think about it more than if they just saw a compile error and fixed without stopping to think why it should be an int64.  I don't know, maybe that's hokey and we should write tests for correctness.  It's just an idea.

pminten: People will most likely not notice these things. It's only after this bites you a few times that you become wary of int sizes.

I'd prefer things to fail during compilation, in line with the principle of failing as early as possible.

soniakeys:
"It's only after this bites you a few times that you become wary of int sizes."  Exactly.  So for the purpose of learning, we should allow people to make the mistake and get bit.  If we specify int64 no one ever gets bit.  I think nitpicking is an acceptable place for people to get bit.

kytrinyx:
I'm going to close this, since @soniakeys updated the test suite to be more flexible in terms of int sizes.

PR #159, opened 2015-04-01
Alphametics

pminten:
See exercism/x-common#64.

I hope the example is somewhat readable, it was a challenging example to make.

### DIFF
--- a/alphametics/example.go
+++ b/alphametics/example.go
@@ -1,0 +1,200 @@
+package alphametics
+
+import "sync"
+
+const TestVersion = 1
+
+// Set this to true to debug the example. It enables several useful logging tools.
+const debugging = false
+
+type accumulator struct {
+	carry    int
+	usedNums uint16       // bit n denotes if digit n is taken
+	solution map[rune]int // partially filled solution map
+}
+
+type ipc struct {
+	resultChan chan map[rune]int
+	abortChan  chan struct{}
+	waitGroup  *sync.WaitGroup
+}
+
+func Solve(a, b, c string) map[rune]int {
+	aRunes, bRunes, cRunes := prepareRunes(a, b, c)
+
+	// Setup the machinery for solving in parallel.
+	resultChan := make(chan map[rune]int)
+	abortChan := make(chan struct{}) // Closed when goroutines can abort
+
+	var waitGroup sync.WaitGroup
+	waitGroup.Add(1)
+	go func() {
+		// This goroutine ensures the main thread will return nil if no answer can
+		// be found instead of blocking forever.
+		waitGroup.Wait()
+		resultChan <- nil
+	}()
+	go solveStep(aRunes, bRunes, cRunes, accumulator{}, &ipc{resultChan, abortChan, &waitGroup})
+
+	result := <-resultChan
+	close(abortChan)
+	return result
+}
+
+func solveStep(aRunes, bRunes, cRunes []rune, acc accumulator, ipc *ipc) {
+	defer ipc.waitGroup.Done()
+	if abortSignalled(ipc) {
+		return
+	}
+	aRune, bRune, cRune := aRunes[0], bRunes[0], cRunes[0]
+	c, cFound := acc.solution[cRune]
+	for _, a := range availableDigits(aRune, acc, 0, 0) {
+		for _, b := range availableDigits(bRune, acc, aRune, a) {
+			if cFound {
+				cCalc := (a + b + acc.carry) % 10
+				if cCalc != c {
+					continue
+				}
+			} else {
+				c = (a + b + acc.carry) % 10
+				if (aRune == cRune && aRune != 0 && c != a) ||
+					(bRune == cRune && bRune != 0 && c != b) {
+					continue
+				}
+				if isUsed(c, acc.usedNums) {
+					continue
+				}
+			}
+			newAcc := acc
+			newAcc.carry = (a + b + acc.carry) / 10
+			newUsed := acc.usedNums
+			// Needed because availableDigits returns []int{0} if the rune is 0.
+			if aRune != 0 {
+				newUsed = setUsed(a, newUsed)
+			}
+			if bRune != 0 {
+				newUsed = setUsed(b, newUsed)
+			}
+			newAcc.usedNums = setUsed(c, newUsed)
+			newAcc.solution = copyMap(acc.solution)
+			newAcc.solution[aRune] = a
+			newAcc.solution[bRune] = b
+			newAcc.solution[cRune] = c
+			if len(aRunes) == 1 {
+				// This was the final iteration, so we've found a result
+
+				// Last check, the left-most (in our case last) character of each part
+				// of the equation cannot be 0.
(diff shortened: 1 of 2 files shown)


soniakeys:
Well this will be a fun one.  Especially fun to benchmark.  How about if we make it return a ten character string though, like "OMY--ENDRS" and put the solutions in the test cases.  We could either restrict test cases to puzzles with zero or one solutions, or we could could handle multiple solutions by allowing any correct solution or requiring a list of all correct solutions.  Part of my motivation is reducing the computations required by the test program.  The more work the test program has to do, the more chance of a bug in the test program.

pminten:
You may be right about the tests being too complex. When I read your comment I realized that there's a bug in the tests, they don't check for uniqueness of digits. So generating valid output using some existing alphametics tools should be a better approach.

And I bet there are plenty of alphametics tools, since this whole problem just screams constraint programming.

pminten:
Aargh, and what's worse, my example code is non-deterministic, it only sometimes gives a correct result.

pminten:
@soniakeys Do you think a bonus exercise that requires the user to find all solutions to math+code=same (156 solutions)  would make sense here?

pminten:
Regarding return value, I kind of like the fact that the test code can print out a nice graphical representation of the solution as it really helps when debugging.

At the same time, if the test code only checks against a list of valid outcomes it only would need a small bit of extra logic to transform a map as it is now into a string like "A=3 C=1 D=5 E=6 H=0 M=2 O=9 S=4 T=7". So that may be a good middle-of-the-road solution.

soniakeys:
It was fun.  I posted a rendition here https://github.com/soniakeys/xgo/commit/5251d037b52516c0f3e1277b63de1711f6ac715f

The example is completely brute force, so it takes several seconds.

kytrinyx:
What's the status of this one?

pminten:
Not clear which solution to merge. @soniakeys solution is IMHO better, but we never got around to making an actual decision.

zmackie-nbc:
This looks very cool!  Was there any progress on making a decision or should we close the PR?

alphametics/example.go
```diff
@@ -0,0 +1,200 @@
...
+	resultChan chan map[rune]int
+	abortChan  chan struct{}
+	waitGroup  *sync.WaitGroup
+}
+
+func Solve(a, b, c string) map[rune]int {
+	aRunes, bRunes, cRunes := prepareRunes(a, b, c)
+
+	// Setup the machinery for solving in parallel.
+	resultChan := make(chan map[rune]int)
```
cpuguy83: The concurrency here is pretty confusing.
You might try handling the waiting here in the caller rather than giving the waitgroup to the step solver.

So something like
```go

go func() {
    solveStep(...)
    waitGroup.Done()
}()
```

Or alternatively handle the concurrency at a lower level may help clean this up a bit as I see you are also adding to the waitgroup in the `solveStep` fn, which is recursive.

Cancellation may also be better served by a `context.WithCancel`.

cpuguy83: It may also be helpful to resolve this w/o concurrency and then add it in if needed only after you have a clean solution.

alphametics/example.go
```diff
@@ -0,0 +1,200 @@
...
+
+func solveStep(aRunes, bRunes, cRunes []rune, acc accumulator, ipc *ipc) {
+	defer ipc.waitGroup.Done()
+	if abortSignalled(ipc) {
+		return
+	}
+	aRune, bRune, cRune := aRunes[0], bRunes[0], cRunes[0]
+	c, cFound := acc.solution[cRune]
+	for _, a := range availableDigits(aRune, acc, 0, 0) {
+		for _, b := range availableDigits(bRune, acc, aRune, a) {
```
cpuguy83: Interestingly, this looks like a case for recursion (which you are doing below), but are not doing here.

alphametics/example.go
```diff
@@ -0,0 +1,200 @@
...
+		resultChan <- nil
+	}()
+	go solveStep(aRunes, bRunes, cRunes, accumulator{}, &ipc{resultChan, abortChan, &waitGroup})
+
+	result := <-resultChan
+	close(abortChan)
+	return result
+}
+
+func solveStep(aRunes, bRunes, cRunes []rune, acc accumulator, ipc *ipc) {
```
cpuguy83: `solveStep` should probably always return a value rather than sending it on a channel, it may help to clean up how the concurrency is handled.

soniakeys:
Hi people,

It seems I deleted my version from last year and I really have no memory of it.  I did just rewrite a single-thread brute force solution that passes tests in about 7 seconds on my machine.  I could repost it...

Something faster would sure be nice.  I confess I haven't studied at Peter's solution.  Is it faster?

Maybe something clever has been posted in one of the other languages that could be ported?  There was a thing that showed a big table of exercises implemented in each language.  Does that still work?

Has anyone looked at this new language picat that won some recent awards?  I tried but their doc is pretty terse if you're not up on concepts of logic programming.  (Prolog was too long ago for me.)  I really don't have good ways to solve constraint problems in Go.  I coded Knuth's DLX once but that's an ugly mess of pointers.  I think I coded a 2-SAT solver once.  None of that helps, I know, just rambling.

petertseng:
About *this* particular PR - since it seems like the original author is not currently to make changes, and the *Allow maintainers to make changes* option did not exist at the time, I think it is not possible to progress with this PR. Of course, if the author returns we will happily reconsider. For now, it seems like the thing to do is to close it and let any other interested party open a PR, since then that person will be able to make changes in response to feedback.

Why not just merge it in its current state, you ask? As I see it, it is just that it needs to be added to config.json - and don't forget that we now have a https://github.com/exercism/x-common/blob/master/exercises/alphametics/canonical-data.json from which this exercise should draw its test cases unless there is very strong reason to deviate.

If someone wishes to make use of the work already done in this PR, that is acceptable as well, with credit to the original author.

> I did just rewrite a single-thread brute force solution that passes tests in about 7 seconds on my machine. I could repost it...

I would welcome it. Caution though, I suspect it may run for quite a long time on the new test case with 10 letters, since that's 10! permutations (you said it was brute force so I guessed at this, I don't know for sure).

> Something faster would sure be nice. I confess I haven't studied at Peter's solution. Is it faster?

I am not 100% sure, but it looks like the Travis result says it is passing all tests in about 0.1 seconds. Again, caveat because we don't know what will happen with the new test cases (in particular, if this code only handles two terms, it can't handle some of the new test cases)

> Maybe something clever has been posted in one of the other languages that could be ported? 

I will point to https://github.com/exercism/xruby/pull/469 - if you don't want to read all the code, I'll explain what goes on in their example.rb. The same thing happens in the example.go here. This only works if all problems are addition. canonical-data.json only has addition, but F# and Lua tracks have non-addition tests so they don't use this technique. Rot13 for those who don't want to be spoiled and figure it out for yourself. rot13(fbyir bayl bar pbyhza ng n gvzr!) 

> There was a thing that showed a big table of exercises implemented in each language. Does that still work?

I remember seeing the table you speak of. It's sliped my memory. These days I only use http://x.exercism.io/problems/alphametics

PR #166, opened 2015-04-25
Add error handling exercise

pminten:
This is a focus exercise designed to make people play with the error handling systems of go. See exercism/x-common#76.

I've put it fairly early on the list of exercises since it's a relatively uncomplicated exercise in that it only requires applying standard idioms and because error handling is quite important so it would be good if someone who only does 10 exercises on this track gets to see this one.


kytrinyx:
I realized that I hadn't merged this. I look forward to seeing how people react to this / solve this.

PR #167, opened 2015-04-25
Generate test cases more flexibly

kytrinyx:
I'm not convinced I have made the right assumptions here, so this is more of a tentative suggestion than anything else.

This makes it possible to generate test cases either from a problem directory, or from the root of the repository. E.g. both of the following now work:
    
        cd xgo
        go run leap/example_gen.go
    
and
    
        cd xgo/leap
        go run example_gen.go
    
This clarifies the assumption about where the x-common repository is, making it relative to the gen/gen.go file rather than relative to the present working directory.
    
The test_cases.go file is generated to the directory containing the example_gen.go file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,20 @@ all at once.
 We like errors in Go.  It's not idiomatic Go to ignore invalid data or have undefined behavior.  Sometimes our
 Go tests require an error return where other language tracks don't.
 
+## Generating test cases
+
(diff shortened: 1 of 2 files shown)


kytrinyx:
@soniakeys Would you have time to take a look at this and see if it makes sense?

kytrinyx:
I'm going to go ahead and merge this; we can alway make changes later.

PR #168, opened 2015-04-25
Add "counter", a write-the-test-suite exercise

pminten:
I'm a bit worried about the exercism cli tool here. I seem to recall that it stops the user accidentally submitting the test file, but in this case the file with the "_test.go" name is the submission.

### DIFF
--- a/counter/example.go
+++ b/counter/example.go
@@ -1,0 +1,59 @@
+package counter
+
+import "testing"
+
+func TestNoAdd(t *testing.T) {
+	counter := makeCounter()
+	if counter.Lines() != 0 {
+		t.Errorf("Lines mismatch: got %d, expected %d", counter.Lines(), 0)
+	}
+	if counter.Letters() != 0 {
+		t.Errorf("Letters mismatch: got %d, expected %d", counter.Letters(), 0)
+	}
+	if counter.Characters() != 0 {
+		t.Errorf("Characters mismatch: got %d, expected %d", counter.Characters(), 0)
+	}
+}
+
+func TestEmptyString(t *testing.T) {
+	counter := makeCounter()
+	counter.AddString("")
+	if counter.Lines() != 0 {
+		t.Errorf("Lines mismatch: got %d, expected %d", counter.Lines(), 0)
+	}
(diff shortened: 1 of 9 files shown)


kytrinyx:
Hm. Yepp, that's still the case:

```
$ exercism submit hamming_test.rb 
2015/04/25 18:43:43 Please submit the solution, not the test file.
```

Maybe we should just take the feature out. Are we being too helpful? When we added that feature we didn't have the "delete submission" feature, but now that's not an issue.

pminten:
We could replace it with a message saying "You're submitting a test file. If this is not what you want you can delete it in this way ...". That's still fairly helpful.

kytrinyx:
Agreed. I like leaving the responsibility to the user rather than trying to be overly helpful.

kytrinyx:
The `--test` behavior is now available in the CLI.

I looked through this again, and the only question I have is whether we should name the test file `example_test.go` rather than `example.go`.

pminten:
I believe I went with `example_test.go` because without the `_test.go` suffix `go test` won't run the tests in it.

kytrinyx:
The current file name is `example.go`, but it has tests in it. (Sorry, I might be misunderstanding your comment).

counter/example.go
```diff
@@ -0,0 +1,59 @@
+package counter
```
kytrinyx: Can we rename the file to `example_test.go`?

pminten: Doesn't that automatically include it in the files sent to the user?

kytrinyx: No, anything with `example` in the path (not just the name of the file) will be excluded from the payload to the user.

pminten: After trying this I remembered why it doesn't work. The example_test.go gets picked up by go test in the travis and fails since you have to explicitly set an environment variable to choose an implementation to test with.

kytrinyx: Ah, right. Yepp, that makes sense! Would you add a comment to that effect in the file itself? I can see myself forgetting this and going through the exact same set of questions six months from now.

pminten:
Added explanation and rebased.

PR #170, opened 2015-05-16
Add react exercise

pminten:
See exercism/x-common#83.

### DIFF
--- a/config.json
+++ b/config.json
@@ -39,6 +39,7 @@
     "say",
     "circular-buffer",
     "robot-name",
+    "react",
     "custom-set",
     "atbash-cipher",
     "phone-number",
--- a/react/example.go
+++ b/react/example.go
@@ -1,0 +1,127 @@
+package react
+
+const TestVersion = 1
+
+type reactor struct {
+	cells []*cell
+}
+
+type cell struct {
+	ord       int // Cell ordinal (identifier)
+	reac      *reactor
+	value     int
+	consumers []int
+	producers []*cell
+	compute   func([]int) int
+	nextKey   int
+	observers map[int]func(int)
+}
+
+func (c *cell) Value() int { return c.value }
+
+func (c *cell) SetValue(v int) {
+	oldVal := c.value
+	c.value = v
+	if v != oldVal {
+		c.reac.trigger(c)
+	}
+}
+
+func (c *cell) AddCallback(cb func(int)) CallbackHandle {
+	key := c.nextKey
+	c.nextKey++
+	c.observers[key] = cb
+	return key
+}
+
+func (c *cell) RemoveCallback(key CallbackHandle) {
+	delete(c.observers, key.(int))
+}
+
+func New() Reactor {
+	return &reactor{}
+}
+
+func (r *reactor) CreateInput(initial int) InputCell {
+	new := &cell{ord: len(r.cells), reac: r, value: initial}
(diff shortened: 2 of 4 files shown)


kytrinyx:
The build is failing for Go tip (not 1.3, not 1.4), but I think it is unrelated to this exercise, but rather something that _tournament_ is triggering:

```plain
--- FAIL: TestTallyHappy-2 (0.00s)
	tournament_test.go:124: Tally for input named "ignore empty lines" returned error "line 5, column 0: wrong number of fields in line". Error not expected.
FAIL
FAIL	github.com/exercism/xgo/tournament	0.005s
```

pminten:
Still works with 1.4.2 source release. Maybe it's best to have this PR wait until the next Go release. It could be that this is just a bug in Go tip that we're running into.

kytrinyx:
Yes, it might be a bug. Let's hold off on it for a bit. I'll retrigger the build periodically to see if it passes on a more recent tip.

dgryski:
If it consistently fails against tip it should be trivial to git-bisect to find the offending commit.

kytrinyx:
True! I forgot that go was in git now. Back when it was in hg I treated the history as a bit of a black box since my hg skills are essentially nil.

dgryski:
Bisect says that the first bad commit is https://github.com/golang/go/commit/2db58f8f2daea1c6f6134584f7811bb229177b28 , which given that it's an encoding/csv change and Tally uses encoding/csv is believable.

pminten:
That explains it. The CSV I use has varying number of fields per record. It's a precondition violation on my side that this commit turned into an actual crash.

I'll fix this tomorrow.

Thanks for your help in hunting this down.

Damian Gryski <notifications@github.com> schreef op 6 juni 2015 21:12:37 CEST:
>Bisect says that the first bad commit is
>https://github.com/golang/go/commit/2db58f8f2daea1c6f6134584f7811bb229177b28
>, which given that it's an encoding/csv change and Tally uses
>encoding/csv is believable.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/exercism/xgo/pull/170#issuecomment-109638354

-- 
Verzonden van mijn Android telefoon met K-9 Mail.

pminten:
Ok, I've figured out where exactly things went wrong. The Go code used to accept empty lines because it checks for whether the `fields []string` slice from parseRecord is nil and if so skips that line. The `fields` slice is zero initialized, meaning `nil`. If no fields are appended it stays nil and thus the line is skipped. However the golang/go@2db58f8 commit preallocates the slice so that even if no fields are found the slice is not `nil`. This causes newlines to no longer be skipped.

I'm going to change the tests a bit to require comment lines instead of newlines to be filtered. This still requires a bit of thought on the side of the user but works with the Go CSV reader (since you can specify a comment char). The bump in TestVersion should hopefully prevent people from running into this issue with code that worked for the old Go but not for the new one. Will make a new PR for this, once that's merged this one should be mergeable as well.

pminten:
Update: just noticed Go supports variable number of fields. Still going to bump testVersion and require handling of comments (since that's just one extra field to set for the user) but will leave newline in.

kytrinyx:
The newlines fix is merged into master and I've retriggered the build.

kytrinyx:
And it's good to go. Thanks, everyone for hunting that down!

PR #171, opened 2015-06-07
Fix newlines, see discussion under #170.

pminten:


kytrinyx:
Tricky, thanks for hunting that down.

PR #173, opened 2015-06-16
Update large number test to increase security

calebthompson:
Nothing in the [README] seems to indicate that numbers larger than 32
wouldn't be in the set of potential handshakes.

[README]: https://github.com/exercism/x-common/blob/master/secret-handshake.md

Practically speaking, a sophisticated attacker could brute-force so few
combinations (31 codes each with 8 possible handshakes = 248), so we'd
want to ensure that larger numbers with seemingly no relation to the
significant binary bits would also be possible codes.

Seriously though, am I missing something in this problem? @kytrinyx @soniakeys

### DIFF
--- a/secret-handshake/example.go
+++ b/secret-handshake/example.go
@@ -5,7 +5,7 @@ var signals = []string{"wink", "double blink", "close your eyes", "jump"}
 // Handshake returns sequence to perform corresponding to the given code.
 func Handshake(code int) (h []string) {
 	switch {
-	case code < 1 || code > 31:
+	case code < 1:
 	case code&16 == 0:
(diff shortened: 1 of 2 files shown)


kytrinyx:
No, you're not missing anything. This was made up in a hurry back in the day.

The test is not compiling, though -- this needs an extra brace after the slice of string in the new test case.

calebthompson:
Fixed, and corrected the example.

kytrinyx:
Thanks :)

kytrinyx:
Oh: I tend to deploy once a day or so. This should go up in the next 24 hours sometime.

PR #174, opened 2015-09-03
Add instruction for symlinking into GOPATH

carlisia:
It is very likely that at least some users, if not most, will have
installed their exercism code (and thus setup the exercism path) in a
directory that is different from where they installed GO.

This PR instructs on what to do to make GO recognize directories outside
of its path.

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -2,3 +2,11 @@ Follow the instructions for your system on the installation page at [golang.org]
 
 Exercism supports Go 1.2 and higher.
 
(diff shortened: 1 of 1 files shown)


kytrinyx:
Thanks!

Does the `xgo` repository need to be within the Go path? There are no imports outside of the standard library, so as long as Go exists on the system, there should be no problem, right?

carlisia:
@kytrinyx aaaaaand you are right! I'l close this. :)

kytrinyx:
I was so sure you were going to tell me about something I had completely missed. I'm glad we had this conversation :)

carlisia:
Ha! Nope. That'll be the day. :)

PR #176, opened 2015-10-17
Add two bracket-push tests

petertseng:
I am adding these two tests because they rule out some incorrect solutions that would have been accepted by the previous set of tests. This would be the class of solution that simply counts the number of open/close brackets and makes sure they're equal. This includes my own solution. I only realized my mistake after looking at others' solutions. These tests would have helped me realize my mistake before submitting.

NOTE: This exercise is also in the javascript track (see http://x.exercism.io/problems/bracket-push ) -- therefore I have opened https://github.com/exercism/xjavascript/pull/118

### DIFF
--- a/bracket-push/bracket-push_test.go
+++ b/bracket-push/bracket-push_test.go
@@ -33,6 +33,10 @@ var testCases = []struct {
 		expected: true,
 	},
 	{
+		input:    "{[}]",
+		expected: false,
+	},
+	{
 		input:    "{[)][]}",
 		expected: false,
 	},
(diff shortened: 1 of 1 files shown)


bracket-push/bracket-push_test.go
```diff
@@ -29,6 +33,10 @@ var testCases = []struct {
 		expected: true,
 	},
 	{
+		input:    "{[}]",
+		expected: false,
```
petertseng: please do sanity check for me so that we agree whether the result should be true or false for this. I argue it should be false because https://github.com/exercism/x-common/blob/master/bracket-push.md says **nested correctly** and `{[}]` is an incorrect nesting.

kytrinyx: Yeah, I totally agree on this one. Nice catch, thank you!

PR #178, opened 2015-10-18
don't do install step, update go version numbers, add comments,

soniakeys:
move configlet out of build matrix, exclude 1.4 rather than include tip, and enable container-based build.

This was all started with an attempt to add a stub solution to one of the problems.  It didn't build because there was a separate install step running that wasn't needed.  I disabled that then noticed the other things could be updated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,14 +1,54 @@
 language: go
+sudo: false
+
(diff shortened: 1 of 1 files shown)


kytrinyx:
This is a great cleanup. Thanks!

PR #180, opened 2015-10-22
circular-buffer: Test overwrite on non-full buffer

petertseng:
This test is useful to make sure implementations don't drop items from
non-full buffers when using Overwrite.

### DIFF
--- a/circular-buffer/circular_buffer_test.go
+++ b/circular-buffer/circular_buffer_test.go
@@ -19,10 +19,11 @@ import (
 	"testing"
 )
 
-const testVersion = 1
+const testVersion = 2
 
 // Retired:
 // (non-versioned) ea3c43868a02d8d899d3a99163380a3b7b3a0a18
+// 1               38c982fc7e6861dfebec2a0023a85e53ea8c1b59
 
 func TestTestVersion(t *testing.T) {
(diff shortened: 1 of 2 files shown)


petertseng:
Examining other language tracks:
http://x.exercism.io/problems/circular-buffer
    
Tracks that include this test:

* ruby (the first version of the exercise I did, and I failed this exact test in that track, which is why I noticed this difference when compared to Go's)
* rust (at my recommendation when the exercise was about to be added to that track)

Tracks that do not include this test:
    
* erlang (but it does test their equivalent of ovewrite on a completely empty buffer, so maybe that is close enough?)
* javascript
* lfe (same comment as erlang)
* lua

So I guess Ruby was the first to actually explicitly test this behavior of Overwrite (what should it do when the buffer isn't full), and I guess now's a good a time as any to ask whether it should be spread to the other tracks with circular-buffer.

petertseng:
Assumption: I should bump `testVersion` for this (because it might cause previously-passing solutions to fail). I will do so and amend the commit.

kytrinyx:
Yeah, I think adding the test is a good idea. @soniakeys do you see any gotchas with this?

soniakeys:
Very nice.  Thank you, Peter!

PR #182, opened 2015-10-23
clock: add stub file

soniakeys:
Leap is currently first in config.json order, clock is second.  So this is still rather verbose although a little different than leap.


kytrinyx:
I like it.

PR #186, opened 2015-10-25
issue-148: add remaining stubs

soniakeys:
There were 8 left, these were all done minimally, including bob and hamming.  Those two appear as first exercises in some tracks but are not listed that early in Go's config.json so I left them minimal.

Also in this commit are some little changes to the stubs for gigasecond and triangle.  Strangely, they were causing error messages.  The build constraint keeps the stubs from being tested, but the top of the file must be parsed to see the build constraint.  It appears that the go tool prints error messages from the partial parse, although it (properly) doesn't then act on the errors.  Originally I had deliberately left some incomplete lines in the stubs not wanting to give too much away and hadn't noticed the error messages in the build log.  In this commit though I fixed/removed stuff until the messages went away, to avoid possible confusion in the future.

### DIFF
--- a/bob/bob.go
+++ b/bob/bob.go
@@ -1,0 +1,5 @@
+// +build !example
+
+package bob // package name must match the package name in bob_test.go
+
+const TestVersion = 1 // same with TestVersion
--- a/circular-buffer/circular_buffer.go
+++ b/circular-buffer/circular_buffer.go
@@ -1,0 +1,5 @@
+// +build !example
+
+package circular
+
+const TestVersion = 2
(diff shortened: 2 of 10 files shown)


bob/bob.go
```diff
@@ -0,0 +1,5 @@
+// +build !example
+
+package bob // package name must match the packgage name in bob_test.go
```
kytrinyx: s/packgage/package/

kytrinyx:
Aside from the one typo, this LGTM.

soniakeys:
Thank you :)

PR #187, opened 2015-10-27
word-count remodel: replace histogram.go with stub,…

soniakeys:
… use common test data,
use TestVersion, improve package name, improve type name,
remove dead code from example, simplify benchmark.


kytrinyx:
This is a great improvement. I'm really, really liking these generators.

PR #188, opened 2015-10-27
React tests

petertseng:
Two commits, the first a small refactor (but does add a small bit of extra functionality) and the second adding new tests. Much more detail provided in the individual commit messages.

### DIFF
--- a/react/react_test.go
+++ b/react/react_test.go
@@ -21,17 +21,20 @@ func TestTestVersion(t *testing.T) {
 	}
 }
 
+func assertCellValue(t *testing.T, c Cell, expected int, explanation string) {
+	observed := c.Value()
+	if observed != expected {
+		t.Fatalf("%s: expected %d, got %d", explanation, expected, observed)
+	}
+}
+
 // Setting the value of an input cell changes the observable Value()
 func TestSetInput(t *testing.T) {
 	r := New()
 	i := r.CreateInput(1)
-	if i.Value() != 1 {
-		t.Fatalf("i.Value() doesn't match initial value")
-	}
+	assertCellValue(t, i, 1, "i.Value() doesn't match initial value")
 	i.SetValue(2)
-	if i.Value() != 2 {
-		t.Fatalf("i.Value() doesn't match changed value")
-	}
+	assertCellValue(t, i, 2, "i.Value() doesn't match changed value")
 }
 
 // The value of a compute cell is determined by the value of the dependencies.
@@ -39,13 +42,9 @@ func TestBasicCompute(t *testing.T) {
 	r := New()
 	i := r.CreateInput(1)
 	c := r.CreateCompute1(i, func(v int) int { return v + 1 })
-	if c.Value() != 2 {
-		t.Fatalf("c.Value() isn't properly computed based on initial input cell value")
-	}
+	assertCellValue(t, c, 2, "c.Value() isn't properly computed based on initial input cell value")
 	i.SetValue(2)
-	if c.Value() != 3 {
-		t.Fatalf("c.Value() isn't properly computed based on changed input cell value")
-	}
+	assertCellValue(t, c, 3, "c.Value() isn't properly computed based on changed input cell value")
 }
 
 // Compute cells can depend on other compute cells.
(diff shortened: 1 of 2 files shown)


react/react_test.go
```diff
@@ -21,47 +24,96 @@ func TestTestVersion(t *testing.T) {
...
+	assertCellValue(t, c, 3, "c.Value() isn't properly computed based on initial input cell values")
+	i1.SetValue(4)
+	assertCellValue(t, c, 6, "c.Value() isn't properly computed when first input cell value changes")
+	i2.SetValue(8)
+	assertCellValue(t, c, 12, "c.Value() isn't properly computed when second input cell value changes")
 }
 
-// Compute cells can depend on other compute cells.
+// Compute 2 cells can depend on compute 1 cells.
 func TestTopology(t *testing.T) {
```
petertseng: Perhaps a name change is in order now, since there are now multiple topologies as opposed to before when this was the only topology test. Perhaps `TestCompute2Diamond` because the topology is kind of in a diamond shape. Suggestions very welcome. Or if you think it's fine I'll leave it, but I suspect I should change.

react/react_test.go
```diff
@@ -21,47 +24,96 @@ func TestTestVersion(t *testing.T) {
...
+		digitToAdd := i
+		c = r.CreateCompute1(c, func(v int) int { return v*10 + digitToAdd })
+	}
+	assertCellValue(t, c, 12345678, "c.Value() isn't properly computed based on initial input cell value")
+	inp.SetValue(9)
+	assertCellValue(t, c, 92345678, "c.Value() isn't properly computed based on changed input cell value")
+}
+
+// Compute 2 cells can depend on other compute 2 cells.
+func TestCompute2Tree(t *testing.T) {
```
petertseng: simple 3 input -> 2 compute2 -> 1 compute2 tree. I suppose I could go pretty deep like I did with the compute1 chain test though... 16 inputs -> 8 compute2 -> 4 compute2 -> 2 compute2 -> 1 compute2 for instance. I'll think about it.

react/react_test.go
petertseng: Renamed.

react/react_test.go
petertseng: I'll hold off. This test already covers that compute2 can depend on compute2 for one level of nesting, and I've tested many levels of nesting in the compute1 chain. I can't imagine someone would make a solution that passes both those tests but that wouldn't support trees with many levels, so I suspect a large tree test would add little value.

react/react_test.go
```diff
@@ -21,47 +24,96 @@ func TestTestVersion(t *testing.T) {
...
+	i1.SetValue(4)
+	assertCellValue(t, c, 6, "c.Value() isn't properly computed when first input cell value changes")
+	i2.SetValue(8)
+	assertCellValue(t, c, 12, "c.Value() isn't properly computed when second input cell value changes")
 }
 
-// Compute cells can depend on other compute cells.
-func TestTopology(t *testing.T) {
+// Compute 2 cells can depend on compute 1 cells.
+func TestCompute2Diamond(t *testing.T) {
```
petertseng: renamed for https://github.com/exercism/xgo/pull/188#discussion_r43110695

kytrinyx:
I think the added granularity here is a huge improvement. @soniakeys do you see any gotchas with this?

soniakeys:
I can't comment right away because I hadn't looked at this exercise before.  It looks rather complex.  I should probably try to solve it on my own first.

soniakeys:
Mm, okay, I coded a solution but it doesn't pass the last test,

```
// Callbacks should only be called once even though
// multiple dependencies have changed.
func TestOnlyCallOnceOnMultipleDepChanges(t *testing.T)
```

The readme, from x-common/react.md, asks for callback mechanism which would be adequate to propagate values, except this last test disallows it.  I'm not sure what the reasoning was there.

petertseng:
That test was long before my time so you'd have to ask the original author for the very best answer, but here is my attempt at a reasoning for that answer (I agree with the presence of that test).

Consider the diamond topology. There's one input. There are two Compute1 cells that depend on that input, let's say they're `c1_1 = input + 1` and `c1_2 = input - 1`. Finally there is a Compute2 cell that is `c2 = c1_1 * c1_2`.

Let's say initially the value of `input` is 0, such that the value of `c2` is -1 (`1 * -1 = -1`). Now we change `input` to 4. We can see that the final value of `c2` should be `5 * 3 = 15`. However, when propagating the values, you might first update the value of `c1_1` based on the input, and update the value of `c2` based on `c1_1`. At that intermediate point in time, `c2`'s value would be `5 * -1 = -5` instead of the final answer of `15`.

So the question becomes: *Should we fire the callback on the intermediate value of -5*?

I argue: "No, we should not". The value of -5 only appears while the system is in a half-updated state, it doesn't truly make sense to notify the callback of that half-updated value. We should only notify the callback of the final answer once all values have propagated, which is 15.

If someone uses this react system to model a circuit, then sure, propagation delays probably mean that the updated signals arrive at ever-so-slightly different times (by necessity of physical considerations), so the value of that multiplier would be in that half-updated state for a fleeting moment, but again I argue that that is not really a value of interest; we really care about the stable values.

That's why I think the multiple dependency test is there. If you like I can open a PR to add a comment explaining that. It could be useful to explain it.

soniakeys:
Well, my issues are unrelated to this PR.  It looks fine to me and I don't want to hold it up.

Issue #189, opened 2015-10-28
React adds extra requirement


soniakeys:
This issue continues discussion that started at https://github.com/exercism/xgo/pull/188.

When I tried solving this exercise I was bothered by the final requirement of the the test program, that a callback on a cell only be called once when both of the cell's inputs change.  My issue is that my reading of the readme not only does not make this requirement, but even worse suggests an implementation that does not work with this requirement.  The readme says "...allow a change to one value to automatically propagate to other values..." and then ..."allow for registering callbacks which get called when the value of a cell changes."  This to me indicates an implementation where changes are propagated by callbacks.

Working the exercise, when I saw the final test fail I understood its intent but I felt I had been led down a garden path.  I felt like the readme had indicated that the problem was to be implemented a certain way.  But then after coding what the readme indicated, the test program says at the very end, "sorry you did it the way the readme suggested, now here's an extra requirement tacked on that invalidates all that, so you must now code something completely different."  I felt cheated and the problem stopped being fun at that point.  Now, this happens a lot in the real world of course.  Stakeholders give you a set of requirements, then QA demands something that expands the scope of the project and that you feel was not within the intent of the original requirements.  It struck me as devious, so I wondered if that was deliberate here, following the theme of the site.

As far as surprising behavior, if I had defined a cell with two inputs and I knew both inputs were changing,  I would be alarmed to get only a single update.  It's a difference of assumptions and the readme doesn't suggest either assumption.

My preferences in order would be

1.  Remove the test program requirement that a callback be called only once when both inputs change.
2.  Update the readme to be more explicit and less potentially misleading.
3.  Leave the problem devious.

And with this choices 2 or 3, perhaps add a new test case where changes to single inputs would produce a different result but changes to both inputs leave the result the same.  Presumably you would not want a callback in this case.  Like with `b = a+1`, `c = a-1`, `d = b-c`, `d` should always be 2 and a callback attached to `d` should not be called when `a` changes.

petertseng:
I understand I lack the authority to decide, but I will post my input in case it can be helpful.
While it might appear that I argue very strongly for a particular solution and have many arguments for the solution I support, I understand that I have no personal stake in the matter and that ultimately I would defer to your reasoned decision.

Post is long.
Points I feel are important are bolded; those are my tl;dr, along with my following statement:
[...]

PR #192, opened 2015-10-30
React tests

petertseng:
Details in individual commit messages. One commit improves error reporting (report line number of failed tests in common function), other adds new tests from #189 (should those two tests be in a single commit, or two?)

(fun fact: my solution failed the test of "callback not called if dependencies change but final computed value didn't change". iteration 3 of my solution fixes it)

### DIFF
--- a/react/react_test.go
+++ b/react/react_test.go
@@ -1,6 +1,9 @@
 package react
 
-import "testing"
+import (
+	"runtime"
+	"testing"
+)
 
 // Define a function New() Reactor and the stuff that follows from
 // implementing Reactor.
@@ -26,8 +29,9 @@ func TestTestVersion(t *testing.T) {
 
 func assertCellValue(t *testing.T, c Cell, expected int, explanation string) {
 	observed := c.Value()
+	_, _, line, _ := runtime.Caller(1)
 	if observed != expected {
-		t.Fatalf("%s: expected %d, got %d", explanation, expected, observed)
+		t.Fatalf("(from line %d) %s: expected %d, got %d", line, explanation, expected, observed)
 	}
 }
 
--- a/react/react_test.go
+++ b/react/react_test.go
@@ -11,9 +11,10 @@ import (
 // Also define an exported TestVersion with a value that matches
 // the internal testVersion here.
 
-const testVersion = 2
+const testVersion = 3
 
 // Retired:
+//  2 9d1c247d698fb68119379ef92f2b15ff46175f7c
 //  1 afa5c1278857457403a30479663d26d4e1c8c496
 
 // This is a compile time check to see if you've properly implemented New().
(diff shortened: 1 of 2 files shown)


react/react_test.go
```diff
@@ -192,6 +197,35 @@ func TestCallbackAddRemove(t *testing.T) {
 	}
 }
 
+func TestMultipleCallbackRemoval(t *testing.T) {
+	r := New()
+	inp := r.CreateInput(1)
+	c := r.CreateCompute1(inp, func(v int) int { return v + 1 })
+
+	calls := make([]int, 5)
```
petertseng: too many magic 5s. extracting.

react/react_test.go
```diff
@@ -192,6 +197,37 @@ func TestCallbackAddRemove(t *testing.T) {
 	}
 }
 
+func TestMultipleCallbackRemoval(t *testing.T) {
+	r := New()
+	inp := r.CreateInput(1)
+	c := r.CreateCompute1(inp, func(v int) int { return v + 1 })
+
+	numCallbacks := 5
```
petertseng: why 5? no real reason; it was arbitrary. 2 is enough to trigger the problem that @soniakeys wanted me to trigger, really.

react/react_test.go
```diff
@@ -192,6 +197,37 @@ func TestCallbackAddRemove(t *testing.T) {
...
+
+	calls := make([]int, numCallbacks)
+	handles := make([]CallbackHandle, numCallbacks)
+	for i := 0; i < numCallbacks; i++ {
+		// Rebind i, otherwise all callbacks will use i = numCallbacks
```
petertseng: maybe the comment (explaining the `i := i`) isn't necessary, maybe if someone is familiar enough with go and they see `i := i` in a loop body they know what it means

Also removing that line will break the tests (they'll all use `calls[5]` which doesn't exist) so maybe that will be compelling enough to indicate that the line shouldn't be removed. Will let reviewers make the call.

petertseng:
While we're thinking about possible tests, here's a possibility, let me know if you want it in:

Something like: 

```go
output := r.CreateCompute1(...)
timesCalled := 0
cb1 := output.AddCallback(func(int) {})
cb2 := output.AddCallback(func(int) { timesCalled++ })
for i := 0; i < 10; i++ {
    output.RemoveCallback(cb1)
}
input.SetValue(...)
// check that callback 2 still gets called i.e. removing an already-removed callback doesn't break other callbacks, as interfaces.go says RemoveCallback "removes a previously added callback, if it exists."
```

I don't know though, this mostly sees whether slice-based callback solutions are doing the right thing, and the multiple add multiple remove test probably already takes care of that...

soniakeys:
This all looks fine to me.  I don't have strong feelings about any of the other points, 5, i := i, or testing idempotency of Remove.

petertseng:
Decided to add idempotence test, leaving the others as is. Found a stray 5 that I forgot to change to numCallbacks as well.

soniakeys:
This all really great.  Thank you!

PR #193, opened 2015-10-30
ledger: Make initial implementation pass test

petertseng:
The readme for the ledger exercise claims that the current code
consistently passes the tests. Unfortunately, on my machine it
consistently does not pass the tests. Every time I try, I get this:

```
--- FAIL: TestFormatLedgerSuccess (0.00s)
        ledger_test.go:267: FormatLedger for input named "credit and
debit" was expected to return...
                Date_______|_Description_______________|_Change
                01/01/2015_|_Buy_present_______________|______($10.00)
                01/02/2015_|_Get_present_______________|_______$10.00_

                ...but returned...
                Date_______|_Description_______________|_Change
                01/02/2015_|_Get_present_______________|_______$10.00_
                01/01/2015_|_Buy_present_______________|______($10.00)
```

Looking into the initial implementation, the implementation is creating
a goroutine for each line of the ledger, and the main goroutine is
collecting the formatted lines from a channel that all the goroutines
are sending to. Unfortunately, there are no guarantees as to the order
in which the goroutines will send on the channel. In fact on my machine
they consistently send the results in the wrong order, so I always fail
the test.

Well, I can fix that, we'll just make each goroutine aware of the index
of the line it's formatting, and the main goroutine needs to reorder
them in the right order. Now it will consistently pass the tests.

The alternative, of course, would be to modify the readme to say that
the ledger might not even pass the tests all the time.

### DIFF
--- a/ledger/ledger.go
+++ b/ledger/ledger.go
@@ -66,13 +66,15 @@ func FormatLedger(currency string, locale string, entries []Entry) (string, erro
 	}
 	// Parallelism, always a great idea
 	co := make(chan struct {
+		i int
 		s string
 		e error
 	})
-	for _, et := range entriesCopy {
-		go func(entry Entry) {
+	for i, et := range entriesCopy {
+		go func(i int, entry Entry) {
 			if len(entry.Date) != 10 {
 				co <- struct {
+					i int
 					s string
 					e error
 				}{e: errors.New("")}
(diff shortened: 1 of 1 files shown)


ledger/ledger.go
```diff
@@ -200,19 +207,24 @@ func FormatLedger(currency string, locale string, entries []Entry) (string, erro
...
 	for _ = range entriesCopy {
 		v := <-co
 		if v.e != nil {
 			return "", v.e
 		} else {
-			s += v.s
+			ss[v.i] = v.s
 		}
 	}
+	for i := 0; i < len(entriesCopy); i++ {
```
petertseng: the initial code is supposed to be badly written so I intentionally am not using `for _, s := range ss` here. Be assured that I would not normally write code like this...

soniakeys:
Yes, it looks like this was the intent of the author.

PR #196, opened 2015-11-01
add exercise pov

soniakeys:
I like graph algorithms so I kind of liked this exercise until now implemented in just Clojure and Haskell.  I used the same test cases.  I added some paragraphs of explanation in the test program.  And I kept the API as flexible as I could, allowing solvers to define their own graph representation.  Also I split the tests into two parts.  Part one is just to create the graph and code a method for inspecting the graph.  Code-wise, this is a significant part of the exercise and yet it's not even the part advertised as the crux of the exercise.  Since part two is the interesting re-rooting algorithm, I did a few more things to try to simplify part one.  The test program promises to build a graph that is tree, to build it from the bottom up, and to pass the root to the the re-rooting method.  To avoid future test case additions or changes breaking these promises, the example program validates them.  This way CI validates the test cases.

Finally I added benchmarks for exploring the relative times of the two parts.

### DIFF
--- a/pov/pov_test.go
+++ b/pov/pov_test.go
@@ -1,0 +1,295 @@
+package pov
+
+import (
+	"reflect"
+	"sort"
+	"testing"
+)
+
+const testVersion = 1
+
+// POV / reparent / change root of a tree
+//
+// API:
+// type Graph
+// func New() *Graph
+// func (*Graph) AddNode(nodeLabel string)
+// func (*Graph) AddArc(from, to string)
+// func (*Graph) ArcList() []string
+// func (*Graph) ChangeRoot(oldRoot, newRoot string) *Graph
+//
+// The type name is Graph because you'll probably be implementing a general
+// directed graph representation, although the test program will only use
+// it to create a tree.  The term "arc" is used here to mean a directed edge.
+//
+// The test program will create a graph with New, then use AddNode to add
+// leaf nodes.  After that it will use AddArc to construct the rest of the tree
+// from the bottom up.  That is, the `to` argument will aways specify a node
+// that has already been added.
+//
+// ArcList is a dump method to let the test program see your graph.  It must
+// return a list of all arcs in the graph.  Format each arc as a single string
+// like "from -> to". The test program can then easily sort the list and
+// compare it to an expected result.  You do not need to bother with sorting
+// the list yourself.
+//
+// All this graph construction and dumping must be working before you start
+// on the interesting part of the exercise, so it is tested separately as
+// a first test.
+//
+// API function ChangeRoot does the interesting part of the exercise.
+// OldRoot is passed (as a convenience) and you must return a graph with
+// newRoot as the root.  You can modify the original graph in place and
+// return it or create a new graph and return that.  If you return a new
+// graph you are free to consume or destroy the original graph.  Of course
+// it's nice to leave it unmodified.
+
+type arc struct{ fr, to string }
+
+type testCase struct {
+	description string
+	leaves      []string
+	arcPairs    []arc
+	root        string
+	arcStrings  []string
+	reRooted    []string
+}
+
+var testCases = []testCase{
+	{
+		description: "singleton",
+		leaves:      []string{"x"},
+		arcPairs:    nil,
+		root:        "x",
+		arcStrings:  nil,
+		reRooted:    nil,
+	},
+	{
+		description: "simple tree",
+		leaves:      []string{"sibling", "x"},
+		arcPairs: []arc{
+			{"parent", "sibling"},
+			{"parent", "x"},
+		},
+		root: "parent",
+		arcStrings: []string{
+			"parent -> sibling",
+			"parent -> x",
+		},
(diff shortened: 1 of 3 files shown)


kytrinyx:
These next two weeks I've got some big, complicated things on my plate, and I don't have the capacity to think this through. @petertseng do you want to work with Sonia on this one?

petertseng:
No problem, I expect I'll review tonight, will probably work through the problem to see if everything seems OK on that front. Someone remind me if you don't hear back from me in 48 hours.

petertseng:
Oops, accidentally pushed my impl on top of this branch, removed it. Meant to put it at my branch https://github.com/petertseng/xgo/commit/ee8d33b971b01199ceef6b0370e5c5b4fe187d5b

Exercise seems reasonable enough (though I admit I already did it in Clojure so maybe I had an advantage because of that). Doesn't have the "trace a path" portion of Clojure/Haskell but I don't suppose that's terribly important since the meat of it (rerooting) is there. I think it's good that the internal representation was left flexible and it just gets inspected via `ArcList`. 

This may be a bit bikesheddy, but any thoughts on Arc vs Edge terminology? I don't care too much about this so if what you currently have is to your preference then go with that.

I'll add a few line comments on places I think the code could use some deduplication but otherwise I generally feel happy about this.

pov/pov_test.go
```diff
@@ -0,0 +1,317 @@
...
+			}
+			t.Fatal("incorrect graph construction")
+		}
+	}
+}
+
+func TestChangeRoot(t *testing.T) {
+	// very similar code above in TestConstruction
+	for _, tc := range testCases {
+		g := New()
```
petertseng: given that L261-L267 are the same as in above, I might suggest adding a function on the `testCase` struct that generates the graph from it, to avoid the code repetition.

(obviously that would need `type testCase struct { ... }` rather than using the anonymous struct as it currently is)

pov/pov_test.go
```diff
@@ -0,0 +1,317 @@
...
+		for _, l := range tc.leaves {
+			g.AddNode(l)
+		}
+		for _, a := range tc.arcPairs {
+			g.AddArc(a.fr, a.to)
+		}
+		got := g.ArcList()
+		gs := append([]string{}, got...)
+		sort.Strings(gs)
+		if !same(gs, tc.arcStrings) {
```
petertseng: Perhaps print`tc.description` somewhere?

pov/pov_test.go
```diff
@@ -0,0 +1,317 @@
...
+	for _, tc := range testCases {
+		g := New()
+		for _, l := range tc.leaves {
+			g.AddNode(l)
+		}
+		for _, a := range tc.arcPairs {
+			g.AddArc(a.fr, a.to)
+		}
+		g = g.ChangeRoot(tc.root, "x")
+		got := g.ArcList()
```
petertseng: I imagine there could be a function called `assertArcListSame(got, want []string, description string)` and it'd be called in a manner like `assertArcListSame(g.ArcList(), tc.reRooted, fmt.Sprintf("incorrect root change for %s", tc.description))` or `assertArcListSame(g.ArcList(), tc.arcStrings, fmt.Sprintf("incorrect graph construction for %s", tc.description))` - it would handle the sorting of `got` and the comparisons. Would greatly reduce duplication.

petertseng:
One more thought: Clojure and Haskell have tested what happens if the `newRoot` does not exist in a call to `graph.ChangeRoot(oldRoot, newRoot)` - any thought on whether we should test that as well? (And if we do, I think it may be at least reasonable to test the input graph isn't destroyed in such a case)

soniakeys:
Sorry, confused for a second and made a mistake with git.

soniakeys:
Duplicate code factored out and "arc" definition moved up.

It bugs me to see "edge" used to mean "directed edge" when I think edge by itself most commonly means "undirected edge".  I know arc is not as widely used as edge but it is used sometimes and I like it better than directed edge.  It's short and easy to type and clearly distinct from "edge".

Also I think it's fine to omit the test for a root not in the tree.  The reason is that there are lots of other curves we could throw.  Why throw just this one?  What if the node we passed to ChangeRoot was in the tree but wasn't the root?  Why even pass the old root as an argument?  What if we gave them a graph that wasn't a tree but had cross edges--or back edges?  How about a forest?  What if the old root and new root were in different connected components?

petertseng:
OK. Given you feel strongly about arcs and I can go either way, let's roll with it.

As for the question of what error cases to test for, I suppose a lot of the time I base my answers to this question on "what mistakes do I observe in people's submissions?" so I suppose I can't speak to this until we see some submissions.

The alternative answer to the question "why aren't we testing all these other error cases?" is to say "well maybe we should test them all!" like poker does: https://github.com/exercism/xgo/blob/master/poker/poker_test.go#L237 . Possible that we could test all the error cases you describe using something like that.

You convinced me, so I can go either way on this as well, so I think we're good to ship this and let people start working on it. I'll leave the merge button for you to click in case you do want to make some change

PR #198, opened 2015-11-03
Tweak stock comments in leap stub solution

kytrinyx:
I just nitpicked three solutions in a row where they had
left the comments in place.

I'm attempting to move the instructions about deleting closer to where
they're writing code, and then also yelling loudly in all caps.

It might help.


soniakeys:
Sure.  We'll have to see how all this goes.  In the end we might have to take all the comments out of the stub files and maybe put the information somewhere else.

Issue #195, opened 2015-10-31
ledger: handling unicode


petertseng:
While doing the ledger exercise, I noticed that both the example solution and initial implementation have a defect in how they count lines that are too long. The "Description" column of the ledger has width 25, so any string that has more than 25 characters should display the first 22 characters followed by "..."

And yet, the example solution and initial implementation both check for length greater than 27 characters, and cut at index 24 instead of 22.
https://github.com/exercism/xgo/blob/master/ledger/example.go#L44
https://github.com/exercism/xgo/blob/master/ledger/ledger.go#L98

You may ask, is it something to do with the padding in the string `" | "` that separates each column? No, it's not even that because there is in fact a test that tests for proper cutoffs: https://github.com/exercism/xgo/blob/master/ledger/ledger_test.go#L128

In fact it's because that is the only test that tests too-long descriptions and it happens to have two unicode codepoints that take up two bytes each, the two ö (C3 B6 in utf-8). Therefore, for that particular string, cutting at the 24th byte is cutting at the 22nd rune, and everything is correct.

I'm not sure if I'm comfortable leaving that code in example.go, because it's very much not generic - only works on strings with two more bytes than runes. I'm very tempted to add tests that more thoroughly test cutoffs, such as https://github.com/exercism/xgo/compare/master...petertseng:ledger-unicode and then fix the code. Both the example and initial implementations currently fail:

```
--- FAIL: TestFormatLedgerSuccess (0.00s)
	ledger_test.go:347: FormatLedger for input named "overlong ascii descriptions" was expected to return...
		Date_______|_Description_______________|_Change
		01/01/2015_|_This_description_is_wa..._|________$1.00_
		
		...but returned...
		Date_______|_Description_______________|_Change
		01/01/2015_|_This_description_is_way_..._|________$1.00_
```

(They cut the string at the wrong place and misalign the columns)

Thing is, if I add the tests, that means the initial implementation has to pass the tests too, and that means I have to also write bad code that happens to pass the tests. I'll consider whether that's a thing I want to do.

So while I think about that, I wanted to open an issue and ask what y'all think about it as well. Worth adding the tests? And ideas on how to write an intentionally-bad implementation that nevertheless passes tests?

[...]

PR #199, opened 2015-11-04
ledger: remove multibyte characters

petertseng:
The discussion we had in #195 indicates that Unicode introduces a whole
host of issues that are better suited for their own exercise. Let's keep
the ledger exercise focused on refactoring. This avoids a potential
source of confusion of "why is it cutting at 24 instead of 22?"

ledger test version is bumped for this.

### DIFF
--- a/ledger/example.go
+++ b/ledger/example.go
@@ -10,7 +10,7 @@ import (
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


soniakeys:
Looks great.

PR #206, opened 2015-12-02
Add helloworld exercise

k4rtik:
This PR is for https://github.com/exercism/todo/issues/32

I am unclear about the role of `testVersion`, included as other Go exercises have it too.

The previous first exercise (leap) has fairly good comments in its stub. Can add some comments for helloworld, if needed, but this is really a very trivial first exercise.

### DIFF
--- a/config.json
+++ b/config.json
@@ -4,6 +4,7 @@
   "repository": "https://github.com/exercism/xgo",
   "active": true,
   "problems": [
+    "helloworld",
     "leap",
     "clock",
     "gigasecond",
--- a/config.json
+++ b/config.json
@@ -4,7 +4,7 @@
   "repository": "https://github.com/exercism/xgo",
   "active": true,
   "problems": [
-    "helloworld",
+    "hello-world",
     "leap",
     "clock",
     "gigasecond",
--- a/hello-world/example.go
+++ b/hello-world/example.go
@@ -1,4 +1,4 @@
-package helloworld
+package hello
 
 const TestVersion = 1
 
--- a/hello-world/example.go
+++ b/hello-world/example.go
@@ -2,6 +2,9 @@ package hello
 
 const TestVersion = 1
 
-func HelloWorld() string {
-	return "Hello, World!"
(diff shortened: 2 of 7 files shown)


kytrinyx:
Cool!

Let's see, the slug needs to be `hello-world`, not `helloworld`, in order to match the metadata here: https://github.com/exercism/x-common/blob/master/hello-world.md

Let's call the package `hello` rather than `helloworld` (the style guide tends towards shorter names).

The `testVersion` is documented here: https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#exercise-versioning

I think we should have a stub file with stub comments since this is the first exercise. We can tone down the stub comments in the leap exercise once we have this exercise in place, since they'll already have worked through this.

@soniakeys thoughts?

k4rtik:
That makes total sense, I will try moving some stub comments from leap to hello-world.

Looking at the [spec](https://github.com/exercism/x-common/blob/master/hello-world.md), I realized the exercise also needs to support taking user's name as input.

kytrinyx:
Ah, good catch. I missed that!

kytrinyx:
@k4rtik Are you still working on this? If not, let me know and I can take a stab at tweaking it to get it integrated into the track.

k4rtik:
@kytrinyx I had worked on it at that time, but did not like the ugliness introduced by needing to pass an empty string to a simple Hello World program. I just pushed my last commit and an additional test. Please feel free to merge or modify as you see fit.

kytrinyx:
OK, thanks, I'll take a look at it.

kytrinyx:
Merged manually in a79a49e78e7d5e0a668d26c4dbd2b99ad08170f9

k4rtik:
Thanks for merging this, I liked it that you abstracted out on the additional test I wrote in a hurry.

kytrinyx:
Thanks for taking the time to make the first draft of it, it made it much easier to tweak it and get it into the track.

PR #208, opened 2015-12-07
triangle: Ensure no pair of triangle types are equal

### LINKED ISSUE #207: Tests for triangle can be solved by an empty implementation.
Hi,

I started working on this problem  (`triangle`), only to notice that after I added stubs for entities in my code, all tests passed. Basically, my "solution" was this:

```
package triangle

func KindFromSides(a, b, c float64) Kind {
	return Kind{}
}

type Kind struct {
}

var Equ = Kind{} // equilateral
var Iso = Kind{} // isosceles
var Sca = Kind{} // scalene
var NaT = Kind{} // not a triangle
```

and all of the tests which I ran locally passed. I presume this is incorrect?

petertseng:
As reported in #207, the triangle tests can be solved by a trivial empty
implementation that simply sets all four triangle types equal to one
another. We should prevent false positives, and we can do so by checking
that Equ != Iso, etc.

I verified that 1) the trivial solution fails the new test, and 2) the
example solution passes the test.

Fixes #207

### DIFF
--- a/triangle/triangle_test.go
+++ b/triangle/triangle_test.go
@@ -47,6 +47,29 @@ func init() {
 	testData = append(testData, nf[1:]...)
 }
 
+// Test that the kinds are not equal to each other.
+// If they are equal, then TestKind will return false positives.
+func TestKindsNotEqual(t *testing.T) {
+	kindsAndNames := []struct {
+		kind Kind
+		name string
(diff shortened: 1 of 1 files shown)


kytrinyx:
This looks good to me.

PR #209, opened 2015-12-13
Add benchmark to clock exercise

deild:


kytrinyx:
The one thing I wonder about here is in the `Add` benchmark. Should we create a single clock outside of the main benchmark loop, and then only call the `Add` function within?

kytrinyx:
Not sure how I missed this. Thanks @deild!

PR #211, opened 2015-12-27
Fix a few typos in the `clock_test` file

thenickcox:
This commit fixes a few typos in the `clock_test` file. It also adds backticks around a function name to distinguish it from plain words in the documentation.

### DIFF
--- a/clock/clock_test.go
+++ b/clock/clock_test.go
@@ -11,14 +11,14 @@ import (
 // (Clock) String() string         // a "stringer"
 // (Clock) Add(minutes int) Clock
 //
-// Add should also handle subtraction by accepting negative values.
+// The Add method should also handle subtraction by accepting negative values.
 // To satisfy the readme requirement about clocks being equal, values of
 // your Clock type need to work with the == operator.
 //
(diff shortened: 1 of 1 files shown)


kytrinyx:
Thanks, nice catch on those typos!

Go comments don't use any formatting like backticks, underscores, etc. From [Effective Go](https://golang.org/doc/effective_go.html#commentary)

> Comments do not need extra formatting such as banners of stars. The generated output may not even be presented in a fixed-width font, so don't depend on spacing for alignment—godoc, like gofmt, takes care of that. The comments are uninterpreted plain text, so HTML and other annotations such as _this_ will reproduce verbatim and should not be used. One adjustment godoc does do is to display indented text in a fixed-width font, suitable for program snippets.

How about something like _The Add method..._ to give the type of context that the backticks would provide in markdown?

thenickcox:
@kytrinyx Works for me! It also gave me a great opportunity to refresh myself on the subtle differences between methods and functions. Thanks, Katrina!

kytrinyx:
Lovely! And thanks for squashing, too.

PR #215, opened 2016-01-19
Fix style violations in bracket push

kytrinyx:
The package name should not have an underscore.
The test file name should only have underscores, not a mix of hyphens and underscores.
Constants should be mixedCaps.
The constants that the reference solution invents should not be exported.

I also gave the constants a prefix for readability.

This change is not backwards compatible due to the change of the package name, so I added a TestVersion.

### DIFF
--- a/bracket-push/bracket_push_test.go
+++ b/bracket-push/bracket_push_test.go
@@ -1,8 +1,10 @@
-package bracket_push
+package brackets
 
 import (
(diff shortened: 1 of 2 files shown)


kytrinyx:
hold on, this isn't actually asserting on the test version.

petertseng:
everything looks good from here :+1:

PR #216, opened 2016-01-19
Fix style violations in scrabble-score

kytrinyx:
The package name should not have an underscore.

I also added a comment to the Score function, for the sake of completeness.


kytrinyx:
hold on, this needs a test version.

petertseng:
Looks good :+1:

PR #217, opened 2016-01-19
Fix style violation in error-handling

kytrinyx:
Package names should not have underscores. I added an assertion for the test version
since this is a backwards-incompatible change.


petertseng:
Code changes look OK.

Note that as of 55de1e8e09f76a1b6b3d41b70b9d81e43ce249bd the error handling exercise is temporarily ignored because of the binary file you hvae deleted in this PR.

I think the file is unnecessary. I ran `go test` with the file deleted and it passed. So I think it is right to delete it.

Can probably un-ignore the exercise now?

kytrinyx:
Oh, right--I forgot about the binary file thing. I can't remember why it was there in the first place, but I did check that the tests passed before I committed the change. Hm. Maybe un-ignore it, then.

petertseng:
Assuming that the commit message in your commit is accurate and is the only reason this exercise has to be ignored, I would suspect we are good. Any way to test?

kytrinyx:
Yeah, let me think for a second. I think the way to test this is to turn it back on, run x-api locally and configure the CLI to fetch from the local version of the x-api. I'll give that a go and report back.

kytrinyx:
Nope, we still have a problem: `JSON::GeneratorError - source sequence is illegal/malformed utf-8`. Since nobody has done this problem I'm going to change the name of the package but not change anything else, and leave it ignored until the CLI can talk to v3 of the x-api (which zips everything instead of sending the contents of the files as a json response).

kytrinyx:
OK, this is odd. I ran the test locally, and that deleted the binary file. Do you remember what the binary file was for? Was it the package compiled with the tests?

kytrinyx:
Huh. Yeah, I think that was it.

Darn, now I want to delete it from the history so that the repo isn't so much bigger than it needs to be, but then we have to jump through hoops for everyone to get on the same page.

kytrinyx:
Another experiment. Running the tests with `error_handling` as the package name leaves the binary. Running it with `erratum` as the package name deletes the binary. I'm pretty confused right now.

kytrinyx:
However: deleting the binary without changing the package name doesn't cause the tests to fail, and it also doesn't recreate the binary (at least not in the working directory, maybe some temp dir somewhere).

kytrinyx:
Removed binary in 46c376fa9f2b9d54f539ebfd73517ec7e0be6af4

petertseng:
That binary business (with it being automatically created and deleted) is super weird, I wonder if we can figure out why that happens? Obviously not anything important, but for curiosity's sake. I can reproduce (testing with `erratum` deletes). Strange.

PR #218, opened 2016-01-19
Remove stutter from API of meetup

kytrinyx:


petertseng:
Yup seems good. One small question - the git shas in the old version log: Should that be "first commit that does NOT comply with this version" or "last commit that DOES comply?"

I only ask because in this PR you've done the former, whereas IIRC I did the latter in my previous PRs. I guess there isn't a hard rule about it but I'll ask in case anyone cares deeply.

kytrinyx:
> One small question - the git shas in the old version log

I have no idea!

> Is that log even useful?

I'm not so sure. I think `git log -p` on the file would be the better way of go.
I think I'll just go through and delete these.

petertseng:
I would be in agreement.

In the meantime, :+1:  on this PR

kytrinyx:
Sweet. Thanks!

PR #226, opened 2016-01-23
Show function argument to failure message in hello-world. Fixes #225

### LINKED ISSUE #225: helloworld: If test fails, print out the input that caused the test to fail
right now, just the expected/observed output is printed. The final missing piece of the puzzle is the input that caused the failure.

Sure, for the hello world exercise this is easy to discern anyway, but good to stay in the habit. Besides, some new problem may decide to copy hello world's tests as a starting point...

kytrinyx:
I had a moment to tackle this one, and it was a quick fix.


petertseng:
Ah good, thanks. Wouldn't have been able to get to this for a few hours.

PR #227, opened 2016-01-23
largest-series-product: generate tests

petertseng:
This commit adds a generator for the common test cases of
largest-series-product. Other effects of this change:

* Cases moved from largest_series_product_test.go to cases_test.go
* example solution needs to reject negative spans
* example solution needs to start at 0 minimum
* test version bump


kytrinyx:
Nice! This looks great, thank you.

PR #224, opened 2016-01-21
Move exercises to subdirectory. Fixes #223

kytrinyx:

PR #228, opened 2016-01-23
import ../../gen in all generators rather than ../gen

petertseng:
In d895e12bc457dad8c024ad457380ade2a5f55aa9 we moved all exercises to
the exercises subdirectory. So, whereas the `gen` directory used to be
one directory above a given generator file, now it is two directories
above. So we need ../../gen rather than ../gen

### DIFF
--- a/exercises/bob/example_gen.go
+++ b/exercises/bob/example_gen.go
@@ -6,7 +6,7 @@ import (
 	"log"
 	"text/template"
 
-	"../gen"
+	"../../gen"
 )
 
 func main() {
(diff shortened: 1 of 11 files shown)


petertseng:
The perils of relative import paths!

But in this case I understand why it is done (to be independent of where the xgo repo is checked out) so we kind of have to bear with it.

kytrinyx:
Yeah, it's really handy to not have this in the GOPATH. (I have all the different tracks checked out together).

petertseng:
> In d895e12

I should say in #224 so we get the reference in the PR.

PR #232, opened 2016-01-30
triangle: Allow triangles where a+b == c

### LINKED ISSUE #231: triangle: incorrect test in some tracks
Please check if there's a test that states that a triangle with sides 2, 4, 2 is invalid. The triangle inequality states that for any triangle, the sum of the lengths of any two sides must be greater than or equal to the length of the remaining side. If this doesn't affect this track, go ahead and just close the issue.

petertseng:
These are degenerate triangles since they're really just lines, but
they're still good enough under the triangle inequality (which
explicitly allows a+b == c) so let's go ahead and test for them.

Test for isosceles and scalene added. Equilateral is impossible.

Closes #231

### DIFF
--- a/exercises/triangle/example.go
+++ b/exercises/triangle/example.go
@@ -26,9 +26,11 @@ func KindFromSides(a, b, c float64) Kind {
 	}
 	// sides are now sorted
 	switch {
+	case math.IsInf(c, 1): // largest side is +Inf, guards against (3, +Inf, +Inf)
+		return NaT
 	case a <= 0:
 		return NaT
-	case a+b <= c: // triangle inequality
+	case a+b < c: // triangle inequality
 		return NaT
 	case a == b:
 		if b == c {
--- a/exercises/triangle/example.go
+++ b/exercises/triangle/example.go
@@ -4,6 +4,8 @@ import "math"
 
 type Kind string
 
+const TestVersion = 1
+
 const (
 	Equ Kind = "equilateral"
 	Iso = "isosceles"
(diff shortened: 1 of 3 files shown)


petertseng:
Going to wait a bit and monitor the issues in the other tracks to see if any other track maintainers have objections since I'd find it better if all tracks are in agreement on this one. If everyone's convinced then we're good to merge.

petertseng:
Ah hold on. Should probably bump test version.

exercises/triangle/example.go
```diff
@@ -26,9 +26,11 @@ func KindFromSides(a, b, c float64) Kind {
 	}
 	// sides are now sorted
 	switch {
+	case math.IsInf(c, 0):
```
petertseng: someone might be confused about why you only need to check `c` and not `a`, `b`. comment?

petertseng:
Test version is added so this is ready to ship.

exercises/triangle/triangle.go
```diff
@@ -2,6 +2,8 @@
 
 package triangle
 
+const TestVersion = 1
```
petertseng: Notice that most other elements of this file (`KindFromSides`, `Kind`, `Equ`, etc...) have some sort of explanatory comment. I chose not to include one for `TestVersion` since students should have gotten the drill already from previous exercises. Let me know if you disagree.

kytrinyx: I agree with this.

kytrinyx:
Thank you!

PR #233, opened 2016-01-30
Triangle cleanup

petertseng:
I know examples don't need to be the best code, but

* this was bugging me while I was looking at the example code for #231 so I decided to fix it
* even if the code is unabashedly not the best, I say there's still no harm in increasing readability (which I argue this does)

### DIFF
--- a/exercises/triangle/example.go
+++ b/exercises/triangle/example.go
@@ -30,12 +30,9 @@ func KindFromSides(a, b, c float64) Kind {
 		return NaT
 	case a+b <= c: // triangle inequality
 		return NaT
-	case a == b:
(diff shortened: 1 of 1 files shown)


petertseng:
Strictly speaking this needn't conflict with #232 but I just tried to merge the two locally and git plays it safe and declares there is a merge conflict. So if you care about merging both in quick succession you can deal with the conflict in the obvious way. I can deal with the rebase if it's OK to merge them a few hours apart, of course (and I argue that it's totally OK to merge this one later)

kytrinyx:
This is all good! Thanks for fixing :)

Issue #143, opened 2015-03-14
Paasio issues


pminten:
So I've been looking at the paasio test and it strikes me as an odd one. Take for example this bit from the example, one of the core functions:

```go
func (wc *writeCounter) Write(p []byte) (int, error) {
	m, err := wc.w.Write(p)
	wc.wrl.Lock()
	wc.nwr += int64(m)
	wc.nwrops++
	wc.wrl.Unlock()
	return m, err
}
```

The test code calls this with a bytes.Buffer to write to (i.e. `wc.w` is a bytes.Buffer). Writing to that is not thread safe, which strongly suggests that io.Writer is not thread safe (otherwise bytes.Buffer would be buggy). But if `wc.w.Write` is not thread-safe then the whole `Write` function is not thread-safe and the use of a mutex in it is useless. Furthermore if this function is going to have race problems they're most likely going to be in the buffer write since incrementing an integer is so quick and costs so little time relatively to the buffer write that you need _really_ bad luck to hit it.

Putting the `wc.w.Write` within the critical section is possible but as I understand the point of the created code is to provide io.Reader and io.Writer wrappers that themselve satisfy io.Reader and io.Writer respectively. In that respect it's a bit odd that the Read and Write methods would be thread-safe if the interface doesn't require it.

Another issue I see with this code is that it's about mutexes and those are somewhat discouraged in Go. They're around for when you need them but you shouldn't use them unless you have a good reason for why channels won't do the job. So is this something we want to teach to newbies?

So I would like to ask, what do you think about this exercise?

petertseng:
> the whole `Write` function is not thread-safe

Well, this problem was long before my time, but if I look at it, I say yes it's true that `Write` is not thread safe... which actually is written in the documentation for it:

https://github.com/exercism/xgo/blob/d895e12bc457dad8c024ad457380ade2a5f55aa9/exercises/paasio/example.go#L11
https://github.com/exercism/xgo/blob/d895e12bc457dad8c024ad457380ade2a5f55aa9/exercises/paasio/example.go#L20
https://github.com/exercism/xgo/blob/d895e12bc457dad8c024ad457380ade2a5f55aa9/exercises/paasio/example.go#L30

> Furthermore if this function is going to have race problems they're most likely going to be in the buffer write since incrementing an integer is so quick and costs so little time relatively to the buffer write that you need *really* bad luck to hit it.

[...]

PR #234, opened 2016-01-30
paasio: Document required concurrency guarantees

petertseng:
As discussed in #143, since the tests are going so far as to ensure that
the counters are still accurate in the face of concurrent operations, it
is wise to document these guarantees as well. This also clarifies that
there need be no guarantees on interleaving of Read/Write calls, another
discussion point in #143.


kytrinyx:
Yes, getting this documented is really nice!

PR #235, opened 2016-01-30
paasio: Actually test ReadWriteCounter

petertseng:
Because it doesn't look good to have commented-out code in examples, nor required interfaces that aren't even getting tested.

### DIFF
--- a/exercises/paasio/example.go
+++ b/exercises/paasio/example.go
@@ -5,7 +5,7 @@ import (
 	"sync"
 )
 
-const TestVersion = 1
+const TestVersion = 2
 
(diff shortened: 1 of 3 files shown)


kytrinyx:
This looks good to me. I'd love to have a second opinion before I merge, though, since I'm not familiar with the paasio exercise.

petertseng:
Is it just you and me these days? Obviously I wouldn't want to be my own second opinion, so I wonder if one is forthcoming

kytrinyx:
Yeah, I think it's just you and me for now. Merging.

PR #236, opened 2016-01-30
paasio: Use channels instead of mutexes

petertseng:
The discussion in #143 suggests that we might like to avoid mutexes
unless we have a good reason for why channels can't do the job. This
commit implements a channel-based approach to counters. A single
goroutine is responsible for updating the bytes/ops counters. All other
goroutines send requests to it on the write channel.

Since an unprotected read of the counter variables would be a data race
(which is rightfully caught by Go's race detector!), reads must be
protected as well. This is done by sending channels into a channel, at
which point the values are sent on the channel.

### DIFF
--- a/exercises/paasio/example.go
+++ b/exercises/paasio/example.go
@@ -2,7 +2,6 @@ package paasio
 
 import (
 	"io"
-	"sync"
 )
 
 const TestVersion = 1
@@ -11,8 +10,8 @@ const TestVersion = 1
 // w.Write() are not guaranteed to be synchronized.
 func NewWriteCounter(w io.Writer) WriteCounter {
 	return &writeCounter{
-		w:   w,
-		wrl: new(sync.Mutex),
+		w:       w,
+		counter: newCounter(),
 	}
 }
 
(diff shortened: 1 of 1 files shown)


petertseng:
Really though, https://github.com/golang/go/wiki/MutexOrChannel suggests that it's perfectly appropriate to use mutexes for protecting state, and number of bytes/ops performed most definitely falls within the realm of state. Even [Effective Go's Concurrency section](https://golang.org/doc/effective_go.html#sharing) states:

> This approach can be taken too far. Reference counts may be best done by putting a mutex around an integer variable, for instance. But as a high-level approach, using channels to control access makes it easier to write clear, correct programs. 

We have (at least) parallel letter frequency to show use of channels (any others?), it doesn't seem so unreasonable to me that we would have at least one exercise where a mutex is appropriate. Good to teach people to use the right tools available to do the job, right?

I slightly vote against merging this PR because I think channels are the wrong way to go about doing this, and since it's my own PR I'm going to close it to prevent any accidents. But if other maintainers want to convince me I wouldn't feel so unhappy about this code being merged.

petertseng:
However note that this PR does demonstrate an interesting point: The Counter portion of ReadCounter and WriteCounter have a lot in common. We could extract the common elements.

petertseng:
reopening to amend commit message, then closing

PR #237, opened 2016-01-30
paasio: Extract common counter code

petertseng:
The "count bytes/operations" portions of the ReadCounter and
WriteCounter were completely analogous, and can be extracted.

### DIFF
--- a/exercises/paasio/example.go
+++ b/exercises/paasio/example.go
@@ -11,8 +11,8 @@ const TestVersion = 1
 // w.Write() are not guaranteed to be synchronized.
 func NewWriteCounter(w io.Writer) WriteCounter {
 	return &writeCounter{
-		w:   w,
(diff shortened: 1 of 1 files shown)


petertseng:
Again, the [contribution guidelines](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md#reference-solution) say that reference solutions don't need to be particularly great code. I balance this against the current situation that they are named "example" which means they're meant to serve as an example... so we should strive to set somewhat good examples, and that means we should at least try to keep the code DRY.

petertseng:
Extracted from #236

petertseng:
I am quite sure that again, git will play it safe and say this will have a merge conflict with #235 . Should be easy to understand since they don't actually modify any same lines (just lines that are very close together), but of course I'm totally ready to handle rebases.

PR #239, opened 2016-02-05
clock: make benchmark implementation non-specific

### LINKED ISSUE #238: clock: benchmark assumes specific implementation
Right now the benchmark in clock assumes that you've implemented the Clock type as an int, but a lot of people start off with a struct with `hour` and `minute`. I think it would be worth changing the benchmark to not assume a specific implementation (perhaps use `Time` instead of `Clock(i)`).

kytrinyx:
The benchmark was forcing the clock type to be an alias to int.
It's more interesting if we leave the implementation open.

This tweaks the benchmark so that other implementations of clock
aren't a compiler error.

Fixes #238

### DIFF
--- a/exercises/clock/clock_test.go
+++ b/exercises/clock/clock_test.go
@@ -65,7 +65,8 @@ func TestCompareClocks(t *testing.T) {
 }
 
 func BenchmarkAddMinutes(b *testing.B) {
-	c := Clock(720)
+	c := Time(12, 0)
+	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
 		for _, a := range addTests {
 			c.Add(a.a)


petertseng:
I think I'm OK with this, but one thing I want to make sure on before merge: Looking at `Time(hour, minute int)`, is anyone assuming that `hour` is in range `[0, 23]` and minutes in range `[0, 59]`? This could put `Time(0, 720)` at risk.

kytrinyx:
Good catch, let me verify that.

petertseng:
On the other hand, `Time(12, 0)` should work for everyone, right?

kytrinyx:
Hah. Yeah, that's true. Sorry--I've been traveling. I'll take care of this today.

petertseng:
Seems good

Issue #126, opened 2015-01-30
Clock: New must return a pointer


TatriX:
The "Clock" exercise wants ```New()``` to return non pointer value and it's wrong in my opinion.
```New()``` must return a pointer or a value which represents pointer like reflect.Value
So if:
```go
c1 := New()
c2 := New()
```
```c1``` cannot be equal to ```c2``` because it's different pointers.

soniakeys:
Working as intended.  See issue https://github.com/exercism/xgo/issues/123.

In PR https://github.com/exercism/xgo/pull/125 I explained that I didn't think TestVersion needed to be bumped, but maybe it should be bumped after all, so we could know if the solver has the new error message and comments available.

As to whether a function named New should always return a pointer, I just posted to the go-nuts mailing list asking for opinion.  It should be interesting.  If the verdict is that a non-pointer New is bad style, I propose renaming our function something like Time or HourMinute.  I'll still be opposed to changing it to return a pointer.

soniakeys:
Huh.  The response was a little disappointing.  Sometimes on issues of style like this people more eagerly voice their opinions.  Out of 95 views, only four people felt compelled to post.  A couple of the responses were a bit tangential, but none said that New should return a pointer, none said that a constructor that returns a non-pointer should not be named New.

I'm still most inclined to leave this as "working as intended."

TatriX:
greping through the standard library shoes that New*() funcs return
1) *struct
2) interface
3) *struct, error
4) interface, error
On the other side structs having non-pointer receivers do not have a New() func and rather have alternative constructors like time.Unix()

kytrinyx:
Even though this is not clear-cut, I find it interesting that we can't find an example in the standard library where a `New()` function returns a non-pointer (aside from interface, which is different).

I'm inclined to follow that unspoken pattern.

soniakeys:
I'm not sure I understand.  What unspoken pattern?

kytrinyx:
The idea that `New()` would probably return a pointer, and to call a constructor that returns a value something else.

I don't feel strongly about it, but I lean slightly towards that side of the argument.

soniakeys:
Okay.  PR soon.

kytrinyx:
Fixed in #127

PR #240, opened 2016-02-05
Clock constructor

kytrinyx:
Note, merge #239 first, this builds on top of that change.

### DIFF
--- a/exercises/clock/clock_test.go
+++ b/exercises/clock/clock_test.go
@@ -7,7 +7,7 @@ import (
 
 // Clock type API:
 //
-// Time(hour, minute int) Clock    // a "constructor"
+// New(hour, minute int) Clock     // a "constructor"
 // (Clock) String() string         // a "stringer"
 // (Clock) Add(minutes int) Clock
 //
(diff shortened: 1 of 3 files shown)


petertseng:
Seems good to remove the duplicate code between `Add` and `New` as you did here. If I were being really nitpicky I would suggest that go in a separate commit. I'm OK with this whenever #239 gets merged; I had something I wnted to make sure of on that one first.

kytrinyx:
I've reworked this a bit. You're right about putting the example.go refactoring in a separate commit.

exercises/clock/clock_test.go
```diff
@@ -7,7 +7,7 @@ import (
 
 // Clock type API:
 //
-// Time(hour, minute int) Clock    // a "constructor"
+// New(hour, minute int) Clock    // a "constructor"
```
petertseng: I know this is super nit, but notice how this `// a "constructor"` was aligned with the `// a "stringer"` comment below, but now it's misaligned by one space because of `Time` -> `New`. Perhaps add a space back?

kytrinyx: Yepp, that's a good idea.

petertseng:
Feel free to merge whenever you fix that small thing.

kytrinyx:
Updated

petertseng:
Note for history: This PR reverses the change made in #126 / #127.

PR #241, opened 2016-02-05
triangle: reorder constants in stub file

kytrinyx:
If someone chooses to use iota for the triangle kinds, then it makes sense to have
NaT represented as 0.

### DIFF
--- a/exercises/triangle/triangle.go
+++ b/exercises/triangle/triangle.go
@@ -11,9 +11,9 @@ func KindFromSides(a, b, c float64) Kind
 type Kind
 
 // Pick values for the following identifiers used by the test program.
+NaT // not a triangle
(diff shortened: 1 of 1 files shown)


petertseng:
Well, this doesn't necessarily mean *we* have to reorder it -- the student is free to do so! But I'm OK with lowering the barrier to such solutions, if this is the effect you think this will have. So :+1:

kytrinyx:
Agreed, we don't have to. I just kind of liked the idea of putting it first. I'm not very tied to either ordering, to be honest.

petertseng:
Yeah me neither, so I'm happy to go either way. I'll leave it to you if you want to merge it.

kytrinyx:
Cool, thanks.

PR #246, opened 2016-02-23
Clarified go test instructions with a period

mrice:
I'll submit an issue too, but this one little thing was a speed bump for me when I was first getting started with the Go exercises. I think it cost me almost an (very distracted by kids) hour and I almost gave up on the exercises. I could be doing it wrong or overlooked something about exercism but Go's GOPATH is confusing enough for a complete newbie (like me) that this simple period was a big help for me.


kytrinyx:
Per the discussion in #247 -- I need to understand the problem before this can be merged.

mrice:
Nevermind. My bad!

Issue #248, opened 2016-02-28
Why is a build constraint in the exercise template?


toolness:
Hi!

I don't know go, but I just did the first Hello World exercise. And I noticed there were some comments in there that said:

```go
// This next comment is a "build constraint."  It's here for, well, kind of
// complicated testing purposes you don't need to worry about now.
// Actually, you can delete it.

// +build !example
```

This made me chuckle. Is there a point to having it in there, if it doesn't explain what it is and then says I can actually delete it? Or is it in there specifically to make readers laugh? :grin:

petertseng:
Good question. My explanation (which may need corrections because these build constraints were added before my time) follows.

In this repo, we have example solutions to each of the exercises. To make sure that the examples are correct, we have [automated tests](https://travis-ci.org/exercism/xgo) that test each of the examples. But when attempting to compile the tests, if we attempted to compile both the example files *and* the stub files, we would get compilation errors for conflicting definitions.

The [Go docs for build constraints](https://golang.org/pkg/go/build/) tell us that we can use build constraints to control what files get built. And so, we put build constraints on our stub files, then [we instruct our automated tests to exclude the stub files](https://github.com/exercism/xgo/blob/master/.travis.yml#L48).

It's good that you ask this question because I imagine many others had this question before but didn't ask. It makes me wonder, can we come up with a solution that doesn't involve the build constraints? 

What if, for example, we just had the automated tests delete the stub files. If we had the convention that for a directory `exercise-slug` the stub file is named `exercise_slug.go` (true for every one so far except for `hello`) we can delete them with something similar to `for dir in *; do stub=$(echo $dir | tr - _); ls $dir/$stub.go; done`. The benefit is that it would be less confusing. The disdavantage is that we would be bound to that constraint for all stub files we ever write. @kytrinyx , what would you think of this?

kytrinyx:
I commented in the PR, but the summary is this:

Yeah, it would be nice to get rid of these, but I would also like to be able to run the full test suite locally without having to delete stub files, since I can absolutely imagine a future where I would accidentally commit back the deleted file.

PR #249, opened 2016-02-29
No build constraints

petertseng:
Instead of build constraints (which may confuse students), this PR proposes that we have Travis remove stub files (which is necessary so as not to have conflicting definitions) before running tests. Further discussion is in individual commit messages and #248.


petertseng:
In my view, the primary question to be answered before deciding whether to merge this is: Does the benefit (less confusing to students) outweigh the disadvantage (track maintainers need to adhere to a naming convention for stub files)?

I say yes because the convention has been documented in the readme, we only have a limited number of stub files, and the fact that (almost) everything already adhered to the convention kind of means that it was already established anyway. But I may be biased since I thought this up, so that's why a second opinion would help.

petertseng:
Correction, both `hello.go -> hello_world.go` and `tree.go -> tree_building.go` renames are needed (I only saw the first). Will fix that.

bin/delete-stub-files
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+for dir in exercises/*; do
+  stub=$(echo $(basename $dir) | tr - _)
+  file="$dir/$stub.go"
+  if [ -f $file ]; then
+    echo "deleting stub $file"
```
petertseng: verbosity so that if ever a file is removed that shouldn't be, or a file isn't removed that should be, looking at travis logs can help diagnose. I think it's better to have this than not, but would like to point it out in case someone disagree

kytrinyx: Agreed. We need to be able to diagnose logs.

kytrinyx:
Yeah, I think that anything to remove noise for people going through the track is a good thing.

The one thing that I wonder about is how to run the full test suite locally without accidentally committing the deleted stub files.

petertseng:
That's a good question. Let's think about it for a few hours and see what comes up.

One suggestion that comes to mind is: Maybe there could be a script that temporarily moves the stub files (so they don't get built when testing), runs the tests, then puts the stub files back. For example it could change their extension to `.notgo` or something so that `go test` doesn't build them.

Seems a bit odd, so let's see if we can't think of something better.

kytrinyx:
Yeah, the `.notgo` thing might work. I've got a bunch of things going on at once today, so I'm not thinking very straight, but if I do think of something I'll comment here.

petertseng:
OK, given it's the current prevailing suggestion, I had `bin/test-without-stubs` script do the renames. This is what Travis uses and it is what will work locally as well.

kytrinyx:
Yeah, it's a bit hacky, but I can't think of anything better, and this is a definite improvement for the user experience. Thanks!

PR #253, opened 2016-03-06
Poker tests: label hands correctly, deduplicate cards, ensure straight flush > 4OAK

petertseng:
Rollup fix of three issues with the tests.

Not included here is a decision on whether hands with duplicate cards should be an error and/or whether we should even test for that at all.


kytrinyx:
Let's not test for hands with duplicate cards. It seems like it would only happen in a completely degenerate situation.

kytrinyx:
Thank you!

PR #254, opened 2016-03-06
README: use bin/test-without-stubs in test instructions

petertseng:
Since #249 was merged, stub files no longer have build constraints. This
means an attempt to run `go test` without removing the build constraints
will fail. So instead, we should direct contributors to run the script
that moves the stub files aside.


kytrinyx:
Yepp, this is much better.

PR #257, opened 2016-03-13
sum-of-multiples: add clarifying comment

### LINKED ISSUE #256: Sum Of Multiples incorrect test case
In the `varTests` test cases for the `Sum Of Multiples` exercise it has:
`{[]int{}, 10000, 0}`

But the accompanying README.md file states: 

> If no set of numbers is given, default to 3 and 5.

Should this test case actually be the following or have I misread the description?
`{[]int{}, 10000, 23331668}`

kytrinyx:
Clarify the discrepancy between the Go implementation and the current README.
This fixes #256.


petertseng:
Yeah looks good, so I'm :+1: on it. Will probably leave you to hit the merge button since I'll occupied for a bit.

kytrinyx:
I appreciate having a second pair of eyes on this. Thanks!

PR #259, opened 2016-03-14
Add additional equality test to clock exercise

notgary:
The existing test suite was not rigorous enough in testing the equality
of two clock values. Specifically, it was unable to compare two clocks
where the times were offset by 24 hours. For example,

  clock{0, 1}
  clock{0, 1441}

Since there are 1440 minutes in a 24 hour period, the time on the second
clock should be wrapped around, making it equal to the first.

This new test will catch this situation.

### DIFF
--- a/exercises/clock/cases_test.go
+++ b/exercises/clock/cases_test.go
@@ -75,4 +75,10 @@ var eqTests = []struct {
 		hm{34, 37},
 		true,
 	},
+	// clocks with minute overflow
+	{
+		hm{0, 1},
+		hm{0, 1441},
+		true,
+	},
 }


kytrinyx:
Thanks, @notgary!

petertseng:
Thanks! This is a good test. It would be best if it were also added to the common tests in https://github.com/exercism/x-common/blob/master/clock.json . This will allow other tracks to benefit from this change as well. Otherwise, I fear there is a risk that this change may be automatically overwritten next time we regenerate tests (tests are automatically generated using that file)

kytrinyx:
Ah, I forgot that this is a generated test. Thanks for catching that, @petertseng.

PR #264, opened 2016-03-24
README: No need to clone to GOPATH

### LINKED ISSUE #263: README/development setup: Probably don't need to clone to GOPATH
The README says to clone the repo to my GOPATH, but this does not appear to be necessary... especially since I have the repo outside my GOPATH and I don't have any problems.

Should have a quick check as to whether anything actually needs it, and if not we can remove that from the instructions and just say clone it anywhere you like.

petertseng:
Works for both kytrinyx and me; we both clone outside of our GOPATH, so
can remove that from the instructions.

Closes #263

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,11 +19,12 @@ installed, ideally either the current release, the previous release, or tip.
 
 You will need a github account and you will need to fork exercism/xgo to your account.
 See [GitHub Help](https://help.github.com/articles/fork-a-repo/) if you are unfamiliar with the process.
-Clone your fork into a Go workspace on your GOPATH.  Read [How to Write Go Code](http://golang.org/doc/code.html)
-if you are unfamiliar with Go workspaces and GOPATH.  As an example, if you have `$HOME/dev` on your GOPATH,
-you should cd to `$HOME/dev/src/github.com/<you>` to type the clone command:
-`git clone https://github.com/<you>/xgo`.  Test your clone by cding to the xgo directory and typing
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,12 +19,12 @@ installed, ideally either the current release, the previous release, or tip.
...
 See [GitHub Help](https://help.github.com/articles/fork-a-repo/) if you are unfamiliar with the process.
-Clone your fork into a Go workspace on your GOPATH.  Read [How to Write Go Code](http://golang.org/doc/code.html)
-if you are unfamiliar with Go workspaces and GOPATH.  As an example, if you have `$HOME/dev` on your GOPATH,
-you should cd to `$HOME/dev/src/github.com/<you>` to type the clone command:
-`git clone https://github.com/<you>/xgo`.  Test your clone by cding to the xgo directory and typing
+Clone your fork with the command: `git clone https://github.com/<you>/xgo`.
+Test your clone by cding to the xgo directory and typing
 `bin/test-without-stubs`. You should see tests pass for all exercises.
 
+Note that unlike most other Go code, it is not necessary to clone this to your GOPATH.
```
petertseng: Can keep the http://golang.org/doc/code.html link but I didn't find it necessary.

Can possibly explain why it's not necessary to clone to GOPATH though?

kytrinyx: We don't use any imports outside of the standard library.

petertseng: And I imagine another reason is nothing needs to import this repo. I added a sentence saying both these things.

kytrinyx: Yepp, good point.

PR #266, opened 2016-03-29
diffie-hellman: remove extra commas in expected signatures

### LINKED ISSUE #265: diffie-hellman: extra commas in signatures in test file
```go
// Step 2:   PublicKey(private, p, *big.Int, g int64) *big.Int
// Step 2.1: NewPair(p, *big.Int, g int64) (private, public *big.Int)
```

the commas between the `p and *big.Int` shouldn't be there

petertseng:
We wouldn't want students to copy those directly and be confused when
they don't work.

Closes #265


kytrinyx:
Thanks!

PR #268, opened 2016-04-04
clock: regenerate test cases

kytrinyx:
The canonical inputs/outputs in clock.json were expanded
recently to handle multiple rollovers in either direction, and
to test more edge cases in the equality tests.

This also bumps the test version.


petertseng:
Yup, seems good.

PR #269, opened 2016-04-06
Move bob to an earlier position in the track.

### LINKED ISSUE #138: Weird order of tests in config.json
config.json starts with leap instead of the traditional bob, but also has parallel-letter-frequency before bob. Bob is a trivial exercise while parallel-letter-frequency is somewhat more advanced (it requires using channels and goroutines).

Is there a particular logic to the order of exercises in this track?

kytrinyx:
It looks like bob, which is a pretty trivial exercise, kept getting
pushed farther and farther towards the end of the track.

This brings it back towards the beginning.

Fixes #138


petertseng:
Yeah seems reasonable, let's do it. I'll let you merge since maybe you can adjust if you decide to tweak the position more, and that doesn't introduce additional delay because even if I merged it would have still needed an x-api bump anyway.

petertseng:
Wonder if we can help detect this across various tracks, but it's difficult because a certain task may have varying difficulty in various languages. The best we may be able to do is probably some rough ranges of estimated difficulty and see whether any tracks have a sudden drop in difficulty at a certain problem.

Not something I'd prioritise immediately because of various questions on how we'd even do it, but a possibility for the future

kytrinyx:
Yeah, I think for now we should just keep our eye on it occasionally.

PR #273, opened 2016-04-07
scrabble-score: capitalize oxyphenbutazone

### LINKED ISSUE #270: scrabble-score: replace 'multibillionaire' with 'oxyphenbutazone'
The word _multibillionaire_ is too long for the scrabble board. _Oxyphenbutazone_, on the other hand, is legal.

Please verify that there is no test for _multibillionaire_ in the scrabble-score in this track. If the word is included in the test data, then it should be replaced with _oxyphenbutazone_. Remember to check the case (if the original is uppercase, then the replacement also should be).

If multibillionaire isn't used, then this issue can safely be closed.

See https://github.com/exercism/x-common/issues/86

petertseng:
A history:

* 1d941e524d25b40d3e8ab452dc1a945423da897b: MULTIBILLIONAIRE
* 9541d7b7b835fb98eef32c281a8931f34b73322b: oxyphenbutazone
  (because multibillionaire was too long!)
* But since MULTIBILLIONAIRE was our only word with capital letters, now
  we've lost the capability to test solutions for them! So let's bring
  it back by putting OXYPHENBUTAZONE in all caps!

Test version is bumped for this.

Closes #270

### DIFF
--- a/exercises/scrabble-score/example.go
+++ b/exercises/scrabble-score/example.go
@@ -5,7 +5,7 @@ import (
 )
 
 // testVersion tracks the version of the exercise.
-const testVersion = 3
+const testVersion = 4
 
 var letterValues = map[rune]int{
 	'a': 1, 'b': 3, 'c': 3, 'd': 2, 'e': 1,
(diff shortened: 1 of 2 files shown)


kytrinyx:
Thank you!

Issue #275, opened 2016-04-07
discussion: Should we check return values when there is an error?


petertseng:
While making #274 for binary I realized that I had accidentally checked the expected output value in the case where an error is expected (Preserved at https://github.com/petertseng/xgo/commit/6ac976c6b2ac73e6dd03ccece31541e96ff03595, as opposed to what's currently on #274). That means if there is an error, you always had to return `0, err` and not something else like `9001, err` since the test will still expect that you return 0 even if there is an error.

But should that matter and should it be checked? I think the usual contract is "if there is a non-nil error, assume nothing about the normal output value", including not assuming it's the zero value for the type. The student shouldn't necessarily have to return any particular value in an error case.

This is probably pretty low priority, since I assume any test case that checks the output value in an error case is checking that it's the zero value and that students will probably independently choose to return the zero value anyway. But it's worth a discussion.

If we agree we shouldn't check the other output in error cases, I'll look through our tests and see which tests should be changed.

kytrinyx:
This is a good question.

> I think the usual contract is "if there is a non-nil error, assume nothing about the normal output value"

Yes, I completely agree. I think that we should make sure that the second return value is an error type, though. I've seen people return something that is not an error where I would have expected it to be an error. And then, of course, check that it is not nil (but I don't think we should make any assumptions about the error beyond that.

petertseng:
Okay. Next thing to do is to write up the list of exercises in which to make this change. I'll likely close this issue and open separate issues for each exercise so each can be closed individually.

kytrinyx:
That sounds good.

petertseng:
Oh! One thing I thought about when deciding how to do something you mentioned:

> I think that we should make sure that the second return value is an error type, though.

Is the work done by the type system sufficient, or is there more to be done here? I believe that if the type is `error`, you wouldn't be able to return, say, a `string` or `int` in its place?

kytrinyx:
That depends--I've we've defined the function or the type somewhere, then yeah, the type system is enough. If the only thing we've specified is that we expect two return values, then we could easily end up with strings or ints there.

petertseng:
Ah you're right, since the tests just checks `== nil` or not! I didn't even think of that. OK, we could get the type system to help us in the tests by explicitly having a `var err error` and assigning the error return value to that - this will ensure that it is indeed an `error` type.

This might warrant an entire-track sweep. I wonder if we should add a mode of operation to blazon that posts one issue for each problem in a single track...

[...]

PR #277, opened 2016-04-09
hamming: Check that err is an error

petertseng:
Tested by first making the example `Distance` return an `*int` instead
of an `error` (without changing the test).

The test passed, indicating that the additional line in the test is
necessary.

Then, the line was added to the test, which caused a compilation error
as hoped. So this is correctly checking that the `err` return value is
indeed an `error`. This is the standard technique in Go for asserting
that a value implements an interface.

This arises out of discussion in #275. This will be an example that will
be used when providing guidance on how to enact the changes for other
exercises.

### DIFF
--- a/exercises/hamming/example.go
+++ b/exercises/hamming/example.go
@@ -2,7 +2,7 @@ package hamming
 
 import "errors"
 
-const testVersion = 3
+const testVersion = 4
 
 func Distance(a, b string) (d int, err error) {
 	if len(b) != len(a) {
--- a/exercises/hamming/hamming.go
+++ b/exercises/hamming/hamming.go
@@ -1,3 +1,3 @@
 package hamming
 
-const testVersion = 3
+const testVersion = 4
--- a/exercises/hamming/hamming_test.go
+++ b/exercises/hamming/hamming_test.go
@@ -2,7 +2,7 @@ package hamming
 
 import "testing"
 
-const targetTestVersion = 3
+const targetTestVersion = 4
 
 func TestHamming(t *testing.T) {
 	if testVersion != targetTestVersion {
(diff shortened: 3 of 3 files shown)


petertseng:
Despite this being a one-line change, now's a great time for us to ask ourselves, "so is this something we want to do in all exercises that have error handling?"

Argument against is that you normally would not see a line like this in a real test. Sure, I might do it for `var _ Database = &fakeDatabase{}` to make sure that my fake database type correctly implements my Database interface, but I wouldn't do such a thing for `error` because in fact usually I just check whether the error is nil or not (or if I do want some other property, the compiler is checking it elsewhere).

Now, this is different because we're asking students to write functions and they're not necessarily familiar with the idioms, so it could be good to show them what's expected.

I don't feel particularly strongly about whether it should be done, so let's discuss.

petertseng:
Forgot to bump test version. Bumped it.

kytrinyx:
I actually saw someone submit a solution that had a non-error as the second value, because they weren't familiar with the idioms.

I think that in this context, adding things to the test to help enforce the idiomatic behavior that we want and the interfaces that we want is a legitimate thing to do, even though it's not something you would usually put in a test.

kytrinyx:
Whether or not we want it in all of the tests, I do think that it makes sense in this test, which I think is the first time we have error handling in the Go track.

petertseng:
> the first time we have error handling in the Go track.

Hmm, I didn't even realize that when I picked it! I'm going to check whether it is the case.

... hmm, `queen-attack` also has error handling, as does `grains`.

Also, `food-chain` and `house` could arguably get error handling added, but that's another story.

Since `queen-attack` is actually the first exercise we have that has error handling, I'm going to add this there too (or we can discuss moving `hamming` earlier). Then we can think about whether it goes everywhere.

petertseng:
As for whether it should go in all tests

I'm still not sure. Given that I'm unsure, I'd err on the side of adding it, since it'll be easy to grep for and remove if we change our minds.

kytrinyx:
I agree. Let's add it to all of them.

PR #278, opened 2016-04-09
queen-attack: Ensure that err is an error

petertseng:
Same as in #277, but good because queen-attack comes before hamming. We
want this to happen ASAP for the first exercise that has error handling.


petertseng:
As an alternative though, I kind of want to propose moving hamming before queen-attack. Give me a bit to file an issue.

kytrinyx:
I agree, we should move hamming before queen-attack. It can go pretty early, I think.

PR #315, opened 2016-04-17
test-without-stubs: Signal failure if a test failed

petertseng:
Otherwise Travis will falsely report test success when instead the test
was failing.
~~We don't immediately exit though; we need to run the cleanup first. This
is noted in a code comment.~~

(Updated to use `trap` instead)

This script was introduced in fa897488609899b977adf95c693f3535c6315321
as part of #249 an initiative to not have build constraints which were
confusing to students.

Luckily we have not had any test failures since then, but it would be
good to restore the right behavior.

### DIFF
--- a/bin/test-without-stubs
+++ b/bin/test-without-stubs
@@ -19,12 +19,18 @@ for dir in exercises/*; do
   fi
 done
 
-go test -cpu 2 $RACE ./...
+cleanup () {
+  echo "moving stub files back"
 
-# could use `rename` here, but it's not uniform among platforms.
-# some renames take two arguments while others take a single regex.
-for file in exercises/*/*.notgo; do
-  dir=$(dirname $file)
-  base=$(basename $file .notgo)
-  mv "$dir/$base.notgo" "$dir/$base.go"
-done
+  # could use `rename` here, but it's not uniform among platforms.
+  # some renames take two arguments while others take a single regex.
+  for file in exercises/*/*.notgo; do
+    dir=$(dirname $file)
+    base=$(basename $file .notgo)
+    mv "$dir/$base.notgo" "$dir/$base.go"
+  done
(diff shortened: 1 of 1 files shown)


petertseng:
I noticed this when testing out various Go versions with Travis: See https://travis-ci.org/exercism/xgo/builds/123735373 which has passing builds even for outputs like 1.0.3's "can't load package: package github.com/exercism/xgo: no Go source files in /home/travis/gopath/src/github.com/exercism/xgo"

Compare against https://travis-ci.org/exercism/xgo/builds/123736271 which correctly fails like I expect

bin/test-without-stubs
```diff
@@ -20,6 +22,11 @@ for dir in exercises/*; do
 done
 
 go test -cpu 2 $RACE ./...
+if [ $? -ne 0 ]; then
+  # The test has failed, but don't immediately exit yet!
+  # We still need to run the cleanup code below!
```
petertseng: since we're in the go repo, this really makes you think "Wow, if only we had `defer` here! Then we could have just `defer`red the cleanup function at the beginning, exited immediately here, and the cleanup function would automatically run.

Turns out we can in fact do such a thing in a shell script with traps... so I may try that.

petertseng:
OK, updated to instead use `trap` for cleanup. This means that `go test` is now the last command in the file, which means that that will be the exit status of the script, which is what we want. Comment added so that hopefully it is less likely to be broken in the future.

petertseng:
Good, https://travis-ci.org/exercism/xgo/jobs/123742784 has both of my expected outcomes:

* "moving stub files back" is outputted
* build fails, so travis shouls it as red

kytrinyx:
Excellent catch! Thanks for digging into this.

PR #316, opened 2016-04-17
Travis versions

petertseng:
Individual commit messages have more detail. The short of is that we now test:

Required:

* 1.6.1
* 1.5.4

Can fail:

* 1.4.3
* tip


kytrinyx:
LGTM!

PR #321, opened 2016-04-25
config.json: deprecate counter

petertseng:
As of https://github.com/exercism/x-common/issues/80 we'd like to keep
the scope of Exercism focused, and counter is the single exercise that
falls under the write-the-test-suite category.

### DIFF
--- a/config.json
+++ b/config.json
@@ -42,7 +42,6 @@
     "circular-buffer",
     "robot-name",
     "react",
-    "counter",
     "custom-set",
     "atbash-cipher",
     "phone-number",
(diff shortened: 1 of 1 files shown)


petertseng:
Note for review: Prepared so it's ready to go at any time at the press of a button, but I expect to wait for discussion on the x-common issue first.

Then again, discussion may not be forthcoming since xgo is the only implementing track of the counter exercise (and I am relatively certain it is the only such exercise), but it is still possible others have ideas on the topic.

kytrinyx:
The discussion wasn't a resounding "yes" or "no", but I ended up concluding that we should reduce the scope of this. I will put a note in the README of the exercise in x-common.

Issue #279, opened 2016-04-09
Check ordering of exercises with stubs vs without stubs


petertseng:
Here are the exercises with stubs, with their order in config.json and the number of lines in the stub:

```
 1      hello-world  23
 2             leap  12
 3            clock  26
 4       gigasecond  10
 5        raindrops   7
 6              bob   3
 7         triangle  17
10       food-chain   3
19          hamming   3
23    crypto-square   3
30       word-count   9
36  circular-buffer   3
40       custom-set   3
61    tree-building 109
64           paasio   3
68       tournament   3
71           ledger 227
```

A few points of interest to me:

* difference-of-squares, secret-handshake, do not have despite coming before food-chain which does
* a big gap between hamming and food-chain: should hamming be moved earlier, or should its stub be deleted?
* it's understandable that tree-building and ledger have stub files since they are refactoring-focused exercises...
* but what about all the others that come after (my arbitrary cutoff of) hamming? And the three-line ones have nothing more than a `testVersion` and `package` name, so don't seem terribly necessary.

If the idea behind the stub files is to give people new to Go a starting point, it makes the most sense that the first N exercises would have stub files and then we can stop. So I wonder if we should close the gaps in the early exercises.

Footnote, script run to produce these results:

```bash
exercises="
    hello-world
    leap
    clock
    gigasecond
    raindrops
    bob
    triangle
    difference-of-squares
    secret-handshake
    food-chain
    house
    pascals-triangle
    series
    queen-attack
    grains
    etl
    scrabble-score
    parallel-letter-frequency
    hamming
    sum-of-multiples
    pythagorean-triplet
    bank-account
    crypto-square
    luhn
    largest-series-product
    sieve
    palindrome-products
    bracket-push
    anagram
    word-count
    binary
    allergies
    rna-transcription
    roman-numerals
    say
    circular-buffer
    robot-name
    react
    counter
    custom-set
    atbash-cipher
    phone-number
    strain
    pig-latin
    prime-factors
    nth-prime
    diffie-hellman
    octal
    beer-song
    ocr-numbers
    wordy
    nucleotide-count
    grade-school
    matrix
    saddle-points
    trinary
    hexadecimal
    meetup
    binary-search
    binary-search-tree
    tree-building
    kindergarten-garden
    simple-cipher
    paasio
    pov
    minesweeper
    robot-simulator
    tournament
    word-search
    connect
    ledger
    poker
"

i=0

for slug in $exercises; do
  i=$(expr $i + 1)
[...]

PR #322, opened 2016-04-28
new exercise to encode/decode variable length quantity

mattetti:
I'm not sure if that shouldn't move this exercise as a problem. I feel that the problem should be more defined so people understand the goal of the exercise.

### DIFF
--- a/exercises/variable-length-quantity/example.go
+++ b/exercises/variable-length-quantity/example.go
@@ -1,0 +1,15 @@
+package variablelengthquantity
+
+// EncodeVarint returns the varint encoding of x.
+func EncodeVarint(x uint32) []byte {
+	// TODO: implement me
+	return nil
+}
+
+// DecodeVarint reads a varint-encoded integer from the slice.
+// It returns the integer and the number of bytes consumed, or
+// zero if there is not enough.
+func DecodeVarint(buf []byte) (x uint32, n int) {
+	// TODO: implement me
+	return 0, 0
+}
--- a/exercises/variable-length-quantity/example.go
+++ b/exercises/variable-length-quantity/example.go
@@ -2,14 +2,55 @@ package variablelengthquantity
 
 // EncodeVarint returns the varint encoding of x.
 func EncodeVarint(x uint32) []byte {
-	// TODO: implement me
-	return nil
+	if x>>7 == 0 {
+		return []byte{
+			byte(x),
+		}
+	}
+
+	if x>>14 == 0 {
+		return []byte{
+			byte(0x80 | x>>7),
+			byte(127 & x),
+		}
+	}
+
+	if x>>21 == 0 {
+		return []byte{
+			byte(0x80 | x>>14),
+			byte(0x80 | x>>7),
+			byte(127 & x),
+		}
+	}
+
+	return []byte{
+		byte(0x80 | x>>21),
+		byte(0x80 | x>>14),
+		byte(0x80 | x>>7),
+		byte(127 & x),
+	}
 }
 
 // DecodeVarint reads a varint-encoded integer from the slice.
 // It returns the integer and the number of bytes consumed, or
 // zero if there is not enough.
 func DecodeVarint(buf []byte) (x uint32, n int) {
-	// TODO: implement me
-	return 0, 0
+	if len(buf) < 1 {
+		return 0, 0
+	}
+
+	if buf[0] <= 0x80 {
+		return uint32(buf[0]), 1
+	}
+
+	var b byte
+	for n, b = range buf {
+		x = x << 7
+		x |= uint32(b) & 0x7F
+		if (b & 0x80) == 0 {
+			return x, n
+		}
+	}
+
+	return x, n
 }
--- a/exercises/variable-length-quantity/variable_length_quantity_test.go
+++ b/exercises/variable-length-quantity/variable_length_quantity_test.go
@@ -1,0 +1,52 @@
+package variablelengthquantity
+
+import (
+	"bytes"
+	"testing"
+)
+
+/*
+The goal of this exercise is to implement
+VLQ encoding/decoding as described here: https://en.wikipedia.org/wiki/Variable-length_quantity
+
+In short, the goal of this encoding is to save encode integer values in a way that would save bytes.
+Only the first 7 bits of each byte is significant (right-justified; sort of like an ASCII byte).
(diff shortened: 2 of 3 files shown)


mattetti:
btw, I'm confused the tests are failing, but they re supposed to :)

IanWhitney:
Your example file needs to provide a working solution. Travis uses it to make sure your tests are valid. Students won't see your example, it's only used as part of the CI process.

exercises/variable-length-quantity/example.go
```diff
@@ -0,0 +1,15 @@
+package variablelengthquantity
```
kytrinyx: The example.go should be a reference solution—something that passes the given test suite. It doesn't get delivered when people `fetch` the exercise, it's just used to make sure that the test suite makes sense.

kytrinyx: Oh. Hah. I see that Ian said the same thing below.

mattetti:
Oops sorry, I missed that when I read the guidelines. Just for my own sake, will the users be provided with an empty test file?

PR update coming

petertseng:
> will the users be provided with an empty implementation file?

They will not, unless such a file is explicitly provided. The rule is any file with `example` in its name doesn't get shown to students, and all other files do.

For example [leap/leap.go](https://github.com/exercism/xgo/blob/master/exercises/leap/leap.go) is given to students, and [leap/example.go](https://github.com/exercism/xgo/blob/master/exercises/leap/example.go) isn't.

As for whether to provide a stub file for this exercise, we discussed in #279 and can revisit, but the general feeling back then was that by the time of reaching the later exercises it isn't really needed anymore.

mattetti:
PR updated, tests are now passing. I listed this exercise as the last one because I wasn't sure where it would fit, fell free to move it around. 

I respect your decision in regards to #279 even tho I disagree. At the end of the day it's matter of perspective, type of exercises and what part of the code you'd want students to focus on. It's all good with me.

exercises/variable-length-quantity/variable_length_quantity_test.go
```diff
@@ -0,0 +1,36 @@
...
+		7: {[]byte{0xC0, 0x80, 0x80, 0x00}, 134217728},
+		8: {[]byte{0xFF, 0xFF, 0xFF, 0x7F}, 268435455},
+
+		9:  {[]byte{0x82, 0x00}, 256},
+		10: {[]byte{0x81, 0x10}, 144},
+	}
+
+	for i, tc := range testCases {
+		t.Logf("test case %d - %#v\n", i, tc.input)
+		if o, _ := DecodeVarint(tc.input); o != tc.output {
```
petertseng: Hmm, given that the second return value is being discarded, it could be anything.

Given that in the example implementation it's the number of bytes consumed, now I want to ponder a few things.

These tests expect a `DecodeVarint` that decodes a *single* number it finds in a buffer.

Should we specify (and therefore test) `DecodeVarint`'s behavior:

* If there are multiple numbers in the buffer? `[]byte{0x00, 0x00}`
* If there is not a complete number in the buffer? `[]byte{0x80, 0x80}`
* If the buffer is empty?

Also, as an extension of this funciton that decodes a single varint, should we ask for an implementation of a function that decodes as many numbers as possible from a buffer as well? It would expand the scope of the exercise, though!

kytrinyx:
I just realized that we never got this merged. @mattetti would you have a moment to look it over and decide if you're happy with it? @petertseng had a few thoughts, but they were more in the "let's ponder this interesting question" than "we need to change this". We could ship it and then see how the exercise does in the wild.

What do you think?

mattetti:
It's been a while indeed! I just went through the exercise and re-read @petertseng's comments. I agree that his suggestions would be nice to add but aren't blockers.

@kytrinyx I think it's ready to be merged with one caveat, I'm not sure that the goal of the exercise might not be clear without more context. However since I'm not a exercism user I am not sure what the expectations are or even where to better describe the goal of the exercise and give more context.

petertseng:
I've reread my comments too.

I do believe that "no we don't need to, so no changes are necessary" are valid answers to my questions of the form "should we specify these behaviours? / add this other function?"

One thing I would like to make sure on - given that the second return value of `DecodeVarint` is never checked by the tests - to me this says it should just be removed. Otherwise students have to stick a dummy value in there, to no benefit to them. (If wanting to keep it, the test should check it)

> I'm not sure that the goal of the exercise might not be clear without more context.

We have this in https://github.com/exercism/x-common/blob/master/exercises/variable-length-quantity/description.md - I know in https://github.com/exercism/x-common/pull/239 there was uncertainty too on whether it was clear. To me, on reading the current description I don't see any specific place where I would say it was unclear, so I think we are good on that front! (But if anyone disagree, that is the place to send the PRs to make it clearer!)

kytrinyx:
> One thing I would like to make sure on - given that the second return value of DecodeVarint is never checked by the tests - to me this says it should just be removed. Otherwise students have to stick a dummy value in there, to no benefit to them. (If wanting to keep it, the test should check it)

That's a good point. What do you think @mattetti?

mattetti:
The API is consistent with the std lib in this case. I would suggest to
update the tests to check the value instead of removing it.
Unfortunately I don't currently have time to do so myself at the moment.

kytrinyx:
> Unfortunately I don't currently have time to do so myself at the moment.

I hear you! :heart:

If someone has the time to make a PR to this PR that would be awesome! (I added the hacktoberfest label, maybe some friendly gophers can help here)

mattetti:
@Ngo-The-Trung updated the tests and I merged their PR against my branch but it doesn't seem to be reflected here. The new GH review does seem interesting but I'm on mobile and super confused. Sorry :(

kytrinyx:
@mattetti I think it must have been caching, because I see it here.

I'm merging this. Thanks @mattetti @Ngo-The-Trung :heart:

petertseng:
The PR https://github.com/mattetti/xgo/pull/1 caused the new commits to show up in your master branch. But this PR was from your `ma-variable-length-quantity` branch, not your master branch. So the commit didn't get merged to this repo... but that's OK, we can just merge it now.

petertseng:
Applied using https://github.com/mattetti/xgo/commit/4b6df38d1f44ad1e9501acae4a81ce16ef383953.patch + `git am`, resulting in https://github.com/exercism/xgo/commit/a5d91a064960cd2495d1ed47e39f07242c657923. Thanks to all involved.

PR #327, opened 2016-05-28
fix comment in bank-account exercise which suggests a wrong input format

alebaffa:
Hi, 
the comment of this exercise is wrong. It suggests to declare the function Deposit with a _uint64_ variable in input, but it should be _int64_ instead.
If the function Deposit is implemented using _uint64_ the execution of the test file (specifically at lines 131, 138, 264 where the function is called with a negative variable) causes

>  constant -3 overflows uint64

This is also documented in the [Golang specification](https://golang.org/ref/spec#Constant_expressions):

> uint(-1)     // -1 cannot be represented as a uint

This PR is just a simple fix to the comment :) .


kytrinyx:
Yeah, that should be `int64`. Thanks!

PR #331, opened 2016-06-18
Copy track icon into language track repository, Fixes #330

### LINKED ISSUE #330: Copy track icon into language track repository
Right now all of the icons used for the language tracks (which can be seen at http://exercism.io/languages) are stored in the exercism/exercism.io repository in `public/img/tracks/`. It would make a lot more sense to keep these images along with all of the other language-specific stuff in each individual language track repository.

There's a pull request that is adding support for serving up the track icon from the x-api, which deals with language-specific stuff.

In order to support this change, each track will need to

- create a directory named `img/` at the root of this repository if it doesn't already exist, then
- find the correct icon in https://github.com/exercism/exercism.io/tree/master/public/img/tracks, and
- stick it in the `img/` directory, and importantly
- **name the image `icon.png`**

In other words, at the end of it you should have the following file:

```
./img/icon.png
```

See https://github.com/exercism/exercism.io/issues/2925 for more details.

manusajith:
Fixes: #330


petertseng:
Seems good to me, thanks.

petertseng:
in case any wonders, to me (and to `diff`) the image seems identical to https://github.com/exercism/exercism.io/blob/master/public/img/tracks/go.png - hope I haven't just given anyone any strange ideas...

PR #335, opened 2016-07-27
Connect use json

jonatas:
I updated from static examples to use from connect.json on x-common project.

It was merged recently: exercism/x-common#297.


petertseng:
Everything here looks OK.

Probably would be good if commit messages all mention "connect" somewhere.

I added two cases to connect.json

I am wondering whether this exercise should get a testVersion now, since solutions made before this PR will not pass the test suite after this PR.

(testVersion like https://github.com/exercism/xgo/blob/master/exercises/leap/leap.go and https://github.com/exercism/xgo/blob/master/exercises/leap/leap_test.go )

exercises/connect/cases_test.go
```diff
@@ -0,0 +1,96 @@
+package connect
+
+// Source: exercism/x-common
+// Commit: 6c6a395 improving description again
+
+const testVersion = 2
```
petertseng: we would want the `testVersion` to go in `example.go` so that submitted solutions have to show their `testVersion` (allowing reviewers of the code to easily see what version of the tests a solution was coded against). Putting it in `cases_test.go` would mean that nobody has to declare it

petertseng:
I'm fine with everything I see here, and I asume if anyone else had anything to say I expect they should have said something earlier since this PR has been here for a few days.

Some final things to check on, and then I'll merge. This will happen within 24 hours, and you can and should feel justified in prodding me if I haven't done that.

Note: I plan to squash all commits into one before I merge, since this only really makes sense as one commit.

PR #336, opened 2016-08-01
travis.yml: 1.6.2 -> 1.6.3

petertseng:
1.6.3 was released on 2016/07/17, so let's update.


petertseng:
1.7.0 is out too. I'll combine

PR #337, opened 2016-08-01
improve accumulate problem's error messages

### LINKED ISSUE #317: accumulate: test failure message prints unhelpful function pointer, go vet complains
I know there was a past work on this in https://github.com/exercism/exercism.io/pull/1260 , but now `go vet` has something to say again:

```
accumulate_test.go:33: arg test.converter in printf call is a function value, not a function call
```

Here is an example of the error message you will see if your implementation fails a test:

```
--- FAIL: TestAccumulate (0.00s)
	accumulate_test.go:35: PASS: echo []
	accumulate_test.go:35: PASS: echo [echo echo echo echo]
	accumulate_test.go:33: Accumulate([first letter only], (func(string) string)(0x47de00)): expected [First Letter Only], actual [first letter only]
```

That `(func(string) string)(0x47de00)` is decidedly unhelpful. I can think of two possibilities to replace:

1. The name of the function - however that appears to need reflection, unless I am mistaken.
2. The description (just like the PASS message does)

bordeltabernacle:
I'm not sure if this is what you were looking for @kytrinyx @petertseng (re. #317) so do let me know if I'm wide of the mark.

### DIFF
--- a/exercises/accumulate/accumulate_test.go
+++ b/exercises/accumulate/accumulate_test.go
@@ -23,14 +23,14 @@ var tests = []struct {
 	{[]string{}, []string{}, echo, "echo"},
 	{[]string{"echo", "echo", "echo", "echo"}, []string{"echo", "echo", "echo", "echo"}, echo, "echo"},
 	{[]string{"First", "Letter", "Only"}, []string{"first", "letter", "only"}, capitalize, "capitalize"},
-	{[]string{"HELLO", "WORLD"}, []string{"hello", "world"}, strings.ToUpper, "upcase"},
+	{[]string{"HELLO", "WORLD"}, []string{"hello", "world"}, strings.ToUpper, "strings.ToUpper"},
 }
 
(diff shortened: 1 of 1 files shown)


bordeltabernacle:
The failure message is now:
```go
--- FAIL: TestAccumulate (0.00s)
	accumulate_test.go:33: Accumulate([hello world], strings.ToUpper): expected [HELLO WORLD], actual []
FAIL
exit status 1
FAIL	github.com/xgo/exercises/accumulate	0.002s
```

kytrinyx:
Yes, this looks much better to me! I'll leave this to get a second opinion before merging.

Thank you :heart:

petertseng:
Confirmed that `go vet` has no complaints now, and that the new error message looks good`, so since we've got two yeses, let's go!

petertseng:
Thank you!

bordeltabernacle:
:+1: Brilliant, thanks!

PR #339, opened 2016-08-03
Added two benchmark tests

roundand:
Added two benchmark tests to allow users to experiment with channel granularity and buffer depth.

### DIFF
--- a/exercises/parallel-letter-frequency/parallel_letter_frequency_test.go
+++ b/exercises/parallel-letter-frequency/parallel_letter_frequency_test.go
@@ -46,23 +46,13 @@ O'er the land of the free and the home of the brave?`
 )
 
 func BenchmarkSequentialFrequency(b *testing.B) {
-	b.StopTimer()
-	b.StartTimer()
-
 	for i := 0; i < b.N; i++ {
 		Frequency(euro + dutch + us)
 	}
-
-	b.StopTimer()
 }
 
 func BenchmarkConcurrentFrequency(b *testing.B) {
-	b.StopTimer()
-	b.StartTimer()
-
 	for i := 0; i < b.N; i++ {
 		ConcurrentFrequency([]string{euro, dutch, us})
 	}
-
-	b.StopTimer()
 }
--- a/exercises/parallel-letter-frequency/parallel_letter_frequency_test.go
+++ b/exercises/parallel-letter-frequency/parallel_letter_frequency_test.go
@@ -44,3 +44,25 @@ Gave proof through the night that our flag was still there;
 O say does that star-spangled banner yet wave,
 O'er the land of the free and the home of the brave?`
 )
+
+func BenchmarkSequentialFrequency(b *testing.B) {
+	b.StopTimer()
+	b.StartTimer()
+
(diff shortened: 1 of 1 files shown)


exercises/parallel-letter-frequency/parallel_letter_frequency_test.go
```diff
@@ -44,3 +44,25 @@ Gave proof through the night that our flag was still there;
 O say does that star-spangled banner yet wave,
 O'er the land of the free and the home of the brave?`
 )
+
+func BenchmarkSequentialFrequency(b *testing.B) {
+	b.StopTimer()
```
petertseng: I'm not familiar with everything about the timer in benchmarking, so you may have to excuse some easy questions - what will happen if these calls to `StopTimer` and `StartTimer` are removed? Will they cause reporting to be inaccurate in some way?

I see that there are no such calls in https://github.com/exercism/xgo/blob/master/exercises/leap/leap_test.go , but there are in (for example) https://github.com/exercism/xgo/blob/master/exercises/connect/connect_test.go where we didn't want to inculde `prepare(tt.board)` in the time recorded.
It doesn't look like we have anything that applies here since the timer just gets started immediately after it gets stopped, with no intervening work in between.

Are we doing something wrong in `leap` (we should call `StopTimer` and `StartTimer` but we aren't), or are the other differences between the situations that I'm not aware of?

exercises/parallel-letter-frequency/parallel_letter_frequency_test.go
```diff
@@ -44,3 +44,25 @@ Gave proof through the night that our flag was still there;
...
+	b.StartTimer()
+
+	for i := 0; i < b.N; i++ {
+		Frequency(euro + dutch + us)
+	}
+
+	b.StopTimer()
```
petertseng: In the same vein, what will happen if this is removed?

roundand: I have to come clean - I copy-pasted that from another test. If you remove the Timer calls, it appears to make no difference at all.

Not too sure what comes next here, but looks like I should be able to "Add more commits by pushing to the master branch on roundand/xgo" so I'll try that.

petertseng: Yup, you figured it out.

roundand: Looks like StartTimer() and StopTimer() are only needed if you want to do stop the clock while doing some expensive setup (https://golang.org/pkg/testing/#B.StartTimer)

petertseng:
Thanks, everything looks good at first glance. If you could please allow me a few hours for some final checks on it when I get to my main computer. If I haven't said anything in about 8 hours you can and should prod me, but I expect I'll be merging this.

petertseng:
Thanks!

roundand:
Thank you, Peter - not only has the experience been just as instructive as I'd hoped, but it's a pleasure to give even a few obvious bytes back to such a helpful project as Exercism.

PR #342, opened 2016-08-12
Fix readability of comments in test file

alebaffa:
Hi, I know this is a very trivial contribution :) but I fetched this exercise and I had quite a hard time trying to figure out what I had to implement reading the test file. I found the comments quite confusing due to a bad formatting, so this contribution tries to improve their readability. 

I hope you find this helpful :)

### DIFF
--- a/exercises/palindrome-products/palindrome_products_test.go
+++ b/exercises/palindrome-products/palindrome_products_test.go
@@ -46,19 +46,32 @@ var bonusData = []struct {
 	pmin, pmax Product
 	errPrefix  string
 }{
-	// You can reach the limit of the largest palindrome made of two 2-digit numbers.
-	{-99, -10, Product{}, Product{}, "Negative limits"},
+	/* This commented test case has the same input of the following uncommented one.
+	Comment/uncomment one of them if you want to try to handle different cases using the same input.
+	WARNING: you will probably have to change your solution in order to handle it.
+	*/
+
+	/* Here you can test that you can reach the limit of the largest palindrome made of two 2-digit numbers.
+	{-99, -10, Product{}, Product{}, "Negative limits"}, */
+
 	// You can still get non-negative products from negative factors.
 	{-99, -10,
 		Product{121, [][2]int{{-11, -11}}},
 		Product{9009, [][2]int{{-99, -91}}},
 		""},
-	// In case you reverse the *digits* you could have the following cases:
-	// - the zero has to be considered
+
+	/* This commented test case has the same input of the following uncommented one.
+	Comment/uncomment one of them if you want to try to handle different cases using the same input.
+	WARNING: you will probably have to change your solution in order to handle it.
+	*/
+
+	/*In case you reverse the *digits* you could have the following cases:
+	- the zero has to be considered
 	{-2, 2,
 		Product{0, [][2]int{{-2, 0}, {-1, 0}, {0, 0}, {0, 1}, {0, 2}}},
 		Product{4, [][2]int{{-2, -2}, {2, 2}}},
-		""},
+		""}, */
+
 	// - you can keep the minus sign in place
 	{-2, 2,
 		Product{-4, [][2]int{{-2, 2}}},
--- a/exercises/palindrome-products/palindrome_products_test.go
+++ b/exercises/palindrome-products/palindrome_products_test.go
@@ -46,10 +46,7 @@ var bonusData = []struct {
 	pmin, pmax Product
 	errPrefix  string
 }{
-	/* This commented test case has the same input of the following uncommented one.
-	Comment/uncomment one of them if you want to try to handle different cases using the same input.
-	WARNING: you will probably have to change your solution in order to handle it.
-	*/
+	// The following two test cases have the same input, but different expectations. Uncomment just one or the other.
 
 	/* Here you can test that you can reach the limit of the largest palindrome made of two 2-digit numbers.
 	{-99, -10, Product{}, Product{}, "Negative limits"}, */
@@ -60,10 +57,7 @@ var bonusData = []struct {
 		Product{9009, [][2]int{{-99, -91}}},
 		""},
 
-	/* This commented test case has the same input of the following uncommented one.
-	Comment/uncomment one of them if you want to try to handle different cases using the same input.
-	WARNING: you will probably have to change your solution in order to handle it.
-	*/
+	// The following two test cases have the same input, but different expectations. Uncomment just one or the other.
 
 	/*In case you reverse the *digits* you could have the following cases:
 	- the zero has to be considered
--- a/exercises/palindrome-products/palindrome_products_test.go
+++ b/exercises/palindrome-products/palindrome_products_test.go
@@ -7,14 +7,15 @@ import (
 	"testing"
 )
 
-// API to impliment:
-// type Product struct {
-// 	Product int // palindromic, of course
-// 	// list of all possible two-factor factorizations of Product, within
-// 	// given limits, in order
-// 	Factorizations [][2]int
-// }
-// func Products(fmin, fmax int) (pmin, pmax Product, error)
+/* API to impliment:
+
+type Product struct {
+	Product int // palindromic, of course
+	Factorizations [][2]int //list of all possible two-factor factorizations of Product, within given limits, in order
+ }
+
+ func Products(fmin, fmax int) (pmin, pmax Product, error)
+*/
 
 var testData = []struct {
 	// input to Products(): range limits for factors of the palindrome
@@ -37,10 +38,16 @@ var testData = []struct {
 		""},
 	{4, 10, Product{}, Product{}, "No palindromes"},
 	{10, 4, Product{}, Product{}, "fmin > fmax"},
-	/* bonus curiosities.  (can a negative number be a palindrome?
-	// most say no
+}
+
+// Bonus curiosities. Can a negative number be a palindrome? Most say no
+var bonusData = []struct {
+	fmin, fmax int
+	pmin, pmax Product
+	errPrefix  string
+}{
 	{-99, -10, Product{}, Product{}, "Negative limits"},
-	// but you can still get non-negative products from negative factors.
+	// but you can still get non-negative products from negative factors
 	{-99, -10,
 		Product{121, [][2]int{{-11, -11}}},
 		Product{9009, [][2]int{{-99, -91}}},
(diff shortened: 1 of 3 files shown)


exercises/palindrome-products/palindrome_products_test.go
```diff
@@ -37,10 +38,14 @@ var testData = []struct {
...
-	/* bonus curiosities.  (can a negative number be a palindrome?
-	// most say no
-	{-99, -10, Product{}, Product{}, "Negative limits"},
-	// but you can still get non-negative products from negative factors.
+
+	/* Bonus curiosities.  Can a negative number be a palindrome? Most say no
+	{-99, -10,
+		Product{},
+		Product{},
+	"Negative limits"},
```
kytrinyx: I find this bit harder to read when it's spread across several lines in this way.

kytrinyx: I wonder if we can put the bonuses in a separate slice, and then have a line that we comment out that would append the bonuses to the testData.

Does that make sense?

exercises/palindrome-products/palindrome_products_test.go
```diff
@@ -7,14 +7,15 @@ import (
...
-// func Products(fmin, fmax int) (pmin, pmax Product, error)
+/* API to impliment:
+
+type Product struct {
+	Product int // palindromic, of course
+	Factorizations [][2]int //list of all possible two-factor factorizations of Product, within given limits, in order
+ }
+
+ func Products(fmin, fmax int) (pmin, pmax Product, error)
+*/
```
kytrinyx: Nice, this feels cleaner and easier to scan than the previous version.

exercises/palindrome-products/palindrome_products_test.go
```diff
@@ -49,13 +54,17 @@ var testData = []struct {
...
 		Product{4, [][2]int{{-2, -2}, {2, 2}}},
 		""},
-	// or you could reverse the *digits*, keeping the minus sign in place.
+
+	or you could reverse the *digits*, keeping the minus sign in place.
 	{-2, 2,
 		Product{-4, [][2]int{{-2, 2}}},
 		Product{4, [][2]int{{-2, -2}, {2, 2}}},
-		""},
+	""},
```
petertseng: the indentation seems inconsistent here, since the last line is unindented, as oppoed to the examples immediately above (where they're the same indentation level as the `Product` lines.

If we were able to uncomment these cases (just not necessarily have them in the default set of cases to test) then we would be able to enlist the help of `go fmt` to choose the consistent style. In that light, 
https://github.com/exercism/xgo/pull/342/files#r74678867 seems like a useful route to go.

exercises/palindrome-products/palindrome_products_test.go
alebaffa: it definitely makes sense. 
I have now put the bonus data in a different slice (called bonusData) and I put this commented line (`//testData = append(testData, bonusData...)`) inside the test function `TestPalindromeProducts`, in the first line.

The bonus slice remains not commented. And doing this I found a compilation error with the bonus data (related to the last element of the bonus slice):

> cannot use ^uint(0) >> 1 (type uint) as type int in field value

The problem now is that if the bonus slice remains uncommented, this line won't make the test compile.

petertseng: doh, that's unfortunate. Another good reason why we'd want to uncomment it, I guess, we wouldn't have been able to catch that.

My first suggestion is... is it possible to convert that value back to an int? `int(^uint(0) >> 1)` or something? Since it's right-shifted by one, that means that should become the largest positive signed int value, which I suppose is the intent of the test (to make it overflow).

I looked to see if there were any constants that can be used, but there's `MaxInt32` and `MaxInt64` and we would need to know the size of `int` before using either of those. I guess that's why `^uint(0) >> 1` is being used. Anyway, from http://stackoverflow.com/questions/6878590/the-maximum-value-for-an-int-type-in-go it looks like `int(^uint(0) >> 1)` is indeed the way to go.

alebaffa: Unfortunately, also casting to `int` the compiler complains and it doesn't even build the entire test file (I guess it's part of the static checks that the Go compiler does at compiling time). Maybe we can leave this element in the slice commented (but then, would it make sense to keep it?)

alebaffa: oh no my mistake, sorry! I casted only the `^uint(0)` to `int(^uint(0))` so leaving out the bit shift and so I had the error.
Your solution works fine! I will update the PR :) Sorry for that.

petertseng: I had a different experience: It compiled fine for me, but the example solution doesn't handle the case.

The first problem with the example code (well, more like a weird thing with this test case, really) that zero is a palindrome, so all pairs whose product is zero was getting appended to the array, which for `0..maxint` is going to be a large number of them (so much that you can easily run out of memory).

The other problem of course is that the example solution doesn't deal with the overflow.

Even if the 0 were changed to a 1, a solution that doesn't detect the overflow would take a long time to run (it does take a long time to make maxint^2 iterations, of course). You can ctrl-C the test, of course, but it can still be puzzling.

I think if this case is kept, it is wise to have a comment warning of the potential problems. I would not mind removing it.

alebaffa: I think it's better to remove it, if you don't mind. I will update the PR. Do you prefer me to squash the commits?

exercises/palindrome-products/palindrome_products_test.go
```diff
@@ -37,10 +38,16 @@ var testData = []struct {
 		""},
 	{4, 10, Product{}, Product{}, "No palindromes"},
 	{10, 4, Product{}, Product{}, "fmin > fmax"},
-	/* bonus curiosities.  (can a negative number be a palindrome?
-	// most say no
+}
+
+// Bonus curiosities. Can a negative number be a palindrome? Most say no
```
kytrinyx: Can we have one comment per test case in the bonus section so that each comment is close to the test case that it refers to?

alebaffa: Good idea, I added two new comments for two bonus test cases (and removed the overflow case, discussed before).

petertseng:
> Do you prefer me to squash the commits?

It will be best if each commit contains "palindrome-products" in it so that when scanning the commit logs it's easy to tell what commit affects what exercise.

My most preferred would be one commit that changes the comment at the top of the file, and a second that deals with the bonus caes.

Only if that is not achievable, a single commit.

---

The bonus cases contradict each other (two possible answers for `{-99, -10}` and `{-2, 2}`) - I guess that's one reason why they were commented originally. Are we OK with this? That means uncommenting the `testData = append(testData, bonusData...)` line will cause no solution to pass

alebaffa:
It would be better to keep just one of the answers and to comment the other one, for both cases. I could keep the first answer of both cases and comment the second ones (putting an additional comment saying to uncomment them in case you want to handle other cases - and so change your code, I guess).
What do you think?

kytrinyx:
> Are we OK with this?

Ouch, I didn't realize that. No I don't think so. I would rather make it so that it's clear that you have to pick one case or the other... but I'm not sure what the best way to go about doing that is.

petertseng:
> I would rather make it so that it's clear that you have to pick one case or the other... but I'm not sure what the best way to go about doing that is.

> I could keep the first answer of both cases and comment the second ones (putting an additional comment saying to uncomment them in case you want to handle other cases - and so change your code, I guess).

Slightly regretful that we will lose the advantage of `go fmt` helping with formatting, but I suppose this is the best we can do. So let's do it.

alebaffa:
Done. Let me know if you think the comments are OK, then I will (try to) squash the commits as you said you prefer :)

exercises/palindrome-products/palindrome_products_test.go
```diff
@@ -37,29 +38,50 @@ var testData = []struct {
...
-	// but you can still get non-negative products from negative factors.
+}
+
+// Bonus curiosities. Can a negative number be a palindrome? Most say no.
+var bonusData = []struct {
+	fmin, fmax int
+	pmin, pmax Product
+	errPrefix  string
+}{
+	/* This commented test case has the same input of the following uncommented one.
```
kytrinyx: I was confused by this. Maybe:

> The following two test cases have the same input, but different expectations.
> Uncomment just one or the other.

alebaffa: I updated the comments following your suggestions: way better :) thank you!

exercises/palindrome-products/palindrome_products_test.go
```diff
@@ -37,29 +38,44 @@ var testData = []struct {
...
+	// The following two test cases have the same input, but different expectations. Uncomment just one or the other.
+
+	/* Here you can test that you can reach the limit of the largest palindrome made of two 2-digit numbers.
+	{-99, -10, Product{}, Product{}, "Negative limits"}, */
+
+	// You can still get non-negative products from negative factors.
 	{-99, -10,
 		Product{121, [][2]int{{-11, -11}}},
 		Product{9009, [][2]int{{-99, -91}}},
 		""},
```
kytrinyx: Sorry to go back to this. It bugs me—probably more than it should. Is this how gofmt would format this?

alebaffa: yes :( I have all the go toolings installed in my IDE and they are activated every time i save. To be more sure I also just used `go fmt` on the file from the command line and the result is the same :(

petertseng: it may help if you express what bugs you.

Is it the fact that the closing brace is indented differently from the opening brace?

Is it the fact that some values (the -99 and -10) are indented differently from the others?

Newlines in the right places will deal with those, of course, but it could be something else not expressed here.

kytrinyx: > it may help if you express what bugs you.

Yeah, sorry, it would definitely help!

I don't like the uneven indentation, and the fact that some values are indented differently from each other. I would also expect to see the closing brace on the same line as values if it were all on the same line.

kytrinyx: e.g.

```
// all on one line
{"a", "b", "c", "d", "e"}

// even indentation
{
    "a", "b",
    "c", "d",
    "e"
}
```

exercises/palindrome-products/palindrome_products_test.go
```diff
@@ -7,14 +7,15 @@ import (
...
 
-// API to impliment:
-// type Product struct {
-// 	Product int // palindromic, of course
-// 	// list of all possible two-factor factorizations of Product, within
-// 	// given limits, in order
-// 	Factorizations [][2]int
-// }
-// func Products(fmin, fmax int) (pmin, pmax Product, error)
+/* API to impliment:
```
kytrinyx: I totally didn't notice this before (so sorry): `s/impliment/implement/`

kytrinyx: I'll actually go ahead and fix this. I think the rest of this is fine, so I'll get this merged today.

kytrinyx:
Merged in 37f896377e9508839986f9ee9ed0142ca96c7236 :heart:

alebaffa:
Cool! I am really sorry if I didn't follow up this PR. I really didn't have time at all to continue on this. Next time I will do a PR i will do more checks on my side and ask more questions instead of doing too many commits.
Thank you, @kytrinyx and @petertseng !

PR #343, opened 2016-08-14
fix typo in etl_test.go

rmetzler:


kytrinyx:
"Tran" is Norwegian for codliver oil. Not a good look in a test suite :) Thanks!

PR #344, opened 2016-08-16
clock: bump test version in stub file

dmgawel:
After reworking clock tests in 0c4961d test version in stub file was not upgraded. This PR fixes that.


petertseng:
Right you are. Thanks.

PR #346, opened 2016-08-22
sort slice before comparing

aarti:
Small change to sort string slices before comparing for equality.

### DIFF
--- a/exercises/allergies/allergies_test.go
+++ b/exercises/allergies/allergies_test.go
@@ -2,6 +2,7 @@ package allergies
 
 import (
 	"fmt"
+	"sort"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


petertseng:
Yup, I'll take it. Have had to do this in other languages as well (see https://github.com/exercism/xrust/pull/134), and it makes sense that the order doesn't matter, the readme only asks for what they're allergic to.

I am going to amend the commit message to include "allergies" in it, so that if we are only scanning through commit logs we can tell what commits are associated with what exercise.

Thanks.

petertseng:
(I don't think a test version bump is required, since solutions before this change will still pass with this change)

kytrinyx:
> I don't think a test version bump is required

Agreed—I think we're good here.

PR #347, opened 2016-08-25
grade-school: Add the minimum required API

jcbwlkr:
When I solved the `grade-school` problem it took me a while to piece together the required API from the tests.

For `Grade` the tests construct literals with `{2, []string{"Aimee"}}` so we need to make it a `struct{int, []string}` but the names of the fields are up to the dev.

For `School` we need the methods documented here but the actual type for `School` is flexible so I didn't include it; I just included a placeholder to say you need to have a type named `School` (line 7). I think this could be more clear but I'm not sure what would be best here.

### DIFF
--- a/exercises/grade-school/grade_school_test.go
+++ b/exercises/grade-school/grade_school_test.go
@@ -1,3 +1,15 @@
+// API:
+//
+// type Grade struct {
+// 	int, []string
+// }
+//
+// type School
+// func New() *School
+// func (s *School) Add(string, int)
+// func (s *School) Grade(int) []string
+// func (s *School) Enrollment() []Grade
+
 package school
 
 import (


petertseng:
Thanks, I think this will be a good idea. Based on your own thoughts, there could be a few additions

> we need to make it a `struct{int, []string}` but the names of the fields are up to the dev.

I *might* consider a "You may add name to these fields if you like" comment above Grade for this.

>  I just included a placeholder to say you need to have a type named `School` (line 7). I think this could be more clear but I'm not sure what would be best here.

It *might* not be necessary, based on other examples I found:

* https://github.com/exercism/xgo/blob/master/exercises/pov/pov_test.go has no additional explanation on `type Graph`
* https://github.com/exercism/xgo/blob/master/exercises/circular-buffer/circular_buffer_test.go has no additional explanation on `type Buffer`
* https://github.com/exercism/xgo/blob/master/exercises/meetup/meetup_test.go also has no explanation for `type WeekSchedule` (which also should be capitalised but isn't, oops)
* https://github.com/exercism/xgo/blob/master/exercises/custom-set/custom_set_test.go simply doesn't include the `type Set` line at all

I certainly wouldn't complain about a line saying "you may define School as whatever type you like", but based on these others it seems we can do without. Unless you think there is a reason we need to be more explicit here (maybe the grade-school exercise comes earlier than all these other exercises, I didn't check that)

jcbwlkr:
The `Grade` struct won't compile as I have it until they name the fields. This exercise comes in about 70% through the track so we probably don't need any additional explanation on either.

petertseng:
OK. In that case, I think you and I both agree this can get merged as-is. I'll give it about six hours in case anyone has anything else to say, and if not then I'll merge it (and you can remind me if I forget).

petertseng:
Thanks!

jcbwlkr:
Awesome 👍

PR #350, opened 2016-08-31
binary-search-tree: Add Minimum Required API

jcbwlkr:
The fields are accessed directly by the tests so they have to be named just like this. Some people put `left` and `right` on the same line as `left, right *SearchTreeData` but it doesn't really matter.

### DIFF
--- a/exercises/binary-search-tree/binary_search_tree_test.go
+++ b/exercises/binary-search-tree/binary_search_tree_test.go
@@ -1,3 +1,16 @@
+// API:
+//
+// type SearchTreeData struct {
+//	left  *SearchTreeData
+//	data  int
+//	right *SearchTreeData
+// }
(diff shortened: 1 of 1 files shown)


petertseng:
Thank you very much, I think this will do as-is.

Listing out the required API as you have has made me realize that it is quite strange that the test constructs e.g. `bst := SearchTreeData{data: 4}`, forcing the student to name the structure `SearchTreeData`. If it were instead always using `bst := Bst(4)` (or I think renaming `Bst` to `New` would be more idiomatic) then the student would have freedom to name the structure as desired.

Similar ideas may be used to allow the members of the struct to be flexible.

Anyway, all that I've said here is not related to this PR, but I would consider the discussed changes welcome in a follow up too!

PR #352, opened 2016-09-02
Robot Simulator - Update 1

jbsmith7741:
Spelling fixes
Changes to API
- func Robot() -> StartRobot()
  - This function is not a robot, but rather it is starting a process that moves a robot around. This function could be called Start() or something else more descriptive
- DirAt -> Robot
   - DirAt is pretty a good description as it is a Direction and a position, but isn't exactly what defines a robot in stage 2.
- Place -> Robot3
    - Since this has a robot name, direction and position, Place doesn't seem like a very good description of the structure.

### DIFF
--- a/exercises/robot-simulator/defs.go
+++ b/exercises/robot-simulator/defs.go
@@ -19,14 +19,14 @@ type Command byte // valid values are 'R', 'L', 'A'
 type RU int
 type Pos struct{ Easting, Northing RU }
 type Rect struct{ Min, Max Pos }
-type DirAt struct {
+type Step2Robot struct {
 	Dir
 	Pos
 }
 
 // additional definition used in step 3
 
-type Place struct {
+type Step3Robot struct {
 	Name string
-	DirAt
+	Step2Robot
 }
--- a/exercises/robot-simulator/example.go
+++ b/exercises/robot-simulator/example.go
@@ -3,7 +3,7 @@ package robot
 import "fmt"
 
 // ======= Step 1
-
+const testVersion = 2
 const (
 	N Dir = iota
 	E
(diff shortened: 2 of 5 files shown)


exercises/robot-simulator/defs.go
```diff
@@ -19,14 +19,14 @@ type Command byte // valid values are 'R', 'L', 'A'
...
-type DirAt struct {
+type Robot struct {
 	Dir
 	Pos
 }
 
 // additional definition used in step 3
 
-type Place struct {
+type Robot3 struct {
```
petertseng: I'll agree that `Place` was not a good name for it. What would you say about `NamedRobot` so that from reading the name we can get an idea of how it differs from `Robot`?

kytrinyx: Hm. Yeah, `Place` isn't great. `Robot3` has the disadvantage that when reading it we can't tell how it's distinct from `Robot` or `Robot2`, though.

`NamedRobot` is missing the idea that it's also located somewhere. This is really tricky!

exercises/robot-simulator/example.go
```diff
@@ -29,14 +29,14 @@ func Advance() {
...
-func Robot(cmd chan Command, act chan Action) {
+func StartRobot(cmd chan Command, act chan Action) {
 	for c := range cmd {
 		act <- Action(c)
 	}
 	close(act)
 }
 
-func Room(extent Rect, place DirAt, act chan Action, report chan DirAt) {
+func Room(extent Rect, place Robot, act chan Action, report chan Robot) {
```
kytrinyx: I wonder if `place` is now misleading.

jbsmith7741: I agree, that should probably just be robot.

exercises/robot-simulator/defs.go
jbsmith7741: The numbers after each robot was supposed to represent each step of the problem. So in Step 3 you would use the Robot3 and in Step 1 you would need Robot1, making it similar to how Room3 and Robot3 were original used. 

NamedRobot is a pretty good name. If we assume that a robot is defined by a position and direction than NamedRobot would imply location as well.

petertseng:
This also reminds me that the Go track's version of this problem (added at https://github.com/exercism/xgo/pull/83) was written for an old version of the readme - the one before https://github.com/exercism/x-common/pull/104 was merged. We could consider slimming down this exercise to match. In particular the room / wall constraints, and the fact that the README for this exercise no longer mentions any steps (1, 2, 3) like the tests do

petertseng:
Although I guess it does take away a bit of the charm of the problem and the teaching (the channels and all). But maybe there is another exercise better suited for that kind of thing

kytrinyx:
I don't know what the right answer is.

If we do keep the steps, I wonder if the names `Step1Robot`, `Step2Robot`, and `Step3Robot` would help clarify. The thing that was tripping me up about `Robot2` vs `Robot3` is that I couldn't tell from reading the names what the meaningful difference between them was.

jbsmith7741:
I like it. What if we just changed Robot2 to Robot and Robot3 to NamedRobot like @petertseng mentioned.

petertseng:
It looks like we settled on Step2Robot and Step3Robot? OK, I don't mind. I assume Katrina would be OK with that too, given Katrina suggested this change first.

I would really appreciate if the spelling fixes came in a different commit than the renames. They are logically separate changes and it's nice to be able to revert, say, the API changes without having to revert the spelling fixes.

It might be good for me to do this exercise to gain some perspective on it. Don't necessarily wait for me to do that though if you want to merge it @kytrinyx .

kytrinyx:
> I would really appreciate if the spelling fixes came in a different commit than the renames.

I'd like to echo this. We've been really good about keeping logically separate changes in separate commits in this track, and it makes a huge difference when searching the git logs, etc.

@jbsmith7741 would you mind tweaking this? If not, let me know and I'll see what I can do on my end to separate them out.

jbsmith7741:
@kytrinyx @petertseng I'll split this into 2 commits, one for just the spelling changing and the other for everything else

kytrinyx:
Thank you so much, @jbsmith7741!

PR #353, opened 2016-09-02
Robot simulator - Update 2

jbsmith7741:
Phase 2 involves changing the global variables X, Y and Facing (Dir) into a named struct. This is mostly to make the step 1 feel more uniform to step 2 and 3

### DIFF
--- a/exercises/robot-simulator/defs.go
+++ b/exercises/robot-simulator/defs.go
@@ -4,10 +4,10 @@ import "fmt"
 
 // definitions used in step 1
 
-var (
+var Step1Robot struct {
 	X, Y   int
-	Facing Dir
-)
+	Dir
+}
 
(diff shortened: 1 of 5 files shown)


petertseng:
I think the main change here is to add a `Robot1` surrounding the formerly-global `X, Y`, and `Facing` (`Dir`), right? I would find that valuable.

petertseng:
(also can we get a rebase on master so that we can see what changes we have remaining to deal with? Thanks)

petertseng:
Yeah this is looking good! As soon as you can change the test version of the example I think we will be ready!

jbsmith7741:
Let me know if anything else needs to be changed

petertseng:
Everything is looking good here! Thanks!

PR #355, opened 2016-09-04
Add Effective Go

rakyll:
Effective Go is by far the most comprehensive official best practices document
for Go. Worth to mention it in the list somewhere before the other resources.


petertseng:
Quick question, it's already in https://github.com/exercism/xgo/blob/master/docs/RESOURCES.md - is it useful to have in both places? Should it be moved from RESOURCES to LEARNING? The answer will depend on whether it is best for new learners or ongoing work.

kytrinyx:
I use Effective Go in an ongoing manner, but it really is a phenomenal document when you're just getting into Go as well. I think it's worth having it in both places.

petertseng:
Sounds to me like we're in favor of merging then.

Thank you @rakyll!

PR #356, opened 2016-09-05
triangle: Add testcase to catch faulty shortcut

eneuschild:
[A solution I came across](http://exercism.io/submissions/f48a48bf39da400b88da10b148022fd4) suggests a check for
triangle equilaterality that can produce false positives.

The newly added testcase triggers this.

### DIFF
--- a/exercises/triangle/example.go
+++ b/exercises/triangle/example.go
@@ -4,7 +4,7 @@ import "math"
 
 type Kind string
 
-const testVersion = 2
+const testVersion = 3
 
 const (
 	Equ Kind = "equilateral"
--- a/exercises/triangle/triangle_test.go
+++ b/exercises/triangle/triangle_test.go
@@ -26,7 +26,8 @@ var testData = []testCase{
 	{Sca, 5, 4, 2},    // descending order
 	{Sca, .4, .6, .3}, // small sides
 	{Sca, 1, 4, 3},    // a "triangle" that is just a line is still OK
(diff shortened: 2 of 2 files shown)


petertseng:
Ah, after I saw the added test and before I read the solution I thought it was going to be `(a + b + c) / 3 = a` or something (Oh, actually, that's equivalent to `b + c = 2a` because multiply by 3 and subtract a).

OK, I feel like we can take this test. It might also be useful to add it to https://github.com/exercism/x-common/blob/master/triangle.json , would you like to do that?

Do you think it would be useful to put `b + c = 2a` in the comment? Otherwise it might not be clear exactly what false positive is being referred to.

Now this is making me wonder, are the tests `4, 5, 6` and `4, 6, 5` needed too, in case someone did the same check but with `a + b = 2c` instead of `b + c = 2a`? What do you think about this?

eneuschild:
Hi,

sure, I can add it to the JSON too.

I've also been wondering how to write the comment so it says what I mean but doesn't get too verbose.
How about `2a == b+c looks like equilateral, but isn't.`?

It probably makes sense to add the other number permutations as well, I agree.

kytrinyx:
> 2a == b+c looks like equilateral, but isn't.

I like that

petertseng:
Thanks, looking good. Last thing will be to make the test version reported by the example solution match the new test version in the test, if you could do that I think we will be ready to merge this.

eneuschild:
You're right of course; done.

Thanks for your feedback and guidance!

petertseng:
Thanks!

PR #358, opened 2016-09-21
move icon attribution from exercism.io/ATTRIBUTION.md to here

catb0t:
a step in fixing https://github.com/exercism/exercism.io/issues/3112.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,3 +125,6 @@ directions being tried.  Your fresh ideas and contributions are welcome.  :spark
 The MIT License (MIT)
 
 Copyright (c) 2014 Katrina Owen, _@kytrinyx.com
+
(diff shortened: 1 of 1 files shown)


petertseng:
Seems we should have done this in conjunction with #331 - good to get it done now.

I verify that the statement added in this PR is the same statement as currently in https://github.com/exercism/exercism.io/blob/master/ATTRIBUTION.md , and furthermore that the statement seems correct from looking at https://blog.golang.org/gopher . Finally, I believe from other PRs associated with https://github.com/exercism/exercism.io/issues/3112 that this action is generally something we want to do.

Merging.

petertseng:
Thanks.

PR #363, opened 2016-09-29
connect: Improve error-handling behavior (close #286)

### LINKED ISSUE #286: connect: Improve error-handling behavior

If the connect exercise does not have error handling, then this ticket can be closed immediately.

```go
var _ error = err
```

dvrkps:


petertseng:
Thanks!

PR #364, opened 2016-10-01
Improved error handling behavior in phone-number

Akasurde:
Fixes: https://github.com/exercism/xgo/issues/302

### DIFF
--- a/exercises/phone-number/phone_number_test.go
+++ b/exercises/phone-number/phone_number_test.go
@@ -4,6 +4,8 @@ import (
 	"testing"
 )
 
+const targetTestVersion = 1
+
 type testCase struct {
 	input     string
 	expected  string
@@ -20,18 +22,23 @@ var numberTests = []testCase{
 }
 
 func TestNumber(t *testing.T) {
+	if testVersion != targetTestVersion {
+		t.Errorf("Found testVersion = %v, want %v.", testVersion, targetTestVersion)
+	}
 	for _, test := range numberTests {
 		actual, actualErr := Number(test.input)
-		if actual != test.expected {
-			t.Errorf("Number(%s): expected [%s], actual: [%s]", test.input, test.expected, actual)
-		}
-		// if we expect an error and there isn't one
-		if test.expectErr && actualErr == nil {
+		if !test.expectErr {
+			if actualErr != nil {
+				// if we don't expect an error and there is one
+				var _ error = actualErr
+				t.Errorf("Number(%s): expected no error, but error is: %s", test.input, actualErr)
+			}
+			if actual != test.expected {
+				t.Errorf("Number(%s): expected [%s], actual: [%s]", test.input, test.expected, actual)
+			}
+		} else if actualErr == nil {
+			// if we expect an error and there isn't one
 			t.Errorf("Number(%s): expected an error, but error is nil", test.input)
-		}
-		// if we don't expect an error and there is one
-		if !test.expectErr && actualErr != nil {
-			t.Errorf("Number(%s): expected no error, but error is: %s", test.input, actualErr)
 		}
 	}
 }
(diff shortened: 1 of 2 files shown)


exercises/phone-number/phone_number_test.go
```diff
@@ -20,6 +22,9 @@ var numberTests = []testCase{
 }
 
 func TestNumber(t *testing.T) {
+	if testVersion != targetTestVersion {
+		t.Errorf("Found testVersion = %v, wanted %v.", testVersion, targetTestVersion)
+	}
 	for _, test := range numberTests {
 		actual, actualErr := Number(test.input)
 		if actual != test.expected {
```
petertseng: could you make it so that this is only checked if `actualErr == nil`? This would cover the "Ignore the normal return value when an error expected" point of #302

petertseng: And also do the same for `testFormat` and `testAreaCode` except I can't put the comments there since they are too many lines away from a changed line

exercises/phone-number/phone_number_test.go
```diff
@@ -20,19 +22,24 @@ var numberTests = []testCase{
...
+				var _ error = actualErr
+				t.Fatalf("Number(%s): expected no error, but error is: %s", test.input, actualErr)
+			}
+			if actual != test.expected {
+				t.Fatalf("Number(%s): expected [%s], actual: [%s]", test.input, test.expected, actual)
+			}
+		} else if actualErr == nil {
+			// if we expect an error and there isn't one
 			t.Errorf("Number(%s): expected an error, but error is nil", test.input)
 		}
```
kytrinyx: If we're expecting an error and we get an error, I believe we should verify that it is the error type. (Would you double-check me on that @petertseng?)

exercises/phone-number/phone_number_test.go
```diff
@@ -20,19 +22,24 @@ var numberTests = []testCase{
...
+				// if we don't expect an error and there is one
+				var _ error = actualErr
+				t.Fatalf("Number(%s): expected no error, but error is: %s", test.input, actualErr)
+			}
+			if actual != test.expected {
+				t.Fatalf("Number(%s): expected [%s], actual: [%s]", test.input, test.expected, actual)
+			}
+		} else if actualErr == nil {
+			// if we expect an error and there isn't one
 			t.Errorf("Number(%s): expected an error, but error is nil", test.input)
```
petertseng: I notice the other checks were changed to use `Fatalf` - should this be change too?

exercises/phone-number/phone_number_test.go
petertseng: Ah right. What you say is true - however, I believe the `var _ error = actualErr` a few lines above has already constrained the error type. It is OK that it is in a separate conditional branch - the compiler still checks it. I think the following Go code example shows what I mean:

```go
package main

func notAnError() int {
    return 5
}

func main() {
	err := notAnError()
	if false {
		var _ error = err
	}
}
```

This *does* fail to compile, so I think that means we are OK on this front.

kytrinyx: Oh, right! Thanks for catching that.

exercises/phone-number/phone_number_test.go
kytrinyx: `Fatalf` will kill the entire test run. I actually think that we should have `Errorf` for all of these.

If we want to skip to the next iteration in the loop ignoring the next checks, then we should use `Errorf` with `continue`.

Akasurde: @kytrinyx I have seen `Fatalf` in the other Exercises' test case, so thought of converting `Errorf` to `Fatalf`
Let me know if you want this change to `Errorf`

kytrinyx: @Akasurde sometimes `Fatalf` makes sense, but in this case we'd like to see the errors for any of the test cases that would, so we need it to be `Errorf`. Thanks!

Akasurde: Ok.  I will edit pr and rebate it.

petertseng:
Thanks!

Akasurde:
@petertseng @kytrinyx Thanks for comments.

PR #365, opened 2016-10-01
word_search : Improved error handling behavior

Akasurde:
Fixes: https://github.com/exercism/xgo/issues/311

### DIFF
--- a/exercises/word-search/word_search_test.go
+++ b/exercises/word-search/word_search_test.go
@@ -10,7 +10,7 @@ import (
 // Also define a testVersion with a value that matches
 // the targetTestVersion here.
 
-const targetTestVersion = 2
+const targetTestVersion = 3
 
 var words = []string{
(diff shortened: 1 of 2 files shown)


exercises/word-search/word_search_test.go
```diff
@@ -51,6 +51,7 @@ func TestSolve(t *testing.T) {
 	}
 	actual, err := Solve(words, puzzle)
 	if err != nil {
+		var _ error = err
 		t.Fatalf("Didn't expect error but got %v", err)
 	}
 	if !reflect.DeepEqual(actual, positions) {
```
petertseng: note for future:

In case any test cases start to actually expect error (I can't imagine what would), in that case we would want to only check this if `err == nil`

petertseng:
Thank you for your contribution! (Amended commit message to be "Improve" rather than "Improved")

PR #368, opened 2016-10-01
Beer song improve err handling

mchoube:
1) Added versioning for beer-song test.
2) Added check for error type.

Please accept pull request.

### DIFF
--- a/exercises/beer-song/beer_test.go
+++ b/exercises/beer-song/beer_test.go
@@ -4,6 +4,14 @@ import (
 	"testing"
 )
 
+const targetTestVersion = 1
+
+func TestTestVersion(t *testing.T) {
+	if testVersion != targetTestVersion {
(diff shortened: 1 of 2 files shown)


kytrinyx:
Thanks @mchoube, this is exactly what we were looking for.

exercises/beer-song/beer_test.go
```diff
@@ -49,6 +57,7 @@ var verseTestCases = []struct {
 func TestBottlesVerse(t *testing.T) {
 	for _, tt := range verseTestCases {
 		actualVerse, err := Verse(tt.verse)
+		var _ error = err
 		if actualVerse != tt.expectedVerse {
```
petertseng: should this only be checked `if !tt.expectErr`? that would be in line with the "Ignore the normal return value when an error expected"

PR #372, opened 2016-10-01
check for error type only when expecting error.

mchoube:
incorporated code review comment.

### DIFF
--- a/exercises/beer-song/beer_test.go
+++ b/exercises/beer-song/beer_test.go
@@ -96,21 +96,23 @@ func TestSeveralVerses(t *testing.T) {
 
 	for _, tt := range versesTestCases {
 		actualVerse, err := Verses(tt.upperBound, tt.lowerBound)
-		if actualVerse != tt.expectedVerse {
-			t.Fatalf("Verses(%d, %d):\nexpected\n%s\nactual\n%s", tt.upperBound, tt.lowerBound, tt.expectedVerse, actualVerse)
-		}
-
 		if tt.expectErr {
+			// check if err is of error type
 			var _ error = err
-		}
 
-		// if we expect an error and there isn't one
-		if tt.expectErr && err == nil {
-			t.Errorf("Verses(%d, %d): expected an error, but error is nil", tt.upperBound, tt.lowerBound)
-		}
-		// if we don't expect an error and there is one
-		if !tt.expectErr && err != nil {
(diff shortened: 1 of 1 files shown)


petertseng:
Two suggestions on how to achieve "Only check expected value if there is no error":

Binary uses a normal `if/else` https://github.com/exercism/xgo/blob/master/exercises/binary/binary_test.go#L53

Trinary uses a `switch/case`: https://github.com/exercism/xgo/blob/master/exercises/trinary/trinary_test.go#L40

mchoube:
what about
if !tt.expectErr && err != nil {

should error be checked when not expected?

petertseng:
> `if !tt.expectErr && err != nil `
> should error be checked when not expected?

That line should stay because if the student submits a solution that gives a non-nil error when error should be nil, the test should fail.

So if the test case does not expect an error, then:

* The student's function should not say there was an error
* We should also check if the non-error return value (`actualVerse`) is correct.

If the test case does expect an error:

* We should check that the student's function also says there is an error (and it is of the error interface)
* We do *not* check the non-error return value (`actualVerse`)

exercises/beer-song/beer_test.go
```diff
@@ -57,18 +57,23 @@ var verseTestCases = []struct {
...
-		}
-
-		// if we expect an error and there isn't one
-		if tt.expectErr && err == nil {
-			t.Errorf("Verse(%d): expected an error, but error is nil", tt.verse)
-		}
-		// if we don't expect an error and there is one
-		if !tt.expectErr && err != nil {
-			t.Errorf("Verse(%d): expected no error, but error is: %s", tt.verse, err)
+		if tt.expectErr {
```
petertseng: Thanks! Looks like what you did here is good!

exercises/beer-song/beer_test.go
```diff
@@ -91,10 +96,14 @@ func TestSeveralVerses(t *testing.T) {
 
 	for _, tt := range versesTestCases {
 		actualVerse, err := Verses(tt.upperBound, tt.lowerBound)
-		var _ error = err
 		if actualVerse != tt.expectedVerse {
```
petertseng: could you do the same for this `if actualVerse != tt.expectedVerse` ?

mchoube: oops missed that. will update and send review.

mchoube: fixed please review.

petertseng:
Thanks! looking good! I will be squashing these three commits into one.

PR #375, opened 2016-10-02
Hexadecimal improve err handling

mchoube:
Improved error handling in hexadecimal test. Please review.

### DIFF
--- a/exercises/hexadecimal/hexadecimal_test.go
+++ b/exercises/hexadecimal/hexadecimal_test.go
@@ -56,8 +56,8 @@ func TestParseHex(t *testing.T) {
 
 			if !strings.Contains(strings.ToLower(err.Error()), test.errCase) {
 				t.Errorf(
-					"ParseHex(%q) returned error %q. Expected error %s.",
-					err, test.errCase)
+					"ParseHex(%q) returned error %q. Expected error containing '%s'.",
+					test.in, err, test.errCase)
 			}
 		} else {
 			if out != test.out {
--- a/exercises/hexadecimal/hexadecimal_test.go
+++ b/exercises/hexadecimal/hexadecimal_test.go
@@ -11,6 +11,7 @@
 package hexadecimal
 
 import (
+	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


exercises/hexadecimal/hexadecimal_test.go
```diff
@@ -42,18 +43,33 @@ var testCases = []struct {
...
+			// check if err is of error type
+			var _ error = err
+
+			// we expect error
+			if err == nil {
+				t.Errorf("ParseHex(%q): expected an error, but error is nil",
+					test.in)
+			}
+
+			if !strings.Contains(strings.ToLower(err.Error()), test.errCase) {
```
petertseng: interesting new addition! I think I like it.

exercises/hexadecimal/hexadecimal_test.go
```diff
@@ -42,18 +43,33 @@ var testCases = []struct {
...
+				t.Errorf(
+					"ParseHex(%q) returned error %q. Expected error %s.",
+					err, test.errCase)
```
petertseng: the format string has three vars (`%q`, `%q`, `%s`) but I only see two vars here (`err, `test.errCase`) - can it be made consistent? Probably needs an `test.in`.

exercises/hexadecimal/hexadecimal_test.go
```diff
@@ -42,18 +43,33 @@ var testCases = []struct {
...
+
+			// we expect error
+			if err == nil {
+				t.Errorf("ParseHex(%q): expected an error, but error is nil",
+					test.in)
+			}
+
+			if !strings.Contains(strings.ToLower(err.Error()), test.errCase) {
+				t.Errorf(
+					"ParseHex(%q) returned error %q. Expected error %s.",
```
petertseng: can I ask that this be clarify to say something like `Expected error containing '%s'` so a student who sees this error message can better understand what is required of them?

Sure, they can see it by reading the source, but if it can be made from the error message let's do that. For example it might be difficult to understand what "Expected error range" means.

mchoube: done.

exercises/hexadecimal/hexadecimal_test.go
mchoube: done.

exercises/hexadecimal/hexadecimal_test.go
```diff
@@ -34,18 +43,33 @@ var testCases = []struct {
...
+			// check if err is of error type
+			var _ error = err
+
+			// we expect error
+			if err == nil {
+				t.Errorf("ParseHex(%q): expected an error, but error is nil",
+					test.in)
+			}
+
+			if !strings.Contains(strings.ToLower(err.Error()), test.errCase) {
```
petertseng: if `err == nil` (student's function didn't return an error when it should have), this calls `.Error()` on `nil` and test will crash. I think the solution is to add a `continue` inside the `if err == nil`, does that seem right to you?

mchoube: oops. need better self review :(.

petertseng:
I'm pretty sure we got everything now! Thanks for working through it!

PR #376, opened 2016-10-02
secret-handshake: Use unsigned integers (close #373)

### LINKED ISSUE #373: secret-handshake: Use unsigned integers
As explained in #359, the secret-handshake should use use the `uint` type, since it deals with bitfields and masking those bits. It currently uses the `int` type, so this should be changed in both the test cases and the example implementation.

Any current test case that has a negative number can be removed.

mkoppmann:
Fixes #373.
Please review :)

### DIFF
--- a/exercises/secret-handshake/example.go
+++ b/exercises/secret-handshake/example.go
@@ -1,9 +1,11 @@
 package secret
 
+const testVersion = 1
+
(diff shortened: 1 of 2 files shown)


petertseng:
Fantastic. Thank you.

PR #377, opened 2016-10-02
Hamming improve err handling

mchoube:
1) Added check for test version.
2) Improved error handling.

Please review.

### DIFF
--- a/exercises/hamming/hamming_test.go
+++ b/exercises/hamming/hamming_test.go
@@ -2,7 +2,7 @@ package hamming
 
 import "testing"
 
-const targetTestVersion = 4
+const targetTestVersion = 5
 
 func TestTestVersion(t *testing.T) {
 	if testVersion != targetTestVersion {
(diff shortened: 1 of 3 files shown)


petertseng:
It looks like this is for #290 - it looks to me like there is already a `var _ error = err`, and the `switch`/`case` takes care of the "Ignore the normal return value when an error expected" part (If an error is expected, `tc.want < 0` and therefore the `got != tc.want` case will not be taken. So all requests of that issue are taken care of.

So if we would need to merge this PR, it would be for readability - I agree that using this `if`/`else` is at least easier to reason about than the `switch`/`case`, shall we prefer this way?

mchoube:
yea if-else block is improving readability.

exercises/hamming/hamming_test.go
```diff
@@ -2,26 +2,40 @@ package hamming
 
 import "testing"
 
-const targetTestVersion = 4
+const targetTestVersion = 5
+
+func TestTestVersion(t *testing.T) {
+	if testVersion != targetTestVersion {
```
petertseng: if this will be added, shall I suggest it get removed from `TestHamming`? We will only need to check it in one place.

petertseng:
Sorry that I keep finding more and more things to change - I was unprepared for the many PRs today. One last thing is that we have a stub file `hamming.go` that also has a `testVersion` constant. That stub file is served to students (unlike the `example.go` which is not shown to students) - can you also bump that to 5, as has been done in this PR?

PR #378, opened 2016-10-02
allergies: Use unsigned integers (close #374)

### LINKED ISSUE #374: allergies: Use unsigned integers
As explained in #359, the allergies exercise should use use the `uint` type, since it deals with bitfields and masking those bits. It currently uses the `int` type, so this should be changed in both the test cases and the example implementation.

Any current test case that has a negative number can be removed.

mkoppmann:
Fixes #374.

The `indexOf` function is still using `int` but I don’t know if it’s necessary to change that to `uint`.

### DIFF
--- a/exercises/allergies/allergies_test.go
+++ b/exercises/allergies/allergies_test.go
@@ -6,9 +6,11 @@ import (
 	"testing"
 )
 
+const targetTestVersion = 1
+
 var allergiesTests = []struct {
 	expected []string
-	input    int
+	input    uint
 }{
 	{[]string{}, 0},
 	{[]string{"eggs"}, 1},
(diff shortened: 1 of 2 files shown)


petertseng:
> The `indexOf` function is still using `int` but I don’t know if it’s necessary to change that to `uint`.

It is OK to leave it; it is not part of the external interface so it can do as it pleases.

Upon reviewing the implementation (which came long before my time)... I consider what it's doing a dirty trick. It's using the fact that `2 ** -1` = 0.5 which rounds to 0, and `i & x > 0` will never be true. Seems to take a bit of thinking to realize why it does what it does, but that is a different item than what we are dealing with in this PR. I believe this PR has done the right thing.

petertseng:
Thanks!

mkoppmann:
You’re right, that trick works but either needs to be documented or changed to be something more intuitive. But that’s in the scope of a new issue/PR.

Thanks for merging.

PR #274, opened 2016-04-07
binary: Add tests for invalid inputs

petertseng:
* alphabetic characters at beginning, middle, end
* invalid digits

Test version is added (was previously unversioned)

Tested by ensuring that a solution that never errors gets a test
failure:

```
--- FAIL: TestParseBinary (0.00s)
        binary_test.go:50: ParseBinary(foo101) returned 5 and no error.  Expected an error.
```

Tested by ensuring that a solution that always errors gets a test
failure:

```
--- FAIL: TestParseBinary (0.00s)
        binary_test.go:46: ParseBinary(1) returned error "hard drive on fire, cannot convert to binary right now".  Error not expected.
```

Closes #271

PR #379, opened 2016-10-02
test-without-stubs: run go fmt and go vet

petertseng:
These are standard tools and it's good to ensure that we are in keeping
with their guidance.

In particular, note that we had a long-standing formatting inconsistency
in the binary exercise's test suite, only recently fixed via
bd156a4b2958e13212480d9755b5bb50b3abd5be. I've since discovered that
that inconsistency was introduced by me, in #274.

In addition, #317 was a problem reported by `go vet`. It seems good to
automatically check for those to prevent any similar errors.

Here is what happens if `bob/example.go` has `yelling != nil` added:

```
example.go:22: comparison of function yelling != nil is always true
please follow the 'go vet' suggestions in exercises/bob
```

Here is what happens if `bob.go` has a few extra spaces added:

```
please run 'go fmt' on exercises/bob
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,11 +54,9 @@ before_script:
 # -cpu 2 is for our problems that involve concurrency to test that they
 # perform as expected on multiple cores.
 script:
-  - ./bin/test-without-stubs fmt
-  # go vet can't be run on Travis with 1.4.3
-  # this `go version | grep` can be removed once 1.4.3 is no longer tested.
-  - if ! go version | grep -q 1.4.3; then ./bin/test-without-stubs vet; fi
-  - GOARCH=$TESTARCH RACE=$RACE ./bin/test-without-stubs
+  - if [ -n "$STATIC_CHECKERS" ]; then ./bin/test-without-stubs fmt; fi
+  - if [ -n "$STATIC_CHECKERS" ]; then ./bin/test-without-stubs vet; fi
+  - if [ -z "$STATIC_CHECKERS" ]; then ./bin/test-without-stubs; fi
 
 # special cases for the build matrix are that go 1.4.3 can't be tested 32 bit
 # and that tip is allowed to fail.  Broken tips are extremely rare these days
@@ -67,6 +65,10 @@ matrix:
   exclude:
     - go: 1.4.3
       env: TESTARCH=386
+  include:
+    # Keep this in sync with the latest version.
+    - go: 1.7
+      env: STATIC_CHECKERS=1
   allow_failures:
     - go: tip
     - go: 1.4.3
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,10 @@ before_script:
 # -cpu 2 is for our problems that involve concurrency to test that they
 # perform as expected on multiple cores.
 script:
+  - ./bin/test-without-stubs fmt
+  # go vet can't be run on Travis with 1.4.3
+  # this `go version | grep` can be removed once 1.4.3 is no longer tested.
+  - if ! go version | grep -q 1.4.3; then ./bin/test-without-stubs vet; fi
   - GOARCH=$TESTARCH RACE=$RACE ./bin/test-without-stubs
 
 # special cases for the build matrix are that go 1.4.3 can't be tested 32 bit
--- a/bin/test-without-stubs
+++ b/bin/test-without-stubs
@@ -19,8 +19,13 @@ for dir in exercises/*; do
   fi
 done
 
+root=$(pwd)
+
 cleanup () {
   echo "moving stub files back"
+  # the cd for go fmt or go vet could leave us in an unexpected place.
+  # cd to a known place.
+  cd "$root"
 
   # could use `rename` here, but it's not uniform among platforms.
   # some renames take two arguments while others take a single regex.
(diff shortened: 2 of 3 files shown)


petertseng:
Oops, guess we can't run `go vet` with a different GOARCH. This will require a bit of messy business.

petertseng:
I saw two options:

1. Make test-without-stubs take an option to determine what it will do (fmt, vet, test). Have the travis runner pass GOARCH to test, but not to fmt or vet. Current solution.

2. Make test-without-stubs check its GOARCH and not run fmt and vet if GOARCH isn't amd64. But this is merely an artifact of our Travis setup, and it prevents anyone whose computer isn't amd64 from running test-without-stubs locally (without modification)

I'm sure there can be others, so if you know any, speak up.

One thing I don't like (but I don't hate either) is that go fmt and go vet are run for every entry in our build matrix, which probably isn't necessary. We can probably just have one entry that is just for fmt + vet, but is that necessary?

petertseng:
Final note: no golint in this PR, because we have a whole ton of golint output (most of it is uncommented exported items)

petertseng:
It looks like 1.4.3 can't use go vet. Might be time to cut it loose.

kytrinyx:
> we can't run `go vet` with a different GOARCH

I'm not sure I understand what you mean here. I thought `go vet` was available on all of the platforms.

kytrinyx:
> It looks like 1.4.3 can't use go vet. Might be time to cut it loose.

Yeah, I think it's totally fine to ditch 1.4.3.

kytrinyx:
> no golint in this PR [...]

Yeah, I've run golint against everything a number of times and fixed some things, but adding comments was going to be a bigger task. Would it be worth creating issues for it and tagging them with good first task and hacktoberfest?

kytrinyx:
> go fmt and go vet are run for every entry in our build matrix, which probably isn't necessary

Yeah, I don't feel to strongly about this one either, since the build is pretty quick. I haven't looked into how easy it would be to have something in the matrix that is just one set of tests. If it's not ridiculous I'd lean towards having a separate entry in the matrix for it.

petertseng:
>>  we can't run go vet with a different GOARCH
> 
> I'm not sure I understand what you mean here. I thought go vet was available on all of the platforms.

Yeah I should have mentioned - here's what happens, at https://travis-ci.org/exercism/xgo/jobs/164369569 : we get

```
vet: import failed: can't find import: fmt
vet: import failed: can't find import: fmt
```

petertseng:
> It looks like 1.4.3 can't use go vet. Might be time to cut it loose.

I would like to provide evidence of that claim: https://travis-ci.org/exercism/xgo/jobs/164373145

```
if go version | grep -q 1.4.3; then go get golang.org/x/tools/cmd/vet ; fi
package golang.org/x/tools/cmd/vet
	imports golang.org/x/tools/cmd/vet
	imports golang.org/x/tools/cmd/vet: cannot find package "golang.org/x/tools/cmd/vet" in any of:
	/home/travis/.gimme/versions/go1.4.3.linux.amd64/src/golang.org/x/tools/cmd/vet (from $GOROOT)
	/home/travis/gopath/src/golang.org/x/tools/cmd/vet (from $GOPATH)
```

petertseng:
> I've run golint against everything a number of times and fixed some things, but adding comments was going to be a bigger task. Would it be worth creating issues for it and tagging them with good first task and hacktoberfest?

Sometimes I feel unsure about those "should have comment or be unexported" lints, since most of the exported things are what is being asked for in the problem. I sometimes wonder if we should have travis CI filter those out and only show the others.

I put all the golint suggestions at https://gist.github.com/petertseng/1bbaaa1e2fce8a2ae86a16801712b903 so we can take a look.

petertseng:
> If it's not ridiculous I'd lean towards having a separate entry in the matrix for it.

I do plan to explore this and report here, so no need to consider merging this PR yet.

petertseng:
> If it's not ridiculous I'd lean towards having a separate entry in the matrix for it.

OK, this is the way I have achieved it. Unfortunately, it was still necessary to specify a Go version, so I have to add a comment to keep it synced with the latest version. The consequence for not doing so is https://travis-ci.org/exercism/xgo/jobs/164830757 : 

```
/home/travis/build.sh: line 339: export: `1.6.3,': not a valid identifier

/home/travis/build.sh: line 339: export: `1.5.4,': not a valid identifier

/home/travis/build.sh: line 339: export: `1.4.3,': not a valid identifier

/home/travis/build.sh: line 339: export: `tip]': not a valid identifier

12.06s$ GIMME_OUTPUT=$(gimme [1.7, "1.6.3", "1.5.4", "1.4.3", "tip"]) && eval "$GIMME_OUTPUT"

I don't have any idea what to do with '[1.7,'.

  (using type 'auto')

The command "GIMME_OUTPUT=$(gimme [1.7, "1.6.3", "1.5.4", "1.4.3", "tip"]) && eval "$GIMME_OUTPUT"" failed and exited with 1 during .
```

petertseng:
Just rebasing on master.

petertseng:
A minor annoyance about having vet/lint as a separate entry is that currently it is the *last* entry in the matrix since I am using `matrix: include:` for it. I'm not sure there's a way to fix that. And if it is the last entry, it gets run last, whereas I'd usually like to get feedback on vets and lints sooner than later.

Of course, I will run vet and lint locally on my machine before I submit any PRs, so this will not really be an issue for me locally. So if it's still worth it to keep them as a separate entry, then full speed ahead.

petertseng:
I think comments are addressed, it is useful to have ths in the face of incoming contributions, and we can always revert "travis: Use a separate entry for static checkers" if we don't like the separate entry.

PR #380, opened 2016-10-02
Kindergarten garden improve err handling

mchoube:
1) Added test versioning.
2) Improved error handling in TestGarden().

I don't think other tests need error type check.
Please review.

### DIFF
--- a/exercises/kindergarten-garden/kindergarten_garden_test.go
+++ b/exercises/kindergarten-garden/kindergarten_garden_test.go
@@ -128,15 +128,7 @@ RVGCCGCV`, append([]string{}, test6names...), true, []lookup{
 func TestGarden(t *testing.T) {
 	for _, test := range tests {
 		g, err := NewGarden(test.diagram, test.children)
-		if !test.ok { // negative tests; expecting error
-			// check err is of error type
-			var _ error = err
-
-			// we expect error
-			if err == nil {
-				t.Fatalf("NewGarden test %d. Expected error but got nil.", test.number)
-			}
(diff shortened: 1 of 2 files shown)


petertseng:
Thanks. Give me some time to review. While waiting, can you amend the commit message to mention the issue that will be closed? So that I don't forget to close it.

I'll also mention I prefer present-tense commit messages "Add" and "Improve" rather than "Added" and "Improved" but this is not a hard requirement for this repo

petertseng:
> I don't think other tests need error type check.

I agree. It only needs to be checked once, so what is here in this PR is sufficient.

exercises/kindergarten-garden/kindergarten_garden_test.go
```diff
@@ -120,24 +128,29 @@ RVGCCGCV`, append([]string{}, test6names...), true, []lookup{
 func TestGarden(t *testing.T) {
 	for _, test := range tests {
 		g, err := NewGarden(test.diagram, test.children)
-		switch {
-		case err != nil:
-			if test.ok {
+		if !test.ok { // negative tests; expecting error
```
kytrinyx: By starting with the negative case, we end up with a double negative in the else. I have more trouble mentally thinking through the logic when there's a double negative. What do you think of swapping to start with `if test.ok`?

kytrinyx:
I think this looks good now—the only question I have is about the double negative in the conditional.

kytrinyx:
I like how this turned out. Thanks so much!

Issue #282, opened 2016-04-09
beer-song: Improve error-handling behavior


petertseng:
As part of #275, we'd like to more closely follow idiomatic Go behavior in the face of errors.

If the beer-song exercise does not have error handling, then this ticket can be closed immediately.

If it does, then follow the below steps:

#### Ensure `err` return value is of type `error`

It's idiomatic for the error return value to be of type `error`.

Most tests simply check whether `err == nil` or not. If we don't verify the type of `err`, it could be any type that has nils.

To lean on the type system, add the following line after assigning err:

```go
var _ error = err
```

See [the hamming tests](https://github.com/exercism/xgo/blob/master/exercises/hamming/hamming_test.go#L14) for an example.

#### Ignore the normal return value when an error expected

The contract in Go is "if there's an error, assume nothing about the normal return value". Thus, we should only check the non-error return value when there is no error.

In some exercises we found that we were checking the normal return value, even when there was an error.

See [the binary tests](https://github.com/exercism/xgo/blob/master/exercises/binary/binary_test.go#L41-L58) for an example of the behavior that we want.

#### Increment the test version

These changes break backwards compatibility. Bump the test version to let reviewers know that the exercise has changed.

This requires changing two constants:

- `targetTestVersion` in the test file (`*_test.go`)

- `testVersion` in the reference solution (`example.go`)

If the exercise does not have a test version, then add one, setting the value to 1.

See [the version test in leap](https://github.com/exercism/xgo/blob/master/exercises/leap/leap_test.go#L13-L15) for an example.

petertseng:
https://github.com/exercism/xgo/pull/368

petertseng:
#368 handled "Ensure `err` return value is of type `error`" but not "Ignore the normal return value when an error expected"

andrerfcsantos:
@petertseng We ran into the `var _ error = err` when doing some updates to the tests for `largest-series-product` and are a bit confused by this bit of code and what is doing. From #381 looks like this was introduced as a compile-time check for the type of `err`, but I'd argue this is not needed. Since `err` is assigned as the second return of functions that return an `error` as the second return value, so even the function returns `nil` as the second value, it'll be a `nil` of type `error`, e.g [this program](https://go.dev/play/p/zT0qaFooDB_0) fails to compile.

> If we don't verify the type of err, it could be any type that has nils.

Can you provide an example of this?

(Sorry to post on an very old closed issue, but I don't know a better place to talk about this)

[...]

PR #381, opened 2016-10-02
Issue #294: largest-series-product: improv error handling.

mchoube:
Improved error handling for largest series product test. Please review.

### DIFF
--- a/exercises/largest-series-product/largest_series_product_test.go
+++ b/exercises/largest-series-product/largest_series_product_test.go
@@ -2,27 +2,39 @@ package lsproduct
 
 import "testing"
 
-const targetTestVersion = 3
+const targetTestVersion = 4
 
-func TestLargestSeriesProduct(t *testing.T) {
+func TestTestVersion(t *testing.T) {
 	if testVersion != targetTestVersion {
 		t.Fatalf("Found testVersion = %v, want %v", testVersion, targetTestVersion)
 	}
+}
+
+func TestLargestSeriesProduct(t *testing.T) {
 	for _, test := range tests {
 		p, err := LargestSeriesProduct(test.digits, test.span)
-		switch {
-		case err != nil:
-			if test.ok {
+		if !test.ok { // expect error
+			// check if err is of error type
+			var _ error = err
+
+			// we expect error
+			if err == nil {
+				t.Fatalf("LargestSeriesProduct(%s, %d) = %d, %v."+
+					"  Expected error got nil",
+					test.digits, test.span, p, err)
+			}
+		} else {
+			// we do not expect error
+			if err != nil {
 				t.Fatalf("LargestSeriesProduct(%s, %d) returned error %q.  "+
 					"Error not expected.",
 					test.digits, test.span, err)
 			}
-		case !test.ok:
-			t.Fatalf("LargestSeriesProduct(%s, %d) = %d, %v.  Expected error",
-				test.digits, test.span, p, err)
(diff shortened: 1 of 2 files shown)


exercises/largest-series-product/largest_series_product_test.go
```diff
@@ -2,27 +2,39 @@ package lsproduct
...
+
+func TestLargestSeriesProduct(t *testing.T) {
 	for _, test := range tests {
 		p, err := LargestSeriesProduct(test.digits, test.span)
-		switch {
-		case err != nil:
-			if test.ok {
+		if !test.ok { // expect error
+			// check if err is of error type
+			var _ error = err
```
kytrinyx: Since we only need to do this once for each error, should we do it before we enter the switch statement?

exercises/largest-series-product/largest_series_product_test.go
```diff
@@ -2,27 +2,39 @@ package lsproduct
...
 	}
+}
+
+func TestLargestSeriesProduct(t *testing.T) {
 	for _, test := range tests {
 		p, err := LargestSeriesProduct(test.digits, test.span)
-		switch {
-		case err != nil:
-			if test.ok {
+		if !test.ok { // expect error
```
kytrinyx: This creates a double negative. How about inverting the conditional so we test `if test.ok {`?

exercises/largest-series-product/largest_series_product_test.go
mchoube: but then it wouldn't fit with "Ignore the normal return value when an error expected" of #282

kytrinyx: s/normal error/normal value/ -- sorry about that!

mchoube: Please check comment from Peter in PR https://github.com/exercism/xgo/pull/368

exercises/largest-series-product/largest_series_product_test.go
mchoube: this is fixed in the new commit.

exercises/largest-series-product/largest_series_product_test.go
mchoube: err should be checked only when error is expected. If the test case is for good path(non-negative) then we are interested in 'p'(product) only. The code change ensures this.

petertseng: > If the test case is for good path(non-negative) then we are interested in 'p'(product) only.

Yup.

> err should be checked only when error is expected

It's true that whether `err` is nil or not should only be checked when an error is expected.

We are free to check whether `err` is of type `error` anywhere though. The argument here is that maybe we should have the `var _ error = err` line earlier so that it is clearer to the student that the `err` needs to be of type `error`, instead of farther down in the file. what do you think about this?

kytrinyx: @petertseng's commend is about `actualVerse`, not about the error.

kytrinyx: To clarify, the type check `var _ error = err` is verified at compile time, and it doesn't matter where in the file it is, as long as `err` is defined in that scope. My thought was that putting it closer to the place where `err` is originally defined would make this easier to read. As it is it doesn't change the behavior by having it here. I'll go ahead and merge this.

kytrinyx:
Thank you!

PR #383, opened 2016-10-02
Acronym: Add acronym excercise

Akasurde:
This example adds excercise for converting a long name into
three letter acronyms

Fixes https://github.com/exercism/xgo/issues/382

### DIFF
--- a/exercises/acronym/acronym_test.go
+++ b/exercises/acronym/acronym_test.go
@@ -1,0 +1,36 @@
+package acronym
+
+import (
+	"testing"
+)
+
+const targetTestVersion = 1
+
+type testCase struct {
+	input    string
+	expected string
+}
+
+var stringTestCases = []testCase{
+	{"Portable Network Graphics", "PNG"},
+	{"HyperText Markup Language", "HTML"},
+	{"Ruby on Rails", "ROR"},
+	{"PHP: Hypertext Preprocessor", "PHP"},
+	{"First In, First Out", "FIFO"},
+	{"Complementary metal-oxide semiconductor", "CMOS"},
+}
+
+func TestTestVersion(t *testing.T) {
+	if testVersion != targetTestVersion {
+		t.Errorf("Found testVersion = %v, want %v.", testVersion, targetTestVersion)
+	}
+}
+
+func TestAcronym(t *testing.T) {
+	for _, test := range stringTestCases {
+		actual := abbreviate(test.input)
+		if actual != test.expected {
+			t.Errorf("Acronym test [%s], expected [%s], actual [%s]", test.input, test.expected, actual)
+		}
+	}
+}
--- a/exercises/acronym/example.go
+++ b/exercises/acronym/example.go
@@ -1,0 +1,21 @@
+package acronym
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+const testVersion = 1
+
+func abbreviate(s string) string {
+	regex := regexp.MustCompile("[A-Z]+[a-z]*|[a-z]+")
+	words := regex.FindAllString(s, -1)
+	abbr := []string{}
+
+	for _, word := range words {
+		abbr = append(abbr, string(word[0]))
+	}
+
+	return fmt.Sprintf("%s", strings.ToUpper(strings.Join(abbr, "")))
+}
--- a/config.json
+++ b/config.json
@@ -10,6 +10,7 @@
     "gigasecond",
     "hamming",
     "raindrops",
+    "acronym",
     "bob",
     "triangle",
     "difference-of-squares",


kytrinyx:
Thank you for this! :heart:

The build is failing with the following:

```
$ bin/configlet .
Evaluating .
-> config.json does not include [acronym].
```

The `configlet` tool does track-level linting (there's a summary of it's purpose in [this documentation about reviewing pull requests](https://github.com/exercism/exercism.io/blob/master/docs/reviewing-a-pull-request.md#consider-the-structural-requirements).

The quick fix is to add `"acronym"` to the `"problems"` array in the `config.json` file which lives in the root of the repository. That array is ordered very roughly by the difficulty of the exercise, so `acronym` should probably be somewhere in the second quarter. It's not too important to get the placement exactly right—we often end up tweaking the order when we get some feedback from people about how difficult it was to solve it.

Something different (and much more minor) is the commit message. It has `acronym` in it which is great—this lets us search on it. It also has the imperative "Add X" (rather than "Adding X") which we also appreciate. The two things I would tweak are:

- it doesn't need to say "go", since everything in this repository is go
- it's an exercise, not an example

The last thing is actually a bit important because each exercise has an example, and so this could be pretty confusing later.

exercises/acronym/acronym_test.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+}
+
+func TestAcronym(t *testing.T) {
+	for _, test := range stringTestCases {
+		actual, actualErr := abbreviate(test.input)
+		if !test.expectErr {
+			if actualErr != nil {
+				// if we don't expected error but got error
+				var _ error = actualErr
```
kytrinyx: After looking at a few of these error handling pull requests I've started to think we should always put this right after the place where the error gets assigned. In this case, right after line 31. @petertseng do you see any issue with that?

exercises/acronym/acronym_test.go
```diff
@@ -0,0 +1,47 @@
...
+				t.Errorf("Acronym test [%s] returned error [%q], Error not expected", test.input, actualErr)
+			}
+			if actual != test.expected {
+				t.Errorf("Acronym test [%s], expected [%s], actual [%s]", test.input, test.expected, actual)
+			}
+		} else if actualErr == nil {
+			//if we expected error, but got nil instead
```
kytrinyx: This comment restates the error message right below it. It can be deleted.

exercises/acronym/example.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+const testVersion = 1
+
+func abbreviate(s string) (string, error) {
+	regex, _ := regexp.Compile("[A-Z]+[a-z]*|[a-z]+")
+	wordList := regex.FindAllString(s, -1)
+	abbr := []string{}
+
+	for _, word := range wordList {
```
kytrinyx: Instead of `wordList`, let's use `words`. Then the loop reads _for singular in range of plural_.

exercises/acronym/acronym_test.go
petertseng: Seems good to me. In fd8572684c4138fda1b2d583bdaea09dec2051cb it may have been deceptive since it was under `case err != nil`, but that's because I couldn't have placed it before the `case`

petertseng: Edit to clarify: Putting it as close as possible seems good to me.

exercises/acronym/acronym_test.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"testing"
+)
+
+const targetTestVersion = 1
+
+type testCase struct {
+	input     string
+	expected  string
+	expectErr bool
```
Akasurde: ok

exercises/acronym/acronym_test.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+var stringTestCases = []testCase{
+	{"Portable Network Graphics", "PNG", false},
+	{"HyperText Markup Language", "HTML", false},
+	{"Ruby on Rails", "ROR", false},
+	{"PHP: Hypertext Preprocessor", "PHP", false},
+	{"First In, First Out", "FIFO", false},
+	{"Complementary metal-oxide semiconductor", "CMOS", false},
```
Akasurde: I could not think any negative testcase at time of writing so kept it for future development.

exercises/acronym/acronym_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+func TestTestVersion(t *testing.T) {
+	if testVersion != targetTestVersion {
+		t.Errorf("Found testVersion = %v, want %v.", testVersion, targetTestVersion)
+	}
+}
+
+func TestAcronym(t *testing.T) {
+	for _, test := range stringTestCases {
+		actual, actualErr := abbreviate(test.input)
```
Akasurde: ok. I will change it to `err`

exercises/acronym/example.go
Akasurde: Sure. I will change `wordList` to `words`

exercises/acronym/acronym_test.go
Akasurde: Done.

exercises/acronym/example.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+const testVersion = 1
+
+func abbreviate(s string) (string, error) {
+	regex, _ := regexp.Compile("[A-Z]+[a-z]*|[a-z]+")
```
petertseng: I suggest [`MustCompile`](https://golang.org/pkg/regexp/#MustCompile) rather than `Compile` and ignore error to signal intent that you do expect this to compile.

exercises/acronym/acronym_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+import (
+	"testing"
+)
+
+const targetTestVersion = 1
+
+type testCase struct {
+	input     string
+	expected  string
```
petertseng: `go fmt` will want these to be aligned with the below

exercises/acronym/acronym_test.go
petertseng: I would probably vote for removing it.

exercises/acronym/acronym_test.go
Akasurde: ok

exercises/acronym/example.go
Akasurde: OK. I will change this to `MustCompile`

exercises/acronym/acronym_test.go
Akasurde: @petertseng I am not following here. Could you please elaborate more ?

petertseng: I would suggest that we remove the `expectsErr` from `testCase`, remove the `, false` from all cases, and make `abbreviate` return just a `string` rather than `(string, error)`. What say you?

What I am thinking is - does the act of forming an acronym perform any operation that could possibly fail? We can look to [`strings`](https://golang.org/pkg/strings/) as guidance - it has no error on [ToUpper](https://golang.org/pkg/strings/#ToUpper) (or indeed any of its functions except those on `Reader` and `Replacer`)

petertseng:
I saw that now there is no `error` expected - I may have missed some of the actoin due to the recent Github outage. I'll take a look now.

petertseng:
Everything looks good to me, so there we have it! The Go track's 73rd exercise.

Akasurde:
@kytrinyx @petertseng Thanks for your valuable comments. I learned a lot of things from this PR.

PR #396, opened 2016-10-06
Pangram: Add Pangram exercise

Akasurde:
This PR adds exercise for determining if a sentence
is a pangram or not.

Fixes https://github.com/exercism/xgo/issues/395

### DIFF
--- a/exercises/pangram/example.go
+++ b/exercises/pangram/example.go
@@ -6,7 +6,7 @@ import (
 
 const testVersion = 1
 
-func isPangram(s string) bool {
+func IsPangram(s string) bool {
 	lowerString := strings.ToLower(s)
 	var check [26]bool
 	var count int
--- a/exercises/pangram/example.go
+++ b/exercises/pangram/example.go
@@ -1,0 +1,32 @@
+package pangram
+
+import (
+	"strings"
+)
+
+const testVersion = 1
+
+func isPangram(s string) bool {
+	lowerString := strings.ToLower(s)
+	var check [26]bool
+	var count int
+	for _, c := range lowerString {
+		if c >= 'a' {
+			if c > 'z' {
+				continue
+			}
+			c -= 97
+		} else {
+			continue
+		}
+
+		if !check[c] {
+			if count == 25 {
+				return true
+			}
+			check[c] = true
+			count++
+		}
+	}
+	return false
+}
--- a/exercises/pangram/pangram_test.go
+++ b/exercises/pangram/pangram_test.go
@@ -32,7 +32,7 @@ func TestTestVersion(t *testing.T) {
 
 func TestPangram(t *testing.T) {
 	for _, test := range testCases {
-		actual := isPangram(test.input)
+		actual := IsPangram(test.input)
 		if actual != test.expected {
 			t.Errorf("Pangram test [%s], expected [%s], actual [%s]", test.input, test.expected, actual)
 			if !test.expected {
--- a/exercises/pangram/pangram_test.go
+++ b/exercises/pangram/pangram_test.go
@@ -1,7 +1,6 @@
 package pangram
 
 import (
-	"fmt"
 	"testing"
 )
 
@@ -36,9 +35,9 @@ func TestPangram(t *testing.T) {
 		actual := isPangram(test.input)
 		if actual != test.expected {
 			t.Errorf("Pangram test [%s], expected [%s], actual [%s]", test.input, test.expected, actual)
-		}
-		if !test.expected {
-			fmt.Printf("Pangram test [%s], failure reason : %s\n", test.input, test.failureReason)
+			if !test.expected {
+				t.Logf("[%s] should not be a pangram because : %s\n", test.input, test.failureReason)
+			}
 		}
 	}
 }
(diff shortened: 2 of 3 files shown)


exercises/pangram/example.go
```diff
@@ -0,0 +1,32 @@
+package pangram
+
+import (
+	"strings"
+)
+
+const testVersion = 1
+
+func is_pangram(s string) bool {
```
kytrinyx: Go uses MixedCaps for names: https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps

exercises/pangram/pangram_test.go
```diff
@@ -0,0 +1,35 @@
...
+	input    string
+	expected bool
+}
+
+var stringTestCases = []testCase{
+	{"quick movement of the enemy will jeopardize five gunboats", false},
+	{"the quick brown fish jumps over the lazy dog", false},
+	{"7h3 qu1ck brown fox jumps ov3r 7h3 lazy dog", false},
+	{"the 1 quick brown fox jumps over the 2 lazy dogs", true},
+	{"The quick brown fox jumps over the lazy dog", true},
```
kytrinyx: We have a canonical set of test data that we encourage all of the tracks to use: https://github.com/exercism/x-common/blob/master/exercises/pangram/canonical-data.json

If you see any edge cases that you've included and the canonical json is missing, please open an issue in https://github.com/exercism/x-common/issues so that we can discuss it.

exercises/pangram/pangram_test.go
```diff
@@ -0,0 +1,35 @@
...
+	if testVersion != targetTestVersion {
+		t.Errorf("Found testVersion = %v, want %v.", testVersion, targetTestVersion)
+	}
+}
+
+func TestPangram(t *testing.T) {
+	for _, test := range stringTestCases {
+		actual := is_pangram(test.input)
+		if actual != test.expected {
+			t.Errorf("Pangram test [%s], expected [%s], actual [%s]", test.input, test.expected, actual)
```
kytrinyx: One of the things we try to do is help people understand _why_ a test fails. How about adding a _description_ field to the test case struct that explains the context of the failure (e.g. "missing character $CHAR")

exercises/pangram/pangram_test.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+const targetTestVersion = 1
+
+type testCase struct {
+	input    string
+	expected bool
+}
+
+var stringTestCases = []testCase{
```
kytrinyx: The `string` prefix is unnecessary here. I'd suggest dropping it.

exercises/pangram/example.go
Akasurde: Ok, I will `is_pangram` to `isPangram`

exercises/pangram/pangram_test.go
Akasurde: Ok, I will change `stringTestCases` to `testCases`

exercises/pangram/pangram_test.go
Akasurde: OK, I will change dataset to link mentioned.

exercises/pangram/pangram_test.go
Akasurde: I am adding `failureReason` in struct and printing the error reason if test is expected to fail.

exercises/pangram/pangram_test.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func TestPangram(t *testing.T) {
+	for _, test := range testCases {
+		actual := isPangram(test.input)
+		if actual != test.expected {
+			t.Errorf("Pangram test [%s], expected [%s], actual [%s]", test.input, test.expected, actual)
+		}
+		if !test.expected {
+			fmt.Printf("Pangram test [%s], failure reason : %s\n", test.input, test.failureReason)
```
Akasurde: Not using `t.Logf` as it will require `-v` flag in `go test`. Let me know if it needs to be changed to `t.Logf`.

petertseng: Can you clarify what you intend? Now if the `isPangram` returns `false` when it should return `false`, this will print a `failure` reason even though the test has not failed, is that OK? Seems like it might be confusing to the student. Should this go inside the `if actual != test.expected` instead?

I also believe `Logf` messages *are* shown if tests fail, so it would have been OK to use `Logf`.

exercises/pangram/pangram_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+func TestTestVersion(t *testing.T) {
+	if testVersion != targetTestVersion {
+		t.Errorf("Found testVersion = %v, want %v.", testVersion, targetTestVersion)
+	}
+}
+
+func TestPangram(t *testing.T) {
+	for _, test := range testCases {
+		actual := isPangram(test.input)
```
alebaffa: I checked for at least 15 exercises in xgo and they are all exported. I am pretty sure they all are. I think we should continue on this line.

petertseng:
Inactivity: Rebased for config.json, and adding commits to address review comments. Will squash + merge shortly.

petertseng:
Thanks @Akasurde !

PR #398, opened 2016-10-08
Twelve-days: Add exercise for Twelve Days

Akasurde:
Exercise that outputs the lyrics to 'The Twelve Days of Christmas'

Fixes https://github.com/exercism/xgo/issues/397

### DIFF
--- a/config.json
+++ b/config.json
@@ -73,6 +73,11 @@
     },
     {
       "difficulty": 1,
+      "slug": "twelve-days",
+      "topics": []
+    },
+    {
+      "difficulty": 1,
       "slug": "difference-of-squares",
       "topics": []
     },
--- a/exercises/twelve-days/example.go
+++ b/exercises/twelve-days/example.go
@@ -1,0 +1,57 @@
+package twelve
+
+import (
+	"fmt"
+)
+
+const testVersion = 1
+
+var verse = map[string]string{
+	"first":    "a Partridge in a Pear Tree.",
+	"twelfth":  "twelve Drummers Drumming",
+	"eleventh": "eleven Pipers Piping",
+	"tenth":    "ten Lords-a-Leaping",
+	"ninth":    "nine Ladies Dancing",
+	"eighth":   "eight Maids-a-Milking",
+	"seventh":  "seven Swans-a-Swimming",
+	"sixth":    "six Geese-a-Laying",
+	"fifth":    "five Gold Rings",
+	"fourth":   "four Calling Birds",
+	"third":    "three French Hens",
+	"second":   "two Turtle Doves",
+}
+
+var wording = map[int]string{
+	1:  "first",
+	2:  "second",
+	3:  "third",
+	4:  "fourth",
+	5:  "fifth",
+	6:  "sixth",
+	7:  "seventh",
+	8:  "eighth",
+	9:  "ninth",
+	10: "tenth",
+	11: "eleventh",
+	12: "twelfth",
+}
+
+func verses(i int) string {
+	var gifts = ""
+	for from := i; from > 0; from-- {
+		if i != 1 && from == 1 {
+			gifts = fmt.Sprintf("%s, and %s", gifts, verse[wording[from]])
+		} else {
+			gifts = fmt.Sprintf("%s, %s", gifts, verse[wording[from]])
+		}
+	}
+	return fmt.Sprintf("On the %s day of Christmas my true love gave to me%s", wording[i], gifts)
+}
+
+func sing() string {
+	var song = ""
+	for i := 1; i <= 12; i++ {
+		song += verses(i) + "\n"
+	}
+	return song
+}
--- a/exercises/twelve-days/example.go
+++ b/exercises/twelve-days/example.go
@@ -36,7 +36,7 @@ var wording = map[int]string{
 	12: "twelfth",
 }
 
-func verses(i int) string {
+func Verse(i int) string {
 	var gifts = ""
 	for from := i; from > 0; from-- {
 		if i != 1 && from == 1 {
@@ -48,10 +48,10 @@ func verses(i int) string {
 	return fmt.Sprintf("On the %s day of Christmas my true love gave to me%s", wording[i], gifts)
 }
 
-func sing() string {
+func Song() string {
 	var song = ""
 	for i := 1; i <= 12; i++ {
-		song += verses(i) + "\n"
+		song += Verse(i) + "\n"
 	}
 	return song
 }
--- a/exercises/twelve-days/twelve_days_test.go
+++ b/exercises/twelve-days/twelve_days_test.go
@@ -1,0 +1,52 @@
+package twelve
+
+import (
+	"testing"
+)
+
+const targetTestVersion = 1
+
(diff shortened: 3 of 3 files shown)


exercises/twelve-days/example.go
```diff
@@ -0,0 +1,58 @@
...
+	8:  "eighth",
+	9:  "ninth",
+	10: "tenth",
+	11: "eleventh",
+	12: "twelfth",
+}
+
+func verses(i int) string {
+	var gifts = ""
+	for frm := i; frm > 0; frm-- {
```
kytrinyx: What is a `frm`?

exercises/twelve-days/twelve_days_test.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+const targetTestVersion = 1
+
+type testCase struct {
+	input    int
+	expected string
+}
+
+var singTestCases = []testCase{
```
kytrinyx: We only have one set of test cases, so the variable can be `testCases` rather than `singTestCases`.

exercises/twelve-days/twelve_days_test.go
```diff
@@ -0,0 +1,53 @@
...
+	{10, "On the tenth day of Christmas my true love gave to me, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."},
+	{11, "On the eleventh day of Christmas my true love gave to me, eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."},
+	{12, "On the twelfth day of Christmas my true love gave to me, twelve Drummers Drumming, eleven Pipers Piping, ten Lords-a-Leaping, nine Ladies Dancing, eight Maids-a-Milking, seven Swans-a-Swimming, six Geese-a-Laying, five Gold Rings, four Calling Birds, three French Hens, two Turtle Doves, and a Partridge in a Pear Tree."},
+}
+
+func TestTestVersion(t *testing.T) {
+	if testVersion != targetTestVersion {
+		t.Errorf("Found testVersion = %v, want %v.", testVersion, targetTestVersion)
+	}
+}
```
kytrinyx: Let's put this test at the very end of the file. The tests will execute in parallel, but it would be nice if this isn't the first thing you see when you read the tests since it's book-keeping.

exercises/twelve-days/twelve_days_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+func TestSing(t *testing.T) {
+	var song = ""
+	for i := 1; i <= 12; i++ {
+		song = fmt.Sprintf("%s\n", verses(i))
```
kytrinyx: This is overwriting `song` 12 times, right?

kytrinyx: `verses` should probably be exported.

exercises/twelve-days/twelve_days_test.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	if song != sing() {
+		t.Errorf("Twelve Days test, error occured in sing test")
```
kytrinyx: This says that there's an error, but there's no error handling in this test. Would you rephrase this to clarify why the test is failing?

exercises/twelve-days/twelve_days_test.go
```diff
@@ -0,0 +1,53 @@
...
+		t.Errorf("Found testVersion = %v, want %v.", testVersion, targetTestVersion)
+	}
+}
+
+func TestSing(t *testing.T) {
+	var song = ""
+	for i := 1; i <= 12; i++ {
+		song = fmt.Sprintf("%s\n", verses(i))
+	}
+	if song != sing() {
```
kytrinyx: `sing()` should probably be exported.

exercises/twelve-days/twelve_days_test.go
```diff
@@ -0,0 +1,53 @@
...
+		t.Errorf("Twelve Days test, error occured in sing test")
+	}
+}
+
+func TestVerse(t *testing.T) {
+	for testCnt, test := range singTestCases {
+		actual := verses(testCnt + 1)
```
kytrinyx: We've got `test.input` which is the correct value. We should use that here instead of `testCnt`.

exercises/twelve-days/twelve_days_test.go
Akasurde: OK. I will change it to `testCases`

exercises/twelve-days/twelve_days_test.go
```diff
@@ -0,0 +1,53 @@
...
+	for i := 1; i <= 12; i++ {
+		song = fmt.Sprintf("%s\n", verses(i))
+	}
+	if song != sing() {
+		t.Errorf("Twelve Days test, error occured in sing test")
+	}
+}
+
+func TestVerse(t *testing.T) {
+	for testCnt, test := range singTestCases {
```
Akasurde: OK.

exercises/twelve-days/example.go
Akasurde: I know, it is bad mnemonic. I converted from to frm, my bad.

config.json
```diff
@@ -13,6 +13,7 @@
     "acronym",
     "bob",
     "triangle",
+    "twelve-days",
```
Akasurde: Could you please explain this more ?

exercises/twelve-days/twelve_days_test.go
Akasurde: OK

exercises/twelve-days/twelve_days_test.go
Akasurde: @kytrinyx  I am not following you here. Can you explain more ?

exercises/twelve-days/twelve_days_test.go
Akasurde: what do you suggests here ?

petertseng:
My only questions are:

* whether we are sure we want to leave `sing` and `verses` unexported.
* should it be `verses(i int)` where the function name is plural but it only takes a single argument? It seems to make more sense to be called `verse` or `Verse`. depending on whether we decide to export

alebaffa:
In the exercise [food-chain](https://github.com/exercism/xgo/blob/master/exercises/food-chain/example.go) it's Verse (singular and exported). 
About exported functions or not: as I said in PR #396 , all the other exercises in xgo are exported. So, I think we should continue on this line.

petertseng:
Because of inactivity: Rebasing for config.json

Rename sing -> Song and verses -> Verse

The food-chain move is going into a separate commit as well, I think.

petertseng:
Happy late Hacktoberfest to @Akasurde , and thanks for the work.

PR #399, opened 2016-10-08
Improve nucleotide count test with better error handling

### LINKED ISSUE #298: nucleotide-count: Improve error-handling behavior

If the nucleotide-count exercise does not have error handling, then this ticket can be closed immediately.

``` go
var _ error = err
```

georgeteo:
Fixes #298

### DIFF
--- a/exercises/nucleotide-count/nucleotide_count_test.go
+++ b/exercises/nucleotide-count/nucleotide_count_test.go
@@ -22,6 +22,22 @@ func (h Histogram) sameMappings(o Histogram) (res bool) {
 	return
 }
 
+func assertError(count, expected int, err error, t *testing.T) {
+	var _ error = err
+	if err == nil {
+		t.Fatalf("Got \"%v\", expected \"%v\". error is nil", count, expected)
+	}
+}
+
+func assertResult(count, expected int, err error, t *testing.T) {
+	if err != nil {
+		t.Fatalf("Raised error %v when expecting none", err)
+	}
+	if count != expected {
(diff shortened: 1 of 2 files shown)


exercises/nucleotide-count/nucleotide_count_test.go
```diff
@@ -20,6 +22,22 @@ func (h Histogram) sameMappings(o Histogram) (res bool) {
 	return
 }
 
+func assertError(count, expected int, err error, t *testing.T) {
+	var _ error = err
```
kytrinyx: Since you've defined a function that takes an `error` type, this line is unnecessary.

That said, I'd rather not have the function, since it's used only once, and it's hard to guess what the order of the arguments are.

exercises/nucleotide-count/nucleotide_count_test.go
```diff
@@ -20,6 +22,22 @@ func (h Histogram) sameMappings(o Histogram) (res bool) {
...
+	}
+}
+
+func assertResult(count, expected int, err error, t *testing.T) {
+	if err != nil {
+		t.Fatalf("Raised error %v when expecting none", err)
+	}
+	if count != expected {
+		t.Fatalf("Got \"%v\", expected \"%v\".", count, expected)
+	}
```
kytrinyx: This is only used twice, and again it's hard to figure out from reading the function call what the arguments mean. I'd prefer to inline it.

petertseng:
Thanks for the work. In the meantime, we had decided to take #407 in light of the requested changes on this PR. But please don't hesitate to make future contributions!

PR #401, opened 2016-10-10
Protein-translation: Add exercise for Protein Translation

Akasurde:
Exercise that translate RNA sequences into proteins

Fixes https://github.com/exercism/xgo/issues/400

### DIFF
--- a/exercises/protein-translation/example.go
+++ b/exercises/protein-translation/example.go
@@ -1,0 +1,65 @@
+package protein
+
+const testVersion = 1
+
+func proteinLookUp(codon string) string {
+	switch codon {
+	case
+		"AUG":
+		return "Methionine"
+	case
+		"UUU",
+		"UUC":
+		return "Phenylalanine"
+	case
+		"UUA",
+		"UUG":
+		return "Leucine"
+	case
+		"UCU",
+		"UCC",
+		"UCA",
+		"UCG":
+		return "Serine"
+	case
+		"UAU",
+		"UAC":
+		return "Tyrosine"
+	case
+		"UGU",
+		"UGC":
+		return "Cysteine"
+	case
+		"UGG":
+		return "Tryptophan"
+	case
+		"UAA",
+		"UAG",
+		"UGA":
+		return "STOP"
+	}
+	return "STOP"
+}
+
+func toProtein(s string) string {
+	listCodon := []rune(s)
+	var res = ""
+	var tempCodon = ""
+	var proteinName = ""
+	for index, codon := range listCodon {
+		res += string(codon)
+		if index > 0 && (index+1)%3 == 0 {
+			tempCodon = proteinLookUp(res)
+			if tempCodon == "STOP" {
+				break
+			}
+			if proteinName == "" {
+				proteinName += tempCodon
+			} else {
+				proteinName += " " + tempCodon
+			}
+			res = ""
+		}
+	}
+	return proteinName
+}
--- a/exercises/protein-translation/example.go
+++ b/exercises/protein-translation/example.go
@@ -41,11 +41,11 @@ func FromCodon(codon string) string {
 	return "STOP"
 }
 
-func FromRNA(s string) string {
+func FromRNA(s string) []string {
 	listCodon := []rune(s)
 	var res = ""
 	var tempCodon = ""
-	var proteinName = ""
+	var proteins []string
 	for index, codon := range listCodon {
 		res += string(codon)
 		if index > 0 && (index+1)%3 == 0 {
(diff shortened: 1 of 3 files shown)


exercises/protein-translation/protein_translation_test.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+var proteinTestCases = []testCase{
+	{"AUGUUUUCUUAAAUG", "Methionine Phenylalanine Serine"},
+	{"AUGUUUUGG", "Methionine Phenylalanine Tryptophan"},
+	{"AUGUUUUAA", "Methionine Phenylalanine"},
+	{"UGGUGUUAUUAAUGGUUU", "Tryptophan Cysteine Tyrosine"},
+}
+
+func TestTestVersion(t *testing.T) {
```
kytrinyx: Let's stick this at the bottom so people don't have to read it before getting to the real tests.

exercises/protein-translation/example.go
```diff
@@ -0,0 +1,65 @@
...
+	case
+		"UGG":
+		return "Tryptophan"
+	case
+		"UAA",
+		"UAG",
+		"UGA":
+		return "STOP"
+	}
+	return "STOP"
```
petertseng: So I guess we haven't specified what should happen when the sequence is one not in the table, such as `AAA`, whether it should be an `error`. I'm OK with that, we can revisit if students ask for it.

exercises/protein-translation/protein_translation_test.go
```diff
@@ -0,0 +1,63 @@
...
+var proteinTestCases = []testCase{
+	{"AUGUUUUCUUAAAUG", "Methionine Phenylalanine Serine"},
+	{"AUGUUUUGG", "Methionine Phenylalanine Tryptophan"},
+	{"AUGUUUUAA", "Methionine Phenylalanine"},
+	{"UGGUGUUAUUAAUGGUUU", "Tryptophan Cysteine Tyrosine"},
+}
+
+func TestCodon(t *testing.T) {
+	for _, test := range codonTestCases {
+		actual := proteinLookUp(test.input)
```
petertseng: Small question here - in most of our exercises we have the functions under test as exported, here it is unexported. Are we OK with this? Is there a specific argument we have for why things should be one way or the other? The only I could find was in the README "In general, adding tests for unexported names is considered an anti-pattern" (with an exception for the `testVersion`)
(But if we do change it to exported, it will be called as `protein.ProteinLookUp` which will be repetitive so if we export it it can just be called `Lookup` or something similar)

petertseng:
Small notes, I remembered the existence of https://github.com/exercism/x-common/issues/268 when looking at this. Since we are adding protein-translation, we will want to NOT add nucleotide-codons to this track (and it appears that we prefer to have protein-translation).

We have nucleotide-**count** already but not nucleotide-**codons**, so no conflict in this track, everything is good.

exercises/protein-translation/protein_translation_test.go
```diff
@@ -0,0 +1,63 @@
...
+	{"UAC", "Tyrosine"},
+	{"UGU", "Cysteine"},
+	{"UGC", "Cysteine"},
+	{"UGG", "Tryptophan"},
+	{"UAA", "STOP"},
+	{"UAG", "STOP"},
+	{"UGA", "STOP"},
+}
+
+var proteinTestCases = []testCase{
```
petertseng: Small question because I am curious, not asking for anything to change: Did you get these test cases from any other track that implemented this exercise, or from elsewhere?

Akasurde: Yes, I got it from [here](https://github.com/exercism/xruby/blob/master/exercises/protein-translation/protein_translation_test.rb#L61)

exercises/protein-translation/protein_translation_test.go
Akasurde: @petertseng I am OK in exporting names. Let me know any example exercise which does this ?

petertseng: Sure. I know that at least https://github.com/exercism/xgo/blob/master/exercises/leap/example.go and https://github.com/exercism/xgo/blob/master/exercises/hamming/example.go do. I haven't checked all exercises to see if we were consistent in this, though it seems to be the case.

petertseng:
Because of inactivity: First I am rebasing to deal with the config.json conflict. Then I'll add a commit to export the name. I'll squash the commits on merge.

petertseng:
Merged, with thanks to @Akasurde !

PR #222, opened 2016-01-21
Remove birthday from gigasecond.

kytrinyx:
After providing a bunch of feedback to solutions to this exercise, I realized that
the whole birthday thing is a distraction. The time.Duration bit is interesting.

Closes #214.

PR #402, opened 2016-10-12
Gigasecond exercise - removes irrelevant comment from test

daniellee:
The PR #222 removed the Birthday variable from the test code, however the comment was not
removed. I got stuck for a few minutes trying to figure out if I needed to pass the variable into the testrunner before I found the PR. This commit simply removes the comment from the test class.


kytrinyx:
Aaaaah, thank you! :heart:

PR #403, opened 2016-10-12
Add isogram exercise

mvader:


exercises/isogram/example.go
```diff
@@ -0,0 +1,13 @@
+package isogram
+
+import "strings"
+
```
Akasurde: Add `testVersion` here, see other exercises for help

exercises/isogram/isogram_test.go
```diff
@@ -0,0 +1,42 @@
+package isogram
+
+import "testing"
+
```
Akasurde: Check `testVersion` with `targetTestVersion` here.

mvader:
@Akasurde done :)

kytrinyx:
Thank you! (And thanks so much for providing a code review, @Akasurde! :heart:)

exercises/isogram/isogram_test.go
```diff
@@ -0,0 +1,50 @@
...
+	{"éléphant", false},
+}
+
+const targetTestVersion = 1
+
+func TestTestVersion(t *testing.T) {
+	if testVersion != targetTestVersion {
+		t.Errorf("Found testVersion = %v, want %v.", testVersion, targetTestVersion)
+	}
+}
```
kytrinyx: Can we put this at the bottom of the test file? The tests execute in parallel, so it doesn't technically matter where it is, but I since it's more like book-keeping than something essential to the problem, I think it's a bit nicer to keep it out of the way for the person reading it.

exercises/isogram/isogram_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+func TestIsIsogram(t *testing.T) {
+	for _, c := range testCases {
+		if IsIsogram(c.word) != c.expected {
+			t.Fatalf("FAIL: Word %q, expected %v, got %v", c.word, c.expected, !c.expected)
+		}
+
+		t.Logf("PASS: Word %q", c.word)
```
kytrinyx: I don't think we've been using "PASS" explicitly. @petertseng what do you think about this?

mvader: Oh, the one I took a look at as an example had it https://github.com/exercism/xgo/blob/master/exercises/anagram/anagram_test.go#L153 so I thought it was the way to do it. I can remove it if you want

exercises/isogram/isogram_test.go
mvader: Yes, moving it

mvader:
@kytrinyx squashed :)

petertseng:
I'm sorry, we should have merged this a while ago. The only question was the Logf PASS question, and we didn't find a compelling case to remove it here, so let's take it as-is. Anyone who feels strongly about having the Logf PASS lines let's discuss in an issue.

I rebased over the conflict in config.json (using the exercises key instead of problems)

petertseng:
Thanks @mvader !

PR #404, opened 2016-10-13
Add all your base exercise

mvader:


exercises/all-your-base/all_your_base_test.go
```diff
@@ -0,0 +1,197 @@
...
+
+import "testing"
+
+const targetTestVersion = 1
+
+func TestTestVersion(t *testing.T) {
+	if testVersion != targetTestVersion {
+		t.Errorf("Found testVersion = %v, want %v.", testVersion, targetTestVersion)
+	}
+}
```
kytrinyx: Would you mind moving this to the bottom of the file? I know we've been inconsistent with this, but I've realized that it's nice to keep it out of the way since it's more book-keeping and not essential to the problem at hand.

exercises/all-your-base/all_your_base_test.go
```diff
@@ -0,0 +1,197 @@
...
+		inputDigits:  []int64{3, 46, 60},
+		outputBase:   73,
+		outputDigits: []int64{6, 10, 45},
+	},
+	{
+		description:  "empty list",
+		inputBase:    2,
+		inputDigits:  []int64{},
+		outputBase:   10,
+		outputDigits: nil,
```
kytrinyx: I think this should probably be an error rather than `nil`. `nil` vs empty slice is super subtle.

kytrinyx: @petertseng do you have an opinion on this?

exercises/all-your-base/all_your_base_test.go
```diff
@@ -0,0 +1,197 @@
...
+	},
+	{
+		description:  "single zero",
+		inputBase:    10,
+		inputDigits:  []int64{0},
+		outputBase:   2,
+		outputDigits: []int64{0},
```
kytrinyx: I agree that however many zeros we have in the input, we should have a single zero in the output.

For reference, the canonical json says `null`, but has comments at the top that each track should decide for itself: https://github.com/exercism/x-common/blob/master/exercises/all-your-base/canonical-data.json#L80

exercises/all-your-base/all_your_base_test.go
```diff
@@ -0,0 +1,197 @@
...
+		inputDigits:  []int64{0, 0, 0},
+		outputBase:   2,
+		outputDigits: []int64{0},
+	},
+	{
+		description:  "leading zeros",
+		inputBase:    7,
+		inputDigits:  []int64{0, 6, 0},
+		outputBase:   10,
+		outputDigits: []int64{4, 2},
```
kytrinyx: Ditto on this one. I think we should ignore leading zeros in Go.

exercises/all-your-base/all_your_base_test.go
```diff
@@ -0,0 +1,197 @@
...
+	},
+	{
+		description:  "negative digit",
+		inputBase:    2,
+		inputDigits:  []int64{1 - 11010},
```
kytrinyx: Would you elaborate on this as compared to the canonical case? https://github.com/exercism/x-common/blob/master/exercises/all-your-base/canonical-data.json#L94-L98

I'm not sure I understand the difference here.

exercises/all-your-base/all_your_base_test.go
```diff
@@ -0,0 +1,197 @@
...
+		outputBase:   10,
+		outputDigits: nil,
+	},
+	{
+		description:  "invalid positive digit",
+		inputBase:    2,
+		inputDigits:  []int64{121010},
+		outputBase:   10,
+		outputDigits: nil,
```
kytrinyx: I think this should have an `error`.

exercises/all-your-base/all_your_base_test.go
```diff
@@ -0,0 +1,197 @@
...
+	},
+	{
+		description:  "first base is one",
+		inputBase:    1,
+		inputDigits:  []int64{},
+		outputBase:   10,
+		outputDigits: nil,
```
kytrinyx: Here, too, I think an error makes sense.

exercises/all-your-base/all_your_base_test.go
```diff
@@ -0,0 +1,197 @@
...
+	},
+	{
+		description:  "second base is one",
+		inputBase:    2,
+		inputDigits:  []int64{101010},
+		outputBase:   1,
+		outputDigits: nil,
```
kytrinyx: `error` here too?

kytrinyx: And also for all the remaining cases (base zero as input, output, and negative bases as input/output).

mvader:
@kytrinyx fixed everything, let me know if you think of any other change or improvement

exercises/all-your-base/all_your_base_test.go
```diff
@@ -0,0 +1,214 @@
...
+	},
+	{
+		description:  "hexadecimal to trinary",
+		inputBase:    16,
+		inputDigits:  []int64{2, 10},
+		outputBase:   3,
+		outputDigits: []int64{1, 1, 2, 0},
+	},
+	{
+		description:  "15-bit int64eger",
```
petertseng: int64eger?

exercises/all-your-base/all_your_base_test.go
petertseng: Does one way mathematically make more sense? Is one way likely lead to more elegant solutions?

to me, empty input list of digits implies that the value is zero. I might suggest that the output be `[]int64{0}`

exercises/all-your-base/all_your_base_test.go
mvader: fixed

exercises/all-your-base/all_your_base_test.go
petertseng: Error seems to make sense here. The only other alternative is to consider this zero or something, but I am not convinced this adds that much to the problem. Base 1 is just that you'd represent 3 as `[]int64{1, 1, 1}`, huh? Yeah probably not that interesting

exercises/all-your-base/all_your_base_test.go
petertseng: I'll suggest that we document how zeroes are to be treated (accept leading zeroes in the input, but strip them from the output unless the value is zero in which case we include a single zero) in this file so that the expectations are clear to the students.

config.json
```diff
@@ -76,7 +76,8 @@
     "connect",
     "ledger",
     "poker",
-    "variable-length-quantity"
+    "variable-length-quantity",
+    "all-your-base"
```
petertseng: are we OK with this placement? all-your-base might not be as involved as, say, connect and poker, maybe it could go before those at least?

(I guess I also think VLQ is placed a bit late, but that can be considered in a place that's not this PR)

petertseng:
Here is my next crazy idea (though this is not the first time I had this crazy idea, as I said it on https://github.com/exercism/xrust/pull/170#issuecomment-233162831)

Right now we are enforcing in tests that negative values in `inputDigits`, `inputBase`, `outputBase` all result in errors. And it will never be the case that there should be a negative value in `outputDigits`.

So, what if we make them all `uint64`? That way we can use the type system to enforce this, not the tests. Yes, this means that we don't have all the tests in canonical-data.json, and that's OK as long as the ones that are being omitted are just the ones that the type system forbids.

exercises/all-your-base/all_your_base_test.go
mvader: Where should I document this behaviour? In the example or the tests file?

config.json
mvader: Do I place it before them then?

mvader:
Moved to type `uint64` as @petertseng suggested and fixed all the other issues. Only thing that's left is knowing where to document the decisions on how to treat zeroes and the position of the exercise in `config.json`

exercises/all-your-base/all_your_base_test.go
petertseng: It would hvae to be in the test file because the example file will not be served to students.

petertseng:
> Only thing that's left is knowing where to document the decisions on how to treat zeroes

Looks like that'll want to go in the test file. the other alternative is some other file that gets served to students (that rules out the example), or HINTS.md to get it included in the README.

petertseng:
> Looks like that'll want to go in the test file.

I have placed it there, due to inactivity on this PR. Merging now, with thanks to @mvader for the work!

PR #405, opened 2016-10-15
TESTS.md: Make bracketed placeholder render correctly

infogulch:
Both github and http://exercism.io/languages/go/tests renders the placeholder `<regexp>` incorrectly when unquoted. This adds backquotes around the raw text which should fix both.

Specifically, github omits it entirely, and excercism adds `</regexp>` after the period. It seems that the MD renderer is trying to be a bit too helpful by "closing" the "open html tag".

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -38,7 +38,7 @@ Your first goal it to get something to compile, even though it fails tests.  For
 
 ### Fixing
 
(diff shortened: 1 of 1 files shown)


petertseng:
Interesting! The `</regexp>` thing is quite amusing. Thanks.

PR #406, opened 2016-10-15
replace 'problems' key with 'exercises' in config.json

cgrayson:
see #345

### DIFF
--- a/config.json
+++ b/config.json
@@ -3,81 +3,6 @@
   "language": "Go",
   "repository": "https://github.com/exercism/xgo",
   "active": true,
-  "problems": [
-    "hello-world",
-    "leap",
-    "clock",
-    "gigasecond",
-    "hamming",
-    "raindrops",
-    "acronym",
-    "bob",
-    "triangle",
-    "difference-of-squares",
-    "secret-handshake",
-    "food-chain",
-    "house",
-    "pascals-triangle",
-    "series",
-    "queen-attack",
-    "grains",
-    "etl",
-    "scrabble-score",
-    "parallel-letter-frequency",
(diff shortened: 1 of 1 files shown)


petertseng:
Can I get a confirm that it is safe to **delete** the problems key?

Yes, it appears I can get it. https://github.com/exercism/x-api/pull/137 has made it happen. Good. One last check before merge (that the order is the same.

petertseng:
Good, empty output from `diff -u <(jq '.["problems"]' < config.json) <(jq '.["exercises"] | map(.["slug"])' < newconfig.json)`

One last question: Should we put the `slug` key before the `difficulty` key?

cgrayson:
I listed the attributes alphabetically, as I'm used to doing in JSON, but if you prefer `slug` first, happy to change it.

petertseng:
I really wished I could point to an example of how it should be done, but my hope was dashed! Because #345 has it BOTH ways ("hello-world" is slug first, and then "anagram" and "forth" are alphabetical!)

My *instinct* was slug first since it is "the most important", but now I have no ground to stand on! @kytrinyx does this matter to you? If not I'll just take it.

petertseng:
We could just take it now and defer the decision. If we want to change it, we can change it when we make the difficulty values non-1, since all the lines will change anyway. Except the ones that stay 1.

kytrinyx:
> does this matter to you

I can't decide one way or the other. My instinct says `slug` first, since it's the identifier for the exercise.

Let's take this and change it later if we decide we care enough!

petertseng:
Thanks everyone.

PR #407, opened 2016-10-17
nucleotide-count: improve error handling fixes #298

### LINKED ISSUE #298: nucleotide-count: Improve error-handling behavior

ferhatelmas:
In addition to improving error handling, I made some changes according to my taste such as map equality check, value printing, etc.

Unless well aligned, I can revert or update w.r.t your comments. Thanks!

### DIFF
--- a/exercises/nucleotide-count/nucleotide_count_test.go
+++ b/exercises/nucleotide-count/nucleotide_count_test.go
@@ -34,8 +34,9 @@ var tallyTests = []struct {
 func TestNucleotideCounts(t *testing.T) {
 	for _, tt := range tallyTests {
 		dna := DNA(tt.strand)
-		count, _ := dna.Count(tt.nucleotide)
-		if count != tt.expected {
+		if count, err := dna.Count(tt.nucleotide); err != nil {
+			t.Fatal(err)
+		} else if count != tt.expected {
 			t.Fatalf("Got \"%v\", expected \"%v\"", count, tt.expected)
 		}
 	}
@@ -43,11 +44,7 @@ func TestNucleotideCounts(t *testing.T) {
 
 func TestHasErrorForInvalidNucleotides(t *testing.T) {
 	dna := DNA("GATTACA")
-	count, err := dna.Count('X')
-	if count != 0 {
-		t.Fatalf("Got \"%v\", expected \"%v\"", count, 0)
-	}
-	if err == nil {
+	if _, err := dna.Count('X'); err == nil {
 		t.Fatalf("X is an invalid nucleotide, but no error was raised")
 	}
 }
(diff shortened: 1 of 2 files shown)


ferhatelmas:
Ah, just saw that there is already an open PR :disappointed:

kytrinyx:
It doesn't look like the author of the other open PR has come back, so I'm happy to push this one forward.

There are a number of changes in here that are not about improving the error handling. I do like the changes—would you be willing to open a separate PR for those and only clean up the error handling in this one?

kytrinyx:
(If you open a PR for that, would you `@`-mention me? I've got a lot going on and could definitely use the specific reminder).

ferhatelmas:
@kytrinyx I removed irrelevant parts from this PR. After this get merged, I will create another PR for other improvements. Thanks!

exercises/nucleotide-count/nucleotide_count_test.go
```diff
@@ -79,6 +83,10 @@ var histogramTests = []histogramTest{
 		Histogram{'A': 0, 'C': 0, 'T': 0, 'G': 8},
 	},
 	{
+		"GGXXGGGG", // skip mutation
```
petertseng: A new case I am a little unsure about! Made harder by the fact that `DNA` doesn't return an `error`. Should it? That (plus this change) might be for a separate PR though.

This is difficult for me to imagine, since x-common doesn' thave canonical-data.json for nucleotide-count. Have other tracks with nucleotide-count similar cases to this?

exercises/nucleotide-count/nucleotide_count_test.go
```diff
@@ -34,20 +34,17 @@ var tallyTests = []struct {
 func TestNucleotideCounts(t *testing.T) {
 	for _, tt := range tallyTests {
 		dna := DNA(tt.strand)
-		count, _ := dna.Count(tt.nucleotide)
-		if count != tt.expected {
+		if count, err := dna.Count(tt.nucleotide); err != nil {
+			t.Fatal(err)
+		} else if count != tt.expected {
```
cpuguy83: Can drop the else since the above is a `Fatal` and will return immediately.

exercises/nucleotide-count/nucleotide_count_test.go
ferhatelmas: Removed this test case. 

I was planning to open another after this one for cleaning as suggested by @kytrinyx [here](https://github.com/exercism/xgo/pull/407#issuecomment-255615703). I can include error and test case there.

ferhatelmas:
@petertseng Removed test case and rebased on master.

exercises/nucleotide-count/nucleotide_count_test.go
petertseng: will have to watch though, `count` might not be in scope since now `count, err :=` is in the `if`.

petertseng:
All right! Thanks for the work.

PR #411, opened 2016-10-30
food-chain: highlight differences in test output

tdilo:
When the tests for this exercise fail, they print whole verses and all
of the song to the terminal. This can be helpful insofar as the desired
solution is shown in its entirety, which serves as a reference as well
as providing a sense of achievement when the length of the generated
output grows. On the other hand, the overall length of the test output
makes it difficult to spot errors. A lot of scrolling and manual
proof-reading may be required, which makes solving the exercise tedious.

This commit changes the tests to analyze the solution line by line and
append a helpful hint at the bottom of the output of each failed test,
where it can be seen without scrolling. Only the first difference is
highlighted, as a missing line would generate a lot of output otherwise.

Example output:

food_chain_test.go:102: Verse (n) =
	go test these three lines
	subtle errer in-between
	it is hard to spot
	    want:
	go test these three lines
	subtle error in-between
	it is hard to spot
	-- first difference in line 2:
	-- got : "subtle errer in-between"
	-- want: "subtle error in-between"

### DIFF
--- a/exercises/food-chain/food_chain_test.go
+++ b/exercises/food-chain/food_chain_test.go
@@ -1,6 +1,7 @@
 package foodchain
 
 import (
+	"fmt"
 	"strings"
 	"testing"
 )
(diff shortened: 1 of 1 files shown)


kytrinyx:
I really like this change, thank you!

PR #412, opened 2016-10-31
food-chain: resolve ambiguity in verses test

tdilo:
The current test for the Verses function checks for two consecutive
verses (1 and 2). There are multiple ways to interpret the function
signature and satisfy the test, all observed in submitted solutions:

* Verses(start, end int) string
    - return consecutive verses from start to end
* Verses(verses ...int) string
    - return n (possibly non-consecutive) verses
* Verses(a, b int) string
    - join two (possibly non-consecutive) verses

This commit changes the tested function arguments to cover two non-
consecutive verses (1 and 3). With the change, only the first solution
returning consecutive verses is accepted. The test version is bumped to
keep existing exercise solutions valid.

### DIFF
--- a/exercises/food-chain/example.go
+++ b/exercises/food-chain/example.go
@@ -1,6 +1,6 @@
 package foodchain
 
-const testVersion = 2
+const testVersion = 3
 
 var verse = []struct{ eaten, comment string }{
 	{"", ""},
--- a/exercises/food-chain/food_chain_test.go
+++ b/exercises/food-chain/food_chain_test.go
@@ -6,7 +6,7 @@ import (
 	"testing"
 )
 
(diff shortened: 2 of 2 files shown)


petertseng:
Thanks. I agree with the semantics that is now (hopefully) unambiguously determined by this test. @tdilo , it would be useful to go to https://github.com/exercism/x-common/blob/master/exercises/food-chain/canonical-data.json with this change as well, since tracks will get their tests from there. Until then, the Go track will be in contravention of the common data, which will naturally cause questions about why the inconsistency.

tdilo:
Thanks for the pointer to to the common test data, pull request is pending. :)

PR #413, opened 2016-11-01
nucleotide-count: Some cleaning

ferhatelmas:
* use reflect.DeepEqual, instead of custom
 * return error if invalid nucleotide
 * add comments here and there

### DIFF
--- a/exercises/nucleotide-count/nucleotide_count_test.go
+++ b/exercises/nucleotide-count/nucleotide_count_test.go
@@ -55,27 +55,41 @@ func TestCountingDoesntChangeCount(t *testing.T) {
 type histogramTest struct {
 	strand   DNA
 	expected Histogram
+	err      bool
 }
 
 var histogramTests = []histogramTest{
 	{
 		"",
 		Histogram{'A': 0, 'C': 0, 'T': 0, 'G': 0},
+		false,
 	},
 	{
 		"GGGGGGGG",
 		Histogram{'A': 0, 'C': 0, 'T': 0, 'G': 8},
+		false,
 	},
 	{
 		"AGCTTTTCATTCTGACTGCAACGGGCAATATGTCTCTGTGTGGATTAAAAAAAGAGTGTCTGATAGCAGC",
 		Histogram{'A': 20, 'C': 12, 'T': 21, 'G': 17},
+		false,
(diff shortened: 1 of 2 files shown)


ferhatelmas:
@petertseng @kytrinyx other improvements that are extracted from previous PR.

exercises/nucleotide-count/nucleotide_count_test.go
```diff
@@ -69,30 +53,43 @@ func TestCountingDoesntChangeCount(t *testing.T) {
...
 func TestSequenceHistograms(t *testing.T) {
 	for _, tt := range histogramTests {
-		dna := DNA(tt.strand)
-		if !dna.Counts().Equal(tt.expected) {
-			t.Fatalf("DNA{ \"%v\" }: Got \"%v\", expected \"%v\"", tt.strand, dna.Counts(), tt.expected)
+		counts, err := tt.strand.Counts()
+		if tt.err && err == nil {
+			t.Fatalf("DNA{ %q }: expected error but didn't get one.", tt.strand)
+		} else if !tt.err && err != nil {
+			t.Fatalf("DNA{ %q }: expected no error but got error", tt.strand)
```
petertseng: here, it will be possible to show what error the student's solution showed (though not strictly necessary)

petertseng:
There are three distinct changes in this one PR, as evidenced by the three bullet points in the description. It would be great if they could be separate commits, since there should be one commit per logical change. This makes the three commits individually easier to understand, instead of having one large commit that is hard to understand (because it contains three unrelated types of changes)

In particular, the "return error if invalid nucleotide" will want a testVersion bump accompanying it (this can go in the same commit as it, since it is essential to it)

petertseng:
(Although I forgot to mention this: I agree with these changes and happily welcome them!)

exercises/nucleotide-count/nucleotide_count_test.go
ferhatelmas: added

ferhatelmas:
@petertseng split it into 3 commits with each logical change. Bumped version and added error message to fatal. Also, rebased. Anything else, please let me know.

petertseng:
Thanks, this is all excellent.

PR #416, opened 2016-11-24
docs: Extend learning resources with awesome go

ferhatelmas:
* e-books and tutorials


petertseng:
Ah cool, the Rust track also added Awesome Rust in https://github.com/exercism/xrust/pull/91, seems good to do the same here.

PR #418, opened 2016-11-25
perfect-numbers: Add perfect-numbers exercise

ferhatelmas:


exercises/perfect-numbers/perfect_numbers_test.go
```diff
@@ -0,0 +1,39 @@
...
+			t.Errorf("%d: Expected no error but got %s", c.input, err)
+		} else if cat != c.expected {
+			t.Errorf("%d: Expected %q, got %q", c.input, c.expected, cat)
+		}
+	}
+}
+
+func TestGivesPositiveRequiredError(t *testing.T) {
+	if _, err := Classify(0); err != ErrOnlyPositive {
+		t.Errorf("Expected eror %q but got %q", ErrOnlyPositive, err)
```
petertseng: Do you think that we need to do the "ensure the `err` return is actually an `error`" of https://github.com/exercism/xgo/issues/275, or would the name `ErrOnlyPositive` starting with `Err` convince the student enough? The goal is simply to encourage idiomatic Go code, so if thinking the name is good enough, we don't need to add an explicit compile-time check.

petertseng: Think that we need to do the "ensure the `err` return is actually an `error`" of https://github.com/exercism/xgo/issues/275, or would the name `ErrOnlyPositive` starting with `Err` convince the student enough? The goal is simply to encourage idiomatic Go code, so if thinking the name is good enough, we don't need to add an explicit compile-time check.

ferhatelmas: Thanks fixed.

exercises/perfect-numbers/perfect_numbers_test.go
```diff
@@ -0,0 +1,48 @@
+package perfect
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestPositiveNumberRequiredErrorDefined(t *testing.T) {
+	if !reflect.TypeOf(ErrOnlyPositive).Implements(reflect.TypeOf((*error)(nil)).Elem()) {
```
petertseng: Do you feel strongly about using this runtime check rather than a compile-time check of `var _ error = ErrOnlyPositive`?

ferhatelmas: No. IMHO, we don't need neither of them.

However, we want to ensure so I tried to make it as a separate achievement and put as the first test but I can change it to compile time check, as you wish.

petertseng: Hmm, if you think we don't need either, I am happy to trust your judgment. I am fine with its name starting with `Err` being indicative enough.

If we have a check anyway, I would prefer the compile time check. And that can just go as a single line in the other error test or on the top level.

So either changing or removing would be OK with me, and then after that I think it will be good to merge.

exercises/perfect-numbers/example.go
```diff
@@ -0,0 +1,41 @@
...
+				return ClassificationAbundant, nil
+			}
+		}
+	}
+	if sum == n {
+		return ClassificationPerfect, nil
+	} else if sum < n {
+		return ClassificationDeficient, nil
+	}
+	return ClassificationAbundant, nil
```
petertseng: I imagine execution will never reach this line, but that is OK, doesn't need to change.

ferhatelmas: Yes, I wrote this line first then added short return but kept it because even if we don't support, for older version of go (<1.4) it's needed, I think.

Issue #414, opened 2016-11-08
accumulate: use %q to make difference between [""] and [] clear


Insti:
"go path secret handshake does not work properly": https://github.com/exercism/exercism.io/issues/3239

----

![](http://i.imgur.com/x9ZVfKz.png)
```go
package secret

import (
	"regexp"
	"strconv"
)

const testVersion = 1

var cases = map[int]string{
	0: "wink",
	1: "double blink",
	2: "close your eyes",
	3: "jump",
	4:"reverse",
}

//Handshake function returns the encrypted human readable handshake
func Handshake(inp uint) []string {
	var in = strconv.Itoa(int(inp))
	if in == "0" {
		return nil
	}
	isDecimal := regexp.MustCompile("[2-9]")
	if isDecimal.MatchString(in) {
		num, _ := strconv.Atoi(in)
		in = strconv.FormatInt(int64(num), 2)
	}
	reverse := false
	out := make([]string, 0, 4)
	ln := len(in) - 1
	for num, val := range in {
		if ln-num == 4 {
			reverse = true
			continue
		}
		if string(val) == "1" {
			out = append(out, cases[ln-num])
		}
	}
	if !reverse {
		for i, j := 0, len(out)-1; i < j; i, j = i+1, j-1 {
			out[i], out[j] = out[j], out[i]
		}
	}
	return out
}
```

petertseng:
I suggest to change %v to %q so that the difference between `[""]` and `[]` is made clear

petertseng:
also consider the following exercises that may expect `[]string`s: allergies, anagram, binary-search-tree, custom-set, grade-school, kindergarten-garden, ocr-numbers, poker, series, strain (I may have missed some, I just grepped for `[]string`, and some listed may already be using %q)

robphoenix:
:tada:

PR #422, opened 2016-11-29
update test formatting, response to #414

bordeltabernacle:


exercises/allergies/allergies_test.go
```diff
@@ -67,9 +67,9 @@ func TestAllergicTo(t *testing.T) {
 	for _, test := range allergicToTests {
 		actual := AllergicTo(test.i, test.allergen)
 		if actual != test.expected {
-			t.Fatalf("FAIL: AllergicTo(%d, %s): expected %t, actual %t", test.i, test.allergen, test.expected, actual)
+			t.Fatalf("FAIL: AllergicTo(%d, %q): expected %t, actual %t", test.i, test.allergen, test.expected, actual)
```
petertseng: this one wouldn't strictly have been necessary since it's just a single string (and we know it will always be non-empty since the test provides it, not the student solution), but won't hurt to have.

exercises/allergies/allergies_test.go
```diff
@@ -67,9 +67,9 @@ func TestAllergicTo(t *testing.T) {
...
 		} else {
-			t.Logf("PASS: AllergicTo(%d, %s) %t", test.i, test.allergen, actual)
+			t.Logf("PASS: AllergicTo(%d, %q) %t", test.i, test.allergen, actual)
```
petertseng: same: this one wouldn't strictly have been necessary since it's just a single string (and we know it will always be non-empty since the test provides it, not the student solution), but won't hurt to have.

bordeltabernacle:
I changed and undid a few in grade-school, seeing what difference it made, I think I `ctrl-z`'d a bit too much there! Anyway, I've fixed it up and changed the `allergies_test` as well. Even though you say it can't hurt, it's not necessary, so probably best to leave as is, plus I learnt how to remove a commit! :smiley:

petertseng:
Thanks!

PR #423, opened 2016-12-16
modify house-test.go comments

bordeltabernacle:
The description at the top of `house_test.go` can be confusing
and ultimately distracting from the problem at hand. this commit
attempts to replace the (unnecessarily) academic language with
something plainer and more to the point.

I noticed I'm not the only one who had this issue:

http://exercism.io/submissions/3b5b6003f03248f6a402859df52b6381


petertseng:
Seems good, thanks. I honestly wonder why `Embed` has to be exported + tested at all. I would remove it.

PR #425, opened 2016-12-17
Fix case of abbreviate function to be capitalized

felixbuenemann:
I found it very irritating, that the `abbreviate` function name was all lowercase instead of beginning with a cap and [others seem to agree](https://github.com/exercism/xgo/issues/382#issuecomment-253994632), given that this is the main function of the package.

This changes the casing to `Abbreviate`.

### DIFF
--- a/exercises/acronym/acronym_test.go
+++ b/exercises/acronym/acronym_test.go
@@ -4,7 +4,7 @@ import (
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


petertseng:
An overdue change since it was discussed and agreed upon but then nobody did it. Thanks for putting in the work.

PR #426, opened 2016-12-19
house: re-implement exercise, fixes #242

### LINKED ISSUE #242: Consider re-implementing house to not guide the solution
Right now the test suite suggests that house should have a recursive solution, but it's quite a lot simpler without it. I'd like to leave the door open for a larger variety of approaches to this problem.

bordeltabernacle:
Remove `Embed()` function, and enable a non-recursive solution. The test
file still suggests using recursion as a method for solving this
problem, and both recursive and non-recursive solutions are documented
in the `example.go` file. Reading suggestions for recursion in Go have
also been added.

Because of this change a test version number and test have been added.

An extra error check has also been added to the Song() test, https://github.com/bordeltabernacle/xgo/blob/re-implement-house/exercises/house/house_test.go#L130-L134, to check
for the number of verses in the solution. This is to avoid unhelpful test
failure messages when a whole verse is missing, such as:

```
house_test.go:156: Song() line 77 = "", want ""
```

now reads:

```
house_test.go:139: Song() has 11 verses, want 12 verses
```

### DIFF
--- a/exercises/house/example.go
+++ b/exercises/house/example.go
@@ -1,39 +1,67 @@
 package house
 
-func Embed(relPhrase, nounPhrase string) string {
-	return relPhrase + " " + nounPhrase
+var songLines = []string{
+	"the horse and the hound and the horn\nthat belonged to",
+	"the farmer sowing his corn\nthat kept",
+	"the rooster that crowed in the morn\nthat woke",
+	"the priest all shaven and shorn\nthat married",
+	"the man all tattered and torn\nthat kissed",
+	"the maiden all forlorn\nthat milked",
+	"the cow with the crumpled horn\nthat tossed",
+	"the dog\nthat worried",
+	"the cat\nthat killed",
+	"the rat\nthat ate",
+	"the malt\nthat lay in",
 }
 
-func Verse(subject string, relPhrases []string, nounPhrase string) string {
-	return subject + " " + recurse(relPhrases, nounPhrase)
+// Recursive Solution
+
+func Verse(v int) (verse string) {
+	v--
+	i := len(songLines) - v
+	verse += buildVerse(songLines[i:], "This is ")
+	verse += "the house that Jack built."
+	return
 }
 
-func recurse(relPhrases []string, nounPhrase string) string {
-	if len(relPhrases) == 0 {
-		return nounPhrase
+func buildVerse(songLines []string, cur string) string {
+	if len(songLines) == 0 {
+		return cur
 	}
-	return Embed(relPhrases[0], recurse(relPhrases[1:], nounPhrase))
+	cur += songLines[0]
+	cur += " "
+	return buildVerse(songLines[1:], cur)
 }
 
-func Song() string {
-	relPhrases := []string{
-		"the horse and the hound and the horn\nthat belonged to",
-		"the farmer sowing his corn\nthat kept",
-		"the rooster that crowed in the morn\nthat woke",
-		"the priest all shaven and shorn\nthat married",
-		"the man all tattered and torn\nthat kissed",
-		"the maiden all forlorn\nthat milked",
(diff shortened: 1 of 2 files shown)


exercises/house/example.go
```diff
@@ -1,39 +1,69 @@
...
+		song += Verse(i + 1)
+		if i < len(songLines) {
+			song += "\n\n"
+		}
 	}
-	return s
+	return
 }
+
+// Iterative Solution
```
petertseng: Takes me wish we could have multiple examples and test both of them, to avoid the possibility that the commented-out code gets out of date.

No change requested in this PR though, since that is not a problem we need to solve in this PR.

exercises/house/house_test.go
```diff
@@ -1,19 +1,13 @@
...
 
-// Song generates the full text of "The House That Jack Built". Of course, you
-// could return a string literal, but humor us; try using Verse as a subroutine.
+// While recursion isn't always the simplest or most efficient solution to a problem,
+// it's a powerful programming technique nonetheless.
+// New to recursion? Here are some useful articles to read up on it:
 //
-//    func Song() string
+// https://www.golang-book.com/books/intro/7#section5
+// https://www.goinggo.net/2013/09/recursion-and-tail-calls-in-go_26.html
```
petertseng: this one is a lot to absorb for someone new to recursion (the supposed target audience for these links). I say this because it describes details such as the stack layout, the difference between a recursive call in tail-call position versus not, and how the compiler might treat each.

The [image showing the call stack](https://www.goinggo.net/images/goinggo/Screen+Shot+2013-09-20+at+6.36.41+PM.png) was useful. I would argue that the rest of the page may not appeal to someone new to recursion. I can't say for sure ([curse of knowledge](https://en.wikipedia.org/wiki/Curse_of_knowledge), etc.) but I was guessing that such a student would be interested simply in "what is recursion? how do I use it? how can I reason through what happens with a recursive function, so that I know how to code one?" and not how to optimise one (tail-call versus not).

What say you? Maybe I'm wrong, and people are interested!

exercises/house/example.go
bordeltabernacle: That would be great. I did think of putting it in a different file.

It is a concern that the commented code will get out of date. In this case is it perhaps best to remove the commented code, leaving the recursive solution, as that is what is suggested to attempt with this problem? Potentially by this exercise the iterative solution shouldn't be too difficult to implement anyway? I'm happy to remove.

exercises/house/house_test.go
bordeltabernacle: tbh with this link I mostly wanted to bring up the idea of tail call recursion. You're right it is a bit heavy for someone new to recursion, and as the article itself says, Go doesn't really optimise for it anyway. I'll remove it, it would turn up in a search for `golang + recursion` anyway if someone is interested in gaining a deeper understanding.

exercises/house/example.go
petertseng: I see the primary purpose of the example as making sure that the tests are in fact solvable, by being used in Travis CI.

If we consider only this purpose, a commented-out example serves no benefit!

Though that's not the whole story, since I do see a secondary purpose: To be (as its name suggests) an example of a working solution, available for students to look at. But they have to work a bit hard to find it, since they have to know about the existence of this git repo and examine this file.

The commented-out example is, of course, useful for this purpose.

I don't feel strongly either way; it surely can be kept.

If there is great interest in actually putting multiple examples though Travis CI (this will help us assure they are good examples for students to look at), that is a separate project.

Haskell has done this at https://github.com/exercism/xhaskell/pull/396 ; it was important for that track for slightly different reasons, since the Haskell track wished to allow its students to submit functions with various type signatures, so the multiple examples help ensure that multiple type signatures are in fact accepted by the tests.

PR #427, opened 2016-12-22
Update/sum of multiples

bordeltabernacle:


bordeltabernacle:
I've left the example function as variadic. I quite like it, and like the original exercise implementer in #53 think it could be a good opportunity to introduce variadic functions. Though I do also realise this goes against the recommendation from @petertseng to just take a slice instead, and could very well be considered inappropriate. I'm happy to change to a slice if preferred.

exercises/sum-of-multiples/example.go
```diff
@@ -1,16 +1,15 @@
...
-		sum := 0
-		for i := 1; i < limit; i++ {
-			for _, d := range divisors {
-				if i%d == 0 {
-					sum += i
-					break
-				}
+// MultipleSummer returns the sum of the multiples of the given divisors
+// up to, but not including, the given limit.
+func MultipleSummer(limit int, divisors ...int) (sum int) {
```
petertseng: it looks like historically it made sense to call this `MultipleSummer` because it would be used like `MultipleSummer(3, 5).sum(4)` or something.

At this point, now it is being used like `MultipleSummer(4, 3, 5)`, in which case I think it is better to give it a verb name than a noun name. I am thinking of `SumMultiples`. `go lint` may complain about `summultiples.SumMultiples` being redundant, but maybe someone else has a better suggestion.

What say you? I can't find justification for this in either https://github.com/golang/go/wiki/CodeReviewComments or https://golang.org/doc/effective_go.html so it may be a personal thing, but I usually see function names as verbs.

exercises/sum-of-multiples/sum_of_multiples_test.go
```diff
@@ -2,72 +2,37 @@ package summultiples
...
-	sum35 := MultipleSummer(3, 5)
-	for _, test := range test35 {
-		s := sum35(test.limit)
-		if s != test.sum {
-			t.Fatalf("Sum to %d returned %d, want %d.", test.limit, s, test.sum)
-		}
-	}
-}
-
 func TestVar(t *testing.T) {
```
petertseng: With the removal of `Test35` we could rename this, since the `Var` is no longer needed to distinguish it from the `35`. I'm not sure that simply `Test` will be accepted, but maybe `TestSum` will suffice. Optional change.

exercises/sum-of-multiples/sum_of_multiples_test.go
```diff
@@ -2,72 +2,37 @@ package summultiples
...
-	sum35 := MultipleSummer(3, 5)
-	b.ResetTimer() // bench just the sum function
-	for i := 0; i < b.N; i++ {
-		for _, test := range test35 {
-			sum35(test.limit)
-		}
-	}
-}
-
 func BenchmarkVar(b *testing.B) {
```
petertseng: same idea, maybe this can get renamed now.

exercises/sum-of-multiples/example.go
bordeltabernacle: Yeah, `summultiples.SumMultiples` isn't perfect but will do for now, no complaints from `go lint`

bordeltabernacle:
oh gosh yeah, I totally overlooked the test version, sorry!
Thanks for merging, happy holidays @petertseng! :fireworks:

Issue #357, opened 2016-09-09
test generators should use post-re-organisation x-common JSON file paths


petertseng:
https://github.com/exercism/x-common/pull/351 has reorganised the JSON filenames in the x-common repo. See https://github.com/exercism/xgo/blob/master/exercises/leap/example_gen.go for an example of `Gen` being used called with `leap.go`, and it appears `Gen` in https://github.com/exercism/xgo/blob/master/gen/gen.go will expect that filename. Before the reorganisation, this file was located at [x-common/leap.json](https://github.com/exercism/x-common/blob/master/leap.json). After the reorganisation, this file is located in e.g. [x-common/exercises/leap/canonical-data.json](https://github.com/exercism/x-common/blob/master/exercises/leap/canonical-data.json) .

This means that if I were to run for example `go run example_gen.go` right now it would not be able to find the file.

The task here is:
- Make `Gen` search for the file in the correct place
- Change all callers of `Gen`, if necessary, depending on the implementation of the above. A suggestion I have is that `Gen` need only take the exercise slug such as `"leap"`

After this, we should be able to run `go run example_gen.go` again.

PR #428, opened 2017-01-02
Update test cases for word-count to match latest meta

turnkey-commerce:
I noticed the test cases for the Go word-count exercises weren't challenging compared to C# and Python (mainly missing apostrophe cases). They were behind the meta cases so I updated them as well as the example to pass.

### DIFF
--- a/exercises/word-count/example.go
+++ b/exercises/word-count/example.go
@@ -5,20 +5,24 @@ import (
 	"strings"
 )
 
-const testVersion = 2
+const testVersion = 3
 
+// Frequency is a map of the frequency of occurrence keyed to the unique word.
 type Frequency map[string]int
 
+// WordCount returns the map of frequency of words based on the input phrase.
 func WordCount(phrase string) Frequency {
 	freq := Frequency{}
 	for _, word := range strings.Fields(normalize(phrase)) {
+		word = strings.Trim(word, "'")
 		freq[word]++
 	}
 	return freq
 }
 
 func normalize(phrase string) string {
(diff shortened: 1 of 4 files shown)


exercises/word-count/example.go
```diff
@@ -5,20 +5,24 @@ import (
...
 	for _, word := range strings.Fields(normalize(phrase)) {
+		word = strings.Trim(word, "'")
 		freq[word]++
 	}
 	return freq
 }
 
 func normalize(phrase string) string {
-	r, _ := regexp.Compile(`[^\w]`)
+	//  Allow for apostrophes in words.
```
bordeltabernacle: I think maybe this comment is specific to the changes, rather than the function, and is probably unnecessary?

turnkey-commerce: I thought the same but it's sometimes hard to understand the regex content, but am OK to remove it if you think it's redundant.

kytrinyx: I think it's worth keeping.

turnkey-commerce: Thanks for the validation on that, I agree it's good to be spare with comments but regex is one place where comments are often needed :).

bordeltabernacle: :+1:

petertseng:
Notes for history: I noted at the time that the presence of `cases_test.go` would indicate that this file had generated tests. In light of this, it would be a good long-term fix to solve this by fixing the generator. The generator was broken at the time though (see #357 ), so this PR was still the right thing to do at the time, I just like keeping track of these things for history. Thanks for the work.

PR #429, opened 2017-01-05
hello-world: add topics in config.json

bordeltabernacle:


petertseng:
This will be a large undertaking (doing this for all 80 of our problems). It is good to see a start on it. Would it be good to mention #345 in PRs of this type, or did you intentionally not? (Sorry that I did it for this one if you intentionally didn't)

bordeltabernacle:
Ah, no, just completely overlooked it, I'll remember to add it in in future.

PR #430, opened 2017-01-05
grains: add topics to config.json, update difficulty

bordeltabernacle:


config.json
```diff
@@ -107,9 +107,15 @@
       "topics": []
     },
     {
-      "difficulty": 1,
+      "difficulty": 2,
```
ferhatelmas: If we change difficulty, don't we need to put it into correct order in this array ?

petertseng: Hmm, where we want to put this exercise just depends on: Do we think `grains` is well-placed in this track, given its difficulty relative to the exercises around it?

Maybe the picture would be easier to see if we have a single PR proposing difficulty numbers for all the exercises, without adding topics to most of them yet (it's what #345 suggests to do).

For example, is there any exercise before `grains` that we would rate a 3? Any exercise after `grains` that we would rate a 1? If so, that is a bit strange, and then maybe some of those exercises should get moved.

config.json
```diff
@@ -107,9 +107,15 @@
...
       "slug": "grains",
-      "topics": []
+      "topics": [
+        "Control-flow (if-else statements)",
+        "Bitwise operations",
+        "Mathematics",
+        "Floating-point numbers",
```
petertseng: uncertain of this one, given that all answers fit in unsigned integers. But did you see someone use floating-point numbers? If so, OK.

bordeltabernacle: It was less the use of floating-point numbers and more the type conversion to a float that occurs; `math.Pow(2, float64(n-1))`
Now I write this, I've just checked the topics list and see there is a Type conversion topic, so I think I'll swap it out for that, what do you think @petertseng @ferhatelmas ?

config.json
bordeltabernacle: For me I've been adding the difficulty and topics at the same time as the process of making those judgements is intrinsically combined. I'll look at other peoples submissions and see if there are any comments indicating peoples feelings regarding the exercise, so that the difficulty rating is not confined to my own experience. This is especially important for exercises that I completed a while back where maybe I've forgotten some of the pain of not knowing the solution. At the same time I'm looking for what topics are being covered. For some exercises the submissions are all very similar, and so the topics list and difficulty is quite easy. For others, such as this one, there can be a number of different approaches, covering different topics. 

I think it would be a more difficult task to rate all the exercises in one fell swoop and then add topics as we go. Would someone need to have completed all the exercises to do this? If they're reviewing the exercise to gauge difficulty, is it not worth making a note of topics simultaneously? Or would the initial difficulty ratings be a quick and simple best guess effort, to be adjusted over time? (My concern with this is starting an exercise's rating with a misleading level of difficulty) My intention from here is to add topics to exercises in the current order they are in, I only added grains out of step as I completed it yesterday and wanted to tackle it while fresh in my mind.

I agree with @petertseng, it's hard to make a judgement on order until there are adjacent exercises to compare against.

Also to consider is the idea of this exercise being simple in one way, but potentially introducing a difficult concept. 

I've updated the config.json in the xpython track in a similar way to this, and my assumption was that any reordering could occur when there's sufficient information to compare the exercises. I wanted to get started on this to at least have a kind of baseline from which to work from. Anyway I've rambled on, sorry.

petertseng: > I think it would be a more difficult task to rate all the exercises in one fell swoop and then add topics as we go. Would someone need to have completed all the exercises to do this? If they're reviewing the exercise to gauge difficulty, is it not worth making a note of topics simultaneously? Or would the initial difficulty ratings be a quick and simple best guess effort, to be adjusted over time?

My suspicion is that if we think the track is already well-ordered, we can take it slow.

My experience in another track: In Haskell, it was slightly urgent to get the track ordered quickly since some problems were in wildly wrong places, so I did everything in one go by roughly guessing how I'd approach the problem and estimating the difficulty of that approach, resulting in https://github.com/exercism/xhaskell/pull/402 . I had not completed every exercise personally, but I was at least familiar with the tests being given in every single exercise, since I had reviewed PRs updating every single one of them. I also cross-referenced with F# as that track has implemented almost every exercise and assigned difficulties to each one. You're right that these guesses could have been wrong, but it was better than the alternative.

That approach may not be necessary for this track. I had previous work checking ordering in #279, but this only checked stub files vs not. I haven't checked other things about track ordering.

At a quick glance, I would say react is probably too early, and variable-length-quantity probably doesn't belong at the very end.

> My intention from here is to add topics to exercises in the current order they are in, I only added grains out of step as I completed it yesterday and wanted to tackle it while fresh in my mind.

Ah, that explains it. Thanks. I was wondering why you suddenly jumped to grains...

bordeltabernacle: so, just so I'm clear, are you & @ferhatelmas OK if I continue adding difficulty & topics at the same time to exercises? I'll mostly go in order, unless I complete another exercise as I go, though it shouldn't take too long to catch up with myself. I'm just not familiar enough with the exercises to tackle all the difficulties at once. I'll also cross-reference with other tracks as I go.

petertseng: > so, just so I'm clear, are you & @ferhatelmas OK if I continue adding difficulty & topics at the same time to exercises?

OK with me

PR #432, opened 2017-01-06
config.json: Deprecate all base-conversions

petertseng:
We added all-your-base in #404 but forgot to deprecate all the
base-conversion exercises!

According to https://github.com/exercism/x-common/issues/279,
all-your-base is meant to deprecate all of these.

### DIFF
--- a/config.json
+++ b/config.json
@@ -5,9 +5,13 @@
   "active": true,
   "deprecated": [
     "accumulate",
+    "binary",
     "bottles",
     "counter",
-    "point-mutations"
+    "hexadecimal",
+    "octal",
+    "point-mutations",
+    "trinary"
   ],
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -5,9 +5,13 @@
   "active": true,
   "deprecated": [
     "accumulate",
+    "binary",
     "bottles",
     "counter",
-    "point-mutations"
```
petertseng: I regret it, but since

1. trailing comma is not allowed
2. we keep this alphabetical

it was necessary.

There is no rule saying we have to keep it alphabetical, but it started out that way and I didn't think commit cleanliness was a sufficient reason to break it. Unless y'all think it is, in which case I will gladly!

PR #442, opened 2017-01-14
Adding acronym.go stub for solution

usmanismail:


bordeltabernacle:
Looks good to me @usmanismail thanks! :+1: 

@petertseng I think it would be good to get the `config.json` more filled out before considering exercise placement, then there'll at least be an agreed rating for comparison. We could also then re-consider `triangle` as the stub cut-off if necessary.

PR #444, opened 2017-01-16
leap: update example_gen.go & test cases

bordeltabernacle:
the filepath needed used by example_gen.go was changed in #224
the test cases were changed in this commit:
https://github.com/exercism/x-common/pull/463

### DIFF
--- a/exercises/leap/example_gen.go
+++ b/exercises/leap/example_gen.go
@@ -15,7 +15,7 @@ func main() {
 		log.Fatal(err)
 	}
 	var j js
-	if err := gen.Gen("leap.json", &j, t); err != nil {
+	if err := gen.Gen("leap", &j, t); err != nil {
 		log.Fatal(err)
 	}
 }
--- a/exercises/leap/cases_test.go
+++ b/exercises/leap/cases_test.go
@@ -1,17 +1,15 @@
 package leap
 
 // Source: exercism/x-common
-// Commit: 945d08e Merge pull request #50 from soniakeys/master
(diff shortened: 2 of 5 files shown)


exercises/leap/example_gen.go
```diff
@@ -15,7 +16,8 @@ func main() {
 		log.Fatal(err)
 	}
 	var j js
-	if err := gen.Gen("leap.json", &j, t); err != nil {
+	f := path.Join("exercises", "leap", "canonical-data.json")
```
petertseng: Also reference #357 here.

petertseng: Edit: I mean in the commit message, not literally "here"

petertseng: Consider the effect of making this change across every single generator - we will be typing `path.Join("exercises", whateverSlug, "canonical-data.json")` a lot. Can/should this line be moved into the `Gen` function, and we just pass `leap` into `Gen`?

I won't oppose this change for this PR, but I'd advise it if working on more generators.

bordeltabernacle: I totally missed this issue, I stumbled upon this through trying to implement a test generator in another exercise.
I was thinking the same thing, I'll look into cancelling my other PR and updating the `Gen` function. Leave this one unmerged and hopefully I can update it along with the others.

petertseng: You probably don't have to cancel it - all the generators will still need to update to pass in only the slug rather than the filename (`leap` rather than `leap.json`), can just repurpose it to do the `Gen` update in that PR. Whichever is easier for you if you're going to work on it though.

PR #446, opened 2017-01-16
update all example_gen.go files and README.md

bordeltabernacle:
reference: #447 

README details on generating tests also updated to reflect moving
exercises into exercises subdirectory

### DIFF
--- a/exercises/clock/example_gen.go
+++ b/exercises/clock/example_gen.go
@@ -15,7 +15,7 @@ func main() {
 		log.Fatal(err)
 	}
 	var j js
-	if err := gen.Gen("clock.json", &j, t); err != nil {
+	if err := gen.Gen("clock", &j, t); err != nil {
 		log.Fatal(err)
 	}
 }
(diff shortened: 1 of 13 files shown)


exercises/clock/example_gen.go
```diff
@@ -15,7 +16,8 @@ func main() {
 		log.Fatal(err)
 	}
 	var j js
-	if err := gen.Gen("clock.json", &j, t); err != nil {
+	f := path.Join("exercises", "clock", "canonical-data.json")
```
petertseng: The `path.Join("exercises", whateverslug, "canonical-data.json")` is common to every single `example_gen.go`. How would you feel about moving it into the `Gen` function?

I would like to be able to say `gen.Gen(whateverSlug, &j, t)`, and have it infer the path from the slug. That means further reorganisational changes like https://github.com/exercism/x-common/pull/351 can be dealt with by just changing `Gen`, rather than every single generator.

petertseng: (I know I kinda repeated myself and you already responded in the other PR, sorry about that)

bordeltabernacle: no worries, I won't repeat myself here, leave this PR for the time being and hopefully I'll have the `gen.Gen` function working this way soon.

Issue #445, opened 2017-01-16
Generating tests references most recent commit, not relevant commit


bordeltabernacle:
In updating the `leap` exercise tests in #444 the `example_gen.go` script adds in the most recent x-common commit:

```
// Commit: cfdae36 Merge pull request #497 from exercism/stkent-patch-1
```

which has no relevance to the updated leap tests, which were changed in [this commit](https://github.com/exercism/x-common/pull/463)

I don't know the test generation code well enough, or git well enough, to fix this currently, but will try and spend some time this week understanding it better.

In the meantime I wanted to check if this was known and deliberate or if someone else was easily able to fix if not?

petertseng:
How does the generator even *know* what commit it is? I'll read the code real quick...


oh, it just does `exec.Command` on `git log`.

Pass the filename of the JSON file into that `git log` command, and it should be all good.

Compare the effects of running these two commands in the `x-common` repo:

```
git log -1 --oneline
git log -1 --oneline exercises/leap/canonical-data.json
```

Should be a good fix to make, because we should keep the commit relevant.

PR #447, opened 2017-01-16
update gen.Gen function

bordeltabernacle:


gen/gen.go
```diff
@@ -44,10 +44,11 @@ func init() {
 	}
 }
 
-func Gen(jFile string, j interface{}, t *template.Template) error {
+func Gen(exercise string, j interface{}, t *template.Template) error {
 	if dirMetadata == "" {
 		return errors.New("unable to determine current path")
 	}
+	jFile := filepath.Join("exercises", exercise, "canonical-data.json")
```
bordeltabernacle: I'm not sure how idiomatic this is. Is it worth moving the "exercises" and "canonical-data.json" strings into variables within the function or even package level so they're further up the file, making them more obvious and potentially easier to change in the future?? or not??

petertseng: For me personally, I would 100% do it if either of the following were true:

1. they were magic numbers; numbers without explanations/names seem much harder to understand
2. A line containing "exercises" and "canonical-data.json" appeared more than once, since I don't like repetition.

I feel less stringent about these... I guess you call these magic strings? I would definitely not oppose creating a `jsonPath(exercise string) string` function (you can name that whatever you want, I'm not insistent on that name), but I also don't mind having it inlined.

Also part of my decision is that it doesn't seem very likely to change, and it is also not very hard to understand what it is doing by reading it.

bordeltabernacle: ok, super, I'll leave it as is, as you say it's not hard to understand nor is it repeated. I do like the idea of a `jsonPath` function but it also feels like maybe one layer of abstraction too many.

bordeltabernacle:
#357 & #445 do both get mentioned in the commits. I've tested this and it works as before. If you're happy with this I'll merge and update my other open PR's.

petertseng:
> #357 & #445 do both get mentioned in the commits. I've tested this and it works as before. If you're happy with this I'll merge and update my other open PR's.

Oops, that's what I get for not expanding the commit messages. I'm usually better about this, I have no excuse (I'm in a bit of a hurry, but no excuse).

OK, given that you've tested it go ahead and merge. I may not get to review your other PRs for a few hours, depending. We'll see.

petertseng:
Oh yeah, those commit messages do a good job of explaining why the change is made. I definitely shouldn't  have missed them!

bordeltabernacle:
hey, no worries, don't feel obliged at all, I'm going to just update them and then work on other things, so whenever you get a chance, and no excuse needed, c'est la vie y'know. :smile:

PR #449, opened 2017-01-16
Stub for pangram problem

usmanismail:
Also adding stub solution for pangram as As referenced in #442


bordeltabernacle:
super, thanks @usmanismail

PR #452, opened 2017-01-17
exercises: remove executable bits from *.go files; gen: Write files non-executable (0666, not 0777)

petertseng:
For reasons not known to me, these files are all marked executable,
though one will find that an attempt to execute them directly will not
succeed (they have no shebang line, so a Unix shell might naively
attempt to execute them as shell script, with hilariously bad results)


petertseng:
Oh, I see, the generator writes them as executable by default.

gen/gen.go
```diff
@@ -83,7 +83,7 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
 		return err
 	}
 	// write output file for the Go test cases.
-	return ioutil.WriteFile(filepath.Join(dirProblem, "cases_test.go"), src, 0777)
+	return ioutil.WriteFile(filepath.Join(dirProblem, "cases_test.go"), src, 0666)
```
petertseng: I can make it 0644 if someone really cares, though I expect people's umasks to mask out group write/world write, so I leave it as 0666. To wit, on my machine the umask is 077, so the file gets created 0600.

bordeltabernacle:
good catch! :+1:

petertseng:
in case someone wonders - this does not affect what will be served to the students, because the CLI has its own idea of what files should be executable (none of them), so this change is only visible to people who install trackler directly or who look at this repo. This helps these people avoid confusion.

Issue #328, opened 2016-06-02
Updated tests for the Custom Set problem


ryanplusplus:
In order to reduce the amount of code required to pass incremental tests (assuming that users pass tests starting from the top), the order of the tests was modified slightly.

Since this track implements Custom Set, please take a look at the new custom-set.json file and see if your track should update its tests.

If you do need to update your tests, please refer to this issue in your PR. That helps us see which tracks still need to update their tests.

If your track is already up to date, go ahead and close this issue.

More details on this change are available in https://github.com/exercism/x-common/pull/257.

bordeltabernacle:
closed by #454

PR #454, opened 2017-01-17
custom-set: update, remove and re-order tests

bordeltabernacle:
updated tests for custom-set #328

for reference:
https://github.com/exercism/x-common/pull/232
https://github.com/exercism/x-common/pull/257

### DIFF
--- a/exercises/custom-set/custom_set_test.go
+++ b/exercises/custom-set/custom_set_test.go
@@ -6,20 +6,16 @@ package stringset
 //
 // New() Set
 // NewFromSlice([]string) Set
-// (s Set) Add(string)         // modify s
-// (s Set) Delete(string)      // modify s
-// (s Set) Has(string) bool
-// (s Set) IsEmpty() bool
-// (s Set) Len() int
-// (s Set) Slice() []string
 // (s Set) String() string
-// Equal(s1, s2 Set) bool
-// Subset(s1, s2 Set) bool     // return s1 ⊆ s2
+// (s Set) IsEmpty() bool
+// (s Set) Has(string) bool
+// Subset(s1, s2 Set) bool
 // Disjoint(s1, s2 Set) bool
+// Equal(s1, s2 Set) bool
+// (s Set) Add(string)
 // Intersection(s1, s2 Set) Set
+// Difference(s1, s2 Set) Set
 // Union(s1, s2 Set) Set
-// Difference(s1, s2 Set) Set  // return s1 ∖ s2
-// SymmetricDifference(s1, s2 Set) Set
 //
 // For Set.String, use '{' and '}', output elements as double-quoted strings
 // safely escaped with Go syntax, and use a comma and a single space between
@@ -28,12 +24,11 @@ package stringset
 
 import (
 	"math/rand"
-	"reflect"
 	"strconv"
 	"testing"
 )
 
-const targetTestVersion = 3
+const targetTestVersion = 4
 
 func TestTestVersion(t *testing.T) {
 	if testVersion != targetTestVersion {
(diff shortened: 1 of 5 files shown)


bordeltabernacle:
"I'll just quickly regenerate these custom-set tests" I thought.

Well, this was an interesting learning exercise. I got a wee bit carried away and I realise I've made some drastic changes to the exercise here, and my brain has melted a little bit, so am very open to any and all criticism and am OK with the possibility of discarding this. :man_shrugging:

exercises/custom-set/custom_set_test.go
```diff
@@ -247,22 +180,6 @@ func TestDifference(t *testing.T) {
 	testBinOp("Difference", Difference, differenceCases, t)
 }
 
-func TestSymmetricDifference(t *testing.T) {
-	testBinOp("SymmetricDifference", SymmetricDifference, symmetricDifferenceCases, t)
-}
-
-func BenchmarkNewFromSlice1e1(b *testing.B) { bench(1e1, b) }
```
petertseng: does the benchmarking not make any sense anymore? or was it jus tmoved?

bordeltabernacle: removing this was just an accident.

bordeltabernacle:
Honestly, I've not done the exercise yet, and this is a rather opinionated change that has not come from a need to improve the exercise from it's current form. What I think I'll do is change it back to using the `String` as before, leaving the structure of what a set is open for interpretation, and then that change can be a separate PR if there's a qualified view that such a change needs to be done. Leaving this PR to just the updates from the canonical-data changes.

bordeltabernacle:
ok, I've updated this so that it just reflects the changes in the canonical data now.

exercises/custom-set/custom_set_test.go
```diff
@@ -194,22 +131,25 @@ func TestHas(t *testing.T) {
...
-	for _, tc := range emptyCases {
+func TestAdd(t *testing.T) {
+	for _, tc := range addCases {
 		s := NewFromSlice(tc.set)
-		got := s.IsEmpty()
-		if got != tc.want {
-			t.Fatalf("%v IsEmpty = %t, want %t", s, got, tc.want)
+		s.Add(tc.ele)
+		want := NewFromSlice(tc.want)
+		if !Equal(s, want) {
```
petertseng: There's an interesting ordering quirk now that we've done this. In the previous version, the `Equal` test came very early. Notice how there is a test that says "With Equal tested, remaining tests use it to judge correctness." Whereas here, a test that uses `Equal` comes before the actual test of `Equal`, so it seems that we miss this property. This, I see, was a guiding desire for the reordering done in https://github.com/exercism/x-common/pull/257. Might be better to keep it that way.  Unless there is a strong reason to do it this way (I think you made it so that all binary operations happen in sequence, etc.)

exercises/custom-set/example_slice.go
```diff
@@ -26,27 +32,26 @@ func NewFromSlice(l []string) Set {
 	return s
 }
 
-func (s Set) Add(v string) {
-	for _, e := range *s.s {
-		if e == v {
-			return
+// String returns a printable representation of s.
+func (s Set) String() string {
```
petertseng: I got a distinct sense of deja vu when reading this file, since it appears you made the exact same change in example.go. Then I realized, wow, we have two example files in here, one with `[]string` and one with `map[string]bool`, and control which one to build with build flags. Though nowhere in Travis do we build with any build flags, so this means this file is never tested, and is susceptible to going out of date. Maybe one day we should test it. But that's a talk for another time. No change needed at this moment

exercises/custom-set/custom_set_test.go
bordeltabernacle: ah, yes, I remember what I did now, I had them all in the order suggested, and then forgot while cleaning things up a bit and grouped all the methods together, silly oversight on my part, thanks again for the catch @petertseng

bordeltabernacle:
ok, I think that should be everything now?

exercises/custom-set/custom_set_test.go
```diff
@@ -6,20 +6,16 @@ package stringset
...
-// (s Set) Has(string) bool
-// (s Set) IsEmpty() bool
-// (s Set) Len() int
-// (s Set) Slice() []string
 // (s Set) String() string
-// Equal(s1, s2 Set) bool
-// Subset(s1, s2 Set) bool     // return s1 ⊆ s2
+// (s Set) IsEmpty() bool
+// (s Set) Has(string) bool
+// (s Set) Add(string)
```
petertseng: Should these be listed in the same order they are tested? I think that was what you were going for. In that case, let's move this one down.

bordeltabernacle:
Cool, done, good to merge?

PR #455, opened 2017-01-17
Implement transpose

bordeltabernacle:


exercises/transpose/cases_test.go
```diff
@@ -0,0 +1,22 @@
...
...uance of their parents' rage,", "Which, but their children's end, naught could remove,", "Is now the two hours' traffic of our stage;", "The which if you with patient ears attend,", "What here shall miss, our toil shall strive to mend."}, []string{"CIFWFAWDTAWITW", "hnrhr hohnhshh", "o oeopotedi ea", "rfmrmash  cn t", ".a e ie fthow ", " ia fr weh,whh", "Trnco miae  ie", "w ciroitr btcr", "oVivtfshfcuhhe", " eeih a uote  ", "hrnl sdtln  is", "oot ttvh tttfh", "un bhaeepihw a", "saglernianeoyl", "e,ro -trsui ol", "h uofcu sarhu ", "owddarrdan o m", "lhg to'egccuwi", "deemasdaeehris", "sr als t  ists", ",ebk 'phool'h,", "  reldi ffd   ", "bweso tb  rtpo", "oea ileutterau", "t kcnoorhhnatr", "hl isvuyee'fi ", " atv es iisfet", "ayoior trr ino", "l  lfsoh  ecti", "ion   vedpn  l", "kuehtteieadoe ", "erwaharrar,fas", "   nekt te  rh", "ismdsehphnnosa", "ncuse ra-tau l", " et  tormsural", "dniuthwea'g t ", "iennwesnr hsts", "g,ycoi tkrttet", "n ,l r s'a anr", "i  ef  'dgcgdi", "t  aol   eoe,v", "y  nei sl,u; e", ",  .sf to l   ", "     e rv d  t", "     ; ie    o", "       f, r   ", "       e  e  m", "       .  m  e", "          o  n", "          v  d", "          e  .", "          ,"}},
```
petertseng: This might be a tall order because this is generated, but what if these were shown such that each element of the slice goes on a separate line?

I don't know if we should do that for all of them, or only texts for which the size exceeds some arbitrary threshold (which includes this case)

exercises/transpose/transpose_test.go
```diff
@@ -0,0 +1,41 @@
+package transpose
+
+import (
+	"fmt"
+	"math"
+	"testing"
+)
+
+// For a more visual representaion of the test data than is provided by
```
petertseng: representation

exercises/transpose/transpose_test.go
```diff
@@ -0,0 +1,41 @@
...
+				}
+			}
+		}
+	}
+}
+
+func BenchmarkTranspose(b *testing.B) {
+	b.StopTimer()
+	for _, test := range testCases {
+		b.StartTimer()
```
petertseng: We might like to do the same as #453 and remove these, since there isn't really any initialisation code required (the usual reason for starting/stopping timer). I suppose one could consider that iterating over the `testCases` is something, but I don't think it has such a large impact. I wish I had examples so that I could know whether I'm right on that.

exercises/transpose/transpose_test.go
```diff
@@ -0,0 +1,41 @@
...
+)
+
+// For a more visual representaion of the test data than is provided by
+// cases_test.go have a look at the original json data here:
+// https://github.com/exercism/x-common/blob/master/exercises/transpose/canonical-data.json
+
+func TestTranspose(t *testing.T) {
+	for _, test := range testCases {
+		actual := Transpose(test.input)
+		if fmt.Sprintf("%s", actual) != fmt.Sprintf("%s", test.expected) {
```
petertseng: any risk of a false positive from this? For example:

```go
package main

import (
	"fmt"
)

func main() {
	a := []string{"a b", "c"}
	b := []string{"a", "b c"}
	fmt.Printf("%s and %s are equal? %t", a, b, fmt.Sprintf("%s", a) == fmt.Sprintf("%s", b))
}
```

https://play.golang.org/p/uvtXMhdJ8W

exercises/transpose/transpose_test.go
```diff
@@ -0,0 +1,41 @@
...
+			// find the shorter of actual and expected, to avoid index out of range errors
+			min := int(math.Min(float64(len(test.expected)), float64(len(actual))))
```
petertseng: interesting, guess you ran into http://stackoverflow.com/questions/27516387/what-is-the-correct-way-to-find-the-min-between-two-integers-in-go

(no change needed, unless you want to go write that `min` function and then use it)

exercises/transpose/cases_test.go
petertseng: editing to add "(and so we don't necessarily have to do it)"

bordeltabernacle: I agree, before I generated the test cases, I had just copied the data over from x-common and it does make a huge difference having it formatted as you suggest, seeing the rows and columns visually gives you a better sense of what the problem is. As you say though, this file is generated, so I'm not sure how we could get around this. this is why I put in the line about looking at the canonical data in the x-common github, which is less than satisfactory.
I'm almost inclined to not generate the test cases in this instance (unless I can find a way for the generation to format them) as I'm leaning towards the value of them being visually formatted being greater.

exercises/transpose/transpose_test.go
bordeltabernacle: ah sheesh, never even occurred to me, thanks!
I can fix it like this right? https://play.golang.org/p/LhVDPUCHG-

exercises/transpose/transpose_test.go
bordeltabernacle: No, I didn't see that SO question, I think I like the min function better though.

exercises/transpose/transpose_test.go
bordeltabernacle: :+1:

exercises/transpose/transpose_test.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// For a more visual representation of the test data than is provided by
+// cases_test.go have a look at the original json data here:
+// https://github.com/exercism/x-common/blob/master/exercises/transpose/canonical-data.json
+
+func TestTranspose(t *testing.T) {
+	for _, test := range testCases {
+		actual := Transpose(test.input)
+		if fmt.Sprintf("%#v", actual) != fmt.Sprintf("%#v", test.expected) {
```
petertseng: yeah sure this seems like it has a better chance of being correct. There could be cases I'm not aware of, in which case we may get forced to use https://golang.org/pkg/reflect/#DeepEqual instead, but I at least can't think of a false positive on this one

exercises/transpose/cases_test.go
petertseng: If you are interested in seeing whether there is a way, it would be good to look at https://golang.org/pkg/text/template/ - from a quick glance the `range` pipeline seems useful, but I have not experimented enough to see.

bordeltabernacle: I've tried changing the template so that it loops over the slice in a range loop, but because the first test is an empty string it has nothing to loop over, and so nothing appears.

petertseng: Hmm, that might be OK to me. For the empty array `[]` I would want want `[]string{}` to appear, and for `["a", "b"]` I would want `[]string{"a", "b"}` to appear. So either way, I need `[]string{}`, and then print each element of the list inside.

Though if the first case does require special handling, perhaps that `{{range pipeline}} T1 {{else}} T0 {{end}}` is helpful?

I'll try to experiment after I finish some unrelated work.

petertseng: Something that seems plausible.

```go
package main

import (
	"os"
	"text/template"
)

type Inventory struct {
	Things []string
}

func main() {
	no := func(err error) {
		if err != nil {
			panic(err)
		}
	}

	tmpl, err := template.New("test").Parse(`[]string{
{{range .Things}}  {{printf "%q" .}},
{{end}}},
`)
	no(err)

	no(tmpl.Execute(os.Stdout, Inventory{[]string{"a", "b"}}))
	no(tmpl.Execute(os.Stdout, Inventory{[]string{}}))
}
```

https://play.golang.org/p/lqNYRaE9yx

bordeltabernacle: ah man, this is perfect, I don't know what I was doing wrong, I had pretty much the same, I think I just had the `[]string{` in the wrong place. I obviously gave up too soon, embarrassing, sorry to make you do the work for me. I'll push up the change now.

exercises/transpose/transpose_test.go
bordeltabernacle: I've changed to to `reflect.DeepEqual` anyway, just to avoid any unforeseen problems and as it may get changed to this in the future anyway.

bordeltabernacle:
Thanks for your help with this @petertseng I really appreciate it. I've learnt a ton from implementing this exercise and from the changes you've pointed out.

PR #456, opened 2017-01-18
queen-attack: Add edge test cases

leenipper:
Queen-attack programs which are "off by 1"
in their "valid rank" test will not fail since
there are no test cases to cover a valid rank of 8
or an invalid rank of 0.

Add these test cases:
 - off board invalid case (a0) - rank 0
 - no attack case (a1, f8)
 - attack cases at opposite corners

Bump testVersion to 2 to reflect new test cases.

### DIFF
--- a/exercises/queen-attack/example.go
+++ b/exercises/queen-attack/example.go
@@ -2,7 +2,7 @@ package queenattack
 
 import "fmt"
 
-const testVersion = 1
+const testVersion = 2
 
 func CanQueenAttack(w, b string) (attack bool, err error) {
 	if err = valSq(w); err != nil {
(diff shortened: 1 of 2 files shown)


leenipper:
RE: whitespace difference: Yeah, the automatic "go fmt" action of my editor accounts for that, and I didn't like the fact that the comment changes obscured the net change.  I also didn't want to commit anything which was not taken through "go fmt" either.  I should have made 2 commits with just a little more effort.

RE: i5 test case:  Yes, good point, I could fold that in as well.

I didn't know about canonical-data.json. I thought the algebraic chess notation was a good choice, and of course will stay for now.

So, is it best to just close this PR without a merge ?  I can prepare a new PR with the 2 commits separated and also add an invalid file test case, ie i5.  Sound good ?

leenipper:
I've got the new PR branch ready, so I will close this one.

petertseng:
(small note, it would have also been OK to force push on the `update-queen-attack-test-cases` which would cause this PR to use the new commits instead of the old, but either way is OK)

PR #458, opened 2017-01-19
Update queen attack edge cases

leenipper:
This PR adds some test case robustness.
Two commits are used to separate test case changes from "go fmt" changes.


leenipper:
I like whitespace changes placed in separate commits too, it is good to know if a change set is only cosmetic.   Thanks!

Issue #448, opened 2017-01-16
clock: recommendation in test file confusing


bordeltabernacle:
In the clock exercise test file there is [the following help](https://github.com/bordeltabernacle/xgo/blob/master/exercises/clock/clock_test.go#L18-L23):

```
// It might help to study the time.Time type in the standard library
// (https://golang.org/pkg/time/#Time) as a model.  See how constructors there
// (Date and Now) return Time values rather than pointers.  Note also how
// most time.Time methods have value receivers rather than pointer receivers.
// For more background on this read
// https://github.com/golang/go/wiki/CodeReviewComments#receiver-type.
```

As the clock exercise doesn't actually utilise the `time` package, though there is perhaps the assumption it does due to the exercise name, this help could be construed as confusing, a possible red herring? I know I was confused by it when I did this exercise, and I recently had [this conversation](http://exercism.io/submissions/170c7069ac1a464c82b03cd21f3b2df7) in a nitpick.

We could maybe

1. re-write it in terms of "While this exercise doesn't necessarily require the time package..." though this may lead people too much
2. use a different package as a suggestion, that doesn't hold as much confusion
3. remove entirely

petertseng:
Trying to see why it was placed there. Guess it was because of #125 (based off of discussion in #121), and the main point to be made is that `New` must return a struct, not a pointer to one, otherwise `==` will not work.

What if we just stated this outright? Something like "this means that if your `New` function returns a pointer, your clocks will probably not work with =="? We could leave the https://github.com/golang/go/wiki/CodeReviewComments#receiver-type link if it seems useful, but since that link tries to answer the question "should I use a pointer or value receiver?" and we just answered that question, it may be redundant. Perhaps we could frame it as "in this instance, you are using a value receiver, but this link tells you in what situations you might make a different choice", or something?

bordeltabernacle:
The PR's you reference are interesting and kinda help to clarify what was trying to be achieved by adding the comments in the first place. I think it's interesting information and useful to a new go programmer, I think the confusion just comes from the `time` package being referenced in an exercise called clock. I'll have a go at trying lightly re-wording this to get across some of the discussion from the original PR's.

PR #460, opened 2017-01-20
clock: clarify comments

bordeltabernacle:
issue: #448
also of note: #125 & #121

### DIFF
--- a/exercises/clock/clock_test.go
+++ b/exercises/clock/clock_test.go
@@ -12,15 +12,20 @@ import (
 // (Clock) Add(minutes int) Clock
 //
 // The Add method should also handle subtraction by accepting negative values.
-// To satisfy the readme requirement about clocks being equal, values of
-// your Clock type need to work with the == operator.
+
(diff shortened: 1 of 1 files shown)


exercises/clock/clock_test.go
```diff
@@ -12,15 +12,20 @@ import (
...
+
+// To satisfy the README requirement about clocks being equal, values of
+// your Clock type need to work with the == operator. This means that if your
+// New function returns a pointer rather than a value, your clocks will
+// probably not work with ==.
+//
+// While the time.Time type in the standard library (https://golang.org/pkg/time/#Time)
+// doesn't necessarily need to be used as a basis for your Clock type, it might
+// help to look at how constructors there (Date and Now) return values rather
+// than pointers. Note also how most time.Time methods have value receivers
```
petertseng: wow, I was expecting to ask this to change to "all" instead of "most", but indeed there are some functions with each. I know `go lint` will advise to keep them consistent, but I'm assuming they have a good reason for donig what they've done.

(No change needed for this comment)

PR #461, opened 2017-01-20
config.json: add topics to etl

bordeltabernacle:
I don't think the dificulty needs to be changed for this. (#345)


bordeltabernacle:
note: this is out of order, as I've just completed it.

Issue #465, opened 2017-01-26
difference-of-squares test has no targetTestVersion


tleen:
[The test for difference-of-squares](https://github.com/exercism/xgo/blob/master/exercises/difference-of-squares/difference_of_squares_test.go) currently does not use the targetTestVersion const to sync up solutions with the tests. I believe a few others do not either. Should all the exercises be using these or is it the sort of thing that you only start using on a second version of a test?

petertseng:
I'll defer to the README's https://github.com/exercism/xgo#problem-versioning section on this point, added in #140 . It says

> Each problem defines a `const targetTestVersion` in the test program

Since it does say "each", then it would appear that any problem that fails to define a `targetTestVersion` is in violation of the README.

Assuming we agree with the reasoning for having test versions (this is someone's opportunity to speak up, if you think they are not useful), they should be added or the sentence in the README should be changed.

robphoenix:
I'd lean towards all exercises starting from a point of having a targetTestVersion that can be incremented as necessary.

tleen:
The point of it is to sync up the exercise with the test to ensure we don't have users running their source file against a test that may cause incorrect fails or passes?

This looks like the kind of thing that may have been dropped then starter files we not provided with the exercises? 

I can see an argument that if there has been no change to the test than the test version doesn't have to be checked (or is implicitly at 0), and if you change the initial test then you have to start using the targetTestVersion at 1. 

Either way it is inconsistent in the code base and should probably be brought into alignment across the exercises. Either explicitly not needing it for the first version or requiring it across the board.

robphoenix:
I agree, it should certainly be consistent across all the exercises. I think starting each exercise off with a `testVersion`, whether 1 or 0, is better than not having one, as the intention is to use it if there is ever a significant change to the exercise or it's tests. It's an indication there is some mechanism in place for accommodating changes to tests, rather than it coming into play only if there is a change.
I think this will also cause less confusion regarding some exercises having a `testVersion` and some not.

tleen:
I can agree with that, explicit better than implicit.

tleen:
I can get on adding all the test vars if if what we want to do.

robphoenix:
Are you happy with this @petertseng?

petertseng:
Yup, we'd be glad to have them!

robphoenix:
Good stuff, thanks for making this happen @tleen

tleen:
Ok will do!

tleen:
[...]

PR #469, opened 2017-01-30
moved target test versioning test before problem specific test

tleen:
I just did accumulate, added test versioning and moved the version test to be the first function in the test file. Wanted to be sure its correct before starting the rest. See #465


tleen:
Based on consensus in #465, not going to move the test version test to the top so this is no good.

Issue #470, opened 2017-01-30
Ensure consistency across exercise tests


robphoenix:
#465 highlighted certain test file inconsistencies between the exercises.
These were:

1. lack of `testVersion/targetTestVersion` for all exercises, initialised at 1, to be incremented upon significant changes
2. separate `TestTestVersion` function, rather than as part of main test function
3. placement of `TestTestVersion` function as the first test function
4. `TestTestVersion` function using `Fatalf` rather than `Errorf`

robphoenix:
- [ ] accumulate
- [ ] acronym
- [ ] all-your-base
- [ ] allergies
- [ ] anagram
- [ ] atbash-cipher
- [ ] bank-account
- [ ] beer-song
- [ ] binary-search-tree
- [ ] binary-search
- [ ] binary
- [ ] bob
- [ ] bracket-push
- [ ] circular-buffer
- [ ] clock
- [ ] connect
- [ ] counter
- [ ] crypto-square
- [ ] custom-set
- [ ] diamond
- [ ] difference-of-squares
- [ ] diffie-hellman
- [ ] error-handling
- [ ] etl
- [ ] food-chain
- [ ] gigasecond
- [ ] grade-school
- [ ] grains
- [ ] hamming
- [ ] hello-world
- [ ] hexadecimal
- [ ] house
- [ ] isogram
- [ ] kindergarten-garden
- [ ] largest-series-product
- [ ] leap
- [ ] ledger
- [ ] luhn
- [ ] matrix
- [ ] meetup
- [ ] minesweeper
- [ ] nth-prime
- [ ] nucleotide-count
- [ ] ocr-numbers
- [ ] octal
- [ ] paasio
- [ ] palindrome-products
- [ ] pangram
- [ ] parallel-letter-frequency
- [ ] pascals-triangle
- [ ] perfect-numbers
- [ ] phone-number
- [ ] pig-latin
- [ ] poker
- [ ] pov
- [ ] prime-factors
- [ ] protein-translation
- [ ] pythagorean-triplet
- [ ] queen-attack
- [ ] raindrops
- [ ] react
- [ ] rna-transcription
- [ ] robot-name
- [ ] robot-simulator
- [ ] roman-numerals
- [ ] saddle-points
- [ ] say
- [ ] scrabble-score
- [ ] secret-handshake
- [ ] series
- [ ] sieve
- [ ] simple-cipher
- [ ] strain
- [ ] sum-of-multiples
- [ ] tournament
[...]

PR #471, opened 2017-01-31
consistent test versioning in accumulate exercise, see #470

tleen:


tleen:
Let me squash these commmits first.

PR #472, opened 2017-01-31
added test versioning consistency to accumulate, see #470

tleen:


tleen:
If this is good we can use it as the model for the other tests.

exercises/accumulate/example.go
```diff
@@ -1,5 +1,8 @@
 package accumulate
 
+// TestVersion constant declaration - part of the boilerplate for exercism.
```
robphoenix: I think we can leave this comment out, it'll be covered in the README and is only necessary in stub files for exercises that have them. This is another thing that is not consistent across the exercises, I'll add it to #470

PR #473, opened 2017-02-01
hello-world: simplify to unconditional "Hello, World!"

petertseng:
The introductory exercise should require as little Go knowledge as
necessary. We'd like to introduce `go test` and `testVersion` here. We
skip introducing the empty string as no string, as this is more specific
knowledge.

https://github.com/exercism/x-common/issues/520

---

I wouldn't recommend merging this until the corresponding change is made in x-common.
Let's get it ready to go before then.

### DIFF
--- a/exercises/hello-world/hello_test.go
+++ b/exercises/hello-world/hello_test.go
@@ -2,25 +2,20 @@ package greeting
 
 import "testing"
 
-// Define a function HelloWorld(string) string.
+// Define a function named HelloWorld that takes no arguments,
+// and returns a string.
+// In other words, define a function with the following signature:
+// HelloWorld() string
 //
 // Also define a testVersion with a value that matches
 // the targetTestVersion here.
 
-const targetTestVersion = 3
+const targetTestVersion = 4
 
 func TestHelloWorld(t *testing.T) {
-	tests := []struct {
-		name, expected string
-	}{
-		{"", "Hello, World!"},
-		{"Gopher", "Hello, Gopher!"},
-		{"ゴーファー", "Hello, ゴーファー!"},
-	}
-	for _, test := range tests {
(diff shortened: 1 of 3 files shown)


exercises/hello-world/hello_test.go
```diff
@@ -2,25 +2,17 @@ package greeting
 
 import "testing"
 
-// Define a function HelloWorld(string) string.
+// Define a function HelloWorld() string.
```
robphoenix: The way this reads is a little confusing because of the return type of the function signature, it takes a second to understand that the word string at the end is part of the function, rather than it saying *Define a function string*.
Maybe moving the function signature to the next line will help?
```
// Define a function with the following signature/the following function (??)
// HelloWorld() string

petertseng: Very nice point, I'll take it.

Another possibility - instead of speaking in terms of the signature, describe the signature in words? "Define a function named Hello World that takes no arguments and returns a string"? Better or worse?

petertseng: > named Hello World

I meant HelloWorld of course. Editing.

robphoenix: Yeah, better definitely, I was uncomfortable with using signature anyway, can confuse the issue further.

petertseng: Since a lot of later exercises do simply show the signature, I'll try to find some way to fit both in, probably, so as to transition the student a bit more smoothly.

petertseng: Now showing both.

robphoenix: Brilliant, nicely introduces what a function signature is.

petertseng:
The time has finally come.

PR #474, opened 2017-02-01
config.json: add topics to triangle, bump difficulty

robphoenix:
see #345

### DIFF
--- a/config.json
+++ b/config.json
@@ -106,9 +106,13 @@
       ]
     },
     {
-      "difficulty": 1,
+      "difficulty": 3,
       "slug": "triangle",
-      "topics": []
+      "topics": [
+        "Control-flow (conditionals)",
+        "Booleans",
+        "Logic"
+      ]
     },
     {
       "difficulty": 1,


robphoenix:
I've removed filtering, your definition gave me some clarity. Really what's going on here is some busy conditional logic that I saw as a kind of filtering process to remove options until one option remained. I think your definition is more accurate, and more in keeping with the idea of an FP filter function.

For each exercise I'm looking at my own experience of completing the exercise, how difficult I found that process, as much as how difficult the eventual solution may be, as even while a solution may be easy once you have it, getting there may not have been. Also whether significant new concepts have been introduced. I also look at a few dozen implementations from others, looking for any significant comments about the exercise, how many variations on the solution there are, and numbers of revisions each person may have gone through, to try and get a sense of what level of challenge the exercise has provided to others. At the same time I'll keep [topics.txt](https://github.com/exercism/x-common/blob/master/TOPICS.txt) open and note down the running themes in all the exercises, checking a few other tracks for what they've added to their config.json.

Mostly I'm trying trying to reach a baseline for the entire config.json and expect multiple reviews and adjustments once it's complete. Of course.

PR #476, opened 2017-02-01
word-search: extract TestTestVersion function (#470)

robphoenix:


robphoenix:
I think also `improve test consistency` is going to cover a possible combination of up to 4 items. For some exercises it could just be changing `Errorf` to `Fatalf`, in others it could be adding in a previously missing `testVersion` const and a `TestTestVersion` function. Maybe `improve test consistency` is too general and it should just reference the specific change and the issue.

robphoenix:
ping @tleen 
This conversation may be of interest as you're working on this issue also.

tleen:
Part of the thing of inconsistency is some exercises will require almost no change, others all the aspects of the test versioning would have to be updated. I can itemize all the changes in one commit with a long message, or I can do each change in separate commits. The latter seems like overkill though. I'm cool with whatever works best for the code history. I'll be tagging #470 in all of them since that is the anchor issue covering all the test versioning changes.

robphoenix:
All the changes can be made in a single commit I think, with as specific a commit message as needed. Obviously if it's a single change the message can be more specific than if multiple changes are made, and I think a general 'improve test consistency` with an extended list in the rest of the commit message is also fine, as you say #470 will be referenced so when log appears more general there is a pointer to more detail.

PR #478, opened 2017-02-01
changed test version check error to fatal in acronym, see #470

tleen:


robphoenix:
Thanks @tleen 

Might be best in future to keep each exercise in a separate PR, just for clarity, and have the first line of the commit be of the form `exercise_name: commit message`

PR #479, opened 2017-02-01
luhn: update test files

robphoenix:
Add an example_gen.go testCases generator to generate the testCases from
the x-common canonical data. Generate the new testCases data and update
the example to pass the new test data. Bump the testVersion to 2.

### DIFF
--- a/exercises/luhn/example.go
+++ b/exercises/luhn/example.go
@@ -1,40 +1,40 @@
 package luhn
 
-const testVersion = 1
+import "strings"
 
-func Valid(n string) bool {
-	if len(n) == 1 {
+const testVersion = 2
+
+func Valid(id string) bool {
+
+	if len(strings.TrimSpace(id)) == 1 {
 		return false
 	}
-	d := extract(n)
-	if len(d) == 0 {
-		return false
+
+	d := make([]int, 0, len(id))
+
+	for _, r := range id {
+		if r == ' ' {
+			continue
+		}
+		if r < '0' || r > '9' {
+			return false
+		}
+		d = append(d, int(r-'0'))
 	}
-	last := len(d) - 1
-	return (check(d[:last])+d[last])%10 == 0
+
+	return sum(d)%10 == 0
 }
 
-func extract(n string) []int {
-	d := make([]int, 0, len(n))
-	for _, r := range n {
-		if r >= '0' && r <= '9' {
-			d = append(d, int(r-'0'))
+func sum(d []int) (s int) {
+	for i, x := range d {
+		j := len(d) - i
+		if j%2 == 0 {
+			x *= 2
+			if x > 9 {
+				x -= 9
+			}
 		}
-	}
-	return d
-}
-
-func check(d []int) int {
-	for i := len(d) - 1; i >= 0; i -= 2 {
-		x := 2 * d[i]
-		if x > 9 {
-			x -= 9
-		}
-		d[i] = x
-	}
-	s := 0
-	for _, x := range d {
 		s += x
 	}
-	return s
+	return
 }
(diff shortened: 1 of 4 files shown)


robphoenix:
I realise exercism/x-common#522 and exercism/x-common#523 are still open and could well affect this PR, so our track should leave this unmerged until they are resolved, and any relevant updates can be made to these commits.

robphoenix:
separated into 2 PRs (#480)

robphoenix:
I've updated this to reflect the merged changes form exercism/x-common#522, the `example.go` has changed again so do make sure you're happy with it before approving.
Still waiting on exercism/x-common#523 before merging

exercises/luhn/example.go
```diff
@@ -1,27 +1,28 @@
...
-	return (check(d[:last])+d[last])%10 == 0
-}
 
-func extract(n string) []int {
-	d := make([]int, 0, len(n))
-	for _, r := range n {
+	d := make([]int, 0, len(id))
+
+	for _, r := range id {
+		if r != ' ' && (r < '0' || r > '9') {
```
petertseng: *personally*, I would strive to avoid the repetition of the '0' and '9', by saying `continue` if `r == ' '`, then the remaining cases are `if`/`else` of each other. Assuming I didn't get my logic wrong. Does it make sense?

exercises/luhn/example.go
```diff
@@ -1,27 +1,28 @@
...
+			return false
+		}
 		if r >= '0' && r <= '9' {
 			d = append(d, int(r-'0'))
 		}
 	}
-	return d
+
+	last := len(d) - 1
+	return (check(d[:last])+d[last])%10 == 0
```
petertseng: I'm assuming `check` returns the sum? Would anything go wrong if we just used `check(d) % 10 == 0`?

E: Oh never mind, this is a same line as above, just moved. I guess we don't need to change it, though if it's going to show up in git diff as a diff anyway, maybe we could consider

exercises/luhn/example.go
robphoenix: that makes better sense, yes

robphoenix:
updated to take into account @petertseng comments. The `sum` function (formerly `check` function) is more deeply nested than I would like but I think makes things a little clearer now that the `Valid` function can return `check(d)%10 == 0` rather than what it did before.

exercises/luhn/example.go
```diff
@@ -1,40 +1,41 @@
...
-	for _, r := range n {
+	d := make([]int, 0, len(id))
+
+	for _, r := range id {
+		if r == ' ' {
+			continue
+		}
 		if r >= '0' && r <= '9' {
 			d = append(d, int(r-'0'))
+		} else {
```
petertseng: this makes me realize there are some alternatives for how to structure this. They all have some advantage or disadvantage, so you can pick one you like after trying out:

1. Don't bother having `if r == ' ' { continue }` above. Instead just have `else if r != ' ' { return false }` here. Pro: Fewer clauses. Cons: May be a little harder to understand in what cases it does neither.

2. Invert the checks, say `if r < '0' || r > '9' { return false }` above. Pro: Now the `d = append` can be unindented. Con: Have to read through two conditions before getting to the actual loop body

petertseng: or obviously 3. leave as is

robphoenix: I chose door no. 2 :+1:

robphoenix:
wait for https://github.com/exercism/x-common/pull/547 & https://github.com/exercism/x-common/pull/549 also.

robphoenix:
I think we can merge this now without waiting on exercism/x-common#523 to be resolved, we can update the exercise when that happens.
Unless there's any objections I think I'll merge this in the next day or so.

robphoenix:
:+1:

PR #482, opened 2017-02-03
pangram: add benchmark test

robphoenix:
With [so many possible solutions](https://twitter.com/kytrinyx/status/826618650979352576) to this I though having a benchmark test would be useful.
I didn't think this warranted a bump in the testVersion but am happy to do so if others think otherwise.

I also checked the test consistency(#470) while I had the file open.


kytrinyx:
:heart: Agreed—if earlier solutions won't fail against the new tests, I don't think we should increment the test version.

PR #484, opened 2017-02-04
anagram: added test versioning, see #470

tleen:


robphoenix:
Awesome, thanks @teen 🎉

PR #494, opened 2017-02-07
forth: Add forth exercise

leenipper:


robphoenix:
This is awesome, thanks @leenipper 

It's going to take me a little time to review and fully digest this, and I'm probably going to defer to more experienced Go devs than myself ( @petertseng @kytrinyx ) so apologies if this doesn't get merged straight away.

leenipper:
I will appreciate all feedback, so no hurry.  Thank you @robphoenix

exercises/forth/cases_test.go
```diff
@@ -0,0 +1,223 @@
...
+	},
+
+	// addition
+	{
+		"can add two numbers",
+		[]string{"1 2 +"},
+		[]int{3},
+	},
+	{
+		"errors if there is nothing on the stack",
```
petertseng: it's a little interesting because there are many tests whose description is "errors if there is nothing on the stack", and it may be hard to tell the difference if I get in my output that one of them has failed. It would make sense if we prepend the section name, making it "addition errors if there is nothing on the stack", but prepending the section name may not make sense for all sections (or maybe it does; I haven't checked all of them). Anyone have a good solution here?

exercises/forth/example.go
```diff
@@ -0,0 +1,243 @@
...
+
+// parse given phrase, returning an operator list, and updating
+// a userDefs map for any user definifion of words in phrase.
+func parse(phrase string, userDefs map[string][]operatorTyp) (oplist []operatorTyp, err error) {
+	words := strings.FieldsFunc(phrase,
+		func(r rune) bool {
+			return unicode.IsSpace(r) || unicode.IsControl(r)
+		})
+
+	var t int // t is token index into words[]
```
petertseng: hmm, any particular reason why this `var t int` line needs to be here? What if it were removed, leaving just the comment `t is token index...` ? would anything bad happen? I think `t` isn't used outside the for loop, but maybe there is something subtle.

exercises/forth/example.go
```diff
@@ -0,0 +1,243 @@
...
+			oplist = append(oplist, udef...)
+		} else if op, ok := builtinOps[w]; ok {
+			if op.id == userDefOp {
+				// Handle user defined word definition.
+				t++
+				if t >= len(words)-2 {
+					return nil, emptyUserDefErr
+				}
+				userword := strings.ToUpper(words[t])
+				_, numerr := strconv.Atoi(userword)
```
petertseng: personal style, so optional: I consider `if _, numerr := strconv.Atoi(userword); numerr == nil` , to make it clear that `numerr` is never used after this, otherwise my brain has to keep track of whether it might be used afterward.

exercises/forth/example.go
```diff
@@ -0,0 +1,243 @@
...
+		return divideByZero
+	}
+	push(stack, v2/v1)
+	return nil
+}
+
+func dup(stack *[]int) error {
+	v1, e1 := pop(stack)
+	if e1 != nil {
+		return notEnoughOperands
```
petertseng: hmm, question, why unconditionally return `notEnoughOperands` here, rather than `e1`? What if `e1` were a different kind of error?

Then again, we know that `pop` will only ever return `notEnoughOperands`, so this question is only academic. So I guess the only benefit to saying `return e1` would be "it saves you the extra typing".

exercises/forth/example_gen.go
```diff
@@ -0,0 +1,77 @@
...
+},
+{{end}}{{end}}
+
+var testCases = []testCase{
+// parsing and numbers{{template "caseSection" .J.ParsingAndNumbers.Cases}}
+// addition{{template "caseSection" .J.Addition.Cases}}
+// subtraction{{template "caseSection" .J.Subtraction.Cases}}
+// multiplication{{template "caseSection" .J.Multiplication.Cases}}
+// division{{template "caseSection" .J.Division.Cases}}
+// combinedArithmetic{{template "caseSection" .J.CombinedArithmetic.Cases}}
```
petertseng: you used "parsing and numbers" above (with spaces), but `combinedArithmetic` (camel) here. any reason for the difference? Obviously it's not too important since it's a comment, but curious.

exercises/forth/forth_test.go
```diff
@@ -0,0 +1,53 @@
...
+
+const targetTestVersion = 1
+
+type testCase struct {
+	description string
+	input       []string
+	expected    []int // nil slice indicates error expected.
+}
+
+func sliceEqual(a, b []int) bool {
```
petertseng: Any strong preferences on using this `sliceEqual` vs https://golang.org/pkg/reflect/#DeepEqual ? Obviously we can tell that `sliceEqual` does exactly what we want and now can avoid having to import `reflect`, but interested to hear.

exercises/forth/forth_test.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	return true
+}
+
+func TestForth(t *testing.T) {
+	if testVersion != targetTestVersion {
+		t.Fatalf("Found testVersion = %v, want %v", testVersion, targetTestVersion)
+	}
+	for _, tc := range testCases {
+		v, err := Forth(tc.input)
```
petertseng: a bit of history here. Back in https://github.com/exercism/xgo/issues/275#issuecomment-206948940 , we decided to make sure that whenever a function asks for an error that we make sure in tests that it is an error by saying `var err error`. Now that effort is completed for current exercises.

Now it is time to decide, do we still want to do that for this exercise? Or do we assume that by the time the student gets to this point the student will probably not do something weird like return something of a different type? It's admitted by me that it's a weird thing to do since it would be unlikely to be in the test suite of a non-TDD project.

exercises/forth/example.go
```diff
@@ -0,0 +1,243 @@
...
+	push(stack, v1+v2)
+	return nil
+}
+
+func subtract(stack *[]int) error {
+	v1, v2, err := pop2(stack)
+	if err != nil {
+		return err
+	}
+	push(stack, v2-v1)
```
petertseng: interesting fact is that this function only differs from `add` and `multiply` by this one line. One might consider deduplicating the code even more...

but we don't necessarily need to see that in the example solution; the example solution can be quite straightforward. Maybe in your submission to the site, then.

exercises/forth/example.go
```diff
@@ -0,0 +1,243 @@
...
+	fn operatorFn
+	id operatorId
+}
+
+func Forth(input []string) (result []int, err error) {
+	if len(input) == 0 {
+		return result, nil
+	}
+
+	stack := make([]int, 0, 8)
```
petertseng: any particular reason for 8, or arbitrary? I think either way, a comment explaining why 8 will help.

exercises/forth/cases_test.go
leenipper: In an earlier version, I had used multiple testCases, one for additionTestCases, subtractionTestCases, and so forth, because the canonical-data.json (read by gen.go and parsed by the template example_gen.go) was "sectioned". I was able to output the section "name", like "addition" when an error occurred so it addressed these duplicate instances of "error if nothing on the stack".  It did make the test program longer since it used different TestXxxx functions, although each called a "runTest" helper to actually do the mechanics. When running go test, the multiple TestXxxx allowed multiple errors to occur, failing at most one in each "section".  I thought it best to keep only one error in front of the student at one time.  So, I redid the text template in example_gen to put all the test cases together, and that made go test stop on first error, allowed a very simple forth_test and also a simple benchmark function as well.  Long story for this, but the multiple "errors if there is nothing on the stack" originates from the descriptions in canonical-data.json.  I'm thinking of two options:  1) Use the "sectioned" test cases again to allow the error output to indicate which section of the failed case, or 2) Update the canonical-data.json in x-common/exercises/forth to use more specific descriptions, like "+ causes error if nothing is on the stack", and "dup causes error if nothing is on the stack".   I think I like option 2 better(different pull request for that repo), but can easily be convinced to do option 1, since I have that code on a different branch already. Also, is there is a way to stop on first error when you have multiple TestXxxx functions?  Maybe that is a minor thing, as I think I encountered it while "exercising" with go.

exercises/forth/example.go
leenipper: I want to populate he slice with _some_ capacity, and simply picked 8.  Can add a comment.

exercises/forth/example.go
leenipper: True, t should be loop scope only.  And the comment can stand alone without the var.  I thought by putting the comment on a line by the var it was helpful, but forgot about the scope extension past the for loop.  Good point.

exercises/forth/example.go
leenipper: Ah, yes, using the short declaration inside the if will keep the scope tighter. Will do.

exercises/forth/example.go
leenipper: Yeah.  I can refactor the add, subtract, and multiply to be one liners which use a binaryOp function. Divide must stand alone due to divideByZero.

exercises/forth/example.go
leenipper: Oversight on my part.  If pop() returns an error, it is cleaner to just return the error, as it removes the reader from wondering about the error being "changed".

exercises/forth/example_gen.go
leenipper: Accidental.  I had tried to match the original section name spelling in canonical-data.json, but missed that one.

exercises/forth/forth_test.go
leenipper: I will use DeepEqual.  I had not used reflect before.  But in doing so, I discovered an error in my example.go, where it was returning a nil slice instead of an empty slice. Glad that was caught. I found out that DeepEqual returns false for two slices a and b, if a is nil but b is an empty slice; however len(a) and len(b) are both zero, and a for loop ranging over either a or b will execute zero times, which was what my sliceEqual was doing.  Learning more about Go all the time.

exercises/forth/forth_test.go
leenipper: After reading #275 I am wondering which of these is preferred:
```
var err error
v, err := Forth(tc.input)
```
or instead like this:
```
v, err := Forth(tc.input)
var _ error = err
``` 
?
Seems that both use the type system to insure 2nd return value is of type error. Any preference ? Looks like the trend in most exercises is the latter form.

leenipper:
I am taking care of each review item, with the exception of the first one about error... nothing on the stack, as that one isn't settled on the choice yet.  So, I will proceed to rebase, and force push soon. Thank you @petertseng  for the review feedback.

exercises/forth/cases_test.go
petertseng: > I thought it best to keep only one error in front of the student at one time.

Yeah, that seems like the way we like to do it here, so if possible we should try to maintain that property.

> Also, is there is a way to stop on first error when you have multiple TestXxxx functions?

Yeah, I hadn't found any I think, because `Fatalf` only stops the current test function, but still lets the other ones run.

> I'm thinking of two options: 1) Use the "sectioned" test cases again to allow the error output to indicate which section of the failed case

Since we've said we want to keep only one error for the student at a time, would these sections all be run from one function?

> 2) Update the canonical-data.json in x-common/exercises/forth to use more specific descriptions, like "+ causes error if nothing is on the stack", and "dup causes error if nothing is on the stack"

The interesting thing is having the descriptions like that make sense for test frameworks that automatically output the test output in a format like...

```
forth
  parsing and numbers
    empty input results in empty stack
    numbers just get pushed onto the stack
    all non-word characters are separators
  addition
    can add two numbers
    errors if there is nothing on the stack
    errors if there is only one value on the stack
  subtraction
    can subtract two numbers
    errors if there is nothing on the stack
    errors if there is only one value on the stack
```

I don't think we necessarily need to get that fancy, but what if, when we generate the test descriptions, we prepend the section to the description? I'll show in the code where I mean.

petertseng: this line might say "addition - errors if there is nothing on the stack"

exercises/forth/example_gen.go
```diff
@@ -0,0 +1,77 @@
...
+
+// Source: {{.Ori}}
+{{if .Commit}}// Commit: {{.Commit}}
+{{end}}
+
+{{/* template for repeated caseSection */}}
+
+{{define "caseSection"}}
+{{range $y, $c := .}}{
+	{{printf "%q" $c.Description}},
```
petertseng: if this line would add the section name. not sure how easy that is though, it is looking like we would have to pass two arguments into the `caseSection` template, the first being the section name, the second being the cases, and I'm not familiar enough with templates to know whether this is doable. You could try to see if you could find a way.

exercises/forth/cases_test.go
leenipper: I see how to do it now. The template will generate separate test case slices, and I will use one main test function, and let it call a helper for the mechanics.

leenipper: Yes.

exercises/forth/example_gen.go
leenipper: I will work with it.

leenipper:
After I've slept, I realize I don't like how the error messages are prefixed.  Instead of reading for example

`
addition - errors if there is nothing on the stack ...
`

it reads

`
additionGroup - errors if there is nothing on the stack ...
`

so I think I'd like to make it match the former, or perhaps even

`
addition(+) - errors if there is nothing on the stack ...
`

This last approach indicates the group of tests and the specific Forth operator, and student should have enough information to address the problem. When the test case group name matches the operator, it will just read for example "dup - errors if ...".  The way I changed the example_gen, I have the flexibility to do it.  Does this sound ok @petertseng ?

exercises/forth/cases_test.go
```diff
@@ -0,0 +1,246 @@
...
+	},
+	{
+		"errors if executing a non-existent word",
+		[]string{"foo"},
+		[]int(nil),
+	},
+}
+
+var testSections = []testcaseSection{
+	{"parsing", &parsingGroup},
```
petertseng: I'd like to suggest leaving the `&` off - since a slice is a pointer+length+capacity (https://blog.golang.org/go-slices-usage-and-internals), we probably don't need to take a pointer to a pointer. (Unless you have specific reason to say it is necessary)

exercises/forth/forth_test.go
```diff
@@ -0,0 +1,63 @@
...
+		t.Fatalf("Found testVersion = %v, want %v", testVersion, targetTestVersion)
+	}
+}
+
+func runTestCases(t *testing.T, sectionName string, testCases *[]testCase) {
+	for _, tc := range *testCases {
+		if v, err := Forth(tc.input); err == nil {
+			var _ error = err
+			if tc.expected == nil {
+				t.Fatalf("%s - %s\n\tForth(%#v) expected an error, got %v",
```
petertseng: when you make your change to the prefixes, you may notice that you are repeating the `%s - %s\n\tForth(%#v)` a lot. you could consider extracting it to a `prefix :=` or something, so you only have to say it once.

exercises/forth/cases_test.go
leenipper: No reason except my leftover C habits!  I did it without even thinking.  Will certainly fix.

exercises/forth/forth_test.go
leenipper: ok.

leenipper:
For this one, these changes:
Use slice instead of pointer to slice in testcaseSection.
In the error output, simplify description formatting.
I omitted the lengthy commit message this time.

PR #498, opened 2017-02-08
Fix bug in Luhn test output

michaelorr:
In the user output string, the args to Fatalf have the `Expected` and
`Got` arguments swapped.


ferhatelmas:
@michaelorr Thanks :heart:

PR #507, opened 2017-02-10
error-handling: Ensure test versioning consistency with other exercises

tleen:
Moved TestTestVersion to be the first test function.

See #470

### DIFF
--- a/exercises/error-handling/error_handling_test.go
+++ b/exercises/error-handling/error_handling_test.go
@@ -27,6 +27,14 @@ func (mr mockResource) Close() error      { return mr.close() }
 func (mr mockResource) Frob(input string) { mr.frob(input) }
 func (mr mockResource) Defrob(tag string) { mr.defrob(tag) }
 
+// If this test fails and you've properly defined testVersion the requirements
+// of the tests have changed since you wrote your submission.
+func TestTestVersion(t *testing.T) {
+	if testVersion != targetTestVersion {
+		t.Fatalf("Found testVersion = %v, want %v", testVersion, targetTestVersion)
+	}
+}
+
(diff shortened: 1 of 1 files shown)


exercises/error-handling/error_handling_test.go
```diff
@@ -27,6 +27,14 @@ func (mr mockResource) Close() error      { return mr.close() }
 func (mr mockResource) Frob(input string) { mr.frob(input) }
 func (mr mockResource) Defrob(tag string) { mr.defrob(tag) }
 
+// If this test fails and you've properly defined testVersion the requirements
```
robphoenix: I wonder if this comment is necessary? I think a comment like this should either be in the `Hello-world` exercise as per #515 or should appear in all exercises rather than just randomly added in here to a later exercise.

tleen: Probably not, but since the previous author added it and its not wrong I left it. Removing those comments can go along with #470 and #515. hello-world should be the definitive documentation on how the test versioning works for all the tests.

kytrinyx: I like that suggestion—it seems consistent with the changes to make hello-world the canonical place for this.

robphoenix: Yes, that was my thought also. Should we leave this here then, if it will be covered in `hello-world`?

tleen: I think to pluck out the comment as redundant once the hello world stuff is better commented. I will add a note to #515.

robphoenix: Cool :+1:

PR #512, opened 2017-02-12
grains: Ensure consistency with other exercises

tleen:
This moves the testVersion verification to be the first test.
There is no error handling, so we do not need to adjust any calls to Error / Fatal.

See #470

### DIFF
--- a/exercises/grains/grains_test.go
+++ b/exercises/grains/grains_test.go
@@ -23,6 +23,12 @@ var squareTests = []struct {
 	{-1, 0, true},
 }
 
+func TestTestVersion(t *testing.T) {
+	if testVersion != targetTestVersion {
+		t.Fatalf("Found testVersion = %v, want %v.", testVersion, targetTestVersion)
+	}
+}
+
 func TestSquare(t *testing.T) {
 	for _, test := range squareTests {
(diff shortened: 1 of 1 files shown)


kytrinyx:
This looks great, @tleen.

I have a bit of feedback in terms of the PR itself (workflow, best practices), rather than the code.

1. I prefer the subject line in the PR to be a one-sentence summary. This is something that should not have the word _and_ (explicitly or implied).
1. The detail that you provide is good, but I'd prefer to have that in the body of the PR.
1. For things that are "closes #XXX" or "See #XXX" I prefer to have that in the body of the PR, simply because the subject line doesn't support markdown, and these don't get turned into links. It makes it harder to follow to go look at the related issue or PR.

Example:

```
grains: ensure consistency with other exercises

This moves the testVersion verification to be the first test.
There is no error handling, so we do not need to adjust any calls to Error / Fatal.

See #470.
```

robphoenix:
This is good feedback for all contributors. Some of the points are missing from the x-common [contributing guide](https://github.com/exercism/x-common/blob/master/CONTRIBUTING.md). I'll look at adding them, as well as I think more specific links to these items in the `xgo` README, so they're not lost in a wall of text.

PR #531, opened 2017-02-18
react: Return Canceler from AddCallback

petertseng:
This makes more sense than returning a CallbackHandle that has to then
get passed into RemoveCallback. It avoids (as in some solutions but not
all) having to cast the CallbackHandle back and forth.

The author of this commit believes this to be more idiomatic Go.

### DIFF
--- a/exercises/react/example.go
+++ b/exercises/react/example.go
@@ -1,6 +1,6 @@
 package react
 
-const testVersion = 4
+const testVersion = 5
 
 type reactor struct {
 	cells []*cell
@@ -17,6 +17,14 @@ type cell struct {
 	observers map[int]func(int)
 }
 
+type canceler struct {
+	cancel func()
+}
+
+func (c *canceler) Cancel() {
+	c.cancel()
+}
+
 func (c *cell) Value() int { return c.value }
 
 func (c *cell) SetValue(v int) {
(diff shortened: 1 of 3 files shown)


ferhatelmas:
@petertseng I agree it's more idiomatic.

petertseng:
Notes:

One might say it is in violation of https://github.com/golang/go/wiki/CodeReviewComments#interfaces because we put the interface in the package producing one. Though the tests need to consume as well, and it's definitely true the interface is there for the tests.

I would be very interested if someone can find examples in the standard library that either support or refute this thought of returning `Canceler`, but otherwise since I have approval I'll merge it soon.

petertseng:
This is not part of the standard library, but see how https://godoc.org/gopkg.in/redis.v5#Client.Subscribe returns a `*PubSub`, on which you might call https://godoc.org/gopkg.in/redis.v5#PubSub.Unsubscribe , so this matches what we do here.

petertseng:
Standard library: https://golang.org/pkg/database/sql/#Open returns `*DB`, on which you might call https://golang.org/pkg/database/sql/#DB.Close

ferhatelmas:
Another example is https://golang.org/pkg/net/#Dialer.Dial which returns `net.Conn` according to given network backed by [different structs](https://golang.org/src/net/dial.go#L519), on that `Cancel` might be called.

PR #533, opened 2017-02-20
travis: Upgrade Go versions (1.8 released, 1.7 -> 1.7.5), no longer test 1.6.x, 1.5.x, 1.4.x

petertseng:
See individual commit messages.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,8 @@ sudo: false
 # but since Travis only runs five jobs at a time throughout all of Exercism,
 # it is better citizenship to not test the additional versions.
 go:
-  - 1.7
-  - 1.6.3
+  - 1.7.5
+  - 1.8
   - tip
 
 # Travis runs in a 64 bit environment but beginning with Go 1.5, building a
@@ -61,7 +61,7 @@ script:
 matrix:
   include:
     # Keep this in sync with the latest version.
-    - go: 1.7
+    - go: 1.8
       env: STATIC_CHECKERS=1
   allow_failures:
     - go: tip
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,15 +14,12 @@ sudo: false
 # to alert us to any real upcoming change in the language that would be
 # incompatible with our existing code.
 #
-# We may choose to test additional versions as time allows,
-# as long as the number of entries in the build matrix doesn't exceed ten.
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -14,15 +14,12 @@ sudo: false
...
-# We may choose to test additional versions as time allows,
-# as long as the number of entries in the build matrix doesn't exceed ten.
-# These additional versions will not be required, so they're advisory only.
-# Since Travis runs five jobs at a time, a limit of ten jobs makes sense.
+# We used to test additional versions beyond the last two,
+# but since Travis only runs five jobs at a time throughout all of Exercism,
+# it is better citizenship to not test the additional versions.
 go:
-  - 1.7
-  - 1.6.3
```
ferhatelmas: What about `1.6.x` ?

.travis.yml
```diff
@@ -14,15 +14,12 @@ sudo: false
...
-  - 1.5.4
-  - 1.4.3
+  - 1.8
+  - 1.7.5
```
ferhatelmas: Order might be better. From oldest to the newest, for example.

petertseng: Ah, okay. I agree (weird to have 1.7.5 between 1.8 and tip) and now's a good time to change.

.travis.yml
petertseng: Hmm, before the change I'm proposing, I would have kept it and just made it optional. We discover in https://github.com/exercism/x-common/issues/539 that the entire Exercism organisation has five concurrent jobs though, so I'm a little doubtful we should keep it, especially if (it being optional) we never care about whether it breaks anyway.

The Go release policy at https://golang.org/doc/devel/release.html is that because 1.8 is released, it is the only supported release, but 1.7 will still receive security fixes. So probably 1.6 will get no support at all.

Should we keep anyway? Having extra build entries is not **too** bad since each of ours only takes a minute or so, and there can be an argument for doing so if there are still users of it.

Now is a chance for us to decide a policy!

ferhatelmas: I'm fine to remove it.  Since it's not written in commit message, I wanted to be sure it's not skipped by mistake.

petertseng: You are right. Then I will make a quick mention of it in the commit message. (The .travis.yml will not change)

ferhatelmas: :+1:

PR #536, opened 2017-02-21
series: Ensure test versioning consistency

robphoenix:
Added test versioning: see #470.

### DIFF
--- a/exercises/series/example.go
+++ b/exercises/series/example.go
@@ -1,5 +1,7 @@
 package slice
 
+const testVersion = 1
+
 func All(n int, s string) (r []string) {
 	for i := 0; n <= len(s); i++ {
 		r = append(r, s[i:n])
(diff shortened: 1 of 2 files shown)


robphoenix:
My assumption is that not having any test versioning in the bonus tests (https://github.com/exercism/xgo/blob/master/exercises/series/first_test.go) is ok? 
I've not added anything as we're unable to redeclare the `testVersion` & `targetTestVersion` constants and I didn't want to start adding different tests for different version constants.

petertseng:
> as we're unable to redeclare the `testVersion` & `targetTestVersion` constants and I didn't want to start adding different tests for different version constants

Yeah, seems like a good reason too.

Just for the record, even if I'm wrong and `go test -tags first` doesn't run the test version test, I'm still OK with just having just the one. We can change just that one even if a chance affects just the bonus tests, I suppose.

PR #539, opened 2017-02-23
strain: add test versioning

robphoenix:
see #470

### DIFF
--- a/exercises/strain/strain_test.go
+++ b/exercises/strain/strain_test.go
@@ -19,6 +19,14 @@ import (
 	"testing"
 )
 
+const targetTestVersion = 1
+
+func TestTestVersion(t *testing.T) {
+	if testVersion != targetTestVersion {
+		t.Fatalf("Found testVersion = %v, want %v", testVersion, targetTestVersion)
+	}
+}
+
 func lt10(x int) bool { return x < 10 }
 func gt10(x int) bool { return x > 10 }
 func odd(x int) bool  { return x&1 == 1 }
--- a/exercises/strain/example.go
+++ b/exercises/strain/example.go
@@ -1,5 +1,7 @@
 package strain
 
+const testVersion = 1
+
 type Ints []int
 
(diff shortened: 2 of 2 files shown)


exercises/strain/strain_test.go
```diff
@@ -19,6 +19,14 @@ import (
 	"testing"
 )
 
+const targetTestVersion = 1
+
+func TestTestVersion(t *testing.T) {
```
ferhatelmas: In some exercises, I realized we put it as a first test but if there are helper functions, they go first. 
So is it supposed to be the first thing in the file or the first test when tests started ? 

Otherwise, fine.

robphoenix: My reasoning was that the helper functions belonged to the exercise tests and I wanted to keep all that logic together rather than split it up with the version test. So I figured, get the version test done and out the way before moving onto the main tests.

ferhatelmas: Makes sense to me. However, we might want to update exercises with other case.

robphoenix: yeah, probably, I'll do that today.

tleen: ah I've been doing it the other way I believe. Making it the first *Test function but leaving any functions used in tests before them.

tleen: In #470 I definatly read 

> placement of TestTestVersion function as the first test function

to indicate we were floating it to the top of the test functions (i.e. functions that start with Test).

robphoenix: Ah ok, no worries, I'm happy either way tbh, I'll fix this PR, as it's probably easier than going back over what's already been done, unless anyone has any specific objections?

tleen: Yep either is good as long as we try and do it the same way.

ferhatelmas: Exactly same for me, too.

PR #540, opened 2017-02-25
ocr-numbers: add test versioning

tleen:
see #470


exercises/ocr-numbers/ocr_numbers_test.go
```diff
@@ -121,6 +123,12 @@ var tests = []struct {
 
 var _ = recognizeDigit // step 1.
 
+func TestTestVersion(t *testing.T) {
```
ferhatelmas: See https://github.com/exercism/xgo/pull/539/files#r102824576 so that we agree on for one way.

PR #542, opened 2017-02-25
paasio: Ensure test versioning consistency with other exercises

tleen:
* Isolated test version testing into own function
* Errorf -> Fatalf

see #470


exercises/paasio/paasio_test.go
```diff
@@ -14,10 +14,13 @@ import (
 // testVersion identifies the API tested by the test program.
```
ferhatelmas: > targetTestVersion identifies ...

PR #543, opened 2017-02-26
strain: fix test order

robphoenix:
see #539

### DIFF
--- a/exercises/strain/strain_test.go
+++ b/exercises/strain/strain_test.go
@@ -21,12 +21,6 @@ import (
 
 const targetTestVersion = 1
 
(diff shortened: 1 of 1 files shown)


robphoenix:
I tried to fix #539 with an interactive rebase but it looked like I was totally going to mess up the repo's history. If fixing my previous commit (#539) is preferred over this further commit I'll find the way.

As an aside we hit 900 commits :tada: :raised_hands: We should totally have a party when we hit 1000 commits. :tada:

petertseng:
(and this circumstance is definitely not special enough, so no need to try to rebase here)

PR #548, opened 2017-03-01
parallel-letter-frequency: Add test versioning

tleen:
See #470

### DIFF
--- a/exercises/parallel-letter-frequency/example.go
+++ b/exercises/parallel-letter-frequency/example.go
@@ -1,5 +1,7 @@
 package letter
 
+const testVersion = 1
+
 func ConcurrentFrequency(l []string) FreqMap {
 	switch len(l) {
 	case 0:
--- a/exercises/parallel-letter-frequency/parallel_letter_frequency_test.go
+++ b/exercises/parallel-letter-frequency/parallel_letter_frequency_test.go
@@ -5,6 +5,14 @@ import (
 	"testing"
 )
 
+const targetTestVersion = 1
+
+func TestTestVersion(t *testing.T) {
+	if testVersion != targetTestVersion {
+		t.Fatalf("Found testVersion = %v, want %v", testVersion, targetTestVersion)
+	}
+}
+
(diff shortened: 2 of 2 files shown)


robphoenix:
Should the var block in the test file be moved up to before the tests? (#543)

tleen:
I definitely have opinions about how I like to see code files structured. there was some struggled with this test file too, like should I have moved [the comment](https://github.com/exercism/xgo/blob/master/exercises/parallel-letter-frequency/parallel_letter_frequency_test.go#L8-L11) up?

After thinking on it I left as is. My logic is this: 

- Our work is focused on the test versioning, and I didn't want to restructure other aspects of the file as it would exceed our mandate.
- Overall structure is opinion based and that can vary (the test versioning stuff we want consistent because it will be in all exercises).
- I try to honor the original code as much as possible, so if its a toss up, leave it.

I can see that the original author wanted to keep the benchmarking var close to the benchmarks. Not sure that is how I would do it, but its perfectly valid so I left it.

You weren't alone in wondering about this!

robphoenix:
> You weren't alone in wondering about this!

:smile: 

> like should I have moved the comment up?

I see, yeah, I wonder if it should even be in the stub file, which would make more sense to me?

> keep the benchmarking var close to the benchmarks.

My understanding is that the var is also there for the main test

```
func TestConcurrentFrequency(t *testing.T) {
	seq := Frequency(euro + dutch + us)
	con := ConcurrentFrequency([]string{euro, dutch, us})
```

> Our work is focused on the test versioning, and I didn't want to restructure other aspects of the file as it would exceed our mandate.

You're right, any changes to the structure should at least be in a separate commit if not a separate PR. 

Personally I think the comment and the var should be moved, but will defer to popular opinion. 
Unless you want to add an additional commit to this, let's merge and an additional PR can be opened if deemed necessary.

tleen:
Yeah that is weird var placement for sure, but to each their own? We can look at stuff like this separately from test, I wouldn't object to an issue being opened on this one specifically and adjusting the location of the var/comment.

PR #557, opened 2017-03-02
forth: Make more idiomatic

ferhatelmas:


exercises/forth/example.go
```diff
@@ -10,10 +10,10 @@ import (
 
 type operatorFn func(stack *[]int) error
 
-type operatorId byte
+type operatorID byte
 
 const (
-	addOp operatorId = iota
```
ferhatelmas: I think they should be `opAdd`, `opSub`, etc. similar to `http.StatusXxx`. However, they aren't exported so it might be fine. What do you think ?

robphoenix: It would also bring it in line with the `err` prefix change. :+1:

ferhatelmas: Done.

PR #563, opened 2017-03-03
config.json: update house

robphoenix:
add topics and change difficulty

### DIFF
--- a/config.json
+++ b/config.json
@@ -139,9 +139,13 @@
       "topics": []
     },
     {
-      "difficulty": 1,
+      "difficulty": 4,
       "slug": "house",
-      "topics": []
+      "topics": [
+        "Strings",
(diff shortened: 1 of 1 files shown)


ferhatelmas:
Don't we need to reorder exercises according to difficulty ?

robphoenix:
We will need to re-order at some point, yes. We can either re-order as we go along or when the `config.json` file is complete. If we order as we go there might be quite a bit of churn as we continue to update `config.json`, though if we wait for `config.json` to be completed we may be waiting a while.
Also, is it generally accepted to order according to difficulty, rather than interspersing a run of difficult exercises with something slightly easier, to keep morale up as it were, or taking into consideration the topics as well so that similar exercises are not too closely grouped?

Issue #555, opened 2017-03-02
status of accumulate, strain & list-ops


robphoenix:
`accumulate` & `strain` are to be deprecated across the exercism curriculum, in favour of the `list-ops` exercise:

https://github.com/exercism/xjava/issues/297
https://github.com/exercism/x-common/issues/553#issuecomment-279275406
https://github.com/exercism/x-common/issues/588#issuecomment-279275431

Currently in this track `accumulate` is deprecated, `strain` is active and `list-ops` has not been implemented.

There has been some recent discussion about this [here](https://github.com/exercism/xgo/issues/18#issuecomment-283222291)
The main points being:

* Although list operations such as map, filter and reduce are actively discouraged in Go, it is still considered a beneficial learning exercise to implement them, though for a specific type rather than generically.
* The suggestion that this track un-deprecate `accumulate`, until `list-ops` is added. If list-ops will be added within the next few days, then there's no point in adding the churn, but if it will be a while in coming, why not show accumulate?

I'd like to implement the `list-ops` exercise but realistically am not going to within the next few days. Please don't take this as my calling dibs on the exercise, I am sure others would probably implement it better than me and from my point of view it would be an opportunity for learning rather than a doubling of efforts.

So, until `list-ops` is implemented do we:

1. deprecate `strain` and un-deprecate `accumulate`
2. leave `strain` and un-deprecate `accumulate`
3. leave both as is until `list-ops` is implemented and then deprecate `strain`

petertseng:
In case I haven't made it clear by my https://github.com/exercism/xgo/issues/18#issuecomment-283488655, I suggest to un-deprecate `accumulate`. We can deprecate them when `list-ops` comes.

robphoenix:
Ok, well, unless there's any objections, let's do that. @ferhatelmas @tleen

PR #567, opened 2017-03-06
accumulate: update tests

robphoenix:
see #414
Also removed unnecessary else block

### DIFF
--- a/exercises/accumulate/accumulate_test.go
+++ b/exercises/accumulate/accumulate_test.go
@@ -37,10 +37,8 @@ func TestTestVersion(t *testing.T) {
 func TestAccumulate(t *testing.T) {
 	for _, test := range tests {
 		actual := Accumulate(test.given, test.converter)
-		if fmt.Sprintf("%s", actual) != fmt.Sprintf("%s", test.expected) {
-			t.Fatalf("Accumulate(%s, %s): expected %s, actual %s", test.given, test.description, test.expected, actual)
-		} else {
-			t.Logf("PASS: %s %v", test.description, test.given)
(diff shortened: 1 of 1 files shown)


robphoenix:
I realise there is a question mark over the status of accumulate at the moment (#555) but I just wanted to be able to close #414

exercises/accumulate/accumulate_test.go
```diff
@@ -38,9 +38,7 @@ func TestAccumulate(t *testing.T) {
 	for _, test := range tests {
 		actual := Accumulate(test.given, test.converter)
 		if fmt.Sprintf("%s", actual) != fmt.Sprintf("%s", test.expected) {
```
petertseng: I might say we have to use `%q` here too, otherwise for the first test (expects an empty list) someone could pass by emitting a list of many empty strings (not that I *expect* anyone to do that, but for completeness...?)

Or can switch to reflect.DeepEqual I suppose?

petertseng: Ah I'm wrong, many empty strings expands to `[    ]` which is different from `[]`. So I guess the risk is someone might say `["echo echo echo echo"]` instead of `["echo", "echo", "echo", "echo"]`

robphoenix: Ah, yes, that is possible.

robphoenix:
now:

```
--- FAIL: TestAccumulate (0.00s)
	accumulate_test.go:41: Accumulate(["echo" "echo" "echo" "echo"], "echo"): expected ["echo" "echo" "echo" "echo"], actual ["echo echo echo echo"]
```

PR #570, opened 2017-03-06
test-without-stubs,exercises: Typo fixes

ferhatelmas:
- used https://github.com/client9/misspell

### DIFF
--- a/bin/test-without-stubs
+++ b/bin/test-without-stubs
@@ -64,6 +64,6 @@ fi
 
 # This is the last command in the file,
 # so the exit status of this script is the exit status of go test.
-# If this ever changes, remember to preserve the exit status acordingly.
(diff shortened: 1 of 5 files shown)


ferhatelmas:
We might want to add this into CI but not sure because we can get false positives. What do you think ?

tleen:
If changes would be rejected because of false positives it sort of has to be a no? We shouldn't have things fail if they are correct. I do think running this process occasionally would help tighten things up.

ferhatelmas:
I agree.

robphoenix:
Is it worth adding a note somewhere, perhaps the README, about this, so it doesn't get lost here?

tleen:
May be worth pointing out the tool, but effectively it will be up to us to periodically check by using it.

PR #581, opened 2017-03-11
robot-name: Add test versioning

tleen:
See #470

### DIFF
--- a/exercises/robot-name/bonus_example.go
+++ b/exercises/robot-name/bonus_example.go
@@ -7,8 +7,6 @@ import (
 	"math/rand"
 )
 
-const testVersion = 1
-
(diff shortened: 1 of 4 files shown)


robphoenix:
Does this need to be considered here also? https://github.com/exercism/xgo/pull/536#issuecomment-281337014

tleen:
Yes I think so. Wasn't quite sure what to do with the testing stuff in bonus, so didn't. What was the final consensus, leave it or strip it?

robphoenix:
strip it, so the test version info is only in the initial/main test files and not in any extra/bonus files.

robphoenix:
:+1: LGTM. Be good to squash the commits for merging.

PR #597, opened 2017-03-19
change: Add change exercise

leenipper:


leenipper:
oops, used change.go instead of example.go. will re-submit.

PR #598, opened 2017-03-19
change: Add exercise

leenipper:


robphoenix:
Hi @leenipper thanks for this, I hope to give it the attention it deserves over the next couple of days.

tleen:
@robphoenix let me know when you get this underway. I'd like to do it alongside you get a feel for how to review and get new exercises into the track. Perhaps over email, don't want to clutter the thread.

robphoenix:
To be honest @tleen I've not properly reviewed a significant PR before, I'd be learning alongside you how to do this. So far I've re-read through some previous PR's such as @leenipper's https://github.com/exercism/xgo/pull/494 which has a thorough and interesting review from @petertseng 
I've pulled down this PR as a new branch on my laptop with the command `git fetch upstream pull/598/head:add-the-change-exercise` and am reading through it.

I suppose I'm looking to see if the exercise works and whether the implementation makes sense to me. I know from having had a new exercise merged it can be helpful just to have a fresh set of eyes on something you may have been thinking about and working on for a lengthy period of time.

exercises/change/cases_test.go
```diff
@@ -0,0 +1,138 @@
+package change
+
+// Source: exercism/x-common
+// Commit: cda8f98 Create new exercises structure
```
robphoenix: I got an error when running the example_gen.go file:

```
> go run example_gen.go 
2017/03/23 14:32:00 unexpected data structure: json: cannot unmarshal number into Go value of type []int
exit status 1
```
which I assume is from the update of the x-common file (https://github.com/exercism/x-common/commits/master/exercises/change/canonical-data.json)

exercises/change/change_test.go
```diff
@@ -0,0 +1,34 @@
...
+	if testVersion != targetTestVersion {
+		t.Fatalf("Found testVersion = %v, want %v", testVersion, targetTestVersion)
+	}
+}
+
+func TestChange(t *testing.T) {
+	for _, tc := range testCases {
+		actual := Change(tc.coins, tc.target)
+		if !reflect.DeepEqual(actual, tc.expected) {
+			t.Fatalf("%s : Change(%v, %d): expected %v, actual %v",
```
robphoenix: While trying this out I changed some code which gave me this failure message:

```
change_test.go:21: no coins make 0 change : Change([1 5 10 21 25], 0): expected [], actual []
```
which reminded me of this issue https://github.com/exercism/xgo/issues/414

though changing `%v` to `%q` didn't help, `%#v` did, giving:

```
change_test.go:21: no coins make 0 change : Change([1 5 10 21 25], 0): expected []int{}, actual []int(nil)
```
I'm not sure if it's worth changing this?

tleen:
Are we supposed to run it through the exercism fetch, answer, submit cycle, or is that not-doable/overkill for accepting a new implementation?

exercises/change/example.go
```diff
@@ -0,0 +1,119 @@
...
+			findSolutions(coins, target, &solutions)
+		}
+		if len(solutions) == 0 {
+			result = []int{-1}
+		} else if len(solutions) == 1 {
+			result = solutions[0]
+		} else {
+			result = chooseBestSolution(solutions)
+		}
+	}
```
robphoenix: I've heard [Frances Campoy](https://www.youtube.com/channel/UC_BzFbxG2za3bp5NRRRXJSw) say an else block can almost be considered a code smell. Personally I'd return from the function early than re-assigning to the result var.

```
	if target == 0 {
		return []int{}
	}
	if target > 0 {
		findSolutions(coins, target, &solutions)
	}
	if len(solutions) == 0 {
		return []int{-1}
	}
	if len(solutions) == 1 {
		return solutions[0]
	}
	return chooseBestSolution(solutions)
```

exercises/change/example.go
```diff
@@ -0,0 +1,119 @@
...
+	b := make([]int, len(a))
+	j := 0
+	for i := len(a) - 1; i >= 0; i-- {
+		b[j] = a[i]
+		j++
+	}
+	return b
+}
+
+const testVersion = 1
```
tleen: The testVersion const is not technically wrong here, but in regards to #470 should probably be moved up to the top.

exercises/change/example.go
```diff
@@ -0,0 +1,119 @@
...
+
+// Change returns a list of coins in increasing value order to make
+// change for target amount from the coins given; an infinite supply of coins is assumed.
+func Change(coins []int, target int) (result []int) {
+	solutions := make([][]int, 0)
+	if target == 0 {
+		result = []int{}
+	} else {
+		if target > 0 {
+			findSolutions(coins, target, &solutions)
```
robphoenix: I was a little confused by this, maybe a comment that makes it clearer that `findSolutions()` updates `solutions`? Or it could just be me being dense! :smile:

exercises/change/example.go
robphoenix: This const should go at the top of the file.

config.json
```diff
@@ -496,6 +496,11 @@
       "difficulty": 1,
       "slug": "forth",
       "topics": []
+    },
+    {
+      "difficulty": 1,
+      "slug": "change",
+      "topics": []
```
robphoenix: It would be awesome if you felt open to adding the topics to this in a second commit, and maybe suggesting a difficulty rating?  There's a list of topics here https://github.com/exercism/x-common/blob/master/TOPICS.txt
Don't worry if not, just a thought.

robphoenix:
I don't think we need to test if the exercise works within the framework of exercism, just whether it works as an implementation of the canonical exercise.

exercises/change/cases_test.go
robphoenix: Hang on, this is my fault, I ran it against an out-of-date x-common. I've rebased and the example_gen.go works fine, sorry! 

It would be good to re-generate the test cases here before merging.

robphoenix:
Looks good to me @leenipper, just a couple of comments. And I'd appreciate the eyes of the other maintainers on this before signing off on it.

exercises/change/cases_test.go
```diff
@@ -0,0 +1,138 @@
...
+	},
+	{
+		"error testing for change smaller than the smallest of coins",
+		[]int{
+			5,
+			10,
+		},
+		3,
+		[]int{
+			-1,
```
petertseng: even though the JSON file https://github.com/exercism/x-common/blob/master/exercises/change/canonical-data.json#L66 uses -1 for this, I might not put it in a slice `[]int{-1}`, since in the JSON file it's just the single integer -1 instead of `[-1]`. Perhaps the `Change` function could return `([]int, error)` or `([]int, bool)` to make it clear when change can't be made, instead of using this sentinel value

exercises/change/example.go
leenipper: ok.

exercises/change/cases_test.go
leenipper: Yes, it runs but the  generated cases_test.go test case array is now empty   I had fudged a bit with a mod to the canonical-data.json and I had forgotten about it.  I will make another attempt to handle the new layout of the data and also the difference between a [ list ] or a -1 in the expected field. The list vs. constant just seems wrong because the object type is different, but JSON is crazy flexible in that way I suppose.

exercises/change/example.go
leenipper: It can use some readability improvement, for sure.

exercises/change/example.go
leenipper: Yes, that is better. I will update.  If I add an error return it will look just a bit different.

exercises/change/example.go
leenipper: ok.

config.json
leenipper: I'm thinking for the "change" exercise, these topics might apply:  Control-flow (loops), Arrays, and Algorithms.   What do you think ?

exercises/change/cases_test.go
leenipper: Yes, returning an error or bool will be cleaner.

exercises/change/cases_test.go
leenipper: Looks like interface{} is my friend to overcome the [list] vs. -1 difference.  Pursuing that.

config.json
leenipper: Perhaps Recursion might apply too, and/or Searching.  One of the newer test cases fails with my example, so must re-consider my algorithm choice! Having fun with it.

tleen: Def the first three. I see Recursion/Searching as possibilities in the solution, and as such using them as topics may steer users down a particular solution path?

leenipper: I agree @tleen .  I'll use the first three.

exercises/change/change_test.go
```diff
@@ -0,0 +1,68 @@
...
+
+const targetTestVersion = 1
+
+func TestTestVersion(t *testing.T) {
+	if testVersion != targetTestVersion {
+		t.Fatalf("Found testVersion = %v, want %v", testVersion, targetTestVersion)
+	}
+}
+
+func determineExpected(expected interface{}) (valid bool, list []int) {
```
petertseng: I personally was thinking this might not be 100% clear to the students. Any way this logic can be moved to the *generator* instead? That ties in with the above question about the `testCase` containing the `[]int` and `bool`.

exercises/change/cases_test.go
```diff
@@ -0,0 +1,126 @@
+package change
+
+// Source: exercism/x-common
+// Commit: 3d8b5b3 change: Fix canonical-data.json formatting
+
+var testCases = []struct {
+	description string
+	coins       []int
+	target      int
+	expected    interface{}
```
petertseng: I think it would be good if we could have an `[]int` and a `bool` here, rather than an `interface{}`, since otherwise it may not be clear to the student what type the function should return

exercises/change/change_test.go
```diff
@@ -0,0 +1,68 @@
...
+	_, isError := expected.(int)
+	if isError {
```
robphoenix: I think this should be `err` rather than `isError`

exercises/change/change_test.go
leenipper: I will attempt a move of the handling code to the generator to make the _test logic more normal.

exercises/change/cases_test.go
leenipper: Yes, I agree.

exercises/change/change_test.go
leenipper: In the type assertion general form,
   t, ok := i.(T)
the 2nd variable is a bool.  So, err did not seem idiomatic for a bool, which is the reason for isError.  Make sense ?

leenipper: Actually, I will use
    _, ok := expected.(int)
since ok is the typical bool result variable for a type assertion.

exercises/change/example_gen.go
```diff
@@ -38,25 +38,59 @@ type OneCase struct {
...
+// determineExpected examines an .Expected interface{} object and determines
+// whether an error is indicated by an int value (actually -1) in the JSON,
+// or whether .Expected is a slice of integer coins.
+func determineExpected(expected interface{}) (valid bool, list []int) {
+	_, ok := expected.(int)
+	if ok {
+		return false, nil
+	}
+	ilist, ok := expected.([]interface{})
+	if ok {
```
petertseng: optional: could make this an `if !ok { return false, nil }`, then many lines of code can be unindented

exercises/change/example.go
```diff
@@ -0,0 +1,191 @@
...
+const testVersion = 1
+
+var (
+	ErrNoSolutionFound = errors.New("No solution")
+	ErrNegativeTarget  = errors.New("Negative Target")
+)
+
+// Change returns a list of coins in increasing value order to make
+// change for target amount from the coins given; an infinite supply of coins is assumed.
+func Change(coins []int, target int) (result []int, err error) {
```
petertseng: optional: it doesn't really look like the `result` and `error` names are used so it could just be `([]int, error)` and probably still be clear enough

exercises/change/example.go
```diff
@@ -0,0 +1,191 @@
...
+		copy(*solution, newsolution)
+		sort.Ints(*solution)
+	}
+}
+
+// partialSolution takes a partial solution, and coin list c looking to meet target
+// while updating solution s to include partial plus the coins meeting target.
+func partialSolution(partial []int, c []int, target int, s *[]int, bestsize *int) {
+	// We want to consider coins in c in largest to smallest order.
+	sort.Sort(sort.Reverse(sort.IntSlice(c)))
```
petertseng: Hmm, I wish there were an easier way to express it, but https://golang.org/pkg/sort/#Ints is reversed, so I suppose what you have is the way.

exercises/change/example.go
```diff
@@ -0,0 +1,191 @@
...
+
+	for e := range c {
```
petertseng: optional: I notice `c[e]` being used a lot in here. Perhaps a name like `coin` could be given to it. Same number of characters.

exercises/change/example.go
```diff
@@ -0,0 +1,191 @@
...
+		// All coins larger than target.
+		return
+	}
+	findBestMixedSolution(coins[0:last+1], target, s)
+}
+
+// findBestMixedSolution attempts to find a best mixed coin amount solution
+// making correct change for the amount target.
+func findBestMixedSolution(coins []int, target int, s *[]int) {
+	// Consider all combinations of the coins.
```
petertseng: an interesting question I thought of - if `partialSolution` would be willing to consider 0 for `k`, could most of this function be removed?

exercises/change/change_test.go
robphoenix: I didn't know this, thanks :+1:

exercises/change/example.go
petertseng: After experimenting: Looks like it works to me, after changing `k >= 1` to `k >= 0`. At this point, `findBestMixedSolution` isn't even necessary. You can let `findSolution` call `partialSolution`. Suggest you give that a try.

petertseng: Oh, actually, don't even need to change `k >= 1` to `k >= 0`.

exercises/change/example.go
```diff
@@ -0,0 +1,191 @@
...
+					// which doesn't include coin c[e].
+					potentialSize := len(partial) + k + (newtarget / largest)
+					if (newtarget % largest) != 0 {
+						potentialSize++
+					}
+					if potentialSize < *bestsize {
+						// Put k instances of c[e] in a new partial solution.
+						newPartial := append(partial, nCoins(k, c[e])...)
+						d := make([]int, len(c))
+						copy(d, c)
```
petertseng: optional: I'm not sure if this makes things any better, but you could just copy `c[:e]` instead of all of `c` since all elements in positions later than `e` are being overwritten anyway.

https://play.golang.org/p/TyAGEkq5L8

petertseng: https://play.golang.org/p/5nGlahE4Hz

petertseng: https://play.golang.org/p/_Y3zos7KZE

exercises/change/example_gen.go
leenipper: Yes, will do.

exercises/change/example.go
leenipper: ok.

exercises/change/example.go
leenipper: It was the most concise way I found anyway.

exercises/change/example.go
leenipper: I think as I improved the efficiency of how the partialSolution operated, I didn't actually realize that it arrived at the best solution without the permutation sets which findBestMixedSolution was intended to provide.
So, yes, it works now with much less code.  Thanks!

exercises/change/example.go
leenipper: ok

exercises/change/example.go
leenipper: Will work with that and see.

leenipper:
Sounds good. Thanks!

petertseng:
Thanks!

exercises/change/change_test.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+}
+
+func TestChange(t *testing.T) {
+	for _, tc := range testCases {
+		actual, err := Change(tc.coins, tc.target)
+		if tc.valid {
+			if err != nil {
+				t.Fatalf("%s : Change(%v, %d): expected %v, got error %s",
+					tc.description, tc.coins, tc.target, tc.expectedChange, err)
```
petertseng: I realized I forgot to ask about #275 here (ensuring the second return value is of type `error`). I'm not sure it's necessary, but we could either add a `var _ error = err` or call `err.Error()` explicitly? Or should we not bother applying #275 here?

I should document #275 somewhere since everyone keeps forgetting it, maybe it's not really necessary if people take it for granted.

PR #601, opened 2017-03-27
accumulate: add initial file with test version and function stub

mmozuras:
Did not find any history of accumulate.go not being here being intentional.

### DIFF
--- a/exercises/accumulate/accumulate.go
+++ b/exercises/accumulate/accumulate.go
@@ -1,0 +1,5 @@
+package accumulate
+
+const testVersion = 1
+
+func Accumulate([]string, func(string) string) []string


petertseng:
I would say the history is in #279 - we provide stub files for the first "few" (arbitrary) exercises, plus the `word-count` to define a type. I'm not going to vote either way on whether there should be an accumulate.go, it can be up for discussion.

mmozuras:
@petertseng ah, here we go, thanks 😄 

I've encountered stubs in the couple of following exercises I did, so that's why I assumed this might be a mistake.

robphoenix:
I think accumulate didn't have a stub because it was deprecated for a while. I'd rather see it with a stub than have a single exercise within the first N exercises without a stub, which I agree can be a little confusing perhaps.

Makes me think a line or two in the [README](https://github.com/exercism/xgo/blob/master/README.md#stub-files) regarding the thinking in #279 might be worthwhile.

tleen:
May not be a bad idea to mention somewhere in the exercise when stubs stop occurring officially on the track to clear up any confusion.

PR #603, opened 2017-03-29
Add note about missing stubfiles to READMEs

### LINKED ISSUE #602: Add note to README regarding exercises with stub files
#601 highlighted some small confusion regarding which exercises should have stub files. This should be made clear in the README. See also #279

tleen:
Based on the discussion in #601 we add a general note in the
development README and in the specific exercise where stub
files stop being generally supplied with exercises.

Resolves #602


tleen:
@robphoenix is this what you had in mind? Is adding a README.md file to the exercise the right way to get the language specific note into an exercise README?

petertseng:
From what I can tell in https://github.com/exercism/trackler/blob/master/lib/trackler/implementation.rb though, any README.md in this repository is **ignored**. https://github.com/exercism/xerlang/issues/144 seems to claim it overrides the generated README completely, but I don't see evidence for that in the code. Would probably have to make a quick test using trackler to prove or disprove that.

So, we **would not** want to use README.md in this repository. If wanting to append to the README, we would want to put it in **HINTS.md**.

tleen:
Yes I thought it was README [looks like its HINTS](https://github.com/exercism/x-api/pull/93).

README.md
```diff
@@ -130,6 +130,8 @@ By convention, the stub file for an exercise with slug `exercise-slug`
 must be named `exercise_slug.go`. This is because CI needs to delete stub files
 to avoid conflicting definitions.
 
+The track stops automatically providing stub files at the [twelve-days-exercise](exercises/twelve-days/).
```
robphoenix: I'm a little uncomfortable with this wording; `The track stops automatically` gives the impression this is a hard stop that we have no control over. My understanding is that while stubs primarily exist as a helpful start in the early exercises for new Go programmers they also offer the opportunity to provide guidance in later exercises where there may be confusion, a particularly difficult concept, or boilerplate code required. If this is generally agreed then it would be good to reflect this in the README. If not then I think it's just the word `automatically` that I don't like, as this is a choice we have manually implemented, and I think it reads just as well without it, though the addition of a sentence about why stub files are no longer provided would be helpful.

exercises/twelve-days/HINTS.md
```diff
@@ -0,0 +1,2 @@
+## No Stub
+This is the first golang track exercise you encounter without a stub, a pre-existing "twelve-days.go" file for your solution. You may not see stubs in the future and should begin to get comfortable with creating your own go files for your solutions.
```
robphoenix: `Go` instead of `golang`

README.md
tleen: I agree will fix.

tleen: I lifted some of your language :smile:

robphoenix:
Good stuff @tleen, squash the commits and good to merge from my point of view :+1:

Issue #604, opened 2017-04-01
Make sure our existing generators are up-to-date


petertseng:
In [x-common](https://github.com/exercism/x-common/tree/master/exercises), we have various exercises with canonical-data.json representing the recommended set of tests for that exercise. We created generators to more easily keep up with changes in x-common. Whenever x-common changes, the task is as simple as re-running the generator.

The current documentation is at https://github.com/exercism/xgo#generating-test-cases - if at any point we find that we should explain something better, we should add any necessary documentation.

Preliminaries:

* I find the list of our generators using `ls exercises/*/cases_test.go | cut -d/ -f2` from the root of the xgo directory.
* I find the list of exercises with canonical-data.json using `ls exercises/*/canonical-data.json | cut -d/ -f2` from the root of the x-common directory.

This issue concerns **exercises that already have a generator**.

Because of recent changes to make all the canonical-data.json files conform to a common scheme (https://github.com/exercism/x-common/issues/625), it might be the case that our current generator no longer runs.

For these exercises we'd make sure the current generator runs, or if it doesn't then we'd update it.

In addition, the three lines at the top of the file:

```
// Source: {{.Ori}}
{{if .Commit}}// Commit: {{.Commit}}
{{end}}
```

can simply be replaced with `{{.Header}}`, which has the effect of:

* Deprecating the `Ori` field, which is confusingly named
* Showing the x-common version in the test file
* Reducing duplicate code

## Relevant exercises

- bob
- bowling
- change
- clock (#620)
- connect
- custom-set (in progress #622)
- forth (#615)
- gigasecond (#619)
- hamming
- largest-series-product
- leap (#606)
- luhn (#618)
- meetup
- raindrops
- rna-transcription
- roman-numerals
- transpose
- word-count

Once all these are complete, we can delete the deprecated `Ori` field.

leenipper:
> recent changes to make all the canonical-data.json files 

The forth exercise generator became askew due to that.  It runs, but test cases are empty. I have a WIP for it.

robphoenix:
I've changed this issue from a checklist to multiple issues. I've left the links to relevant PR's in the original comment but I don't think it's necessary to continue adding these.

[...]

PR #606, opened 2017-04-01
gen: Parse version out of JSON file; leap: Show version in generated file

petertseng:


exercises/leap/cases_test.go
```diff
@@ -1,7 +1,7 @@
 package leap
 
 // Source: exercism/x-common
-// Commit: be6fa53 leap: Rewrite the test cases and their descriptions
+// x-common version: 1.0.0
```
petertseng: Well, there are advantages and disadvantages.

Advantage of commit: We'll know if there's **any** change, even if someone forgot to change the version when test contents change (I think x-common reviewers have been good at reminding people to change version though)

Advantage of version: We know how important the change is, if there is one - for example in a patch-level change (the last number) it isn't strictly necessary to rerun the generator since no test content will change.

petertseng: So let me know whether you think it's better to only see version, only see commit, or see both.

robphoenix: Nice, I like having the version, I agree with the advantages of it, but I like it as an addition rather than a replacement, I think the commit pinpoints it to a specific point in time, and includes a brief but useful snippet of info regarding any recent changes.
Though if I had to choose, I'd go for the Version.

robphoenix: An additional thought: I'm just starting a PR to try pulling the canonical-data.json down from Github instead of/prior to doing what it does (I just keep forgetting to update my copy of the x-common repo before regenerating! :man_shrugging: ) and as far as I can see this would mean not being able to include the commit message in the generated file. And if the generator does take the route of being able to get the current json data via http as well as from an offline repo I assume we wouldn't want to sometimes have the commit message, sometimes not?

petertseng: > pulling the canonical-data.json down from Github

The only track I know of that tried to do this is https://github.com/exercism/xcrystal/pull/57, guess that hasn't gotten completed though.

> this would mean not being able to include the commit message in the generated file

How about using https://api.github.com/repos/exercism/x-common/commits?path=exercises/acronym/canonical-data.json  ?

> I assume we wouldn't want to sometimes have the commit message, sometimes not

Yeah seems confusing. I would like to always have it.

But since it seems possible to always have it, looks like it's possible to keep it, yeah?

petertseng: If you wonder why I linked to Crystal, it's just for reference to see if you can get any ideas from them.

petertseng:
Based on our discussion (everything in https://github.com/exercism/xgo/pull/606#discussion_r109292703 is still relevant, so expand that outdated diff) I am going to show **both** commit and version...

and that honestly makes me think that there should just be a single function that shows this header.

exercises/leap/example_gen.go
```diff
@@ -32,9 +32,7 @@ type js struct {
 // template applied to above data structure generates the Go test cases
 var tmpl = `package leap
 
-// Source: {{.Ori}}
-{{if .Commit}}// Commit: {{.Commit}}
-{{end}}
+{{.Header}}
```
petertseng: One thing I was talking about in https://github.com/exercism/xgo/issues/605#issuecomment-290891332 (and I should have talked about it in #604 as well): 

> If, in the course of writing a generator, we find that there is significant duplicate code between generators that could be extracted into https://github.com/exercism/xgo/blob/master/gen/gen.go, we should do so, so that writing each generator is as little work as possible.

This is one thing I was talking about - we keep writing this header and it's the same three lines every time (and it we add the version, it's going to be four lines). Since it's so common, I propose it be just one line.

robphoenix: 👍

robphoenix:
Is the deprecation of Ori for Origin noted anywhere? That could be part of #604 
Looks great @petertseng

petertseng:
Updated #604 to say that those lines should be replaced with `{{.Header}}` and Ori should be removed after all exercises in #604 are done.

PR #609, opened 2017-04-01
change: Require err to be error by calling Error

petertseng:
As part of #275

Good opportunity to discuss: Do we still want to keep doing things like this to ensure that return values are of type `error`, or are we OK with assuming the student understands this by the time they get to a certain point in the track?

### DIFF
--- a/exercises/change/change_test.go
+++ b/exercises/change/change_test.go
@@ -19,7 +19,7 @@ func TestChange(t *testing.T) {
 		if tc.valid {
 			if err != nil {
 				t.Fatalf("%s : Change(%v, %d): expected %v, got error %s",
-					tc.description, tc.coins, tc.target, tc.expectedChange, err)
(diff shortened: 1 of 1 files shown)


exercises/change/change_test.go
```diff
@@ -19,7 +19,7 @@ func TestChange(t *testing.T) {
 		if tc.valid {
 			if err != nil {
 				t.Fatalf("%s : Change(%v, %d): expected %v, got error %s",
-					tc.description, tc.coins, tc.target, tc.expectedChange, err)
+					tc.description, tc.coins, tc.target, tc.expectedChange, err.Error())
```
leenipper: So this is another way to ensure the 'error' type is used. It is more compact than other options like:
> var _ error = err
I like it.

robphoenix:
My initial feeling is that the tests should continue checking this, that the tests should continue to reflect real world expectations where they can, rather than start to make assumptions after a certain point. But I'm not set on this and am happy to go with the consensus.

PR #610, opened 2017-04-03
Rename test generator files

### LINKED ISSUE #607: Is the example_gen.go file appropriately named?
One thing that I've always been a little unsure of is the filename `example_gen.go`. It's never struck me as descriptively reflective of what it is. Unless I'm missing something, my assumption is that originally it was an example of a generator for an exercise and an appropriate naming convention was never decided upon? 

At this point of going through the process of updating/creating generators I'd like to suggest a potential change of filename. The current name gives the impression it generates the `example.go` file. So, I was thinking something like `test_cases_gen.go` or `<exercise_name>_gen.go`, though it's probably easier to keep the name the same across the exercises.

robphoenix:
closes #607


README.md
```diff
@@ -61,16 +62,20 @@ each file and briefly describe what it is.
...
-for more information.
-
 **example.go** - This is a reference solution for the exercise.
 
 **leap.go** - This is a *stub file*, and will only be present in some exercises.
 
 **leap_test.go** - This is the main test file for the exercise.
 
+**.meta/** - The *.meta* directory can be used to contain files
+that are not meant to be included when a user fetches an exercise, for instance
```
petertseng: ironically, now that this file says `.meta` isn't included, it may imply (solely by omission) that `example.go` *is* delivered to the student - I suggest it would be good to make clear that `exmaple.go` is also not delivered.

README.md
```diff
@@ -61,16 +62,20 @@ each file and briefly describe what it is.
...
+test case generators.
+
+**gen.go** - This file generates the *cases_test.go* file, and will only
```
petertseng: is any confusion possible about whether `gen.go` is in `.meta`? I suppose the tree makes it clear enough

README.md
```diff
@@ -163,17 +168,21 @@ tracks don't.
...
+directory within each exercise that makes use of a test cases generator. This
+*.meta* directory will be ignored when a user fetches an exercise.
+
 Whenever the shared JSON data changes, the test cases will need to be regenerated.
 To do this, make sure that the **x-common** repository has been cloned in the same
 parent-directory as the **xgo** repository. Then navigate into the **xgo**
-directory and run `go run exercises/<problem>/example_gen.go`. You should see
-that the `<problem>/test_cases.go` file has changed. Commit the change.
+directory and run `go run exercises/<exercise>/.meta/gen.go`. You should see
+that the `<exercise>/test_cases.go` file has changed. Commit the change.
```
petertseng: funnily enough, it's `cases_test.go` (simply so that it has the `_test.go` suffix), so while you're changing this line from problem->exercise, probably good to correct this too?

README.md
robphoenix: added a note

gen/gen.go
```diff
@@ -26,11 +26,10 @@ import (
...
 var dirMetadata string
 
-// dirProblem is the location that the test cases should be generated to.
-// This assumes that the generator script lives in the same directory as
-// the problem.
-// Falls back to the present working directory.
-var dirProblem string
+// dirExercise is the location that the test cases should be generated to.
+// This assumes that the generator script lives in the .meta directory within
+// the exercise directory. Falls back to the present working directory.
```
petertseng: "to the present working directory" is no longer accurate now that it's "..", right?

gen/gen.go
```diff
@@ -58,13 +57,14 @@ func init() {
 		dirMetadata = filepath.Join(path, "..", "..", "..", "x-common")
 	}
 	if _, path, _, ok := runtime.Caller(2); ok {
-		dirProblem = filepath.Join(path, "..")
+		dirExercise = filepath.Join(path, "..", "..")
 	}
-	if dirProblem == "" {
-		dirProblem = "."
+	if dirExercise == "" {
+		dirExercise = ".."
```
petertseng: unsure this really needs to change, but since we never get here for any current generator I can't complain since there is no correct answer nor incorrect answer

README.md
petertseng: Another possibility was including it in `example.go`'s explanation. I'm OK with either, so keep the one you like more.

gen/gen.go
robphoenix: ooh, good catch. Left it as is and changed back line 63 in response to your other comment.

gen/gen.go
robphoenix: I feel it's probably better to leave as is.

gen/gen.go
```diff
@@ -60,7 +60,7 @@ func init() {
 		dirExercise = filepath.Join(path, "..", "..")
 	}
 	if dirExercise == "" {
-		dirExercise = ".."
+		dirExercise = "."
```
petertseng: you put this change into the README commit, though it should probably go in the "fix path" commit?

robphoenix: oh fudge.

robphoenix: fixed :heavy_check_mark:

PR #612, opened 2017-04-04
gen.go: collect canonical-data.json remotely

robphoenix:
see https://github.com/exercism/xgo/pull/606#discussion_r109292703

This is a very rough WIP, but I wanted to open it up to get feedback as this is new territory for me.
As it is now, the `Gen()` function will opt to retrieve the canonical data from a local source, but if this isn't available will then try to retrieve it remotely from the x-common Github repo. Is this best? Should it try remotely first?

Feel free to recommend better methods/style etc.

### DIFF
--- a/gen/gen.go
+++ b/gen/gen.go
@@ -7,11 +7,14 @@ import (
 	"fmt"
 	"go/format"
 	"io/ioutil"
+	"log"
+	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"runtime"
 	"text/template"
+	"time"
 )
 
 // dirMetadata is the location of the x-common repository
@@ -30,6 +33,19 @@ var dirMetadata string
 // This assumes that the generator script lives in the .meta directory within
 // the exercise directory. Falls back to the present working directory.
 var dirExercise string
+
+// genClient creates an http client with a 10 second timeout so we don't get
+// stuck waiting for a response.
+var genClient = &http.Client{Timeout: 10 * time.Second}
+
+const (
+	// canonicalDataURL is the URL for the raw canonical-data.json data,
+	// requires exercise name.
+	canonicalDataURL = "https://raw.githubusercontent.com/exercism/x-common/master/exercises/%s/canonical-data.json"
+	// commitsURL is the GitHub api endpoint for the canonical-data.json
+	// file commit history, requires exercise name.
+	commitsURL = "https://api.github.com/repos/exercism/x-common/commits?path=exercises/%s/canonical-data.json"
+)
 
 // Header tells how the test data was generated, for display in the header of cases_test.go
 type Header struct {
@@ -70,11 +86,21 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
 		return errors.New("unable to determine current path")
 	}
 	jFile := filepath.Join("exercises", exercise, "canonical-data.json")
-	// find and read the json source file
-	jPath, jOrigin, jCommit := getPath(jFile)
+	// try to find and read the local json source file
+	log.Printf("[LOCAL] fetching %s test data\n", exercise)
+	jPath, jOrigin, jCommit := getLocal(jFile)
+	if jPath != "" {
+		log.Printf("[LOCAL] source: %s\n", jPath)
+	}
 	jSrc, err := ioutil.ReadFile(filepath.Join(jPath, jFile))
 	if err != nil {
-		return err
+		// fetch json data remotely if there's no local file
+		log.Println("[LOCAL] No test data found")
+		log.Printf("[REMOTE] fetching %s test data\n", exercise)
+		jSrc, jOrigin, jCommit, err = getRemote(exercise)
+		if err != nil {
+			return err
+		}
 	}
 
 	// unmarshal the json source to a Go structure
@@ -119,7 +145,7 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
 	return ioutil.WriteFile(filepath.Join(dirExercise, "cases_test.go"), src, 0666)
 }
 
-func getPath(jFile string) (jPath, jOrigin, jCommit string) {
+func getLocal(jFile string) (jPath, jOrigin, jCommit string) {
 	// Ideally draw from a .json which is pulled from the official x-common
 	// repository.  For development however, accept a file in current directory
 	// if there is no .json in source control.  Also allow an override in any
@@ -138,4 +164,48 @@ func getPath(jFile string) (jPath, jOrigin, jCommit string) {
 	}
 	// good.  return source control dir and commit.
 	return c.Dir, "exercism/x-common", string(bytes.TrimSpace(ori))
+}
+
+func getRemote(exercise string) (body []byte, jOrigin string, jCommit string, err error) {
+	url := fmt.Sprintf(canonicalDataURL, exercise)
+	resp, err := genClient.Get(url)
+	if err != nil {
+		return []byte{}, "", "", err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return []byte{}, "", "", fmt.Errorf("error fetching remote data: %s", resp.Status)
+	}
+	defer resp.Body.Close()
+	body, err = ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return []byte{}, "", "", err
+	}
+	c, err := getRemoteCommit(exercise)
+	if err != nil {
+		// we always expect to have the commit in the cases_test.go
+		// file, so return the error if we can't fetch it
+		return []byte{}, "", "", err
+	}
+	log.Printf("[REMOTE] source: %s\n", url)
+	return body, "exercism/x-common", c, nil
+}
+
+func getRemoteCommit(exercise string) (string, error) {
+	type Commits struct {
+		Sha    string
+		Commit struct {
+			Message string
+		}
+	}
+	resp, err := genClient.Get(fmt.Sprintf(commitsURL, exercise))
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+	var c []Commits
+	err = json.NewDecoder(resp.Body).Decode(&c)
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("%s %s", c[0].Sha[0:7], c[0].Commit.Message), nil
 }
--- a/gen/gen.go
+++ b/gen/gen.go
@@ -17,11 +17,10 @@ import (
 	"time"
 )
 
-// dirMetadata is the location of the x-common repository
-// on the filesystem.
-// We're making the assumption that the x-common repository
-// has been cloned to the same parent directory as the xgo
-// repository. E.g.
+// dirMetadata is the location of the x-common repository on the filesystem.
+// We're making the assumption that the x-common repository has been cloned to
+// the same parent directory as the xgo repository.
+// E.g.
 //
 //     $ tree -L 1 .
 //     .
@@ -47,7 +46,8 @@ const (
 	commitsURL = "https://api.github.com/repos/exercism/x-common/commits?path=exercises/%s/canonical-data.json"
 )
 
-// Header tells how the test data was generated, for display in the header of cases_test.go
+// Header tells how the test data was generated, for display in the header of
+// cases_test.go
 type Header struct {
 	// Ori is a deprecated short name for Origin.
 	// TODO: Remove Ori once everything switches to Origin.
@@ -117,7 +117,7 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
 		Version string
 	}
 	if err := json.Unmarshal(jSrc, &commonMetadata); err != nil {
-		return fmt.Errorf(`Didn't contain version: %v`, err)
+		return fmt.Errorf(`didn't contain version: %v`, err)
 	}
 
 	// package up a little meta data
@@ -133,7 +133,7 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
 
 	// render the Go test cases
 	var b bytes.Buffer
-	if err = t.Execute(&b, &d); err != nil {
+	if err := t.Execute(&b, &d); err != nil {
 		return err
 	}
 	// clean it up
@@ -150,12 +150,12 @@ func getLocal(jFile string) (jPath, jOrigin, jCommit string) {
 	// repository.  For development however, accept a file in current directory
 	// if there is no .json in source control.  Also allow an override in any
 	// case by environment variable.
-	if jPath = os.Getenv("EXTEST"); jPath > "" {
+	if jPath := os.Getenv("EXTEST"); jPath > "" {
 		return jPath, "local file", "" // override
 	}
 	c := exec.Command("git", "log", "-1", "--oneline", jFile)
 	c.Dir = dirMetadata
-	ori, err := c.Output()
+	origin, err := c.Output()
 	if err != nil {
 		return "", "local file", "" // no source control
 	}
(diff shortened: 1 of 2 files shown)


gen/gen.go
```diff
@@ -119,23 +140,67 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
 	if err != nil {
 		return "", "local file", "" // no source control
 	}
 	if _, err = os.Stat(filepath.Join(c.Dir, jFile)); err != nil {
 		return "", "local file", "" // not in source control
 	}
+
 	// good.  return source control dir and commit.
-	return c.Dir, "exercism/x-common", string(bytes.TrimSpace(ori))
+	fmt.Printf("Fetching canonical data from a local source.\nHave you ensured it is up to date?\n")
```
robphoenix: I don't know about you, but I forget to do this all the time! A little reminder here won't go amiss. :smile: I also thought it would be good to indicate if the generator is collecting locally or remotely.

gen/gen.go
```diff
@@ -119,23 +140,67 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
+	fmt.Println("Collecting remote data.")
+	resp, err := genClient.Get(fmt.Sprintf(canonicalDataURL, exercise))
+	if err != nil {
+		return []byte{}, "", "", err
+	}
+	if resp.Status != "200 OK" {
+		return []byte{}, "", "", fmt.Errorf("error fetching remote data: %s", resp.Status)
+	}
+	defer resp.Body.Close()
+	body, err := ioutil.ReadAll(resp.Body)
```
robphoenix: I'd like to change this to a `json.NewDecoder(resp.Body).Decode(&j) but didn't manage it before the end of the day.

gen/gen.go
```diff
@@ -69,31 +84,45 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
 		return errors.New("unable to determine current path")
 	}
+
 	jFile := filepath.Join("exercises", exercise, "canonical-data.json")
 	// find and read the json source file
-	jPath, jOrigin, jCommit := getPath(jFile)
-	jSrc, err := ioutil.ReadFile(filepath.Join(jPath, jFile))
-	if err != nil {
-		return err
+	jPath, jOrigin, jCommit := getLocal(jFile)
```
robphoenix: I think the `getLocal` function could be refined a bit, moving `jFile` into it and returning the `ioutil.Readfile` in line 101. It then needs a different way to decide if there is a valid local source to use. :thinking:

gen/gen.go
```diff
@@ -119,23 +148,67 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
 		return "", "local file", "" // not in source control
 	}
+
 	// good.  return source control dir and commit.
-	return c.Dir, "exercism/x-common", string(bytes.TrimSpace(ori))
+	fmt.Printf("Fetching canonical data from a local source.\nHave you ensured it is up to date?\n")
+	return c.Dir, "local exercism/x-common", string(bytes.TrimSpace(origin))
+}
+
+func getRemote(exercise string) (body []byte, jOrigin string, jCommit string, err error) {
```
robphoenix: I'm not sure about the error handling in this function.

robphoenix:
This is still a bit of a mess, but getting better, I think?

gen/gen.go
```diff
@@ -69,31 +84,45 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
-		// in the case of invalid JSON.  Use a real validator tool if you can't
-		// spot the problem right away.
-		return fmt.Errorf(`unexpected data structure: %v`, err)
-	}
-
-	// These fields are guaranteed to be in every problem
-	var commonMetadata struct {
-		Version string
+	// Decode the json
+	err = json.NewDecoder(bytes.NewReader(body)).Decode(&j)
```
robphoenix: I initially changed to using `json.NewDecoder` over `json.Unmarshal` because that was the recommended way to deal with the response from an `http.Get`. However as I'm essentially converting the `io.Reader` into a `[]byte` and then back again, as we have to decode it twice, once for the `j` struct and once for the `commonMetadata` struct, it should probably stay as it originally was, using `json.Unmarshal`, no?

petertseng: > it should probably stay as it originally was, using `json.Unmarshal`, no?

Right, doesn't seem to make much sense to go from `io.Reader` to `[]byte` back to `io.Reader`. Seems to make sense to keep it as `json.Unmarshal` or return an `io.Reader` from `getRemote` and `getLocal`. Probably good hygiene to close that `Reader` too (which implies it has to be a `ReadCloser`.

gen/gen.go
```diff
@@ -119,23 +148,67 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
+		return []byte{}, "", "", fmt.Errorf("error fetching remote data: %s", resp.Status)
+	}
+	defer resp.Body.Close()
+	body, err = ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return []byte{}, "", "", err
+	}
+	c, err := getRemoteCommit(exercise)
+	if err != nil {
+		return body, "github.com/exercism/x-common", "", nil
```
petertseng: should this return the error?

gen/gen.go
```diff
@@ -69,31 +84,45 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
-	jPath, jOrigin, jCommit := getPath(jFile)
-	jSrc, err := ioutil.ReadFile(filepath.Join(jPath, jFile))
-	if err != nil {
-		return err
+	jPath, jOrigin, jCommit := getLocal(jFile)
+	var body []byte
+	fullPath := filepath.Join(jPath, jFile)
+	_, err := os.Stat(fullPath)
+	switch {
+	case err != nil:
```
petertseng: Thoughts on  `switch` vs `if err != nil`?

gen/gen.go
robphoenix: If it does then, the the whole `Gen()` function will return the error, when all that's happened is the commit hasn't been fetched. So, under these circumstances, it can either generate the *cases_test.go* file without the commit, or if this is unacceptable, error out completely?

petertseng: I would suppose that depends on on whether the users (that'd be us) deem it acceptable.

For me I think I always expect to see the commit there.

But I'll probably keep using the x-common from the filesystem instead of from the web so I suppose my opinion can be overruled.

robphoenix: Ok, if the commit is always expected, then it should return the error.

> But I'll probably keep using the x-common from the filesystem instead of from the web 

Just out of interest, what are your reasons for this?

petertseng: >>  But I'll probably keep using the x-common from the filesystem instead of from the web
> Just out of interest, what are your reasons for this?

I thought about this and I think the most likely scenario to come up is if I'm working on a new generator and I want to run it very often to see whether it works. I wouldn't want to have to continually fetch the new JSON file from the network every time I try that, since it will lengthen the feedback loop and that makes development less pleasant.

robphoenix: That makes a lot of sense, it is the biggest issue with fetching from the network, especially if you have to deal with rural internet! Good reason to keep this using the local first, before resorting to the remote fetch. Thanks for clarifying this for me.

robphoenix:
I'm on holiday currently, which is why this PR has stalled, but I hope t get it finished off next week.

robphoenix:
updated commit message also

gen/gen.go
```diff
@@ -119,23 +146,69 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
 	if err != nil {
 		return "", "local file", "" // no source control
 	}
 	if _, err = os.Stat(filepath.Join(c.Dir, jFile)); err != nil {
 		return "", "local file", "" // not in source control
 	}
+
 	// good.  return source control dir and commit.
-	return c.Dir, "exercism/x-common", string(bytes.TrimSpace(ori))
+	return c.Dir, "local exercism/x-common", string(bytes.TrimSpace(origin))
```
leenipper: Optional.  Wondering if the "local " prefix is necessary. It makes the generated file be different, and perhaps the only difference, if cases_test.go is already up to date with x-common. I suppose the same situation exists for the commit line even though this exercise did not get any changes in canonical-data.json.

Should there be a standard way to run the tests (say with an x-common git tree local) in order to let the cases_test.go have a consistent "// Source: " comment ?  Then, the "local " prefix could stay I suppose.

gen/gen.go
```diff
@@ -119,23 +146,69 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
 }
 
-func getPath(jFile string) (jPath, jOrigin, jCommit string) {
+func getLocal(jFile string) (jPath, jOrigin, jCommit string) {
 	// Ideally draw from a .json which is pulled from the official x-common
 	// repository.  For development however, accept a file in current directory
 	// if there is no .json in source control.  Also allow an override in any
 	// case by environment variable.
-	if jPath = os.Getenv("EXTEST"); jPath > "" {
+	if jPath := os.Getenv("EXTEST"); jPath > "" {
```
tleen: If the caller has set EXTEST in env, but it (accidently) references a nonexistent file it will then keep rolling on to the remote. It does log which is helpful so you can see that the directory EXTEST references is no good as a data source. Should it be rolling on to the remote at that point or fail loud? I would think if EXTEST is defined than there was some intention of using it, in which case not having that file present may be worth shouting about?

gen/gen.go
```diff
@@ -7,18 +7,20 @@ import (
...
 	"text/template"
+	"time"
 )
 
-// dirMetadata is the location of the x-common repository
-// on the filesystem.
-// We're making the assumption that the x-common repository
-// has been cloned to the same parent directory as the xgo
-// repository. E.g.
+// dirMetadata is the location of the x-common repository on the filesystem.
```
petertseng: It would be good if these were in a separate commit since they have nothing to do with collecting canonical data remotely.

gen/gen.go
```diff
@@ -69,16 +85,27 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
+		log.Printf("[REMOTE] fetching %s test data\n", exercise)
+		jSrc, jOrigin, jCommit, err = getRemote(exercise)
+		if err != nil {
+			return err
+		}
 	}
 
 	// unmarshal the json source to a Go structure
-	if err = json.Unmarshal(jSrc, j); err != nil {
+	if err := json.Unmarshal(jSrc, j); err != nil {
```
petertseng: same with these `=` -> `:=`

gen/gen.go
```diff
@@ -91,7 +118,7 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
 		Version string
 	}
 	if err := json.Unmarshal(jSrc, &commonMetadata); err != nil {
-		return fmt.Errorf(`Didn't contain version: %v`, err)
+		return fmt.Errorf(`didn't contain version: %v`, err)
```
petertseng: same

gen/gen.go
petertseng: I'm also not sure local is the amount of detail we want to expose to potential readers of the generated files. including students. what difference does it make to them (or us) whether the file was sourced locally or remotely? what actions might one take differently as a result?

robphoenix: This is a good point, I think I was just focused on separating local and remote initially and didn't consider it beyond that. Will change.

gen/gen.go
robphoenix: I hadn't considered this, good point.
As it is now, if this happened you would get output like so:
```
> EXTEST="doesnt_exist" go run exercises/leap/.meta/gen.go 
2017/04/15 07:00:36 [LOCAL] fetching leap test data
2017/04/15 07:00:36 [LOCAL] source: doesnt_exist
2017/04/15 07:00:36 [LOCAL] No test data found
2017/04/15 07:00:36 [REMOTE] fetching leap test data
2017/04/15 07:00:37 [REMOTE] source: github.com/exercism/x-common
```
which let's you know that there is no test data in the file so you can then check it and fix the env if necessary. Though maybe a quick check in the `getLocal` function, that the file exists would be helpful. I'll look into it, thanks @tleen

gen/gen.go
robphoenix: :+1:

robphoenix:
I did consider a remote flag, wasn't sure if that was overkill or not. I'll add it as a separate commit/PR if I do add it.

gen/gen.go
tleen: Yeah I'm not sure it's worth adding a bunch of (if flag and file not exist) stuff when it's already in the output that we have bailed to remote.

gen/gen.go
```diff
@@ -31,7 +34,21 @@ var dirMetadata string
 // the exercise directory. Falls back to the present working directory.
 var dirExercise string
 
-// Header tells how the test data was generated, for display in the header of cases_test.go
```
petertseng: what would you say about moving the whitespace change to the formatting change commit too?

gen/gen.go
```diff
@@ -107,7 +135,7 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
 
 	// render the Go test cases
 	var b bytes.Buffer
-	if err = t.Execute(&b, &d); err != nil {
+	if err := t.Execute(&b, &d); err != nil {
```
petertseng: probably the `=` -> `:=` as well (into the formatting change commit)

gen/gen.go
```diff
@@ -119,23 +147,69 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
 	// case by environment variable.
-	if jPath = os.Getenv("EXTEST"); jPath > "" {
+	if jPath := os.Getenv("EXTEST"); jPath > "" {
 		return jPath, "local file", "" // override
 	}
+
 	c := exec.Command("git", "log", "-1", "--oneline", jFile)
 	c.Dir = dirMetadata
-	ori, err := c.Output()
+	origin, err := c.Output()
```
petertseng: probably the `ori` -> `origin` too (into the formatting change commit)

gen/gen.go
```diff
@@ -119,23 +147,69 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
+	return c.Dir, "exercism/x-common", string(bytes.TrimSpace(origin))
+}
+
+func getRemote(exercise string) (body []byte, jOrigin string, jCommit string, err error) {
+	url := fmt.Sprintf(canonicalDataURL, exercise)
+	resp, err := genClient.Get(url)
+	if err != nil {
+		return []byte{}, "", "", err
+	}
+	if resp.Status != "200 OK" {
```
petertseng: I don't particularly care which of these you do, but there was the possibility of `resp.StatusCode` and `http.StatusOK`

gen/gen.go
petertseng: Was there anything specific you were not sure about? We previously talked about what the error for the commit message, and I will point out one more thing about checking the status code, but the current way is OK. Anything else we should look at?

robphoenix:
Nope don't mind any and all suggestions. I think I've got everything, but do say if I've missed anything. And there's nothing with the `getRemote` errors that I'm concerned about now.
Thanks for the top notch reviews everyone. I'll merge this this evening if there's no objections. :tada:

gen/gen.go
```diff
@@ -119,14 +146,15 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
 }
 
-func getPath(jFile string) (jPath, jOrigin, jCommit string) {
+func getLocal(jFile string) (jPath, jOrigin, jCommit string) {
 	// Ideally draw from a .json which is pulled from the official x-common
 	// repository.  For development however, accept a file in current directory
 	// if there is no .json in source control.  Also allow an override in any
 	// case by environment variable.
-	if jPath = os.Getenv("EXTEST"); jPath > "" {
+	if jPath := os.Getenv("EXTEST"); jPath > "" {
```
petertseng: Don't forget to move this one to the formatting commit

*maybe* the added newlines too

robphoenix: ok

robphoenix:
oh, and I almost forgot, I'll add a little detail regarding this to the README.

PR #616, opened 2017-04-07
bank-account: Fix typo in the function signature

parroty:
It's minor one, but as it may be used as scaffold for function definition.


tleen:
Thanks!

PR #617, opened 2017-04-08
forth: update generator - 2nd

leenipper:
For issue #604.

- Update generator:
  Use .Header in template.

- Update test program:
  Reference testGroups containing the generated test cases.
  Annotate the FAIL output with test group name & test case description.
  Add PASS output for successful cases.


leenipper:
Sorry about the 2nd go round here.  I accidentally removed the branch against which the 1st pull request was based, instead of force pushing. I confirmed that the changes are equivalent to the now closed 615 PR.

PR #618, opened 2017-04-08
luhn: update generator

leenipper:
For issue #604.

Change Valid => Cases to align template with latest canonical-data.json.


leenipper:
Thanks, and no problem with review late.  I realize one gets to it when time is available.

PR #619, opened 2017-04-08
gigasecond: update generator

leenipper:
For issue #604.

- Update generator:
  Add test description to generated cases.

- Update test program:
  Output test description.
  Add PASS output for successful cases.

### DIFF
--- a/exercises/gigasecond/.meta/gen.go
+++ b/exercises/gigasecond/.meta/gen.go
@@ -22,11 +22,12 @@ func main() {
 
 // The JSON structure we expect to be able to unmarshal into
 type js struct {
-	Add struct {
-		Description []string
+	Cases []struct {
+		Description string
 		Cases       []struct {
-			Input    string
-			Expected string
+			Description string
+			Input       string
+			Expected    string
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


exercises/gigasecond/.meta/gen.go
```diff
@@ -22,29 +22,32 @@ func main() {
...
 var tmpl = `package gigasecond
 
-// Source: {{.Ori}}
-{{if .Commit}}// Commit: {{.Commit}}
-{{end}}
-{{range .J.Add.Description}}// {{.}}
-{{end}}var addCases = []struct {
+{{.Header}}
+
+{{range .J.Cases}}// {{.Description}}
```
petertseng: I hope there won't ever be more than one element in that `.J.Cases`; that would cause odd things to happen (two `var addCases`).

We can deal with that whe nit happens in the future. Maybe it will never happen.

leenipper: ok.

leenipper:
> split into two separate commits.

Should the test program change be in a different pull request, or simply be another commit within this PR ?
And would this be an instance where a squash would not be done, if the latter.

petertseng:
> Should the test program change be in a different pull request, or simply be another commit within this PR ?

Either seems reasonable. They're closely related (tests for gigasecond) so one can justify putting them in the same PR.

> And would this be an instance where a squash would not be done, if the latter.

Agreed.

PR #620, opened 2017-04-11
clock: update generator

leenipper:
For issue #604.

- Update generator:

### DIFF
--- a/exercises/clock/.meta/gen.go
+++ b/exercises/clock/.meta/gen.go
@@ -22,10 +22,12 @@ func main() {
 
 // The JSON structure we expect to be able to umarshal into
 type js struct {
-	Cases []struct {
-		Description string
-		Cases       []OneCase
-	}
+	Groups TestGroups `json:"Cases"`
+}
+
+type TestGroups []struct {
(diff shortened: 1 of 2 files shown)


exercises/clock/.meta/gen.go
```diff
@@ -22,71 +22,60 @@ func main() {
...
 var tmpl = `package clock
 
-// Source: {{.Ori}}
-{{if .Commit}}// Commit: {{.Commit}}
-{{end}}
+{{.Header}}
 
-{{range .J.Create.Description}}// {{.}}
-{{end}}var timeTests = []struct {
+// Test creating a new clock with an initial time.
```
petertseng: hmm, wish it were possible to get this from the `description` in JSON, but seems difficult.

leenipper: I wondered whether it added much generality, as there are 3 fixed group of test cases.  I suppose if the description of a group changes in canonical, a little extra logic that picked that up can carry the updated description into cases_test.go (instead of my static description comment).

I can work with it a bit.

PR #622, opened 2017-04-18
custom-set: update generator

leenipper:
For issue #604.

Use a OneCase struct type to contain the whole
set of possible json objects for the various
test groups.
Add istrSlice as conversion helper for .Expected.

### DIFF
--- a/exercises/custom-set/.meta/gen.go
+++ b/exercises/custom-set/.meta/gen.go
@@ -89,7 +89,7 @@ type OneCase struct {
 	Description string
 	Property    string
 	Set         []int       // "empty"/"contains"/"add" cases
-	Set1        []int       // "empty"/"contains"/"add" cases
+	Set1        []int       // "subset"/"disjoint"/"equal"/"difference"/"intersection"/"union" cases
 	Set2        []int       // "subset"/"disjoint"/"equal"/"difference"/"intersection"/"union" cases
 	Element     int         // "contains"/"add" cases
 	Expected    interface{} // bool or []int
--- a/exercises/custom-set/.meta/gen.go
+++ b/exercises/custom-set/.meta/gen.go
@@ -3,6 +3,7 @@
 package main
 
 import (
+	"errors"
 	"log"
 	"strconv"
 	"text/template"
@@ -12,8 +13,10 @@ import (
 
 func main() {
 	t := template.New("").Funcs(template.FuncMap{
-		"str":  str,
-		"strs": strSlice,
+		"str":   str,
+		"strs":  strSlice,
+		"istrs": istrSlice,
+		"dict":  dict,
 	})
 	t, err := t.Parse(tmpl)
 	if err != nil {
@@ -25,6 +28,7 @@ func main() {
 	}
 }
 
+// str converts an integer 1..26 to a letter 'a'..'z'.
 func str(n int) string {
 	if n >= 1 && n <= 26 {
 		return string('a' - 1 + n)
@@ -32,6 +36,7 @@ func str(n int) string {
 	return strconv.Itoa(n)
 }
 
+// strSlice converts a slice of int to a slice of string.
 func strSlice(ns []int) []string {
 	s := make([]string, len(ns))
 	for i, n := range ns {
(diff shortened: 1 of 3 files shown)


leenipper:
This generator also uses TestGroups and GroupComment, like the clock exercise generator, and also introduces a PropertyMatch function which may also be reusable.  I thought it best to hold any re-factor of clock and custom-set generator with a change to gen/gen.go until later. Handle with separate issue perhaps.

exercises/custom-set/.meta/gen.go
```diff
@@ -108,74 +95,100 @@ type binaryOp struct {
...
 },
-{{end}}}
+{{- end}}{{end}}{{end}} }
 
-// {{.J.Subset.Description}}
-var subsetCases{{template "binaryBool" .J.Subset.Cases}}
+{{with .J.Groups}}
+    // {{ .GroupComment "subset"}}
+{{end}} var subsetCases = []binBoolCase{ {{range .J.Groups}} {{range .Cases}}
+{{if .PropertyMatch "subset"}} { // {{.Description}}
```
robphoenix: Is it possible to pass `"subset"`/`"disjoint"`/`"equal"` etc. to the template so they can be nested as they currently are?

robphoenix:
see #629

exercises/custom-set/.meta/gen.go
```diff
@@ -40,66 +41,52 @@ func strSlice(ns []int) []string {
...
-type binaryOp struct {
-	Description string
-	Cases       []struct {
-		Description string
-		Set1        []int
-		Set2        []int
-		Expected    []int
+func (c OneCase) PropertyMatch(property string) bool { return c.Property == property }
+
+func (groups TestGroups) GroupComment(property string) string {
```
leenipper: Sure.

exercises/custom-set/.meta/gen.go
leenipper: My attempts to do exactly that have not met with success yet.  I have certainly tried to do it, since there is a nice pattern there.  I will study it more and see.

robphoenix: I assumed you had :smile: I'll have a look too if I get time. As you say there is a nice pattern there.

exercises/custom-set/.meta/gen.go
```diff
@@ -40,66 +45,76 @@ func strSlice(ns []int) []string {
...
-	Description string
-	Cases       []struct {
-		Description string
-		Set         []int
-		Expected    bool
+// dict sets up and returns a map for pairs of items given;
+// it is used in order to pass multiple parameters in a template call.
+func dict(values ...interface{}) (map[string]interface{}, error) {
+	if len(values)%2 != 0 {
+		return nil, errors.New("invalid dict call")
```
robphoenix: is it worth having a more explicit error message? `invalid dict call: odd number of values given` or somesuch?

exercises/custom-set/.meta/gen.go
```diff
@@ -40,66 +45,76 @@ func strSlice(ns []int) []string {
...
-		Description string
-		Set1        []int
-		Set2        []int
-		Expected    []int
+func (c OneCase) PropertyMatch(property string) bool { return c.Property == property }
+
+// GroupComment looks in each of the test case groups to find the
+// group for which every test case has the .Property matching given property;
+// it returns the .Description field for the matching property group,
+// or a 'Note: ...' if no test group consistently matches given property.
```
robphoenix: :+1:

exercises/custom-set/.meta/gen.go
```diff
@@ -110,72 +125,78 @@ type binaryOp struct {
...
+{{/* These template calls utilize a dict helper function in order to pass multiple parameters. */}}
+{{template "unaryBool"  dict "PropertyType" "empty"        "Groups" .J.Groups}}
+{{template "eleBool"    dict "PropertyType" "contains"     "Groups" .J.Groups}}
+{{template "binaryBool" dict "PropertyType" "subset"       "Groups" .J.Groups}}
+{{template "binaryBool" dict "PropertyType" "disjoint"     "Groups" .J.Groups}}
+{{template "binaryBool" dict "PropertyType" "equal"        "Groups" .J.Groups}}
+{{template "eleOp"      dict "PropertyType" "add"          "Groups" .J.Groups}}
+{{template "binaryOp"   dict "PropertyType" "intersection" "Groups" .J.Groups}}
+{{template "binaryOp"   dict "PropertyType" "difference"   "Groups" .J.Groups}}
+{{template "binaryOp"   dict "PropertyType" "union"        "Groups" .J.Groups}}
```
robphoenix: You found a way! :tada: You must be pleased with this! :smile:

robphoenix:
LGTM :+1: though again I'd appreciate a second reviewer.

exercises/custom-set/.meta/gen.go
leenipper: > invalid dict call: odd number of values given

That is better. Even if only for readers of the code.

exercises/custom-set/.meta/gen.go
leenipper: Others discovered this solution prior; some googling helped me find it. Looks odd at first, but solves the problem well.

leenipper:
Should this series be squashed and force pushed again, using "For issue #629" in commit message ?

exercises/custom-set/.meta/gen.go
```diff
@@ -40,66 +45,76 @@ func strSlice(ns []int) []string {
...
 	Description string
-	Cases       []struct {
-		Description string
-		Set         []int
-		Element     int
-		Expected    []int
-	}
+	Property    string
+	Set         []int       // "empty"/"contains"/"add" cases
+	Set1        []int       // "empty"/"contains"/"add" cases
```
petertseng: I think the cases for Set1 should be the same as for Set2, not Set?

leenipper: Indeed, you are correct.

PR #643, opened 2017-04-18
Update stub files

robphoenix:
I noticed the stub file for gigasecond had a `?` for the `testVersion` const, which I though was a bit odd in comparison to the other stub files, and then ended up with an attempt to make the stubs a little more consistent.


tleen:
Gah! How did this get past me in #470! :disappointed:

robphoenix:
ah don't worry about it @tleen I miss stuff like this all the time. C'est la vie :smile:

Issue #621, opened 2017-04-14
Example tests are not fetched by the exercism client


robphoenix:
We recently added example tests to this track, #420, #568, however it's been overlooked that because they have the word `example` in the filename these example tests are not being fetched by the exercism client.

I'm not sure where the constraint for this is located within the exercism codebase, so don't know about a solution that could allow us to keep these files named as they are.

Otherwise we can just name them something else, or move the example tests into the main test files, and they'll still run.

petertseng:
> where the constraint for this is located within the exercism codebase

https://github.com/exercism/trackler/blob/master/lib/trackler/implementation.rb#L9

robphoenix:
ah, ok, thanks @petertseng 

Ideally it would be good to keep the example tests in a separate file named `example_test.go` as this appears to be the convention used in the language. This would mean changing the above regex to something like `example(?!_test)` (see https://regex101.com/r/ifNpuG/1). I don't know what unintended side effects on other languages this would have. Do you know @petertseng @kytrinyx? Would it be worthwhile opening an issue regarding this in the trackler or discussions repos?

If we can't go this route then we have these two options:

- use a different filename, for instance `sample_test.go` or `ex_test.go` (these are not good, I know, purely illustrative) which is problematic because it introduces a convention that is exercism specific, not Go specific.
- keep them in the standard test file `<exercise>_test.go`, which is preferable, but in exercises where there are example tests alongside the standard tests, currently only `hello-world` and `clock`, this could make those test files rather long, though this is not a problem for any future exercises that only have example tests.

kytrinyx:
We can use a custom regex in the `config.json` file at the root of the repository.

https://github.com/exercism/request-new-language-track/blob/master/config.json#L6

robphoenix:
Ah, amazing, thanks @kytrinyx. I assume that regex isn't a complete replacement for the one in Trackler, so in the `config.json`, should it be the filename, ie `example_test` or `example_*_test`? Does it need the file extension?

kytrinyx:
@robphoenix The regex is a complete replacement for the one in Trackler. If the config has a regex, then we use it. If not, we fall back to `/test/i`.

It doesn't add any further restrictions at the boundaries (unless the regex in the config does), so you can leave the extension off if you want.

Not all tests will have `example` in them, so you'd need a regex that handles both example_tests and other tests.

robphoenix:
Sorry, but I think I'm a little confused.  

[...]

PR #645, opened 2017-04-20
Add ignore pattern

robphoenix:
see #621
see also exercism/trackler#33

This adds a repo specific ignore pattern. The default pattern
`/example/` excludes example test files, such as
`example_helloworld_test.go`, which should not be ignored.

This pattern ignores a filename that contains the word `example`, unless
the word `example` is followed by the word `test`, with any number of
characters inbetween.

For instance, these files will be ignored:

`example.go`
`example_slice.go` (from the `custom-set` exercise)

however these files will not be ignored:

`example_test.go`
`example_clock_test.go`
`example_helloworld_test.go`


robphoenix:
@Insti I'd appreciate if you could give a quick look over this, just to make sure I've not misunderstood your intentions. Thanks.

robphoenix:
closes #621

Insti:
> Ideally the track should put everything it wants ignored into the .meta directory and set ignore_pattern to ""

This is more a general concept rather than a specific criticism of this PR.
(And I need to check what happens if you set the ignore_pattern to "")

petertseng:
> (And I need to check what happens if you set the ignore_pattern to "")

Incomplete piece of the puzzle:

```
ruby -e 'puts Regexp.new("").match?("arbitrary_path")'
true
```

So if both are true:

* a track sets ignore_pattern to ""
* there is no other special handling of ""

then all paths are ignored.

kytrinyx:
> then all paths are ignored.

Yeah, my assumption was that if you don't want to ignore paths, your ignore_path is not present (or is explicitly null) and you get the default. We can probably safely assume that "" means that they want the default, because ignoring all the paths seems like a pretty silly approach...

kytrinyx:
> We can probably safely assume

I don't actually know whether to explicitly change trackler to protect tracks from defining `//` as the ignore path.

robphoenix:
> Ideally the track should put everything it wants ignored into the .meta directory and set ignore_pattern to ""

Yes, I agree. I haven't done this as we would then have to move the `example.go` file out of the `.meta` dir for it to be picked up by the tests. Though, that shouldn't happen a lot, only during development, not for a user, and we have to either remove or comment out any repeated code in a stub file if there is one. 

@petertseng I'd like to get your opinion on whether we should add this ignore pattern or move the `example.go` file into `.meta`. Essentially, is this PR good to merge?

Insti:
> I haven't done this as we would then have to move the example.go file out of the .meta dir for it to be picked up by the tests. 

Or update the tests to look at `.meta/solutions/example.go` for the example solution.

robphoenix:
Yeah, looking at the test-without-stubs file that Travis uses, it may be less of an issue than I initially thought.

petertseng:
>  Will mean contributors to this track can no longer `go test` to test the example

Yeah I know some *other* tracks don't have this property either, but might as while keep it while we can. If we need to give it up, need to allow `test_without_stubs` to take a single exercise instead of all of them; wouldn't want to have to wait for all the tests to run if only testing one exercise.l

PR #649, opened 2017-04-24
pascals-triangle: add benchmarks

tleen:
Added two benchmarks to the pascals triangle test.

One notable side-effect is that I moved the size of the test from hard-coded to
an independent variable so that it may be used as the size for the fixed length
test and upper bound for the incremental size test.

### DIFF
--- a/exercises/pascals-triangle/pascals_triangle_test.go
+++ b/exercises/pascals-triangle/pascals_triangle_test.go
@@ -8,6 +8,9 @@ import (
 
 const targetTestVersion = 1
 
+// testSize should reflect the length of t20
+const testSize = 20
+
 var t20 = [][]int{
 	{1},
 	{1, 1},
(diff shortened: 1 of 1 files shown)


exercises/pascals-triangle/pascals_triangle_test.go
```diff
@@ -38,15 +41,15 @@ func TestTestVersion(t *testing.T) {
 }
```
leenipper: Could use

`var testSize = len(triangleTestCases)`

instead of the const testSize above.

exercises/pascals-triangle/pascals_triangle_test.go
```diff
@@ -8,6 +8,9 @@ import (
 
 const targetTestVersion = 1
 
+// testSize should reflect the length of t20
+const testSize = 20
+
 var t20 = [][]int{
```
robphoenix: :+1:

exercises/pascals-triangle/pascals_triangle_test.go
robphoenix: :+1:

tleen: Much better!

PR #650, opened 2017-04-25
hamming: update generator

leenipper:
For issue #632

Use .Header in template.
The generated cases_test.go has a new test case from canonical-data.json,
so update the test program and example solution to bump
testVersion and targetTestVersion to 6.


petertseng:
So an interesting question out of curiosity: Would it be helpful to take advantage of https://help.github.com/articles/closing-issues-via-commit-messages/ ? That way you would not have to manually go and close #632 after this gets merged. Well, all good if you prefer to do it manually, but I'm obligated to make sure the option has been explicitly known and rejected if it is to be rejected.

PR #659, opened 2017-04-28
config.json: add info for pascals triangle

robphoenix:
see #345
change difficulty
add topics

### DIFF
--- a/config.json
+++ b/config.json
@@ -159,9 +159,12 @@
       ]
     },
     {
-      "difficulty": 1,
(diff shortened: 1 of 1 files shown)


tleen:
Let's make a mental note to use "Mathematics" and not "Math" for topics, that would be an easy place for us to colloquially diverge.

robphoenix:
It's in the topics list as `Mathematics` - https://github.com/exercism/x-common/blob/master/TOPICS.txt#L57

PR #660, opened 2017-04-28
config.json: add info for series

robphoenix:
see #345
change difficulty
add topics

### DIFF
--- a/config.json
+++ b/config.json
@@ -164,9 +164,13 @@
       "topics": []
     },
     {
-      "difficulty": 1,
+      "difficulty": 3,
       "slug": "series",
-      "topics": []
+      "topics": [
+        "Strings",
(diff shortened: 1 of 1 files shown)


tleen:
Perhaps we should add "tests" as a topic for this one because it introduces the concept of failing tests and writing more advanced functions to pass tests in the bonus material. I'd say add "slices" but that sort of gives away the most common solution.

robphoenix:
I see what you mean, but I think `tests` could probably be considered a topic for all exercises, with exercism being focused on TDD. I think the bonus material is perhaps more about slice safety/error reporting/function returns. I'd be tempted to add `Refactoring` actually, what do you think?

tleen:
Refactoring is more accurate, its a good add.

PR #661, opened 2017-04-28
config.json: add info for queen-attack

robphoenix:
see #345
change difficulty
add topics

### DIFF
--- a/config.json
+++ b/config.json
@@ -169,9 +169,14 @@
       "topics": []
     },
     {
-      "difficulty": 1,
+      "difficulty": 5,
       "slug": "queen-attack",
-      "topics": []
+      "topics": [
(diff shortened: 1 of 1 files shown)


robphoenix:
I'm a little uncertain about these topics, any thoughts?

tleen:
Why booleans? Is is because the result is a true/false? 

One keyword that may be appropriate is something along the lines of "matrix" or "multi-dimensional (array)" since that is the conceptual visualization?

tleen:
Perhaps "Errors" would be a good topic since this (and hamming) is an early exercise requiring errors as part of the function sig.

robphoenix:
I like both these suggestions, I'll add them in. :+1:

PR #662, opened 2017-04-28
series: change package name to better reflect exercise name

tleen:
This exercise had a package name which was odd compared to the other
exercises in the track: it did not contain any part of the exercise name.
In fact it seems to be the name of the core concept behind the probable
solution in Go.

This also updates the test version since the test would now be in a
different package.


tleen:
Not too pedantic?

leenipper:
> Not too pedantic?

No. I think it adds conformity and should be less confusing to students.

PR #666, opened 2017-05-05
gen/gen.go: improve error support

leenipper:
Update gen/gen.go package to give additional information
when an error occurs.

This change outputs an additional log output line to indicate
the context of the error, either during template Execute
or during the source Format step.

Also, when the error occurs during the source Format,
go ahead and output the generated source to the cases_test.go
file anyway with an extra line appended at the end which contains the
error message from the source Format.
    
Being able to examine the erroneous source text makes it easier
to understand the specific issue with the template/generator output.
    
Show either [SUCCESS] or [ERROR] output: /path/to/cases_test.go
in a log message after saving the generated source.

Four cases of generator run output (3 error types, and one success) are below:
>$ go run .meta/gen.go 
>2017/05/06 20:11:32 template: :11: unexpected "}" in operand
>exit status 1

>$ go run .meta/gen.go
>2017/05/06 20:13:31 [LOCAL] fetching bowling test data
>2017/05/06 20:13:31 [LOCAL] source: /home/lee/x-common/exercises/bowling/canonical-data.json
>2017/05/06 20:13:31 [ERROR] template.Execute failed. The template has a semantic error.
>2017/05/06 20:13:31 template: :16:16: executing "" at <.Scores>: can't evaluate field Scores in type main.OneCase
>exit status 1

>$ go run .meta/gen.go 
>2017/05/06 20:17:05 [LOCAL] fetching bowling test data
>2017/05/06 20:17:05 [LOCAL] source: /home/lee/x-common/exercises/bowling/canonical-data.json
>2017/05/06 20:17:05 [ERROR] format.Source failed. The generated source has a syntax error.
>2017/05/06 20:17:05 [ERROR] output: /home/lee/xgo/exercises/bowling/cases_test.go
>2017/05/06 20:17:05 16:49: missing ',' before newline in composite literal (and 10 more errors)
>exit status 1

>$ go run .meta/gen.go 
>2017/05/06 20:21:04 [LOCAL] fetching bowling test data
>2017/05/06 20:21:04 [LOCAL] source: /home/lee/x-common/exercises/bowling/canonical-data.json
>2017/05/06 20:21:04 [SUCCESS] output: /home/xgo/exercises/bowling/cases_test.go

### DIFF
--- a/gen/gen.go
+++ b/gen/gen.go
@@ -77,6 +77,18 @@ func init() {
 	}
 }
 
+// outputSource puts the src text into given fileName
+// and outputs a log message with given [status].
+func outputSource(status string, fileName string, src []byte) error {
+	ioerr := ioutil.WriteFile(fileName, src, 0666)
+	if ioerr != nil {
+		log.Printf("[FAILED] %s\n", ioerr)
+		return ioerr
+	}
+	log.Printf("[%s] output: %s\n", status, fileName)
+	return nil
+}
+
 // Gen generates the exercise cases_test.go file from the relevant canonical-data.json
 func Gen(exercise string, j interface{}, t *template.Template) error {
 	if dirMetadata == "" {
@@ -128,18 +140,30 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
 		Version: commonMetadata.Version,
 	}, j}
 
+	casesFileName := filepath.Join(dirExercise, "cases_test.go")
+
 	// render the Go test cases
 	var b bytes.Buffer
 	if err := t.Execute(&b, &d); err != nil {
+		log.Print("[ERROR] template.Execute failed. The template has a semantic error.")
 		return err
 	}
 	// clean it up
-	src, err := format.Source(b.Bytes())
+	srcBuf := b.Bytes()
+	src, err := format.Source(srcBuf)
 	if err != nil {
+		log.Print("[ERROR] format.Source failed. The generated source has a syntax error.")
+		b.Reset()
+		_, _ = b.Write(srcBuf)
+		_, _ = b.Write([]byte(
+			"// !NOTE: Error during source formatting: Line:Column " + fmt.Sprint(err) + "\n"))
+		src = b.Bytes()
+		// Save the raw unformatted, error-containing source for purposes of debugging the generator.
+		_ = outputSource("ERROR", casesFileName, src)
 		return err
 	}
 	// write output file for the Go test cases.
-	return ioutil.WriteFile(filepath.Join(dirExercise, "cases_test.go"), src, 0666)
+	return outputSource("SUCCESS", casesFileName, src)
 }
 
 func getLocal(jFile string) (jPath, jOrigin, jCommit string) {


gen/gen.go
```diff
@@ -128,18 +128,47 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
 		Version: commonMetadata.Version,
 	}, j}
 
+	// Use two potential output files, one for success, and one for format.Source failure.
+	casesFileName := filepath.Join(dirExercise, "cases_test.go")
+	casesErrorFileName := filepath.Join(dirExercise, "cases_test.go-error")
```
robphoenix: Perhaps a different name, such as `failed_cases_test.go`, or `error_cases_test.go`, just so it can still be picked up in an editor as a `go` file?

leenipper: That is certainly worth considering.  I first sided on making it something that will not be a valid .go file since Format has already failed on it.  But editors can also show syntax errors too quite easily, so making it a .go helps that happen.  With a non-standard extension, I felt free to use the .gitignore with that specific file name quite safely.  If it has a valid .go name, then I didn't feel as comfortable putting _that_ name in a .gitignore file.

gen/gen.go
```diff
@@ -128,18 +128,47 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
-	src, err := format.Source(b.Bytes())
+	srcBuf := b.Bytes()
+	src, err := format.Source(srcBuf)
 	if err != nil {
+		log.Print("Error during format.Source")
+		b.Reset()
+		_, _ = b.Write(srcBuf)
+		_, _ = b.Write([]byte("// !NOTE: Error during source formatting: Line:Column " + fmt.Sprint(err) + "\n"))
+		src = b.Bytes()
+		// Remove any stale cases_test.go
```
leenipper: No, the comment itself is stale! Thanks.

gen/gen.go
```diff
@@ -128,18 +128,47 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
+		log.Print("For details, examine file ", casesErrorFileName)
+		// Save the raw unformatted, error-containing source for purposes of debugging the generator.
+		_ = ioutil.WriteFile(casesErrorFileName, src, 0666)
 		return err
 	}
 	// write output file for the Go test cases.
-	return ioutil.WriteFile(filepath.Join(dirExercise, "cases_test.go"), src, 0666)
+	ioerr := ioutil.WriteFile(casesFileName, src, 0666)
+	if err != nil {
```
leenipper: Ah, you are correct.  I think this logic applied with an earlier version when the coded did not return in the section prior, but now it doesn't help at all.  I will revert that change.

leenipper:
Thanks for the feedback @robphoenix.  I've been using a version of this during recent generator changes.  When the generator template is more complex, it can really help out to understand the problem.

gen/gen.go
robphoenix: Yes, that makes sense, I'm not set on this by any means. I wonder though if someone does want to access it as a normal go file and so they change the extension to `.go` and then it's not protected from committing by the `.gitignore` file, but then I don't suppose we're going to accidentally let a commit with a failing test file into the codebase? 
Another idea might be to write the error output to `stdout` and then it's open to piping to a file if so desired? This does then put the onus on the committer to not version control it. I've certainly accidentally version controlled files I don't want to before, maybe that's more trouble. Just thinking aloud here really. :smile:

robphoenix:
Yes, I fully agree with the idea, especially any improvements you've found with all the work you've done with the generators. Just a couple of things to make sure we make the most appropriate decision regarding the implementation.

gen/gen.go
leenipper: Another idea which I have rolling around in my head on this filename issue is to add a command line option (handled within gen package, I suppose) to route raw source to filename of user's choice.  So for instance:

`go run .meta/gen.go  -o  tmp.go`
would route the raw source to file named tmp.go. Then user can view it in editor.
And then perhaps:
`go run .meta/gen.go  -o  -`
would simpy output to stdout (terminal by default).
Running without the -o option would simply be similar to existing generator where you either get the cases_test.go produced (on success), or just a log error message on failure.
But running with the -o option would output the source (error or no error) to given file(or stdout).

With this approach, I would drop the .gitignore in exercies altogether, since the filename choice (temporary usually) is left with developer, who could even put the errored source in /tmp/junk.go or something completely out of the way.  What do you think ?  I'll work with it some.

petertseng:
Trying to understand the possible errors:

* Template-language syntax error: `2017/05/06 03:20:50 template: :3: unexpected "}" in operand` (this behaviour is unchanged)
* Template is syntactically correct but semantically incorrect, for example references an unknown field: `2017/05/06 03:22:03 template: :4:2: executing "" at <.GARBAGEY>: can't evaluate field GARBAGEY in type *struct { gen.Header; J interface {} }` (behaviour mostly unchanged, but also prints "Error during template Execute", seems OK)
* Template is correct, but the generated source code is invalid Go, so `format.Source` fails. `2017/05/06 03:16:29 9:17: expected ';', found ',' (and 2 more errors)`. Go source is left in an go-error file. This is the behaviour change. Previously it wouldn't show the Go source at all.

I think I understand the change. Nice idea.

petertseng:
> any side effects of adding this to the exercises dir

Seems unlikely, right? Given that TRACK_HINTS.md is there. Trackler only looks for exercises we define in our config file, right? But I wouldn't know, you would have to look in the Trackler repository to find out the answer.

gen/gen.go
leenipper: After thinking about it a little more, I have an even simpler idea.  I can just retain the source in error on the cases_test.go file, and output ERROR log messages to clearly indicate the error. It is really no worse than retaining the old cases_test.go when the template/generator has a problem as is done now.  At least if the developer doesn't notice the error message output, another syntax error will be shown when doing the "go test".  Is there any strong reason not to output the source which has the syntax error onto the cases_test.go file anyway?  The developer's file changes (generator & cases_test.go) are in an incomplete state when a generator error occurs, so I'm thinking this is the simplest change that achieves what is needed (source available to examine for better hints) and also not be too confusing (at least to me).  No new option is necessary, and the .gitignore file is unnecessary.   Comments ?

robphoenix:
So, if I understand correctly, this now writes the failed generation output to `cases_test.go`, but that if someone wanted to they could pass a different filename in the `gen/gen.go` file, [here](https://github.com/leenipper/xgo/blob/c9e89f095cb2a7c1cdbdebb86def6d4b626c57aa/gen/gen.go#L162)?
If I've not misunderstood then this looks good @leenipper. I have no objections to this being merged. :+1: 

Also, as @petertseng mentioned [previously](https://github.com/exercism/xgo/pull/612#pullrequestreview-32914910) the error filename could end up being passed to `gen/gen.go` as an argument.

leenipper:
> but that if someone wanted to they could pass a different filename in the gen/gen.go file, here?

Yes. But the gen/gen.go is the package used by .meta/gen.go generator of each exercise, so it won't typically be done, if I get your point.

> Also, as @petertseng mentioned previously the error filename could end up being passed to gen/gen.go as an argument.

I am a bit confused by that reference.  Context seems to be something else.

I edited this pull request initial comments (at the beginning of the PR) to indicate the way it works for the 3 different error types with the updated commit pushed for the PR.  That might have been a poor idea; but see if it makes sense @robphoenix .  Right now, this PR commit will output the generated source to cases_test.go even if it has a syntax error. I had explained the reasoning in my last comments on the previous outdated commit.

robphoenix:
Sorry, I missed your updated comments completely, ignore my reference to flags. What you say makes total sense, simplicity is always good. 😀

PR #668, opened 2017-05-05
Test for binary-search-tree now handles missing nodes correctly

tapank:
This is a fix for https://github.com/exercism/xgo/issues/667. The issue can be reproduced by using the code provided in the github issue.


robphoenix:
Thanks for this @tapank and for raising the issue. I don't have the time this week to look over this, sorry, but I'll certainly give it the attention it deserves first thing next week.

tapank:
Thanks @robphoenix, please take your time. Let me know if you need any more information, or any updates to the PR.

exercises/binary-search-tree/binary_search_tree_test.go
```diff
@@ -45,8 +45,12 @@ func TestInsertingLess(t *testing.T) {
 		t.Errorf("bst.data: %d, want %d.", actual, expected)
 	}
 
-	actual = bst.left.data
 	expected = 2
+	if bst.left == nil {
+		t.Errorf("bst.left: nil, want %d.", expected)
```
petertseng: could consider using `t.Fatalf` instead of `t.Errorf` + `return` (applies to other lines too but I don't want to repeat myself)

exercises/binary-search-tree/binary_search_tree_test.go
```diff
@@ -101,38 +113,62 @@ func TestComplexTree(t *testing.T) {
...
+		return
```
petertseng: I have a situation to share, but I am not sure if it is a problem worth solving. Here goes.

Let's say in response to this test, I make a `bst` whose `data` is 4 and both `left` and `right` are nil.

The old test would panic, this is obviously bad.
The new test doesn't panic, tells me `bst.left` is `nil`, much better!

The question is: can/should the test also go on to tell me that in addition to `bst.left` being `nil`, `bst.right` is also nil?

The interesting thing is that if previously I had created a tree whose node values were all 0, because there were no `return` nor `Fatalf`, the test would have gone through and told me **everything wrong with the tree**. Now, if we run into a `nil`, we stop looking for any other problems with the tree.

Feel free to decide that the behaviour in this PR is fine, since the proposed addition of finding everything wrong with the tree would probably involve putting more things in conditionals and the indenting will make the change larger.

tapank:
Thank @petertseng for your comments, I learned something new.  Indeed t.Fatalf is cleaner.  I have made the change (everywhere it applies).

> can/should the test also go on to tell me that in addition to bst.left being nil, bst.right is also nil?

Yes, that would be better than not checking the 'other' node.  But as you said, that is probably a larger fix and can be taken up separately.  Ideally, there can be a deep compare function that checks the following:

1. Check if one of the two bsts is nil
2. Check if both trees have the same shape. Currently, we only check what is expected, but leave out what is not expected, i.e. an extra node somewhere will not be detected.
3. Check if all nodes of both bsts have same values. (This is already being done).

BTW, @petertseng , after I updated the pull request, I do not see your comments.  Apologies if I should have submitted the new change as a new commit, rather than overwriting history.

petertseng:
> I  do not see your comments

They do still exist, but one must use "Show outdated" to show them.

petertseng:
Thanks for your work!

PR #670, opened 2017-05-10
declare compliance with x-common 1.0.0, where applicable

petertseng:
For all non-generated exercises, I examined the x-common canonical data and determined whether the track was already compliant. If so, I marked it as such.


petertseng:
Maybe don't need to show students "converted from x-common by hand"

petertseng:
Actually I honestly don't think we need to merge these. These are informational, to tell us which exercises are up to date, but I can't think of a reason why students would need to know (tell me if you can think of one).

Issue #605, opened 2017-04-01
Write generators for all exercises


petertseng:
In [x-common](https://github.com/exercism/x-common/tree/master/exercises), we have various exercises with canonical-data.json representing the recommended set of tests for that exercise. We created generators to more easily keep up with changes in x-common. Whenever x-common changes, the task is as simple as re-running the generator.

The current documentation is at https://github.com/exercism/xgo#generating-test-cases - if at any point we find that we should explain something better, we should add any necessary documentation.

Preliminaries:

* I find the list of our generators using `ls exercises/*/cases_test.go | cut -d/ -f2` from the root of the xgo directory.
* I find the list of exercises with canonical-data.json using `ls exercises/*/canonical-data.json | cut -d/ -f2` from the root of the x-common directory.

This issue concerns **exercises that don't yet have a generator**.

For each exercise, first we should decide **whether we even want a generator**.
An example where we don't want one is diamond, since property-based tests aren't represented in canonical-data.json.
So don't be afraid to suggest that we pass on having a generator if it creates more work than it saves.
After all, the point is to make maintenance of this track easy.

Or, we may decide to lazy-evaluate the generator:
If our current tests already match x-common, there is no current need to have a generator since it will make no difference currently.
So we may decide that we'll make a generator the next time the x-common JSON changes, and until then we'll stick with what we have.

If we decide we **do** want a generator for a given exercise:
First, we should check whether there are any cases that our track has that the common JSON doesn't have.
For each of those, we should consider the options:

* Propose adding it to the common JSON if we feel it can be valuable to other tracks
* Do nothing and just run the new generator (thus losing the extra case) if we feel the extra case does not contribute anything meaningful to this track
* Have our track's generator add the case after the JSON is parsed if we feel it adds value to the track but wouldn't add value to other tracks

- [x] acronym ( #691 )
- [x] allergies ( #938 )
- [x] all-your-base ( #945 )
- [x] anagram
- [x] atbash-cipher ( #673 )
- [x] beer-song
- [x] ~~binary~~ Is deprecated in this track.
- [x] binary-search ( #957 )
- [x] bracket-push ( #672 )
- [ ] circular-buffer **Warning**: Stateful test.
- [x] collatz-conjecture ( #958 )
- [ ] crypto-square Please note https://github.com/exercism/x-common/issues/750 before generating.
- [x] ~~diamond~~ Uses property-based tests, so won't be generated.
- [x] difference-of-squares
- [x] etl
- [x] food-chain
[...]

PR #674, opened 2017-05-18
variable-length-quantity: create test case generator

leenipper:
For #605

Generate two test case arrays, one for encode and
one for decode, from the canonical-data.json.

Update example solution to pass the tests.
The API changed to use slices in order to match
the canonical-data.json test cases.
Retain the old functions but rename to decodeInt and encodeInt
and utilize these as helpers to support the new slice based API.

Update test program to use generated test case arrays.
Put FAIL and PASS in test result output.

Increment testVersion and targetTestVersion in
example solution and test program.

### DIFF
--- a/exercises/variable-length-quantity/.meta/gen.go
+++ b/exercises/variable-length-quantity/.meta/gen.go
@@ -1,0 +1,113 @@
+package main
+
+import (
+	"log"
+	"text/template"
+
+	"../../../gen"
+)
+
+func main() {
+	t := template.New("").Funcs(template.FuncMap{
+		"GroupComment": GroupComment,
+		"byteSlice":    byteSlice,
+		"lengthSlice":  lengthSlice,
+	})
+	t, err := t.Parse(tmpl)
+	if err != nil {
+		log.Fatal(err)
+	}
+	var j js
+	if err := gen.Gen("variable-length-quantity", &j, t); err != nil {
+		log.Fatal(err)
+	}
+}
+
+// byteSlice converts a slice of uint32 to a byte slice.
+func byteSlice(ns []uint32) []byte {
+	b := make([]byte, len(ns))
+	for i, n := range ns {
+		b[i] = byte(n)
+	}
+	return b
+}
+
+// lengthSlice returns the length of given slice.
+func lengthSlice(ns []uint32) int {
+	return len(ns)
+}
+
+// The JSON structure we expect to be able to unmarshal into
+type js struct {
+	Groups []TestGroup `json:"Cases"`
+}
+
+type TestGroup struct {
+	Description string
+	Cases       []OneCase
+}
+
+type OneCase struct {
+	Description string
+	Property    string   // "encode" or "decode"
+	Input       []uint32 // supports both []byte and []uint32 in JSON.
+	Expected    []uint32 // supports []byte, []uint32, or null in JSON.
+}
+
+// PropertyMatch returns true when given test case c has .Property field matching property;
+// this serves as a filter to put test cases with "like" property into the same group.
+func (c OneCase) PropertyMatch(property string) bool { return c.Property == property }
+
+// GroupComment looks in each of the test case groups to find the
+// group for which every test case has the .Property matching given property;
+// it returns the .Description field for the matching property group,
+// or a 'Note: ...' if no test group consistently matches given property.
+func GroupComment(groups []TestGroup, property string) string {
+	for _, group := range groups {
+		propertyGroupMatch := true
+		for _, testcase := range group.Cases {
+			if !testcase.PropertyMatch(property) {
+				propertyGroupMatch = false
+				break
+			}
+		}
+		if propertyGroupMatch {
+			return group.Description
+		}
+	}
+	return "Note: Apparent inconsistent use of \"property\": \"" + property + "\" within test case group!"
+}
+
+// template applied to above data structure generates the Go test cases
+var tmpl = `package variablelengthquantity
+
+{{.Header}}
+
+// {{GroupComment .J.Groups "encode"}}
+var encodeTestCases = []struct {
+	description string
+	input	[]uint32
+	output	[]byte
+}{ {{range .J.Groups}} {{range .Cases}}
+	{{if .PropertyMatch "encode"}} {
+		{{printf "%q" .Description}},
+		{{printf "%#v" .Input }},
+		{{byteSlice .Expected | printf "%#v" }},
+	},{{- end}}{{end}}{{end}}
+}
+
+// {{GroupComment .J.Groups "decode"}}
+var decodeTestCases = []struct {
+	description string
+	input	[]byte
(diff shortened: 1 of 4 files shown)


exercises/variable-length-quantity/.meta/gen.go
```diff
@@ -0,0 +1,146 @@
...
+type TestGroup struct {
+	Description string
+	Cases       []OneCase
+}
+
+type OneCase struct {
+	Description string
+	Property    string      // "encode" or "decode"
+	Input       []uint32    // supports both []byte and []uint32 in JSON.
+	Expected    interface{} // supports []byte, []uint32, or null in JSON.
```
petertseng: two thing stand out to me:

1. I was a bit surprised: I thought that since `nil` is assignable to `[]uint32` that `uint32` can be used here
2. I feel like the schema is not doing well for us. In the old way, where `encode` and `decode` are keys, we can have a different struct for each kind of case. But since now we can only tell by looking at `property`, we have to make our `OneCase` struct a union of all possible cases. Since you have written more generators than I have, I would ask you: How much does this bother you?

petertseng: > I thought that since nil is assignable to []uint32 that uint32 can be used here

Sorry I meant `[]uint32` of course.

petertseng: > How much does this bother you?

And I guess more importantly, does it bother enough that it's worth proposing a change?

leenipper: > []uint32 can be used here

Actually, I think you are right.  I jumped to the more general interface{}, likely because I had to previously for a different generator.

> In the old way, where encode and decode are keys, we can have a different struct for each kind of case.

This made it easier, I agree.

> does it bother enough that it's worth proposing a change?

Well, I've not thought much about what the change might be. But I do like simplicity, and the newer approach in the JSON is adding complexity to the generators. Perhaps there is a compromise.  Do you know the rational for the new way in the JSON ?

petertseng: > Do you know the rational for the new way in the JSON ?

I didn't know off the top of my head, so I went searching. The clearest explanation is probably: https://github.com/exercism/x-common/issues/336#issuecomment-279880708 : 

> In most of the test suites with more than one type of test, the test's type is encoded
in a property-key, in an object describing a test group.
> 
> There are two problem with that approach:
>
> * It mixes two different concepts regarding the tests: grouping and identification
> * It doesn't allow nesting of test groups. That would be nice to have, but is not really needed.
> * It doesn't allow grouping of test of different types, which would be really great.
> 
> Moving the **test type** near the **test data**, we solved all the above problem easily. It is theoretically sound and adds functionality.

I wonder if it would be too much to ask to have the keys again.

I'm wondering what other statically-typed language has generators. There's https://github.com/exercism/xscala/tree/master/testgen/src/main/scala , might need to take a look at what they're doing.

petertseng: and https://github.com/exercism/xocaml/tree/master/tools/test-generator

leenipper:
I pushed an update which has a simpler generator.

exercises/variable-length-quantity/.meta/gen.go
petertseng: I considered another alternative, which is... can we deal with this on the Go side without needing a change in x-common? I tried it in #677.

exercises/variable-length-quantity/example.go
```diff
@@ -25,34 +29,76 @@ func EncodeVarint(x uint32) []byte {
...
 	var b byte
 	for n, b = range buf {
 		x = x << 7
-		x |= uint32(b) & 0x7F
+		x |= uint32(b) & 0x7f
 		if (b & 0x80) == 0 {
-			return x, n + 1
+			return x, n + 1, nil
+		} else if n == len(buf)-1 {
+			// error case
```
petertseng: since this is the last iteration, this isn't necessary, is it?

petertseng: yeah, still works if this is removed.

exercises/variable-length-quantity/.meta/gen.go
```diff
@@ -0,0 +1,115 @@
+// +build ignore
```
petertseng: we don't need to `build ignore` generators anymore, now that they're in the separate directory, right? They originally were to stop them from being caught when we `go test`, but now no longer necessary.

Doesn't need to be in this PR, maybe I'll make a separate one to remove them.

exercises/variable-length-quantity/example.go
petertseng: I just checked - the example solution still passes the tests even if these lines are removed, so I would suggest removing them.

leenipper: Yes. Will remote those lines.

exercises/variable-length-quantity/.meta/gen.go
leenipper: Ok, can be done with one sweeping commit on all the .meta/gen.go files easily later.

exercises/variable-length-quantity/.meta/gen.go
leenipper: Yes, I've been looking at that one.

exercises/variable-length-quantity/.meta/gen.go
petertseng: I'll make an issue so I don't forget (or if someone else wants to take it)

leenipper:
Well, I did wonder about the 0-length array, whether it was clear enough.  Do you think it is worth adding an errorExpected bool for clarity in the test data ?  Will be fairly easy to do.

petertseng:
I would add it - I'm looking at the example of https://github.com/exercism/xgo/blob/master/exercises/largest-series-product/cases_test.go where we could find out via `expected < 0` whether it's an error, but we still add an `ok` boolean anyway.

However, just to make sure I'm not falling to confirmation bias, I looked for a counterexample, and I found it in Forth: https://github.com/exercism/xgo/blob/master/exercises/forth/cases_test.go#L15 there is a comment there:

> // nil slice indicates error expected.

So it looks like there are two possible ways of making it clearer, so we have a choice.

leenipper:
Ok. I chose to add the nil slice comment, and I updated test program to compare against nil as well.

petertseng:
I just remembered something... forgot to check that `err` is an `error`? (#275)

Maybe we should only check it in the first few exercises that have `error` and not worry about it for the rest.

exercises/variable-length-quantity/.meta/gen.go
petertseng: ok now that #682 is closed, gonna ask you to remove it from this one

leenipper:
Accidental close earlier. Sorry. 
I updated the PR commit without the +build tag and with the var _ error check.

Issue #420, opened 2016-11-28
Consider using Example tests for some exercises


kytrinyx:
One thing that would perhaps lower the barrier even further on the Hello World exercise is to use Example tests, make it a `main` package with a `main` function, and just have people write to standard out.

What do you think?

/cc @ardan-bkennedy for the suggestion
/cc @petertseng @ferhatelmas as maintainers on the track

petertseng:
By example tests, does this mean a separate function to test each case (currently "" and "Gopher") rather than one function performing a table-based test?

Or does it mean https://golang.org/pkg/testing/#hdr-Examples which I did not know existed until I searched for it just now?

Should any other exercises do this? It is good to get exposed to other techniques. It might be a bit jarring to students if hello world does it one way and then all other exercises do it another way. So should this be used in some other exercises too?

ardan-bkennedy:
I don't think it needs to be an all or nothing with the type of tests. There are many exercises where printing to stdout can make the exercise simpler. Using an example test can simplify everything. Just writing out "Hello, World" and on the next line "Hello, Gopher" would be simple.

When it comes to more of the concurrency exercises, writing to stdout could also allow for more interesting exercises.

kytrinyx:
@petertseng I meant the second one (https://golang.org/pkg/testing/#hdr-Examples).

> Should any other exercises do this?

Yeah, I think we should consider using it for some of the other exercises as well. I think it's worth exposing people to the different types of testing that the standard library has.

robphoenix:
I was initially quite confused about Example tests, and started off implementing them in a couple of exercises as a replacement for the current table-driven tests. This didn't initially feel quite right as doing this would mean we wouldn't be able to take advantage of the test cases generation. 

So, my understanding of the example tests, helped enormously by [this](https://blog.golang.org/examples), is that they're similar to Python's doctests and can be useful for clarifying, and verifying, the API expected of an exercise. This can be very useful for exercises such as clock, where I've seen confusion before regarding what is expected (#448).

As such I've opened #568. I don't know if we want to use only example tests, in place of the current mostly table driven tests, in some exercises? I don't expect we will need or want Example tests in all exercises, as this may make things too obvious. I hope I've not totally misunderstood and that #568 introduces Example tests in a meaningful way. If nothing else it helped me to understand what they are and how to use them (surely there's a blog post in that! :thinking: )

kytrinyx:
> surely there's a blog post in that!

Indeed!

[...]

PR #675, opened 2017-05-19
[Do Not Merge] acronym: replace test file with example test

robphoenix:
see #420


robphoenix:
Ok, so I'm actually presenting this as a case against replacing the standard test files with example tests. Why? Well I think they present a degraded feedback loop when errors occur and this is not the intended use for them. If a user get's something wrong the error returned will be something like this:

```
--- FAIL: ExampleAbbreviate (0.00s)
got:
PNG
ROR
HTML
FIFO
PHP
GIMP
CMOS
want:
WRONG
PNG
ROR
HTML
FIFO
PHP
GIMP
CMOS
FAIL
exit status 1
FAIL	github.com/robphoenix/xgo/exercises/acronym	0.003s
```

Which can become a confusing wall of text in all but examples with a small output, rather than a helpful message. In standard tests we have a lot more control over how we test the code and report back on any errors that occur. Even in small exercises like this I don't think having only example tests is beneficial. If we want to introduce users to this other way of verifying code I believe it would be better done as an addition to a standard test file. 
I suggest we close #420 and ensure we properly document the potential for using example tests as a helpful addition in exercises where there may be confusion, as in `clock`.

tleen:
I think @robphoenix is right here. The more I work with Examples the more it seems like we would be overloading their purpose to use them as actual tests against the solutions. Ideally the tests fail early and fail informatively. Example tests fail monolithically and generically.

petertseng:
Yeah, not really seeing their usefulness here. It would kind of make sense if there were one line of output per test, but in that case I don't also see the benefit over a table-based test.

robphoenix:
OK, it seems there's a general agreement that Example tests are not here to replace table-based tests, rather their usefulness is in clarifying how an exercise's API works. Thanks everyone for chipping in on this, I'll ensure the README is clear on this. :+1:

PR #676, opened 2017-05-20
config.json: add topics to sum-of-multiples

tleen:
See #345


tleen:
I set it to 5? I figure unless math is your thing this is a difficult problem.

PR #677, opened 2017-05-21
clock: restore type-safety for different types of cases

petertseng:
~~This PR should not be merged yet... but it should still be reviewed, to see how close to being mergeable it is.~~

This PR is probably ready to merge, but I'll clean up some commits and move the history to a different branch.

The question to ask is: "Would I rather write a generator using this interface than the existing one?"

---

Seeing how this looks. I'm not convinced it's better, but maybe it can be used as a starting point to make things better. That is why I say do not merge, but if we get a good idea out of it it could be.

Clock updated to 1.0.1 and I have code that works for it too, but I want to compare clock 1.0.0 code vs clock 1.0.0 code, so I reverted the 1.0.1 changes (they can be applied later)

### DIFF
--- a/exercises/clock/.meta/gen.go
+++ b/exercises/clock/.meta/gen.go
@@ -70,8 +70,10 @@ var tmpl = `package clock
 	h, m int
 	want string
 }{ {{range .J.Groups}} {{range .Cases}}
-{{if .IsTimeCase}}{ {{.Hour}}, {{.Minute}}, {{.Expected | printf "%#v"}}}, // {{.Description}}
-{{- end}}{{end}}{{end}} }
+{{- if .IsTimeCase}}
+	{ {{.Hour}}, {{.Minute}}, {{.Expected | printf "%#v"}}}, // {{.Description}}
+{{- end}}{{end}}{{end}}
+}
 
 {{with .J.Groups}}
     // {{ .GroupComment "add"}}
@@ -79,8 +81,10 @@ var tmpl = `package clock
 	h, m, a int
 	want string
 }{ {{range .J.Groups}} {{range .Cases}}
-{{if .IsAddCase}}{ {{.Hour}}, {{.Minute}}, {{.Add}}, {{.Expected | printf "%#v"}}}, // {{.Description}}
-{{- end}}{{end}}{{end}} }
+{{- if .IsAddCase}}
+	{ {{.Hour}}, {{.Minute}}, {{.Add}}, {{.Expected | printf "%#v"}}}, // {{.Description}}
+{{- end}}{{end}}{{end}}
+}
 
 {{with .J.Groups}}
     // {{ .GroupComment "equal"}}
@@ -90,7 +94,8 @@ var eqTests = []struct {
 	c1, c2 hm
 	want   bool
 }{ {{range .J.Groups}} {{range .Cases}}
-{{if .IsEqualCase}} // {{.Description}}
+{{- if .IsEqualCase}}
+// {{.Description}}
 {
 	hm{ {{.Clock1.Hour}}, {{.Clock1.Minute}}},
 	hm{ {{.Clock2.Hour}}, {{.Clock2.Minute}}},
--- a/exercises/clock/.meta/gen.go
+++ b/exercises/clock/.meta/gen.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"log"
 	"text/template"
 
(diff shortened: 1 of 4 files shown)


exercises/clock/.meta/gen.go
```diff
@@ -15,88 +16,119 @@ func main() {
...
 	var j js
-	if err := gen.Gen("clock", &j, t); err != nil {
+
+	must := func(err error) {
+		if err != nil {
+			log.Fatal(err.Error())
+		}
+	}
+
+	if err := gen.Gen("clock", &j, t, func() {
```
petertseng: could have used `must` here.

exercises/clock/cases_test.go
```diff
@@ -28,15 +28,13 @@ var timeTests = []struct {
 	{1, -4820, "16:40"},    // negative minutes roll over continuously
 	{-25, -160, "20:20"},   // negative hour and minutes both roll over
 	{-121, -5810, "22:10"}, // negative hour and minutes both roll over continuously
-
 }
 
 // Test adding and subtracting minutes.
-var addTests = []struct {
+var testTests = []struct {
```
petertseng: no

gen/gen.go
```diff
@@ -90,7 +90,7 @@ func outputSource(status string, fileName string, src []byte) error {
 }
 
 // Gen generates the exercise cases_test.go file from the relevant canonical-data.json
-func Gen(exercise string, j interface{}, t *template.Template) error {
+func Gen(exercise string, j interface{}, t *template.Template, tmp func()) error {
```
petertseng: ah right, all other calls to `Gen` are going to need to change here, unless the `tmp` becomes unnecessary. And `tmp` is going to need a better name (`postProcess` or something like that)

petertseng:
sad to see that this increased the number of lines, a bit unfortunate.

petertseng:
Trying out reflection: Does that make it better?

petertseng:
OK, now the clock-specific generator is about the same length (slightly shorter), which means this could be acceptable...

gen/property.go
```diff
@@ -0,0 +1,53 @@
...
+	// TODO: return err if there is no raw?
+
+	for _, raw := range raws {
+		var prop struct {
+			Property string
+		}
+		if err := json.Unmarshal(raw, &prop); err != nil {
+			return err
+		}
+		tgt := properties[prop.Property]
```
petertseng: TODO (only bother if the idea of `Classify` is deemed acceptable): these names (tgt, z, zi, zv) all are terrible, make better ones

petertseng:
lol more reflection

gen/property.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"reflect"
+)
+
+const tagName = "property"
+
+func Classify(group interface{}) error {
```
petertseng: `Classify` doesn't need to be exported anymore.

gen/property.go
```diff
@@ -0,0 +1,64 @@
...
+			switch field.Kind() {
+			case reflect.Slice:
+				elem := field.Type().Elem()
+				if elem.Kind() == reflect.Struct {
+					for j := 0; j < field.Len(); j++ {
+						classify(field.Index(j).Addr().Interface())
+					}
+				}
+			case reflect.Struct:
+				classify(field.Addr().Interface())
```
petertseng: this line of code not tested in the clock generator... but thought it might necessary for other generators. it's if we have cases nested in other cases

gen/gen.go
```diff
@@ -130,6 +130,8 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
 		return fmt.Errorf(`didn't contain version: %v`, err)
 	}
 
+	classify(j)
```
petertseng: don't ignore error here.

gen/property.go
```diff
@@ -0,0 +1,64 @@
...
+			raws = field.Interface().([]json.RawMessage)
+		} else if tag != "" {
+			properties[tag] = field
+		} else {
+			switch field.Kind() {
+			case reflect.Slice:
+				elem := field.Type().Elem()
+				if elem.Kind() == reflect.Struct {
+					for j := 0; j < field.Len(); j++ {
+						classify(field.Index(j).Addr().Interface())
```
petertseng: don't ignore error.

gen/property.go
petertseng: don't ignore error.

tleen:
Are you looking at this as a way to do the other generations more generically? The existing generation structure is fairly easy to follow, this new one gets thick quick. 

I've been thinking on the generators, this morning's comment on #605 has me wondering if we could restructure the local gen to be the struct for the json and the output template only. The tops of the files seem to be boilerplate.

Just thinking aloud here...

petertseng:
The primary goal: To see if this makes generating tests from JSON files with multiple values of `property` easier and/or makes the resulting generators more understandable.

I like that the three possible structs resulting from values of `property` are explicit; I prefer it much more over unioning all of them.

I don't like the added `{{ if .AddCases }}`. Kind of makes me wonder whether the generator should flatten everything into one level, but how will the group-level `description` be preserved?

Secondary goal: Determine whether added complexity caused by the reflection is understandable.

leenipper:
To help get a feel for this generator approach, I made the changes to the custom-set generator
based on the code of this PR, and it is here:  https://github.com/leenipper/xgo/tree/clock-maybe-plus-custom-set

The custom-set generator is very likely the most complex one at present, and that is why I chose it for comparison.

I think the modified generator is much simpler, and it would have been easier to develop with this new approach. The length of the generator is a bit longer, but much of that is because the current one uses the define template mechanism which helps it be shorter.  In any case, I like the simplicity and it would have saved me time and headache in the update pass done with #622, although I learned more about templates during that update.

So I think the approach has value, for the multi-property canonical-data.json instances.

Following an existing example which uses the approach makes it fairly palatable. Doing it without an example would require studying the gen/gen.go code, and of course there is a different flavor of complexity there with reflection being employed. 

> easier and/or makes the resulting generators more understandable.

Yes, easier and I think more understandable.

> whether added complexity caused by the reflection is understandable.

It is understandable with study; but I think understanding it thoroughly may not be necessary if one follows a previous example when developing a new generator.

petertseng:
Interesting. It might be possible to reduce some duplication by defining a type for the binaryBool and binaryOp cases and using it (like how the tests do), though I haven't tried that.

I did mention I dislike the `{{ if .AddCases }}` however removing them might require more drastic changes than I like to make right now, and I'm not sure it will actually make the generators better.

This `"property"` tag only needs to be used in tests with more than one property, of course; we wouldn't use it for `leap` or the like.

It will be advantageous to ensure that existing generators without the `"property"` tag still work; I haven't done that yet though I expect they will work. Beyond that, if there isn't anything else we can think of to change perhaps I'll clean up the commits.

petertseng:
ran `for i in */.meta; do cd $i; go run gen.go; cd ../../; done` , success, which means we are backwards compatible (do not break generators that do NOT use `property`) and thus can only use `property` when we need to.

I need to clean up some commit history before this is suitable.

petertseng:
I tried #696 (`map[string]interface{}`) but did not find it convincing enough

leenipper:
> we are backwards compatible (do not break generators that do NOT use property) and thus can use property only when we need to.

Very good. I had hoped that was the case.

petertseng:
OK, commits cleaned up. I did keep the three steps (explicitly call a function, explicitly call `Classify`, automatically call `classify`) since they could be useful.

petertseng:
and made the "remove extraneous spaces" its own commit too

petertseng:
Ah I get it. Nice idea. I'll do that. The history can be kept in its own branch; I doubt it'll ever be needed. I'll do it end of this week I hope. It is a busy week. I'm not in a terrible hurry to merge it since a generator needing it has not come up yet since the idea came up.

petertseng:
> perhaps consider isolating the gen/gen.go & gen/property.go changes into their own squashed commit, and then isolating the clock/.meta changes into its own commit ?

OK did it.

PR #686, opened 2017-05-25
poker: create test case generator

leenipper:
Generate validTestCases array from the canonical-data.json.
Continue to use the invalidTestCases array in the test program,
since the .json had no invalid cases.

Update example solution to pass the test case with
an Ace .. 5 straight being lower value than a 2 .. 6 straight.

Update test program to use the generated validTestCases array.

Increment testVersion and targetTestVersion in
example solution and test program.

For #605.


leenipper:
> in a separate commit if desired

Since the generator picked up the new test cases, it seemed to be reasonable to put the change for example in the same commit. I agree that it could go separately for clarity.  I had put changes due to new test cases in same commit previously for adding a generator - see #664, so following that same line here.

I think I will let it stay for this one, and in the future use a separate commit for example update due only to test case change.

petertseng:
👍

PR #689, opened 2017-06-02
config.json: add info for pythagorean-triplet

tleen:
See #345


tleen:
Difficulty too high? I always find these algo/pure math conceptual problems to be difficult.

leenipper:
I agree with the topics chosen.
I'm thinking perhaps a difficulty of 5 at most. I'm basing that primarily on the length of the solution. Some exercise solutions need 100 to 300 lines of code, but this one's example solution weighs in at 34 lines. As a comparison, transpose exercise has difficulty of 5 and has 27 lines in the example solution.

I'm surprised how few exercises have an assigned difficulty other than 1.  I have not given much  attention to this detail.  Perhaps we need one new issue that lists them all with a candidate difficulty on each so we can make a comparative evaluation and assignment.

tleen:
I think it may be good to have a separate issue for difficulty. Its very subjective and a combination of two things: inherent difficulty of the problem and ability to implement it in the language.

I agree with your reasoning and will adjust!

robphoenix:
I think the biggest problem with difficulty is that there's no benchmark to compare against, nor any criteria to consider. Personally, for the exercises I have updated in `config.json` I have taken into consideration the following: my own experience of completing the exercise, an experience that is different to how I view the exercise once I have found a solution I am happy with, this is important as it can often happen that once you can solve a problem that problem loses some of that sense of how difficult it was to navigate it, a difference that is important for new programmers; I also spend some time getting a sense of other users solutions, whether they are similar, as in is there perhaps one or two obvious solutions, or if there is a myriad of possible solutions, and whether there are any comments indicating confusion or frustration.

My belief is that currently it's ok for the difficulty settings to be fairly rough. We don't use them for anything, and while we have no formal criteria for evaluation there's no way to get away from their subjectivity. If we do choose to use them to re-order exercises we will have to still consider what kind of path we want to set, easy to hard, start off easy getting harder but interspersed with easier exercises to sustain morale, consider topic ordering alongside difficulty, etc. And finally I believe it's only going to be once every exercise has a difficulty rating and topics that we are going to be able to adjust the difficulty more accurately. What may seem a `3` on it's own, compared with another `3` or a `4` may not make sense. 
We've got ~55 exercises with no considered difficulty settings, all exercises with topics have had their difficulty considered and so if they are a `1` then that is not accidental. I think tackling topics and difficulty at the same time, per exercise, as has currently been done is fine, and once the config.json is complete we can start the process of re-evaluating exercise ordering and exercise difficulty comparitively rather than isolation.

tleen:
I've been doing the track (slowly!) and updating exercise config data after I submit my own. We can always adjust difficulty as we go along, nothing is set in stone here. Even if we aren't using difficulty now, I would love to get it all done so if exercism ever uses it as a whole the Go track is ready to...go.

PR #692, opened 2017-06-03
config.json: add info for isogram

tleen:
See #345

### DIFF
--- a/config.json
+++ b/config.json
@@ -235,9 +235,12 @@
       "topics": []
     },
     {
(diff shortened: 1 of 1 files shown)


tleen:
I'm continuing on with our current topics structure, unsure if or when [the issue of capitalization](https://github.com/exercism/x-common/issues/820) will be resolved.

leenipper:
I agree with topics and difficulty chosen.

PR #694, opened 2017-06-04
anagram: Fix test case for case sensitivity

ferhatelmas:
Response should return actual input.

### DIFF
--- a/exercises/anagram/anagram_test.go
+++ b/exercises/anagram/anagram_test.go
@@ -6,7 +6,7 @@ import (
 	"testing"
 )
 
-const targetTestVersion = 1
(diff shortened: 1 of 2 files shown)


kytrinyx:
`tc` was probably test case. I would probably go with `tests` and `test` now rather than `testCases` and `tc`

ferhatelmas:
I had meaned `test candidate` with `tc` since `c` was used for `candidate`. Do you want me to change it to `lc` or `test` or anything else ?

petertseng:
I'm good with this as-is. I suppose it isn't a test case since it's in an example file

ferhatelmas:
I agree.

PR #695, opened 2017-06-04
anagram: create test case generator

ferhatelmas:
For #605.


ferhatelmas:
Tests will fail see updated example in #694.

exercises/anagram/.meta/gen.go
```diff
@@ -0,0 +1,50 @@
...
+  subject string
+  candidates []string
+  expected []string
+  description string
+}
+
+var testCases = []testcase{
+  {{range .J.Cases}}
+	{
+	  subject: {{printf "%q" .Subject}},
```
petertseng: I would take a look at the whitespace at these four lines, looks like there are some tabs and spaces mixed. It does not matter in the end since the end result gets formatted anyway, but helps to use just one (I don't care which) in case tab widths differ.

leenipper: Using tabs will be consistent with rest of file.

exercises/anagram/.meta/gen.go
```diff
@@ -0,0 +1,50 @@
+package main
+
+import (
+	"text/template"
+
+	"github.com/exercism/xgo/gen"
```
leenipper: The import path used in other generators for referencing the gen package is:

>       "../../../gen"

Will remove the need to have github.com/ ... in the xgo tree/workspace.

tleen:
I think #874 may be the reason this failed on Travis.

Going to close this as #869 duplicates it.

PR #696, opened 2017-06-06
[Do Not Merge] clock: anything goes, it's now a `map[string]interface{}`

petertseng:
Well, this is what it would look like if we didn't bother declaring what structure the JSON will take, and just use `map[string]interface{}`.

It is possible we can say there is less boilerplate here.

It's a little unfortunate we have to type `index . ` all the time.

We should only consider this solution if we can extract some common code using the `map[string]interface{}`

### DIFF
--- a/exercises/clock/.meta/gen.go
+++ b/exercises/clock/.meta/gen.go
@@ -28,27 +28,15 @@ type TestGroups []struct {
 	Cases       []OneCase
 }
 
-type OneCase struct {
-	Description string
-	Property    string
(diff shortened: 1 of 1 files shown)


petertseng:
Hold up, we don't have to use `index .`. Reopening so I can demonstrate.

petertseng:
There we go.

But this causes complications when we want to postprocess.

petertseng:
Well, feel free to argue for this if you really want to have it, I'm just closing it so that we don't accidentally get it merged if nobody wants it.

PR #699, opened 2017-06-10
Delete ignored key from config.json

kytrinyx:
Since the exercise implementations are all in the exercises directory
we no longer need to ignore any non-exercise directories in the root
of the track.

See https://github.com/exercism/meta/issues/3 for context.

### DIFF
--- a/config.json
+++ b/config.json
@@ -13,11 +13,6 @@
     "point-mutations",
     "trinary"
   ],
-  "ignored": [
-    "img",
-    "gen",
-    "docs"
-  ],
   "foregone": [
     "space-age",
     "linked-list",
--- a/config.json
+++ b/config.json
@@ -14,7 +14,6 @@
     "trinary"
   ],
   "ignored": [
-    "error-handling",
     "img",
     "gen",
     "docs"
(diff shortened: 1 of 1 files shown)


kytrinyx:
I've created another PR, #700, which normalizes all of the whitespace in the config.json so
that the change in this PR is minimized and easier to see.

If you merge the other PR first, I'll go ahead and rebase this one onto master.

Sorry about the noise!

petertseng:
An interesting question comes up. The build is broken but that is because `master` would be broken right now too. Since the `ignored` key was removed and it was being used to ignore error-handling (https://github.com/exercism/xgo/commit/55de1e8e09f76a1b6b3d41b70b9d81e43ce249bd) now error-handling is not in any category.

The reasonable options are to add it to any other section.

* `exercises`
* `deprecated`
* `foregone`

I let y'all choose. There is no more reason to ignore it since the reason for doing so is gone. So I suggest placing it in `exercises` again to let students get at it. The binary is gone at https://github.com/exercism/xgo/commit/46c376fa9f2b9d54f539ebfd73517ec7e0be6af4

kytrinyx:
You make a good case for putting it back in `exercises`. I think we should go with your suggestion.

petertseng:
Also, whitespace changes removed (squashed in the reverse of #700) so that this PR is only about removing ignored

kytrinyx:
Thanks @petertseng!

PR #700, opened 2017-06-10
Normalize format of config file

kytrinyx:
This will let us script changes to it with less noise in the diffs.

### DIFF
--- a/config.json
+++ b/config.json
@@ -29,10 +29,10 @@
       "difficulty": 1,
       "slug": "hello-world",
       "topics": [
-          "Control-flow (conditionals)",
-          "Optional values",
-          "Strings",
-          "Text formatting"
+        "Control-flow (conditionals)",
+        "Optional values",
+        "Strings",
+        "Text formatting"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


petertseng:
We fixed error-handling in #699. Thus if this were rebased on master, the branch CI will pass. The PR CI already passes since PR CI checks the result of merging the PR into master which (since I hit rebuilt at a point in time after #699 was merged) result in `error-handling` being in the right place. Branch CI checks the branch without modification.

In addition, the normalisation is currently not applied to `error-handling`.

Katrina knows the below, but this is for the benefit of others who make the decision for this repo:

It does not matter to me whether this is merged. If you ask me, I probably take the indentation changes (tabs->spaces in sum-of-multiples, fewer spaces in hello-world) but not the `[]` -> `[\n\n]` sincewe will naturally deal with that as we add topics and it avoids an unnecessary newline being added then removed. Any whitespace changes from auto-generated commits can be dealt with easily (squash in the reverse of the diff of this PR, basically).

kytrinyx:
I'm going to go ahead and close it. I'll think harder about it if I need to script more changes.

PR #704, opened 2017-07-03
config.json: add info for error-handling

tleen:
See #345

### DIFF
--- a/config.json
+++ b/config.json
@@ -135,9 +135,11 @@
       ]
     },
     {
-      "difficulty": 1,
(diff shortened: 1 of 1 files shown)


tleen:
I thought this one rated a 5, you would need to know about named return values which is not common early in the language use. Doing a panic recover with a named return value is sort of tricky?

kytrinyx:
> you would need to know about named return values which is not common early in the language use

Agreed, this is pretty tricky stuff.

PR #706, opened 2017-07-04
clock 1.0.1

### LINKED ISSUE #705: clock 1.0.1 generated cases_test.go fails build
Committed cases_test.go is based on 1.0.0 so the current code is ok.

Using latest x-common clock/canonical-data.json 1.0.1, the generated code fails to build. It breaks b/c two addTests generate due to two different groups having the "add" property.

> ./cases_test.go:61: addTests redeclared in this block

The subtract cases are in a different group but still have "add" property.

From #677 ...
> Clock updated to 1.0.1 and I have code that works for it too, but I want to compare clock 1.0.0 code vs clock 1.0.0 code, so I reverted the 1.0.1 changes (they can be applied later)

@petertseng Perhaps those 1.0.1 changes combined the matching properties differently? I didn't want to dig further if the reverted code handled it already and might be helpful.

But taking a different slant, it might be cleaner if a "subtract" property were used in clock/canonical-data.json, since combining multiple groups into one seems a bit awkward.  And is this case something which classifyByProperty should detect ?

petertseng:
Closes #705

### DIFF
--- a/exercises/clock/.meta/gen.go
+++ b/exercises/clock/.meta/gen.go
@@ -3,6 +3,7 @@ package main
 import (
 	"encoding/json"
 	"log"
+	"strings"
 	"text/template"
 
 	"../../../gen"
@@ -44,6 +45,10 @@ type testGroup struct {
 	} `property:"equal"`
 }
 
+func (groups testGroup) GroupShortName() string {
+	return strings.ToLower(strings.Fields(groups.Description)[0])
+}
+
 var tmpl = `package clock
 
 {{.Header}}
@@ -63,7 +68,7 @@ var tmpl = `package clock
 	{{- end }}
 
 	{{- if .AddCases }}
-		var addTests = []struct {
+		var {{ .GroupShortName }}Tests = []struct {
 			h, m, a int
 			want string
 		}{
--- a/exercises/clock/clock_test.go
+++ b/exercises/clock/clock_test.go
@@ -51,6 +51,12 @@ func TestAddMinutes(t *testing.T) {
 				a.h, a.m, a.a, got, a.want)
 		}
 	}
+	for _, a := range subtractTests {
+		if got := New(a.h, a.m).Add(a.a); got.String() != a.want {
+			t.Fatalf("New(%d, %d).Add(%d) = %q, want %q",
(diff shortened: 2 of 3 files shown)


exercises/clock/clock_test.go
```diff
@@ -51,6 +51,13 @@ func TestAddMinutes(t *testing.T) {
 				a.h, a.m, a.a, got, a.want)
 		}
 	}
+	// TODO: Don't duplicate.
```
leenipper: Not clear on the "duplicate" meaning.  The subtractTests are part of testing Add, so all seems ok here.

petertseng: OK sure, if you are OK with it, I will simply remove the comment and we will be done with it. It will be fine to have this for now. I was dissatisfied with the fact that lines 56 - 59 are the same as lines 49-52.

exercises/clock/.meta/gen.go
```diff
@@ -44,6 +45,10 @@ type testGroup struct {
 	} `property:"equal"`
 }
 
+func (groups testGroup) GroupShortName() string {
+	return strings.ToLower(strings.Fields(groups.Description)[0])
```
petertseng: be careful here with a variable name dependent on description - for example if someone changes the description to be something like "the following tests will do something" then the variable name becomes `theTests` which doesn't look good. Perhaps we will defer on dealing with that when it happens.

petertseng:
I will make the commit comment actually descriptive.

petertseng:
I wonder whether to split this up into two commits (I have now done so). The first commit is only the generator change. However it is not 100% compatible with clock 1.0.0; it would change the test name to `testTests`. So maybe they should just be one commit. If you have an opinion, you should say so, otherwise I'll squash on merge because the intermediate state is not useful. The history can remain in this PR in case one needs to be reverted separately.

exercises/clock/.meta/gen.go
petertseng: I have acknowledge this weakness in the commit message and noted that I find it better than hard-coding the names `addTests` and `subtractTests`. We will deal with it if it becomes a problem later on.

PR #711, opened 2017-07-10
bank-account: add benchmarks

tleen:
Given that locking seems to be the solution for the problem, I think
benchmarking could be a real interesting add-on to this specific
exercise.

### DIFF
--- a/exercises/bank-account/bank_account_test.go
+++ b/exercises/bank-account/bank_account_test.go
@@ -295,3 +295,23 @@ func TestConcDeposit(t *testing.T) {
 			"a.Balance = %d, %t.  Want 0, true", p, ok)
 	}
 }
+
+func BenchmarkAccountOperations(b *testing.B) {
+	a := Open(0)
+	defer a.Close()
+	for n := 0; n < b.N; n++ {
+		a.Deposit(10)
+		a.Deposit(-10)
+	}
+}
+
+func BenchmarkAccountOperationsConcurrent(b *testing.B) {
+	a := Open(0)
+	defer a.Close()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			a.Deposit(10)
+			a.Deposit(-10)
+		}
+	})
+}
--- a/exercises/bank-account/bank_account_test.go
+++ b/exercises/bank-account/bank_account_test.go
@@ -296,7 +296,7 @@ func TestConcDeposit(t *testing.T) {
 	}
 }
 
-// The benchmark operations are here to allow you to try different
+// The benchmark operations are here to encourage you to try different
 // implementations to see which ones perform better. These are worth
 // exploring after the tests pass.
 //
--- a/exercises/bank-account/bank_account_test.go
+++ b/exercises/bank-account/bank_account_test.go
@@ -296,6 +296,31 @@ func TestConcDeposit(t *testing.T) {
 	}
 }
 
+// The benchmark operations are here to allow you to try different
+// implementations to see which ones perform better. These are worth
+// exploring after the tests pass.
+//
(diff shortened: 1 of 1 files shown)


tleen:
This is as much for discussion of the benchmarks as anything. They may be too naive to be actually helpful in benchmarking.

robphoenix:
I've not seen/used this before, but it certainly looks like you've implemented it correctly. 

**TLDR:** I'd question the benefit of throwing these benchmarks out there without relevant comments/documentation.

---

My only concern is that it may cause more confusion. Running the benchmarks gives me the following result:

```
> go test -bench=.       
BenchmarkAccountOperations-4             	10000000	       171 ns/op
BenchmarkAccountOperationsConcurrent-4   	 5000000	       284 ns/op
PASS
ok  	github.com/robphoenix/exercism-go/exercises/bank-account	3.640s
```

Thinking back to when I first spent time with benchmarks and concurrency in an exercism exercise, (I can't think which exercise now, I'll link to it if I find a suitable reference), here I would be confused as to why the concurrent operations are slower, `284 ns/op` vs. `171 ns/op`, because I would assume they would be faster. I realise this is down to my lack of understanding about concurrency.  I don't think adding benchmarks necessarily adds anything to the understanding of concurrency here, unless we can add some documentation that can explain these results or link to some helpful text, and these results may lead someone to think they're `doing it wrong`. Or maybe I'm just mistaken in my understanding of concurrency benchmarks, if so it's probably still useful to include an explanation to avoid others being similarly mistaken.

tleen:
I would think they are slower because of the overhead of doing the concurrency compared to the not-complex nature of the actual operation that is being performed concurrently. If it was some crazy intensive task I think concurrency would be a performance improvement but here the operation is simple. The locks we are doing here act as a protective measure against the concurrency. The idea being for performance-based concurrency you want to make concurrent an operation that is so intensive that it justifies whatever overhead slowdown occurs with managing the concurrent resources.

Not really 100% sure either, don't have a lot of experience with concurrency in Go. :smile:

robphoenix:
> I would think they are slower because of the overhead of doing the concurrency compared to the not-complex nature of the actual operation that is being performed concurrently.

I think if we include a comment along these lines, then this would be good to merge.

> don't have a lot of experience with concurrency in Go. 

That makes two of us :smile:

tleen:
I'm not sure if these should be included, perhaps someone with more exp in this area can take a look? @petertseng?

The goal with this is to let people play with how they lock the vars to see if there is a performance impact. I would have to add more comments. Prob in the README.

robphoenix:
I don't see a problem with including it as long as it's accompanied by some comments, much as you've explained in this thread @tleen

exercises/bank-account/bank_account_test.go
```diff
@@ -296,6 +296,31 @@ func TestConcDeposit(t *testing.T) {
 	}
 }
 
+// The benchmark operations are here to allow you to try different
```
robphoenix: maybe change allow to help
allow doesn't feel quite right, this is more assistance than permission (nitpicky :smile: )

robphoenix:
Just wondering if this needs to be 2 commits or if squashing them into a single commit would be better. I'm not so concerned but it might make for a cleaner commit history? Totally up to you. :+1:

robphoenix:
**encourage** - even better!

exercises/bank-account/bank_account_test.go
tleen: Taking another look I think "encourage" is the right word. 

(I appreciate the feedback, more feedback == better commits!)

tleen:
:smile: I think one squashed commit: a benchmarks all or nothing.

PR #713, opened 2017-07-11
Fixed perfect_number false positive and equaling types. Closes #712

### LINKED ISSUE #712: The tests for perfect numbers are too vague
This code here right now makes all of the tests pass for perfect numbers.

```go
package perfect

const testVersion = 1

var ErrOnlyPositive error

type Classification struct {
}

var ClassificationDeficient Classification
var ClassificationAbundant Classification
var ClassificationPerfect Classification

func Classify(n uint64) (Classification, error) {
	return ClassificationPerfect, ErrOnlyPositive
}
```

Output:

```bash
❯ go test
PASS
ok  	_/Users/nyehehe/exercism/go/perfect-numbers	0.007s
```

Notice how there is no logic in there what so ever. ;)

The types are a nice addition, but they make the test too fragile. It needs to be made explicit so code like this doesn't make it pass.

Skarlso:
As issue #712 lays out, this test makes sure that the constants created from the Classification interface, will not equal each other providing a false positive for `go test`.

### DIFF
--- a/exercises/perfect-numbers/perfect_numbers_test.go
+++ b/exercises/perfect-numbers/perfect_numbers_test.go
@@ -40,3 +40,25 @@ func TestClassifiesCorrectly(t *testing.T) {
 		}
 	}
 }
+
+// Test that the classifications are not equal to each other.
+// If they are equal, then the tests will return false positives.
+func TestClassificationsNotEqual(t *testing.T) {
+	classifications := []struct {
+		class Classification
+		name  string
+	}{
+		{ClassificationAbundant, "ClassificationAbundant"},
+		{ClassificationDeficient, "ClassificationDeficient"},
+		{ClassificationPerfect, "ClassificationPerfect"},
+	}
+
+	for i, pair1 := range classifications {
+		for j := i + 1; j < len(classifications); j++ {
+			pair2 := classifications[j]
+			if pair1.class == pair2.class {
+				t.Fatalf("%s should not be equal to %s", pair1.name, pair2.name)
+			}
+		}
+	}
+}


exercises/perfect-numbers/perfect_numbers_test.go
```diff
@@ -40,3 +40,25 @@ func TestClassifiesCorrectly(t *testing.T) {
...
+		{ClassificationAbundant, 0},
+		{ClassificationDeficient, 1},
+		{ClassificationPerfect, 2},
+	}
+
+	for i, pair1 := range classifications {
+		for j := i + 1; j < len(classifications); j++ {
+			pair2 := classifications[j]
+			if pair1.class == pair2.class {
+				t.Fatalf("%d should not be equal to %d", pair1.val, pair2.val)
```
petertseng: given the sample solution shown in #712 

```go
package perfect

const testVersion = 1

var ErrOnlyPositive error

type Classification struct {
}

var ClassificationDeficient Classification
var ClassificationAbundant Classification
var ClassificationPerfect Classification

func Classify(n uint64) (Classification, error) {
	return ClassificationPerfect, ErrOnlyPositive
}
```

This would cause the test to output `0 should not be equal to 1`. That would be a confusing error message because zero is already not equal to one and implies something is wrong with math instead of the submitted code

Skarlso:
@petertseng right you are.

Fixed.

```bash
❯ go test
--- FAIL: TestClassificationsNotEqual (0.00s)
	perfect_numbers_test.go:60: ClassificationAbundant should not be equal to ClassificationDeficient
FAIL
exit status 1
FAIL	_/Users/user/exercism/go/perfect-numbers	0.008s
```

Skarlso:
@petertseng fixed it, nice spot!

tleen:
Thanks @Skarlso!

Skarlso:
Np. :-)

PR #715, opened 2017-07-14
config.json: add info for luhn

tleen:
See #345

### DIFF
--- a/config.json
+++ b/config.json
@@ -4,15 +4,6 @@
   "repository": "https://github.com/exercism/go",
   "active": true,
   "ignore_pattern": "example(?!.*test)",
-  "deprecated": [
-    "binary",
-    "bottles",
-    "counter",
-    "hexadecimal",
-    "octal",
-    "point-mutations",
-    "trinary"
-  ],
   "foregone": [
     "space-age",
     "linked-list",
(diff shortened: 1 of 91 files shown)


tleen:
Difficulty of 1 is intentional.

config.json
```diff
@@ -259,7 +259,11 @@
     {
       "difficulty": 1,
       "slug": "luhn",
-      "topics": []
+      "topics": [
+	"Algorithms",
+	"Integers",
+	"Strings"
```
robphoenix: These could do with being indented consistently with the rest of the config.json file.

tleen: [My solution](http://exercism.io/submissions/3638377f9a8c4e869b0c2a0c724ff02f). It could be a 2, if we look at 1 as the most basic "hello world" stuff only. I'm not sure what else would qualify as a 1 in the track though.

tleen: Yeah I have tabs should be spaces. Mind if I add an [.editorconfig](http://editorconfig.org/) to .gitignore, or perhaps we should have one in the project root for everyone?

robphoenix: I'd push it up to a 2 at least, if not a 3. While the solution can be fairly step-by-step, there's still a fair amount of steps to consider to build up the necessary algorithm. This is a lot more difficult than hello-world, and not something I think would be given to a new programmer alongside hello-world, I'd be concerned if this exercise appeared earlier in the track, being considered almost as easy as hello-world, as it has a lot to wrap your head around. Also, I noticed you requested help for that exercise? Was that because you were stuck? If so, I think that would indicate the difficulty should be higher than a 1 perhaps.

tleen: I can agree with that. Do you think anything in the track besides hello world would be a 1? Or should we look at anything more complex than that as at least a 2.

(I requested help just to see what would happen. I hadn't noticed that button before. All it does is put that flag on the solution submission?)

robphoenix: I thought it might be a case of pressing a button to see what happened. We've all been there! :smile:  As far as I know it flags up someone who wants help, so if you're browsing other people's solutions you can see if anyone needs help, not sure if that will change with Nextercism.

The following exercises are also 1:
 - Leap
 - Gigasecond
 - Hamming
 - Raindrops
 - Accumulate

I intentionally left them as such when I added topics, as I thought they were appropriate ratings, to be visited later once the `config.json` was completed. I think right now, it's perhaps more important to just get difficulty ratings in, and we can adjust them once we have that baseline in place, and a better idea of what constitutes a 1. We could then perhaps articulate in words what each rating reflects, and adjust accordingly?

tleen:
This looks wrong.

robphoenix:
It looks like you've included a bunch of other commits, maybe when you merged a branch? Did you rebase on master?

```
git fetch origin master
git rebase origin master
```

If that doesn't help I'd either try an interactive rebase and drop the other commits, or my goto git solution with small changes: delete the branch and start again!

tleen:
Yeah, the config file is so different and my changes so minor I may delete all of the config.json changes and redo off the new config.

PR #714, opened 2017-07-13
config.json: add info for crypto-square

tleen:
See #345

PR #723, opened 2017-07-17
Add .editorconfig file

### LINKED ISSUE #722: Add .editorconfig?
Soliciting opinions on adding an [.editorconfig](http://editorconfig.org/) file to the root of the project. Gofmt makes it unnecessary for the .go files, but it would be helpful for things like config.json to keep us all on the same page regardless of our default settings for js/json.

tleen:
The [.editorconfig file](http://editorconfig.org/)  is compatible with most development environments and is used to setup project-based consistent text editing style. For this repo, gofmt already handles this for the .go files. The .editorconfig file is for consistency in the the [config.json](config.json) file. This would solve the tabs/spaces problem (see #714) automatically for anyone with an editorconfig aware environment.

Closes #722

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,5 @@
+root = true
+
+[config.json]
+indent_style = space
+indent_size = 2
--- a/README.md
+++ b/README.md
@@ -29,6 +29,8 @@ This is because this repo only imports from the standard library and isn't expec
 There is a [misspelling tool](https://github.com/client9/misspell). You can install and occasionally run it to
 find low hanging typo problems. [#570](https://github.com/exercism/go/pull/570) It's not added into CI since it could give false positives.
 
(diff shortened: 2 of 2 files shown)


robphoenix:
This is relevant: https://github.com/exercism/configlet/issues/18

tleen:
Ok let's kill this for now and I will add it  to .gitignore so if people want to use it locally they can.

robphoenix:
> only solves the problem for those who use it, which I don't,

@tleen re-reading this, I realise my comment sounds awful, as if I'm saying I don't use it therefore it's no good! This isn't what I meant, sorry! More that I'd like to see a solution for the repo rather than for a subset of it's users.

tleen:
LOL that's not how I read it I get what you were saying. Agreed it would probably be best checked/reformatted at the configlet level for consistency. I'm just terribly embarrassed I submitted tab indenting to a space indented file! I've become over-reliant on things like editorconfig and don't think about it much anymore. We can put a note in the README for maintainers about the preference and point them in the editorconfig direction. If we somehow get more stuff needing formatting then it may make sense to revisit. 

All in all this keeps showing why gofmt is so awesome!

PR #724, opened 2017-07-17
Ignore .editorconfig file

### LINKED ISSUE #722: Add .editorconfig?

tleen:
This allows contributors to use local .editorconfig files to bring
their editors into compliance with repo style. See tabs/spaces in #714.

Closes #722

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.swp
 .DS_Store
+.editorconfig
 tmp
 bin/configlet
 bin/configlet.exe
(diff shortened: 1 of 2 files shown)


robphoenix:
Cool, is it worth adding a *If you use editorconfig, add these settings...* to the README?

tleen:
I think so, will add to this commit then squash later.

robphoenix:
I didn't realise there was an exercism.io `.editorconfig`, that's cool. I assume `config.json` will be covered by `[*]`, is it worth adding your settings [here](https://github.com/exercism/go/pull/723/files) to the file, so it's more specific?

LGTM :shipit:

tleen:
I think we can just key off the * in the exercism.io as the global setting since it works for the config.json as we have it now.

Issue #731, opened 2017-07-25
Update track metadata as outlined in #717


robphoenix:
From #717:

If you haven't already, now would be a good time to do the following:

- [ ] add a rough estimate of difficulty to each exercise (scale: 1-10)
- [ ] add topics to each exercise
- [ ] choose *at most 20 exercises* to be core exercises (set `core: true`, and delete the `unlocked_by` key)
- [ ] for each exercise that is not core, decide which exercise is the prerequisite (max 1)

If possible, leave 3 or 4 simple exercises as (core: false, unlocked_by: null), as this will provide new participants with some exercises that they can tackle even if they have not finished the first core exercise.

robphoenix:
I've started trying to fill in all the exercise topics and difficulty ratings, roughly, so we can get on to choosing ~10 core exercises.

Though we could probably choose our core exercises now, as well as the 3 or 4 simple exercises that will be outside this hierarchy.

tleen:
[Hello World](../tree/master/exercises/hello-world) should be our first right?

tleen:
What is the overall opinion of core? Is it sort of a mini-track and the unlocks are further explorations? 

Exercises I think differentiate Go from other tracks:
 
* [Scrabble Score](../tree/master/exercises/scrabble-score) - Because it deals with runes which are a unique Go pain point for people transitioning into it
* [Bank Account](../tree/master/exercises/bank-account) - For the structs, variable protection and concurrency
* [Error Handling](../tree/master/exercises/error-handling) - It has proven to be a difficult exercise and introduces some core Go stuff like panics, recovery, errors and named return values. 

These are from the relatively small pool of exercises I have already completed. Many more to go :smile:.

robphoenix:
> Hello World should be our first right?

right :+1:

tleen:
@robphoenix I've been working the config file bottom up so we don't clash.

robphoenix:
Okay all,@tleen @leenipper @ferhatelmas @petertseng, we're getting down to the last few exercises! :tada: 
So we don't duplicate effort I think it's worth *'claiming'* an exercise when updating it in the `config.json`.

Here are the last exercises to be updated:

- [ ] prime-factors (#764)
- [ ] strain (#762)
- [ ] diffie-hellman (#766)
- [ ] beer-song (#772)
- [ ] ocr-numbers (#773)
- [ ] wordy
- [ ] nucleotide-count
- [ ] grade-school
- [ ] matrix
- [ ] saddle-points
[...]

PR #748, opened 2017-07-27
config.json: add topics & difficulty for diamond

robphoenix:
See #345 & #731

### DIFF
--- a/config.json
+++ b/config.json
@@ -528,9 +528,13 @@
       "slug": "diamond",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 4,
       "topics": [
-
+        "Strings",
+        "Text formatting",
+        "Algorithms",
+        "Control-flow (conditionals)",
+        "Control-flow (loops)"
       ]
     },
     {


config.json
```diff
@@ -528,9 +528,10 @@
       "slug": "diamond",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 7,
```
leenipper: 7 seems high for diamond.  I prefer 4.

ferhatelmas: I agree. Since tests seem involving due to how they are written (property-based) but solution to exercise isn't that difficult.

config.json
```diff
@@ -528,9 +528,10 @@
...
       "topics": [
-
+        "Strings",
+        "Text formatting"
       ]
```
ferhatelmas: How about `Control-flow (conditionals)` and `Control-flow (loops)` ?

config.json
robphoenix: Okay, I've not done the exercise, I'm happy to go with your recommendations.

robphoenix:
Updated @leenipper @tleen @ferhatelmas, thanks :+1:

PR #751, opened 2017-07-29
Make myself an alumnus

kytrinyx:
I've not contributed in a meaningful way to the Go track in a long time.
Thank you so much to all of you who have made this possible.


kytrinyx:
Thank you ✨

PR #754, opened 2017-07-30
config.json: modify topics & add difficulty for bowling

tleen:
Added "Algorithms" due to the complexity of the scoring algorithm,
changed topics ordering to alphabetical.

See #345 & #731


tleen:
None of these difficulties are set in stone. I imagine we may re-evaluate them once we have all of them set initially.

PR #755, opened 2017-07-30
config.json: modify topics & add difficulty for change

tleen:
Added "Mathematics" to the topics list since the solution involves
some mathematical thinking + operations. Set topics to be in
alphabetical order.

See #345 & #731

### DIFF
--- a/config.json
+++ b/config.json
@@ -890,11 +890,12 @@
       "slug": "change",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 3,
       "topics": [
-        "Control-flow (loops)",
+        "Algorithms",
         "Arrays",
-        "Algorithms"
+        "Control-flow (loops)",
+        "Mathematics"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -890,11 +890,12 @@
       "slug": "change",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 3,
       "topics": [
```
leenipper: Consider adding "Searching", since the goal is to search for the smallest set of coins based on the input set.

config.json
```diff
@@ -890,11 +890,12 @@
       "slug": "change",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 3,
```
leenipper: I'm thinking at least 4, perhaps 5 for difficulty.

ferhatelmas: I second this.

config.json
ferhatelmas: This too.

PR #762, opened 2017-07-31
config.json: Add difficulty & topics for strain

robphoenix:
See #345 & #731


robphoenix:
I feel like the operation being done here is a filtering one rather than actually transforming anything.

PR #766, opened 2017-07-31
config.json: Add difficulty & topics for diffie-hellman

robphoenix:
See #345 & #731

### DIFF
--- a/config.json
+++ b/config.json
@@ -642,9 +642,12 @@
       "slug": "diffie-hellman",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 5,
       "topics": [
-
+        "Cryptography",
+        "Integers",
+        "Algorithms",
+        "Transforming"
       ]
     },
     {


config.json
```diff
@@ -642,9 +642,12 @@
...
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 5,
       "topics": [
-
+        "Cryptography",
+        "Integers",
+        "Algorithms",
+        "Transforming"
```
leenipper: Since the big.Int is a central part of the solution, I would suggest Interfaces. Not in the Go "interface" sense, but in the "make use an existing interface/package" sense.  What do you think ?

robphoenix: I see where you're coming from, but I think `interfaces` will only suggest the use of Go interfaces. Maybe something like `Standard library`? Or perhaps this is something that should be in the `README` or test suite?

leenipper: Yeah, we should avoid any confusion regarding Go interface.  May we choose a topic which isn't in the TOPICS.txt ?  If so, then Standard library sounds good.

tleen: I think if there are topics we would like to include we should probably propose them to the TOPICS.txt file? I think with nextercism and the way it is managing topics that putting our own in may mess things up? "Standard library" would be a good one, but we would have to be careful not to overuse, since you use some standard library functions a ton. Just where it is unique and not the ubiquitous use of the "fmt" lib.

robphoenix: Honestly I'm torn on this one. For now I'm going to leave it as is. I think the conversation over having a `Standard library` topic is something separate.

tleen: Yeah I think that belongs at [problem-specifications](https://github.com/exercism/problem-specifications).

PR #767, opened 2017-07-31
config.json: Add topics & difficulty for variable-length-quantity

tleen:
See #345 & #731

### DIFF
--- a/config.json
+++ b/config.json
@@ -884,8 +884,8 @@
       "unlocked_by": null,
       "difficulty": 6,
       "topics": [
-        "Integers",
         "Bitwise operations",
+        "Integers",
         "Transforming"
       ]
     },
--- a/config.json
+++ b/config.json
@@ -882,7 +882,7 @@
       "slug": "variable-length-quantity",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 6,
+      "difficulty": 5,
       "topics": [
         "Bitwise operations",
         "Integers",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -882,9 +882,11 @@
       "slug": "variable-length-quantity",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 6,
```
leenipper: Seems high.  Maybe 5.  It's all subjective. I may be too picky here.

config.json
```diff
@@ -882,9 +882,11 @@
...
       "topics": [
-
+        "Integers",
+        "Bitwise operations",
+        "Transforming"
```
leenipper: I would add "Arrays".  The Go term is Slices, but we are picking from the [Topics](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt).

config.json
tleen: It's really hard to tell until we see them all. I have not done a lot of these so I am estimating based on how it looks and for some problems how difficult they were on another track. As I go through the track as a user I will review each of these again from that perspective. The push right now is to get the track ready for nextercism which means we can't take the slower route :frowning_face:.

For this exercise I set the difficulty rather high mostly because for a lot of beginning users any bitwise stuff can be like hitting a conceptual brick wall. In an early exercise [secret-handshake](../tree/master/exercises/secret-handshake) (IIRC) practically announces that it is a bitwise problem yet there are still solutions submitted which try to avoid the concept. 

These are still loose numbers, better than 1 though!

PR #768, opened 2017-07-31
config.json: Add topics & difficulty for poker

tleen:
See #345 & #731

### DIFF
--- a/config.json
+++ b/config.json
@@ -874,10 +874,11 @@
       "unlocked_by": null,
       "difficulty": 5,
       "topics": [
-        "Domain-specific languages",
         "Equality",
         "Games",
         "Parsing",
+        "Pattern matching",
+        "Sequences",
         "Strings"
       ]
     },
--- a/config.json
+++ b/config.json
@@ -872,9 +872,13 @@
       "slug": "poker",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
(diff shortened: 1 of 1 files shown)


tleen:
"Domain-specific languages" may be a bit of a stretch. I figure using the combination of card values as a string with a relative value is a small but very domain-specific language.

config.json
```diff
@@ -872,9 +872,13 @@
       "slug": "poker",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 5,
```
leenipper: Difficulty seems about right.  Might go with 6.

config.json
```diff
@@ -872,9 +872,13 @@
...
       "topics": [
-
+        "Domain-specific languages",
+        "Equality",
+        "Games",
+        "Parsing",
+        "Strings"
```
leenipper: What do you think about "Pattern matching", and "Sequences" as candidate topics ?
A poker "straight" is like a fixed size sequence. A full-house is like a pattern 3 of something and 2 of something else.

I agree that Domain-specific languages is stretching it a bit.  There are objects/nouns (the cards), but no actions/verbs for "the language".

tleen: Yep I think those two are a better descriptive replacement for DSL.

PR #771, opened 2017-08-01
Upping test version

paul-nelson-baker:
I've looked at the history of the last change to this exercise. It
doesn't look like much was changed to expect a different implementation.
I've updated the version number to fix the exercise


robphoenix:
This must have been missed in https://github.com/exercism/go/commit/7bc43d0e2bb592b2419cac17da5bc7966308928f
Thanks for catching it, and fixing it! Much appreciated @paul-nelson-baker :tada:

PR #775, opened 2017-08-01
config.json: Add topics & difficulty for nucleotide-count

robphoenix:
See #345 & #731


robphoenix:
Agreed on `sorting`, removed. Not sure about `filtering`, I think weeding out invalid values is more an input validation task than a filtering task. I'll add `maps` also.

PR #777, opened 2017-08-02
config.json: Add topics & difficulty for connect

tleen:
See #345 & #731

### DIFF
--- a/config.json
+++ b/config.json
@@ -866,9 +866,11 @@
       "slug": "connect",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 9,
       "topics": [
-
+        "Control-flow (loops)",
+        "Games",
+        "Graphs"
       ]
     },
     {
--- a/config.json
+++ b/config.json
@@ -868,9 +868,11 @@
       "unlocked_by": null,
       "difficulty": 9,
       "topics": [
+        "Arrays",
         "Control-flow (loops)",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -866,9 +866,11 @@
       "slug": "connect",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 9,
       "topics": [
-
+        "Control-flow (loops)",
+        "Games",
+        "Graphs"
```
leenipper: "Arrays" applies I think.
Also "Searching", since the program searches for a winner if one exists.

I didn't think that Graphs applied, but maybe it is a valid perspective.

robphoenix: I agree with `Arrays`. I can reasoning for both `searching` and `graphs`.

tleen: I was thinking the board really ends up being a straight-up [graph](https://en.wikipedia.org/wiki/Graph_(abstract_data_type)) you need to traverse.

leenipper:
> a straight-up graph you need to traverse.

That makes sense.  The solutions I've seen didn't construct a graph data structure, but they could.

Looks fine.

tleen:
The rule of thumb I have been using for graphs/matrices: if its a fixed size multidimensional structure is a matrix, if its edges have weight or the sizing is not fixed its a graph. Kinda rough, but we can revise as we get better at this.

PR #779, opened 2017-08-02
config.json: Add difficulty & topics for word-search

tleen:
See #345 & #731

### DIFF
--- a/config.json
+++ b/config.json
@@ -846,9 +846,11 @@
       "slug": "word-search",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 4,
       "topics": [
-
+        "Matrices",
+        "Searching",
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -846,9 +846,11 @@
       "slug": "word-search",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 4,
       "topics": [
-
+        "Matrices",
+        "Searching",
+        "Strings"
```
leenipper: + "Control-flow (loops)"

Or do we limit that topic to intro exercises ?

robphoenix: I don't think we should limit it to intro exercises.

tleen: There is probably looping in everything, I've been trying to limit it to where it may be particularly intense.

PR #781, opened 2017-08-02
config.json: Add topics & difficulty for robot-simulator

tleen:
See #345 & #731

### DIFF
--- a/config.json
+++ b/config.json
@@ -832,9 +832,13 @@
       "slug": "robot-simulator",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 5,
       "topics": [
-
+        "Concurrency",
+        "Control-flow (loops)",
+        "Sequences",
+        "Strings",
+        "Structs"
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -832,9 +832,13 @@
...
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 5,
       "topics": [
-
+        "Concurrency",
+        "Control-flow (loops)",
+        "Sequences",
+        "Strings",
+        "Structs"
```
leenipper: These topics look good.

config.json
```diff
@@ -832,9 +832,13 @@
       "slug": "robot-simulator",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 5,
```
leenipper: Because of the concurrency, I'm thinking at least 6 or 7 for difficulty.

tleen: Ok, wasn't sure with that. When I start seeing channels and stuff I tend to think at least 5+ unless it is utterly trivial. It is one of those unique Go things that trips people up.

PR #783, opened 2017-08-02
config.json: Add topics & difficulty for pov

tleen:
See #345 & #731

### DIFF
--- a/config.json
+++ b/config.json
@@ -814,9 +814,12 @@
       "slug": "pov",
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 7,
       "topics": [
-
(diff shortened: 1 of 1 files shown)


tleen:
Was not sure if we should consider "Graphs" and "Trees" redundant. Going by the way nextercism is shaping up I erred on the side of more topics being better than less.

config.json
```diff
@@ -814,9 +814,12 @@
...
       "core": false,
       "unlocked_by": null,
-      "difficulty": 1,
+      "difficulty": 7,
       "topics": [
-
+        "Graphs",
+        "Structs",
+        "Transforming",
+        "Trees"
```
tleen: Agreed.

PR #784, opened 2017-08-02
config.json: Add difficulty & topics for meetup

leenipper:
See #345 & #731


leenipper:
> perhaps be "Transforming"

Will add it.   I'll leave the "Type conversion" since there is an actual need to use Go type conversion syntax in at least one place in order to do the "Transform".

PR #790, opened 2017-08-03
config.json: Add topics & difficulty for binary-search-tree

tleen:
See #345 & #731


tleen:
Yep thats a good one.

Issue #798, opened 2017-08-08
Experiment with descriptive configuration field


tleen:
The hardest part of the config re-org thus far has been remembering the contents/goals of all the exercises.  The slug is good, but even then I find myself having to go the source and re-examine all the exercise info to figure it out. Over time more we work on the track the more we will recall each exercise, however we need to get the track properly configured for nextercism and do so on the quick side.

WIth this as a background I opened [an issue in discussions](https://github.com/exercism/discussions/issues/178) about adding some sort of descriptive field to the config.json entries for each exercise. In almost any other format we could do this via a comment but JSON does not support commenting and I would like to think this field should be available to the consumers of the JSON file, especially for config visualization tooling.

It seems there is no prohibition on adding extra fields to the config info so looks like if we are up to it we can try the concept out. I envisioned that this would be more focused on information that would be helpful for the maintainers to remember the purpose and intent of each exercise. The exercise info in [problem-specifications](https://github.com/exercism/problem-specifications) tell us the "what" of the exercises but not the "why", and the "why" could be very helpful in mapping out the track hierarchy. This information could be specific to each implementation so it would belong in the local config, not in the problem-specifications.

For example [etl](../tree/master/exercises/etl/) is probably the easiest exercise in the track someone would be forced to use the Map data type because it defines an interface requiring map use. This would be good to know in configuring the track. If any other exercises involve maps this may be a good prerequisite. You can't pickup this information from the topics/slug alone. 

I originally suggested the field may be named "purpose" but thinking on it more that seems limiting. Because it is a consumable field taking the place of a comment I think perhaps we should name it "comment". That leaves some flexibility in the field contents.

Thoughts?

tleen:
The experiment is in #800, any discussion is best handled there.

Issue #794, opened 2017-08-03
Choose core exercises structure


robphoenix:
From #717:

If you haven't already, now would be a good time to do the following:

~~- [x] add a rough estimate of difficulty to each exercise (scale: 1-10)~~
~~- [x] add topics to each exercise~~
- [ ] choose *at most 20 exercises* to be core exercises (set `core: true`, and delete the `unlocked_by` key)
- [ ] for each exercise that is not core, decide which exercise is the prerequisite (max 1)

If possible, leave 3 or 4 simple exercises as (core: false, unlocked_by: null), as this will provide new participants with some exercises that they can tackle even if they have not finished the first core exercise.

robphoenix:
If you've not already seen this, it should be useful https://github.com/exercism/discussions/issues/175#issuecomment-318243990

tleen:
I'm going to be mulling these over in my head for awhile as we have a ton of exercises. For starters what does everyone think do you think our most simple exercises are? 

These are our difficulty 1 exercises:

- [accumulate](../tree/master/exercises/accumulate)
- [etl](../tree/master/exercises/etl)
- [gigasecond](../tree/master/exercises/gigasecond)
- [hamming](../tree/master/exercises/hamming)
- [hello-world](../tree/master/exercises/hello-world)
- [leap](../tree/master/exercises/leap)
- [raindrops](../tree/master/exercises/raindrops)
- [scrabble-score](../tree/master/exercises/scrabble-score)

tleen:
I'm gonna say our easiest exercises are:
* hello-world - to introduce the environment
* hamming - simple looping
* raindrops - conditionals
 
Then very similar:
* gigasecond - slightly more complex than hello world, but not too different either
* leap - another mathy but quick problem (leap and gigasecond should prob have one unlock the other?)

A little harder but also similar:
* scrabble-score - should use map style lookups (but doesn't have to)
* etl - should unlock from scrabble? another map exercise transforming the map instead of reading from it
* accumulate - to introduce functions as args

This is my rough read on the first few exercises.

robphoenix:
tbh I'm finding this quite hard to conceptualise in my head how to sort the exercises, and have been kind of avoiding it. I'm going to need to sit down with a pen and paper and map it out I think.

tleen:
[...]

PR #800, opened 2017-08-14
Restructure configfile for Nextercism, difficulties 1-2

tleen:
Our initial stab at the reconfiguration for Nextercism incorporating
elements of #794 and #798. The "comment" field is an experimental
addition to the configuration to make remembering each exercise a
little easier while working on the file.

Analysis of the first two levels of difficulty led to some
re-assigning of those values. Exercises with stubs moved to
difficulty. Exercises without a stub moved to 2.

### DIFF
--- a/config.json
+++ b/config.json
@@ -270,7 +270,7 @@
     {
       "uuid": "20cbd8f9-f8e3-4767-a0f5-94810ba3f902",
       "slug": "scrabble-score",
-      "core": true,
+      "core": false,
       "unlocked_by": "hamming",
       "difficulty": 2,
       "topics": [
--- a/config.json
+++ b/config.json
@@ -12,25 +12,27 @@
     {
       "uuid": "19957346-dedf-441e-85ea-656cac0d96d8",
       "slug": "hello-world",
-      "core": false,
+      "core": true,
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
         "Strings"
-      ]
+      ],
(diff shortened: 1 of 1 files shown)


tleen:
This is a rather big one, but I thought a lot of these changes were interdependent. After taking a holistic look at the levels I moved a few around. I found that the exercises with stubs, by having stubs were much easier than the non-stubbed exercises and that made a good delineation between the lowest levels.

This is much easier to view in the [visualizer](https://jonmcalder.shinyapps.io/exercism-config-viz/). 

This is a work in progress for sure.

tleen:
Also, I have been using a Google Docs Sheet to try and track some of this stuff. If anyone wants the link let me know.

robphoenix:
> Also, I have been using a Google Docs Sheet to try and track some of this stuff. If anyone wants the link let me know.

Yes, that would be amazing! I've been away, so I'm going to give this some time this week and form a considered reply.

tleen:
Invite sent

robphoenix:
@tleen I've opened #805, I only intended to try and get an idea for what might work as core exercises and got carried away. I'm sure there'll be a lot of moving exercises about, but for me at least it's finally given me a proper grasp of the problem. 
I think we should focus on choosing core exercises first, and then we can configure the others around them, what do you think?

tleen:
As your PR handles the whole track and this one only the level 1&2, it supersedes this PR.

Issue #786, opened 2017-08-02
simple-cipher missing step #3? 


tleen:
I may be mistaken (I'm blitzing through the exercises quickly) but does the this track's [simple-cipher](../tree/master/exercises/simple-cipher) not require the user to do the third step in the README with the random key/ArgumentError? This may be a case where we need to override the generated README?

leenipper:
> override the generated README

Agreed.

From the test code in simple-cipher:

> // Steps 2 and 3 seem to be describing the Vigenère cipher (Google it)
> // so let's do that too.  The **random thing, don't worry about**.  There is
> // no test for that.

tleen:
Ok we can fix that, probably worth keeping an eye out on the other autogenerated READMEs they may be out of sync with the implementations.

tleen:
Now I am wondering if we should take over the generated README to fix that one little thing about randomness that the test file says to ignore. We could

- Leave as is - let the test file explain that the random element of step #3 is unnecessary
- Take over the README - remove the randomness text describing step 3
- Fulfill the README - add the randomness requirement into the exercise

robphoenix:
Ideally I think we should fulfill the README, this could perhaps be a separate issue. In the meantime, I like moving notes like these from the test suite into the README.

PR #802, opened 2017-08-15
pangram: clarify language in README

### LINKED ISSUE #801: Pangram README claims input will only have ASCII
The README says:

    The alphabet used consists of ASCII letters `a` to `z`, inclusive, and is case
    insensitive. Input will not contain non-ASCII symbols.

There are German characters in one test which causes failures if not ignored or counted. But even then tests fail if code does not handle characters outside of ASCII. Specifically, Russian.

    --- FAIL: TestPangram (0.00s)
	pangram_test.go:37: Pangram test [Широкая электрификация южных губерний даст мощный толчок подъёму сельского хозяйства.], expected [false], actual [true]
	pangram_test.go:39: [Широкая электрификация южных губерний даст мощный толчок подъёму сельского хозяйства.] should not be a pangram because : Panagram in alphabet other than ASCII

I am OK with having to check if a rune is ASCII or not. But the README explicitly says not to worry about it. Or is 'symbols' here meant to mean things other than letters like braces, emoji, and the like? Either way the README is at best not clear and at worst wrong.

tleen:
Made README clearer in regards to allowed inputs for pangram / allowed
inputs for interface.

Added the term 'rune' into the problem context.

Resolves #801

### DIFF
--- a/exercises/pangram/README.md
+++ b/exercises/pangram/README.md
@@ -5,8 +5,9 @@ Determine if a sentence is a pangram. A pangram (Greek: παν γράμμα, pan
 The best known English pangram is: 
 > The quick brown fox jumps over the lazy dog.
 
-The alphabet used consists of ASCII letters `a` to `z`, inclusive, and is case
-insensitive. Input will not contain non-ASCII symbols.
+The alphabet used to determine pangram status consists of the ASCII letters `a`
+to `z`, inclusive, and is case insensitive. Input may contain runes not in this
+set.
 
 ## Running the tests
 


robphoenix:
I'm not fully up on generating the READMEs, is this in danger of being overwritten?

tleen:
I don't think so, I believe that was a one time thing? I do have to look into it more, it's part of the reason why #786 has been sitting around. If that ends up happening we can redo in the local metafile or wherever.

tleen:
Yes, it would be overwritten but by us via configlet, automation in README's is a thing of the past. See the comment in 4a9e2a05b1a5e4ee9b9c31bea2fe3128b8935bcb.

robphoenix:
Okay, yeah, unless we change the description.md in the `problem-specifications` repo, then anything we add to `hints.md` will be additional. To change this we need to re-write it here. I suppose we just need to be considerate of this if we ever want to regenerate READMEs in the future and not just do them all.

tleen:
This may be a good example of why we moved to locally generated READMEs in the first place.

robphoenix:
I wonder if we're addressing the wrong problem here. Looking at https://github.com/exercism/problem-specifications/commit/ff2abe7ed8ba8021f6b1b22f57703dd4225ff6a4 & https://github.com/exercism/problem-specifications/issues/842 it may be that we should update the test suite to reflect the `canonical-data.json` instead of rewriting the README. And then add a generator when we can.

tleen:
Yeah, I think that is better since we are the divergent track here.

robphoenix:
Okay, do you want to update the `config.json` manually now until we implement the generator?

tleen:
I was gonna yank those two entries out of the test suite?

robphoenix:
Sounds good :+1:

PR #807, opened 2017-08-21
hello-world: Simplify to be less overwhelming

kytrinyx:
I recently realized how incredibly overwhelming the massive wall of text is in the hello world exercise.
    
This simplifies it down to the barest of bare bones, which I think feels friendlier and easier to complete.
    
In addition to removing comments, this also removes the testVersion constant and the example test.
    
My reasoning for removing the testVersion is primarily that most people don't give feedback on the Hello, World exercise, and if they do, then they probably have seen it enough to know that it got simplified.
    
Secondarily, when we launch the new v2 prototype publically, we will be able to get rid of the testVersions entirely, since we will always show the mentors the exact test suite that the user downloaded.
    
I don't anticipate needing to change hello-world between now and then, so we shouldn't see any confusion amongst the few people who may give feedback on hello-world.

Also, I'm going to implement two-fer (the exercise that replaced the conditional version of hello-world), and give it an example test.


kytrinyx:
Edit: I will implement Two Fer and give it an example test. I will not add the testVersion to two-fer, but rather to Leap. This should give us a more gradual ramp-up of concepts.

robphoenix:
Out of interest, do you have a timeline for launching v2 publicly? It would be good to have a freeze on significant changes to exercises shortly prior so that we can remove all the test versioning code.

And one more thing, when you implement `two-fer`, I think it could be the second `core` exercise, and we could perhaps have `hamming`, `leap` and `gigasecond` either unlocked by it or separate to the core exercise structure, and move `bob` to be unlocked by `two-fer`. (If you don't have time to implement `two-fer` right away, open an issue and I'll get to it before the end of the week.)

kytrinyx:
> do you have a timeline for launching v2 publicly?

Not really. If we have no significant blockers, it could be 3-4 weeks.

kytrinyx:
More details about the criteria for public launch from the v2-feedback README: https://github.com/exercism/v2-feedback/blob/master/README.md#criteria-for-public-launch

PR #808, opened 2017-08-21
Add two-fer exercise

kytrinyx:
This adds the exercise that replaced the old conditional version of Hello World.

I also added the Example Test here. I considered using _only_ an example test, but figured I'd play it safe and go with what we used to have for Hello World.

I did not add the testVersion–we're going to be able to get rid of this book-keeping in the near future, and my fingers are crossed hoping that we won't need to update two-fer in the meantime.

This gives us a fairly gentle ramp-up:

- `hello-world`:, just wire everything together and get a test passing
- `two-fer`: some more explanatory comments in in the stub file and an example test
- `leap`: introduce `testVersion`

### DIFF
--- a/config.json
+++ b/config.json
@@ -20,6 +20,17 @@
       ]
     },
     {
+      "uuid": "dc4ad67d-0fe4-0880-5a13-3aebb8759bcc3095de1",
+      "slug": "two-fer",
+      "core": true,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "Control-flow (conditionals)",
+        "Strings"
+      ]
+    },
+    {
       "uuid": "52582415-8f7f-4ac5-857f-5c160ec48134",
       "slug": "hamming",
       "core": true,
--- a/exercises/two-fer/README.md
+++ b/exercises/two-fer/README.md
@@ -1,0 +1,66 @@
+# Two Fer
+
+`Two-fer` or `2-fer` is short for two for one. One for you and one for me.
+
+```
+"One for X, one for me."
+```
+
+When X is a name or "you".
+
+If the given name is "Alice", the result should be "One for Alice, one for me."
+If no name is given, the result should be "One for you, one for me."
+
+
+## Test-Driven Development
+
+As programmers mature, they eventually want to test their code.
+
+Here at Exercism we simulate [Test-Driven
+Development](http://en.wikipedia.org/wiki/Test-driven_development) (TDD), where
+you write your tests before writing any functionality. The simulation comes in
+the form of a pre-written test suite, which will signal that you have solved
+the problem.
+
+It will also provide you with a safety net to explore other solutions without
+breaking the functionality.
+
+### A typical TDD workflow on Exercism:
+
+1. Run the test file and pick one test that's failing.
+2. Write some code to fix the test you picked.
+3. Re-run the tests to confirm the test is now passing.
+4. Repeat from step 1.
+5. Submit your solution (`exercism submit /path/to/file`)
+
+## Instructions
+
+Submissions are encouraged to be general, within reason. Having said that, it's
+also important not to over-engineer a solution.
+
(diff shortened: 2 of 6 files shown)


config.json
```diff
@@ -20,6 +20,17 @@
       ]
     },
     {
+      "uuid": "dc4ad67d-0fe4-0880-5a13-3aebb8759bcc3095de1",
+      "slug": "two-fer",
+      "core": false,
```
robphoenix: Do you want to leave this exercise out of the new core structure?

I think with these changes it would be good to have this as the second `core` exercise, and `leap` as the third, making sure that any exercises unlocked by them don't contain anything not already explained.

If you want to merge this as is, I'd be happy to update the config.json file and any changes to exercises that might result?

kytrinyx: I was actually thinking that this would be a nice "bonus" exercise that you get as soon as you submit Hello World.

By not including it as core and not having it be unlocked by anything, it's immediately available to work on in addition to Leap.

robphoenix: My only concern with having this as a bonus rather than a core exercise, is it contains info introducing a couple of things, stubs and example tests, and that these may get bypassed if the exercise doesn't need completing, does that make sense? 

> once you click on "complete" for hello world, you would also have Leap available.

Would this be with Leap as the second core exercise, or as another `bonus` exercise.
Currently the core structure for the first two exercises is this
```
hello-world
hamming
  +-- leap
  +-- gigasecond
```

> a nice "bonus" exercise

Yes, I think we could have a couple more of those, maybe `gigasecond` and `bob`, I'll look into it, I can think of a couple of changes to the config.json already now the structure is starting to settle in my head.

kytrinyx: > My only concern with having this as a bonus rather than a core exercise, is it contains info introducing a couple of things, stubs and example tests, and that these may get bypassed if the exercise doesn't need completing

Oh! Yeah, totally. That is an excellent point. You just changed my mind.

kytrinyx: Or... how about sticking the example test in one of the other core exercises? And then also introducing stubs in the first core exercise after Hello World?

robphoenix: Interestingly I've just been reading through https://github.com/exercism/v2-feedback/issues/36
and I'm going to update the config.json to include more side exercises.

kytrinyx: Yeah, this thread has really emphasized for us how important the side exercises are.

We're going to need to do actual product design work to make sure we're never blocking our power users, but even for non-power users the side exercises are going to be really central.

robphoenix: > how about sticking the example test in one of the other core exercises

Yep, I think only exercise with example tests currently is `clock`, so it could even be odd to have this big explanation of example tests and then not have any. The example tests in `clock` were added because the exercise was confusing, maybe for now we could add the note to that exercise and then move it to an earlier core exercise if we add more example tests?

> introducing stubs in the first core exercise after Hello World?

Yep, I think `two-fer` should be the second core exercise, and then we can rejig the next few, make a few of them bonus exercises. I can open a PR for that tomorrow.

robphoenix: Just so I'm clear in my head about the bonus exercises, those that are `core:false, unlocked_by:null` are available as soon as `hello-world` has been submitted? Is that any different to having them be `unlocked_by: hello-world`?

kytrinyx: > Yep, I think two-fer should be the second core exercise

ten four. I'll make the change!

> Is that any different to having them be unlocked_by: hello-world?

Slightly. If they're unlocked_by: null then you get them as soon as you submit hello world. If they are unlocked by hello world, then you have to go through the completion ritual first, before you get them.

We should probably have a few of each.

kytrinyx:
Alrighty. Updated this to be the second core exercise.

config.json
robphoenix: > We should probably have a few of each.

sweet, will do, I'll get on it in the morning.

robphoenix:
Excellent, thanks @kytrinyx! :tada: 
I'll merge it in.

PR #809, opened 2017-08-21
leap: explain testVersion

kytrinyx:
In a separate pull request, I'm simplifying hello-world. That removed
the explanation of what testVersion is.

This adds it to leap exercise so that people get the explanation.

### DIFF
--- a/exercises/leap/leap_test.go
+++ b/exercises/leap/leap_test.go
@@ -2,11 +2,15 @@ package leap
 
 import "testing"
 
-// Define a function with the following signature:
-// IsLeapYear(int) bool
-
+// targetTestVersion is used to ensure that a solution is only evaluated
+// against the correct version of the test suite.
 const targetTestVersion = 3
 
+// TestTestVersion compares the targetTestVersion defined above with
+// a testVersion value.
+// This is a common convention throughout the Go Exercism track.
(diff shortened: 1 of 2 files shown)


exercises/leap/leap_test.go
```diff
@@ -2,17 +2,23 @@ package leap
...
+// To make a test like this test pass, define 'testVersion' in your solution.
+// We've done this for you in the ./leap.go file provided.
 func TestTestVersion(t *testing.T) {
 	if testVersion != targetTestVersion {
 		t.Fatalf("Found testVersion = %v, want %v", testVersion, targetTestVersion)
 	}
 }
 
+// Define a function with the following signature:
+// IsLeapYear(int) bool
```
robphoenix: I'm not sure this is necessary as the function signature is already defined in the stub file. https://github.com/exercism/go/blob/master/exercises/leap/leap.go#L5

And, in fact, I think the  [comment within the function](https://github.com/exercism/go/blob/master/exercises/leap/leap.go#L6) could probably be got rid of as it is introduced in `two-fer`.

kytrinyx: Yeah, agreed. I'm going to go ahead and drop both.

PR #811, opened 2017-08-22
gen: change x-common references to problem-specifications

robphoenix:


robphoenix:
This is a simple change, and I could use it to update some of the exercises while I'm making changes to the `config.json` so I hope no-one minds but I'm going to go ahead and merge this on green.

Issue #703, opened 2017-06-30
Clarify problem statement for error-handling


tleen:
Based on [user feedback](http://exercism.io/submissions/dcbbef5ad09e447fb261ba8a13bc211b) and my own experience, the problem definition for error-handling can use some clarification. This is actually a fairly tricky topic at this stage if you are new to Go and I think in this case a little more hand-holding will be helpful.

I'll handle this one, but am open to feedback first. Should it be explained more?

robphoenix:
I've just attempted this exercise and while it's fundamentally not that complicated, the explanation is utterly confusing and doesn't help clarify the problem and expected solution in any way. I think #707 goes a long way to rectify this, :+1: @tleen
I've made some notes as I was doing the exercise, knowing you recently opened this issue. I think I have a couple of things to add, and I'd like to move the commentary into the README. I'll try and get it done over the weekend, but it might not be til Tuesday.

tleen:
:+1: I've been trying to go through the track as a newbie (which I am) and seeing where suddenly things get weird/hard. It is a valuable way of experiencing things. This exercise is a jump. In nextercism this may be rectified, but for now I think shepherding is a good approach to keep people from just giving up completely.

Thus far when I look at solutions I look for:

- can I do this? - can I figure out the problem quickly from what is available via docs
- common approaches - is it obvious how to handle this and a good number of solutions don't go that way
- blatant copies - copying the example.go file (perhaps with  a simple str replace) to move on
- give ups - pure submissions with no attempt to solve the problem

This problem hit a lot of those buttons and I think moving explanations to the README will be that much more helpful. Now that we have a new way to modify local READMEs we should shift the test commentary to those. #728 should eventually move to README too.

PR #838, opened 2017-08-31
error-handling: move detailed exercise description to README

### LINKED ISSUE #833: error-handling: improve documentation
From https://github.com/exercism/go/issues/829#issuecomment-324587887

>  it's just a little confusing, with some clearer explanation in the README it could be easier to digest. So :+1: with the caveat that the documentation is improved.

tleen:
Originally added for #703, moved the longer description for the
exercise to the README using the new `.meta/` files enabled via
`configlet` v3+. Links have been moved into markdown format, some
language has been updated as well.

Resolves #833

### DIFF
--- a/exercises/error-handling/.meta/description.md
+++ b/exercises/error-handling/.meta/description.md
@@ -1,19 +1,39 @@
 Implement various kinds of error handling and resource management.
 
-An important point of programming is how to handle errors and close resources even if errors occur.
+An important point of programming is how to handle errors and close resources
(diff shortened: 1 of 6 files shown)


tleen:
I still think we should have a simpler error handling exercise in core, early on, that introduces the idiomatic handling pattern of:

```
if foo, err := bar(); e != nil {
//process foo
```
This exercise is good, but perhaps tries to do too much with panics and type assertions on errors.

PR #840, opened 2017-09-03
Improve Handshake Benchmark Test

dustingraham:
It would seem to be a better speed analysis of the routine if we try more than just `Handshake(31)`

This is especially true when you consider there could be two code paths, or significant overhead depending on how the reverse functionality is handled.

I suspect do not need to bump the `targetTestVersion` since the test cases remain the same.

### DIFF
--- a/exercises/secret-handshake/secret_handshake_test.go
+++ b/exercises/secret-handshake/secret_handshake_test.go
@@ -28,7 +28,10 @@ func TestHandshake(t *testing.T) {
 }
 
 func BenchmarkHandshake(b *testing.B) {
+	var j uint
 	for i := 0; i < b.N; i++ {
-		Handshake(31)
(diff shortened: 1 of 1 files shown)


exercises/secret-handshake/secret_handshake_test.go
```diff
@@ -28,7 +28,10 @@ func TestHandshake(t *testing.T) {
 }
 
 func BenchmarkHandshake(b *testing.B) {
+	var j uint
 	for i := 0; i < b.N; i++ {
-		Handshake(31)
+		for j = 0; j < 32; j++ {
```
leenipper: Consider, more simply:
> for j := uint(0); j < 32; j++ {

since j is needed only inside the for loop, and then we can drop the var outside.

dustingraham:
Thanks @leenipper, updated.

PR #841, opened 2017-09-03
phone-number: update README.md with Format

### LINKED ISSUE #839: phone-number: `AreaCode()` and `Format()` are not in description
The problem description says to return the phone number as one long number. But then there is a `Format()` expectation looking for '(NXX)-NXX-XXXX' which is not documented and goes against the description.

I like the addition of `Format()` but it needs to be documented at a minimum in the tests.

leenipper:
Document the expectation for implementing
a Format function, and the example output it should produce.

Closes #839.

### DIFF
--- a/exercises/phone-number/README.md
+++ b/exercises/phone-number/README.md
@@ -29,6 +29,10 @@ Additionally, a Format() function should produce the output
 
 `(613) 995-0253`
 
+and an AreaCode() function should output
+
+`613`
+
 **Note:** As this exercise only deals with telephone numbers used in NANP-countries, only 1 is considered a valid country code.
 
 ## Running the tests
--- a/exercises/phone-number/README.md
+++ b/exercises/phone-number/README.md
@@ -25,7 +25,7 @@ should all produce the output below from a Number() function.
 
 `6139950253`
 
-Additionally, a Format() function should produce the outout
+Additionally, a Format() function should produce the output
 
 `(613) 995-0253`
 
--- a/exercises/phone-number/README.md
+++ b/exercises/phone-number/README.md
@@ -21,9 +21,13 @@ For example, the inputs
 - `1 613 995 0253`
 - `613.995.0253`
 
-should all produce the output
+should all produce the output below from a Number() function.
 
 `6139950253`
+
+Additionally, a Format() function should produce the outout
+
+`(613) 995-0253`
 
 **Note:** As this exercise only deals with telephone numbers used in NANP-countries, only 1 is considered a valid country code.
 
--- a/exercises/phone-number/.meta/description.md
+++ b/exercises/phone-number/.meta/description.md
@@ -23,7 +23,7 @@ should all produce the output below from a Number() function.
 
 `6139950253`
 
-Additionally, a Format() function should produce the outout
+Additionally, a Format() function should produce the output
 
 `(613) 995-0253`
 
--- a/exercises/phone-number/.meta/description.md
+++ b/exercises/phone-number/.meta/description.md
@@ -1,0 +1,30 @@
+Clean up user-entered phone numbers so that they can be sent SMS messages.
+
+The **North American Numbering Plan (NANP)** is a telephone numbering system used by many countries in North America like the United States, Canada or Bermuda. All NANP-countries share the same international country code: `1`.
+
+NANP numbers are ten-digit numbers consisting of a three-digit Numbering Plan Area code, commonly known as *area code*, followed by a seven-digit local number. The first three digits of the local number represent the *exchange code*, followed by the unique four-digit number which is the *subscriber number*.
+
+
+The format is usually represented as
+```
+(NXX)-NXX-XXXX
+```
+where `N` is any digit from 2 through 9 and `X` is any digit from 0 through 9.
+
+Your task is to clean up differently formated telephone numbers by removing punctuation and the country code (1) if present.
+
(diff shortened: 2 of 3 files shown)


robphoenix:
This looks good @leenipper 

Though I'm still not sure if we should be manually editing the README or putting something like this is the `description.md` or `hints.md` files? I think you're clearer on this @tleen

tleen:
Directly editing the README [is now discouraged](https://github.com/exercism/docs/commit/eabb325e8c7404aaba256175d9f4b7f0c0203121). I believe the proper way going forward is to put it in a `<exercise>/.meta/description.md` or `<exercise>/.meta/hints.md` file, depending on context. Then use [configlet](https://github.com/exercism/configlet) to generate the new README from that. (It requires a near current version (3+) of configlet which can be obtained via the [bin/fetch-configlet](https://github.com/exercism/go/blob/master/bin/fetch-configlet) script.)

Ideally going forward we use the .meta/ directory files and configlet to generate local README's that have partial local overriding of information instead of overwriting the whole thing. This way we don't lose the other parts of the README from being part of [a global template](https://github.com/exercism/go/blob/master/config/exercise_readme.go.tmpl). Things like [how to run the exercises](https://github.com/exercism/go/blob/master/docs/EXERCISE_README_INSERT.md), if that ever changes we could update all the README's at once with `configlet`. If this updated information was directly encoded in the README we would lose it if we did a global re-generation.

I'm seeing it as anything specific to our local exercise goes into `.meta/hints.md` (like pointing out useful Go stdlib topics to use during the exercise). But if something in the problem-specification's README is contradictory or incorrect you can replace it whole via the `.meta/description.md` file. This update could be either, I tend to defer to `hints.md` this way we wouldn't miss any description update in the problem-specifications repo.

tleen:
Also there is a current [issue](https://github.com/exercism/configlet/issues/65) in `configlet` that requires a copy of the exercise's `problem-specifications` `metadata.yml` file to be in the local .meta directory to have any of the overrides take effect.

leenipper:
Ah, I have some catching up to do on the new mechanics. This one is on hold for a while.

tleen:
@leenipper it took me a little while to get the hang of it. The whole thing makes sense now after some tinkering. It's part of the reason these README changes are backing up a bit. I think #838 was the first one we did in this repo using the new system.

robphoenix:
> Ah, I have some catching up to do on the new mechanics.

You and me, both, @leenipper :smile:

leenipper:
See if this is going the right direction.  @tleen  I know you favored using hints.md in .meta and I can do that instead if you think strongly on it for preserving description.md changes coming from problem-specifications.  hints.md seemed wrong to "adjust" the text of the description.md, but maybe I just need to think more freely about it.

A question which came up during this, was ... do we periodically update the committed exercise README.md files when they become out of date wrt problem-specification repo files?  I observed that both palindrome-products and simple-cipher are already out-of-date in this fashion.  Seems similar to canonical-data.json making a cases_test.go out of date; so I expect the answer is "yes, when needed".

tleen:
I'm good with doing in the description if it makes sense. It is sort of arbitrary if the contents make more sense inline in the description than in a separate section below it. The more I am updating READMEs the more I am finding myself taking over the description too, just because there is drift between the problem-specification description and the Go track's. I am in the middle of  updating `simple-cipher` right now and ended up doing both description and hints, because I felt the description needed to be more accurate towards our exercise and used hints to describe implementation details like the interfaces/functions that I plucked out of the test file comments. In this case the implementation detail is so minor it may be better served right there in the description.

We should be periodically looking at updates to the descriptions and deciding on their inclusion. I wouldn't think it is an automatic update because changes to the problem-specification description may have moved it further away from our track's implementation. We would have to look at each update individually. So in a nutshell "yes, when needed". :grin:

exercises/phone-number/README.md
```diff
@@ -21,10 +21,14 @@ For example, the inputs
 - `1 613 995 0253`
 - `613.995.0253`
 
-should all produce the output
+should all produce the output below from a Number() function.
 
 `6139950253`
 
+Additionally, a Format() function should produce the outout
```
tleen: Typo here 'outout'.

leenipper: Oops.

shaleh:
@tleen perhaps something like the `testVersion` could be added to the READMEs? `docVersion` perhaps? If the local READMEs version does not match the central repos version then a script to mass tag bugs, complain when output, something useful :-)

tleen:
@shaleh I've been using the [configlet tool](https://github.com/exercism/configlet) generate on a throwaway branch to do a mass update and see what changes. By default it seeks to update all the READMEs. That may be enough? `testVersion` is slated to go away soon as IIRC nextercism will show the tests used with the solutions automagically. I think.

shaleh:
hmmm. I have looked for other people's solutions here on GitHub after I wrote mine. There are a couple from around 2 years back. Seeing the testVersion was how I could see their solution did not match the version of the problem I have now. But I do not want to derail this thread any further :-)

tleen:
@shaleh I'm not 100% sure that the test version is going away.

leenipper:
I just noticed that the #839 mentions in its title that Area(), actually AreaCode(), is also omitted from the README.md.  To really close #839, AreaCode() should be mentioned, so I will add another commit to this series.

PR #843, opened 2017-09-06
simple-cipher: update README to reflect exercise implementation 

### LINKED ISSUE #786: simple-cipher missing step #3?

tleen:
Clear up the expected implementation details for the simple-cipher
exercise in the README. Provide more explanation of the Vigenère
cipher. Moved README updates into .meta/ directory and regenerated the
README.

Resolves #786

### DIFF
--- a/exercises/simple-cipher/.meta/hints.md
+++ b/exercises/simple-cipher/.meta/hints.md
@@ -1,0 +1,35 @@
+## Implementation
+
+The definition of the Cipher interface is located in
+[cipher.go](./cipher.go).
+
+Your implementations should conform to the Cipher interface.
+
+```go
+type Cipher interface {
+    Encode(string) string
+    Decode(string) string
+}
+```
+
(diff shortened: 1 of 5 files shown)


tleen:
I have to squash all these into one before commit.

exercises/simple-cipher/.meta/hints.md
```diff
@@ -0,0 +1,35 @@
+## Implementation
```
leenipper: I really like the "## Implementation" section for hints.md.  It works very well to bridge from the description content.

tleen: Me too. if we find it useful we may want to put it in the template.

PR #845, opened 2017-09-07
roman-numerals: remove non-canonical test data

### LINKED ISSUE #836: roman-numerals: README says no bigger than 3000 but has 3999 as test case
So which is it? No bigger than 3000? Or no bigger than 3999?

tleen:
Removed extra test cases to bring the roman-numeral exercise into
consistency with the problem-specifications description. Bumped test
version accordingly.

Resolves #836

### DIFF
--- a/exercises/roman-numerals/roman_numerals_test.go
+++ b/exercises/roman-numerals/roman_numerals_test.go
@@ -2,7 +2,7 @@ package romannumerals
 
 import "testing"
 
-const targetTestVersion = 3
+const targetTestVersion = 4
 
 func TestTestVersion(t *testing.T) {
 	if testVersion != targetTestVersion {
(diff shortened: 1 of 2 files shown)


exercises/roman-numerals/roman_numerals_test.go
```diff
@@ -11,13 +11,7 @@ func TestTestVersion(t *testing.T) {
 }
 
 func TestRomanNumerals(t *testing.T) {
-	tc := append(romanNumeralTests, []romanNumeralTest{
-		{0, "", true},
-		{-1, "", true},
-		{4000, "", true},
```
leenipper: And then favor changing the 4000 case to be 3001, which should also return an error. That would require updating the line shown below in example.go as well to change its invalid range to use 3001 instead of 4000 for the high boundary test. But it would be in line with the README and yet preserve the usefulness of the error return value.
> if input <= 0 || input >= 4000 {

shaleh: Agreed.

tleen: This is all spot on, incorporated.

PR #850, opened 2017-09-09
gen/gen.go: correct location for cases_test.go

### LINKED ISSUE #848: gen/gen.go with Go 1.9 always outputs cases_test.go to directory above current
When running a test case generator on Linux with Go 1.9, I have observed that it always shows this line of output:
> 2017/09/07 23:23:02 [SUCCESS] output: ../cases_test.go

instead of this line of output for roman-numerals (as with Go 1.8.3):

> 2017/09/07 23:26:05 [SUCCESS] output: /home/lee/exercism/go/exercises/roman-numerals/cases_test.go

Since ".." is relative to your "current working directory", it *may* not be correct, depending on your current directory location.  Running from within <exercise-dir>/.meta will work, but is not very convenient.

I realize exercism/go has not embraced 1.9 yet officially, but this is a heads up.

BTW, I have correction in preparation that works for both earlier versions and also 1.9.

leenipper:
Using Go 1.9 with test case generators breaks
the placement of the output file, cases_test.go.
Specifically, on Linux it is output to ../cases_test.go which is
relative to your current directory.

The bug is related to the fact that with Go 1.9, this line

 	if _, path, _, ok := runtime.Caller(2); ok {

when called from init() within a package
has a different result than 1.8.3 and earlier.
Path is returned as ```<autogenerated>``` with 1.9,
and dirExercise ends up being set to "..".

The correct solution for both new and previous Go versions
is to utilize the same technique of runtime.Caller(),
but from a known call stack position apart from the init() function.

The change moves the logic to set dirExercise
from init() into the Gen() function in order to
set a correct value for the exercise directory regardless
of the current working directory when running the generator.

Also, add some commentary explanation of how dirExercise is determined.

Resolves #848

### DIFF
--- a/gen/gen.go
+++ b/gen/gen.go
@@ -70,12 +70,6 @@ func init() {
 	if _, path, _, ok := runtime.Caller(0); ok {
 		dirMetadata = filepath.Join(path, "..", "..", "..", "problem-specifications")
 	}
-	if _, path, _, ok := runtime.Caller(2); ok {
-		dirExercise = filepath.Join(path, "..", "..")
-	}
-	if dirExercise == "" {
-		dirExercise = "."
-	}
 }
 
 // outputSource puts the src text into given fileName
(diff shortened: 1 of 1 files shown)


tleen:
This is a quality commit message.

PR #858, opened 2017-09-12
robot-name: Add hints.md file

### LINKED ISSUE #832: robot-name: merge bonus tests into main test suite
From https://github.com/exercism/go/issues/829#issuecomment-324587887

> it's probably worth bringing the bonus tests into the main test suite. I don't think anything's gained from splitting them. If needs be we can describe in the readme how to only run certain tests with go test so the user can work on them incrementally.

robphoenix:
This describes the bonus test and how to run it.

closes #832


robphoenix:
I did suggest in #832 that we should combine the test files, but I decided against that. How it is now with the build constraints is a better way than filtering tests as I originally thought.

PR #860, opened 2017-09-13
[nextercism] add, update & remove stub files

robphoenix:


exercises/triangle/triangle.go
```diff
@@ -1,16 +1,29 @@
...
 // Notice KindFromSides() returns this type. Pick a suitable data type.
 type Kind
 
-// Pick values for the following identifiers used by the test program.
-NaT // not a triangle
-Equ // equilateral
-Iso // isosceles
-Sca // scalene
+const (
+    // Pick values for the following identifiers used by the test program.
```
tleen: Wonder if we should link somewhere in the hints about [const values and iota](https://splice.com/blog/iota-elegant-constants-golang/)?

robphoenix: Good idea. I remember when I discovered this through Exercism, mind :boom:

exercises/triangle/.meta/hints.md
```diff
@@ -0,0 +1,10 @@
+## Contant Declarations
```
robphoenix: @tleen thoughts?

exercises/triangle/.meta/hints.md
```diff
@@ -0,0 +1,10 @@
+## Constant Declarations
```
robphoenix: @tleen thoughts?

tleen: LGTM

Issue #847, opened 2017-09-08
pov: why is root not in leaves or otherwise used?


shaleh:
The sequence is:

    graph := New()
    for _, l := range leaves {
        graph.AddNode(l)
    }
    for _, p := range arcPairs {
        graph.AddArc(p.from, p.to)
    }

No where in there is the value for "root" added. So when `AddArc` is called the value of `from` is unknown.

The singleton test which is the first test has one leaf which matches the value of root. Most of the others use "parent" which is never explicitly defined.

The comments in the test file say that `to` will always exist but is quiet on `from`. It seems odd that the expectation is to create missing `from`. Or did I missed something?

tleen:
If it is unclear to you than it is probably unclear and we should fix.

robphoenix:
@shaleh do you have a fix for this?

I have to admit I don't fully understand what the issue is, would you be able to expand on it so we can outline what a fix for this might look like. Thanks.

ferhatelmas:
Actually, every node is root. IMHO, it's the point of the exercise. It changes according to where you're looking from. 

However, you're right. Current implementation has hidden assumptions and isn't aligned with [problem canonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/pov/canonical-data.json)

Anyway, we will need to write a generator for it so we can rework. @shaleh do you want to give a try ?

shaleh:
I could write a generator for this. I am at least a week out from having time for that. If that is ok, then yes consider it assigned to me.

ferhatelmas:
Assigned it to you. Thanks!

PR #869, opened 2017-09-22
anagram: Add test generator

shaleh:
For #605

I did not have a way to wake up #695 so I started a new one.

This generator mimics the output of the connect generator so the output is
a little nicer IMHO. I bumped the `testVersion`.

### DIFF
--- a/exercises/anagram/.gitignore
+++ b/exercises/anagram/.gitignore
@@ -1,0 +1,1 @@
+.meta/gen
--- a/exercises/anagram/.meta/gen.go
+++ b/exercises/anagram/.meta/gen.go
@@ -1,0 +1,55 @@
+package main
+
+import (
+	"log"
+	"text/template"
+
+	"../../../gen"
+)
+
+func main() {
+	t, err := template.New("").Parse(tmpl)
+	if err != nil {
+		log.Fatal(err)
+	}
+	var j js
+	if err := gen.Gen("anagram", &j, t); err != nil {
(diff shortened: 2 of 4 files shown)


shaleh:
Did I break a check? The logs are pretty empty.

robphoenix:
thanks @shaleh I'll have to have a proper look at this on Monday.

> Did I break a check? The logs are pretty empty.

It looks like Travis is complaining about the `testVersion`. We've just removed all test versioning in the `nextercism` branch.

```
# github.com/exercism/go/exercises/anagram
exercises/anagram/anagram_test.go:22: undefined: testVersion
exercises/anagram/anagram_test.go:23: undefined: testVersion
```

shaleh:
Ah. I see. That is disappointing.

shaleh:
All better @robphoenix, thx

tleen:
backlogged will review ASAP.

tleen:
Found #847 while verifying this one. Looks ok in and of itself but this specific exercise does make the bug show up as it has a [multiline most recent remote commit](https://github.com/exercism/problem-specifications/commit/196fc1a2e8274ee082c35442ea4e8b9c8872d9c1) on its canonical-data.json file.

tleen:
I don't see how this passes travis if #847 is a real bug. I'm gonna mess with the build a bit before committing this.

tleen:
Oh I see why, I don't think Travis runs the generators!

PR #870, opened 2017-09-29
triangle: Update README.md

navossoc:
I just "formatted" the document, nothing special.
I'm not pretty sure about it, but the markdown preview on the site exercism.io behaves like on GitHub or the <br> are really needed?

Not related:
I'm still doing all the exercises, but later, can I edit and submit more test cases?
Should I wait the beta goes live or doesn't matter?

Let me know.

Thanks.

### DIFF
--- a/exercises/triangle/README.md
+++ b/exercises/triangle/README.md
@@ -12,14 +12,14 @@ A _scalene_ triangle has all sides of different lengths.
 
 ## Note
 
-For a shape to be a triangle at all, all sides have to be of length > 0, and 
-the sum of the lengths of any two sides must be greater than or equal to the 
+For a shape to be a triangle at all, all sides have to be of length > 0, and
+the sum of the lengths of any two sides must be greater than or equal to the
 length of the third side. See [Triangle Inequality](https://en.wikipedia.org/wiki/Triangle_inequality).
 
 ## Dig Deeper
 
-The case where the sum of the lengths of two sides _equals_ that of the 
-third is known as a _degenerate_ triangle - it has zero area and looks like 
+The case where the sum of the lengths of two sides _equals_ that of the
+third is known as a _degenerate_ triangle - it has zero area and looks like
 a single line. Feel free to add your own code/tests to check for degenerate triangles.
 
 ## Running the tests


tleen:
You can definitely modify exercises. We suggest adding any modifications to the `nextercism` branch, since that is our focus now. If you find anything while going through the track that is worth mentioning at least add an issue right away. This way the reasoning is fresh for you and it can be discussed & handled from that point. It is fair that as a user if you sense something is wrong with an exercise as you get to it it is probably worth bringing up.

tleen:
READMEs in the track are odd. We actually shouldn't be modifying them directly. They are *auto-generated* from [a tool](https://github.com/exercism/configlet), using base information from the [problem-specifications](https://github.com/exercism/problem-specifications/tree/master/exercises/triangle) repository. 

Without a doubt your changes here are valid **but** if we accept them in the README itself they will be overwritten the next time the tool is run!

These changes should be made in two places to make them permanent:

 - Most of it is in the [problem-specifications/exercises/triangle/description.md](https://github.com/exercism/problem-specifications/blob/master/exercises/triangle/description.md) file.

- The lack of spacing under "Submitting Incomplete Solutions" is local to this repository and actually exists for every README. A good catch by you! That can be fixed in our track's [README template](https://github.com/exercism/go/blob/master/config/exercise_readme.go.tmpl). (In fact I think it is present in [other language repos](https://github.com/exercism/csharp/blob/master/config/exercise_readme.go.tmpl) as well!) Once it is fixed in our README template we can automatically fix it in all the READMEs in the track. Tooling is handy that way. :smile:

If you make the changes in the permanent locations you can run the configlet tool to modify all the READMEs at once and we can add it all as a single commit. Sound good?

navossoc:
Gotcha! I'm doing this...
I think I get it how it works.

I did a fork of the `problem-specifications` repository.

I'm sending all the changes to here:
https://github.com/navossoc/problem-specifications/commits/readme

When it is is done, I will create another pull request and close this one.
(for now, I'm using this one for questions 👍 )

Is there any way to preview the changes "on the site" or something like?
Just to be sure I'm not breaking anything.

tleen:
You can push up the README to your fork and see how it looks there?

navossoc:
@tleen Sure, so the behavior is the same as the github?
I know github has some small subtleties on the markdown language.

(don't remember any now, but I already experienced some)

Anyway, if looks good on github, should I assume it is ok?

[]'s

tleen:
The spacing issue for the local template doesn't matter for github since it adds the spacing under the header anyway when it turns the markdown into HTML so it is more about how the README looks as text.

tleen:
> I know github has some small subtleties on the markdown language.

FYI: it is GFM ([GitHub Flavored Markdown](https://github.github.com/gfm/)). Most differences are around un-ambiguating the syntax, and all the specific auto-linking of stuff.

tleen:
@navossoc I found a [neat program](https://github.com/joeyespo/grip) that will render GFM locally for previewing.

navossoc:
@tleen I got the PR accepted on the `exercism/problem-specifications`.

So, the next step will be fix the [template](https://github.com/exercism/go/blob/master/config/exercise_readme.go.tmpl) and generate the questions again.

Now, should we do that on `master` or  `nextercism` branch?
Let me know and I will make a new PR and we can close this one later.

Also, thanks for the tip about the `grip` markdown previewer ;)

tleen:
`nextercism` for now, we are trying to keep all the updates there so we can merge it all in when the time comes.

PR #881, opened 2017-10-03
anagram: remove .gitignore

robphoenix:

PR #885, opened 2017-10-05
say: generate tests

### LINKED ISSUE #837: Say README suggests to test -1 but the unit tests expect uint64

shaleh:
This brings the Go track in line with others. `Say` now returns `(string, bool)` so it can indicate when a value is out of range.

Part of #605.
Closes #837

### DIFF
--- a/exercises/say/.gitignore
+++ b/exercises/say/.gitignore
@@ -1,0 +1,1 @@
+.meta/gen
--- a/exercises/say/.gitignore
+++ b/exercises/say/.gitignore
@@ -1,1 +1,0 @@
-.meta/gen
--- a/exercises/say/example.go
+++ b/exercises/say/example.go
@@ -8,7 +8,14 @@ var tens = []string{"ones", "ten", "twenty", "thirty", "forty",
 var scale = []string{"thousand", "million", "billion",
 	"trillion", "quadrillion", "quintillion"}
 
-func Say(n uint64) string {
+func Say(n int64) (value string, ok bool) {
+	if n < 0 || n > 999999999999 {
+		return
+	}
+	return say(n), true
+}
+
+func say(n int64) string {
 	switch {
 	case n < 20:
 		return small[n]
@@ -23,18 +30,19 @@ func Say(n uint64) string {
 		h := small[n/100] + " hundred"
 		s := n % 100
 		if s > 0 {
-			h += " " + Say(s)
+			h += " " + say(s)
 		}
 		return h
 	}
 	sx := ""
+
 	if p := n % 1000; p > 0 {
-		sx = Say(p)
+		sx = say(p)
 	}
 	for i := 0; n >= 1000; i++ {
 		n /= 1000
 		if p := n % 1000; p > 0 {
-			ix := Say(p) + " " + scale[i]
+			ix := say(p) + " " + scale[i]
 			if sx > "" {
 				ix += " " + sx
 			}
--- a/exercises/say/say_test.go
+++ b/exercises/say/say_test.go
@@ -13,8 +13,9 @@ func TestSay(t *testing.T) {
 			}
 		} else if !ok {
 			t.Fatalf("FAIL: %s\nDid not expect an error", tc.description)
+		} else if actual != tc.expected {
+			t.Fatalf("FAIL: %s\nExpected: %v\nActual: %v", tc.description, tc.expected, actual)
 		}
-
 		t.Logf("PASS: %s", tc.description)
 	}
 }
--- a/exercises/say/say_test.go
+++ b/exercises/say/say_test.go
@@ -1,47 +1,28 @@
 package say
 
-// The steps are interesting, but all that matters is the final exam.
-
 import (
-	"math"
 	"testing"
 )
 
-var tests = []struct {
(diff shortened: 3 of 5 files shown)


exercises/say/.gitignore
```diff
@@ -0,0 +1 @@
+.meta/gen
```
tleen: Why ignore .meta/gen here?

tleen:
My feeling on it is that unless you have more information to convey than worked/didn't work you use `err`, otherwise `ok`.

shaleh:
habit. Checking in binaries is usually frowned on.

On October 6, 2017 7:07:03 AM PDT, Tom Leen <notifications@github.com> wrote:
>tleen commented on this pull request.
>
>
>
>> @@ -0,0 +1 @@
>+.meta/gen
>
>Why ignore .meta/gen here?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/exercism/go/pull/885#pullrequestreview-67678621

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

shaleh:
Some over achiever might want to report too big or too small. I think ok is fine here 

On October 6, 2017 7:11:36 AM PDT, Tom Leen <notifications@github.com> wrote:
>My feeling on it is that unless you have more information to convey
>than worked/didn't work you use `err`, otherwise `ok`.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/exercism/go/pull/885#issuecomment-334766362

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

tleen:
Good habit! I have it too! I only noticed because it [slipped by me](https://github.com/exercism/go/pull/881) in anagram. 

I'm ok with `ok` it's probably not a bad thing to bring into some exercises since it is a pattern in Go. @robphoenix @leenipper thoughts?

shaleh:
Will do. Probably not today.

On October 6, 2017 7:49:07 AM PDT, Tom Leen <notifications@github.com> wrote:
>tleen requested changes on this pull request.
>
>Def remove the gitignore file so we can track gen.
>
>
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/exercism/go/pull/885#pullrequestreview-67692292

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

ferhatelmas:
> Def remove the gitignore file so we can track gen.

@tleen I don't understand your above comment because even with `gitignore`, `gen.go` is tracked and we don't want to track artifact `gen`. I think ignore is good since it would prevent errors but place isn't the best. If we want to provide, it should be in the repo level for all exercises.

tleen:
Sorry @ferhatelmas I wasn't clear. We wouldn't be adding gen anyway, as you can run the gen without building it. I'm actually ok with it getting added to a commit so we can catch that error. The local .gitignore is a nice safety, but as you said should be done as a global policy, not at the exercise level.

ferhatelmas:
@tleen Thanks for clarification. We're on the same page then.

exercises/say/say_test.go
```diff
@@ -1,47 +1,28 @@
...
-		"six hundred fifteen"},
+func TestSay(t *testing.T) {
+	for _, tc := range testCases {
+		actual, ok := Say(tc.input)
+		if tc.expectError {
+			if ok {
+				t.Fatalf("FAIL: %s\nExpected error but received: %v", tc.description, actual)
+			}
+		} else if !ok {
+			t.Fatalf("FAIL: %s\nDid not expect an error", tc.description)
```
leenipper: Suggest adding:
```
+               } else if actual != tc.expected {
+                       t.Fatalf("FAIL: %s\nExpected: %v\nActual: %v", tc.description, tc.expected, actual)
```

exercises/say/example.go
```diff
@@ -8,7 +8,14 @@ var tens = []string{"ones", "ten", "twenty", "thirty", "forty",
 var scale = []string{"thousand", "million", "billion",
 	"trillion", "quadrillion", "quintillion"}
 
-func Say(n uint64) string {
+func Say(n int64) (value string, ok bool) {
```
leenipper: I'm good with ```ok bool``` here.  The standard library has instances like it.

tleen: Does anybody know if/where there is some sort of community accepted definition or guideline as for when to use an `err` vs an `ok`?

leenipper: I have not found a straight out "accepted definition".  There is discussion. Here is one from 2015: https://groups.google.com/forum/#!topic/golang-nuts/e5qIlnhpgg0
The two points I note are:
1. when only one possible error condition, a bool is fine.
2. using an error is more flexible for future updates.

I think for Say, it meets no. 1 (out of range is the !ok case), and doesn't warrant no. 2.

tleen: Thanks @leenipper that was sort of my take on it too. Good to see I was not alone in it!

shaleh: @leenipper thanks for catching that. New patch on the way soon.

Issue #868, opened 2017-09-22
Commentary on nextercism config.json


shaleh:
As promised, @robphoenix.

 * robot-name: I used a singleton pattern and Go's `once` construction. This will also be the first task to use methods and receive by reference. None of this is reflected in the topics. The difficulty is not increased. For someone well verses in Go this is likely a trivial task.
 * error-handling: This is one of the first uses of `defer` and `recover`. Again, no Go specific topics there. Perhaps callbacks? Exception handling?
 * bank-account: again `defer`. Also, maybe OO design?
 * triangle: unlocked by leap this is the first to really introduce types, enums, etc. No mention in topics
 * pascals-triangle: topics: recursion, algorithms
 * queenattack: topics: mathematics
 * anagram: topics: sorting
 * react: topics: interfaces, callbacks
 * phone-numbe: topics: regexp?
 * nth-prime & prime-factors: I used the same core code to solve both problems
 * saddle-points: with the matrix code written, this was a snap
 * word-search: for some programmers this could be more difficult than a 4
 * change: the large number test bumps up the difficulty. I started with greedy, then used a simple either or. When I hit that test I had to write a smarter solution. Perhaps make it an option to the developer?

All in all the unlocks and general order appears pretty sane.

tleen:
We are free to add our own topics, at the same time if something we think should be a topic and is not, I feel it should be proposed first in the [TOPICS.txt](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt) file. Sometimes [things don't get included](https://github.com/exercism/problem-specifications/pull/886), which should not preclude it from our track but we should give the problem-spec contributors a fair crack at thinking on the topic as a global before including it locally.

> saddle-points: with the matrix code written, this was a snap

This is really why I wanted all the matrix stuff together. It feels like if you solved one the rest follow fairly easily.

> queenattack: topics: mathematics

It doesn't seem obvious to me. Is it graph coordinate stuff?

> error-handling: This is one of the first uses of defer and recover. Again, no Go specific topics there. Perhaps callbacks? Exception handling?

I think this is our best exercise when it comes to teaching people how go works. It is an exercise that problem-specifications is very general about, so if we would have custom topics it would be here.

> robot-name: I used a singleton pattern and Go's `once` construction. 

If you think `once` is useful enough here we should probably include it in the HINTS.md file for the exercise so users can explore it before implementation.

Everything else seems ok, if you want we can `nextercism` PR them into existence (or at least a deeper debate!).

[...]

PR #890, opened 2017-10-05
react: update topics

shaleh:
Add callbacks and interfaces as per #868.


shaleh:
@robphoenix if you are confused by the addition of callbacks, read the example provided.

PR #894, opened 2017-10-11
flatten-array: add exercise with test generator

ferhatelmas:

### DIFF
--- a/config.json
+++ b/config.json
@@ -241,6 +241,17 @@
       "uuid": "cd05b63b-df20-4c8d-8743-53033acf7696"
     },
     {
+      "core": false,
+      "difficulty": 3,
+      "slug": "flatten-array",
+      "topics": [
+        "arrays",
+        "recursion"
+      ],
+      "unlocked_by": "error-handling",
+      "uuid": "40cc79a6-a6ea-45a0-944e-b4f3baabebc2"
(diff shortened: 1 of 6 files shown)


ferhatelmas:
Note that I didn't add it into config since I have no idea regarding order, difficulty, etc. I hope we can tune them while reviewing.

tleen:
Looks good so far but we should get it in config.json so it can pass the configlet lint in Travis. 

If I was going to ballpark difficulty I would say 3. My gut feeling is that it should be unlocked by `error-handling` seems like it would group well with the other unlocks `circular-buffer` and `custom-set`. 

Topics-wise probably: recursion, arrays?

ferhatelmas:
@tleen your suggestions seem good to me. I will add them but I have one question regarding UUID generation. We're supposed to use `configlet` right ?

If so, it currently generates weird ones (one example: `42a9ee65-00c7-9680-2bd3-42155d751e1124c902a`) but issue is related to library and I will create an issue there.

ferhatelmas:
@tleen updated as your suggestions.

However, I really suggest changing `uuid` library in `configlet`. Let me know if I'm missing anything.

ferhatelmas:
Let's merge then.

PR #898, opened 2017-10-15
Implement exercise collatz-conjecture

magic003:
Implement exercise collatz-conjecture.

### DIFF
--- a/exercises/collatz-conjecture/collatz_conjecture_test.go
+++ b/exercises/collatz-conjecture/collatz_conjecture_test.go
@@ -4,8 +4,6 @@ import (
 	"testing"
 )
 
-const targetTestVersion = 1
-
 var testCases = []struct {
 	description string
 	input       int
@@ -48,12 +46,6 @@ var testCases = []struct {
 	},
 }
 
-func TestTestVersion(t *testing.T) {
-	if testVersion != targetTestVersion {
-		t.Fatalf("Found testVersion = %v, want %v", testVersion, targetTestVersion)
-	}
-}
-
 func TestCollatzConjecture(t *testing.T) {
 	for _, testCase := range testCases {
 		steps, err := CollatzConjecture(testCase.input)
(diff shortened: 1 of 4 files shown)


tleen:
Thanks for starting on this @magic003!

magic003:
Hi @tleen, no problem. It is ready to review now.

exercises/collatz-conjecture/collatz_conjecture_test.go
```diff
@@ -0,0 +1,81 @@
...
+		expectError: true,
+	},
+	{
+		description: "negative value is an error",
+		input:       -15,
+		expectError: true,
+	},
+}
+
+func TestTestVersion(t *testing.T) {
```
leenipper: on branch nextercism, use of testVersion has been dropped.

exercises/collatz-conjecture/collatz_conjecture_test.go
```diff
@@ -0,0 +1,81 @@
...
+func TestCollatzConjecture(t *testing.T) {
+	for _, testCase := range testCases {
+		steps, err := CollatzConjecture(testCase.input)
+		if testCase.expectError {
+			if err == nil {
+				t.Fatalf("FAIL: %s\n\tCollatzConjecture(%v) expected an error, got %v",
+					testCase.description, testCase.input, steps)
+			}
+		} else {
+			if steps != testCase.expected {
```
leenipper: It is unlikely in this exercise, but it is standard handling to test for err != nil when no error was expected and if true call t.Fatalf.

exercises/collatz-conjecture/collatz_conjecture_test.go
```diff
@@ -0,0 +1,77 @@
...
+				t.Fatalf("FAIL: %s\n\tCollatzConjecture(%v) expected %v, got %v",
+					testCase.description, testCase.input, testCase.expected, steps)
+			}
+			if err != nil {
```
leenipper: Suggest moving the test for err != nil before the test on steps != testCase.expected.
The reason is that one should test for err before the normal return value. Generally in Go, the return value does not necessarily have a useful value if an error is returned.  It may seem a bit strict here, but it is good to establish that way of thinking about testing for error, and for the benefit of anyone reading the test code as well.

magic003:
Updated. Could you take another look? @leenipper

leenipper:
Looks good.

PR #903, opened 2017-10-19
isbn-verifier: add new exercise

creaaa:
Dear tleen,

Hello, I'm Masa.
I implemented a new exercise, so could you review and consider to merge this PR?
I'm happy to fix them if there's any defect.

Sincerely,
Masa

### DIFF
--- a/config.json
+++ b/config.json
@@ -17,7 +17,7 @@
       "difficulty": 1,
       "slug": "two-fer",
       "topics": [
-        "control_flow_conditionals",
+        "control-flow-(conditionals)",
         "strings"
       ],
       "unlocked_by": null,
@@ -29,7 +29,7 @@
       "slug": "leap",
       "topics": [
         "booleans",
-        "control_flow_conditionals",
+        "control-flow-(conditionals)",
         "integers",
         "logic"
       ],
@@ -43,7 +43,7 @@
       "topics": [
         "equality",
         "mathematics",
-        "text_formatting",
+        "text-formatting",
         "time"
       ],
       "unlocked_by": null,
@@ -64,7 +64,7 @@
       "difficulty": 1,
       "slug": "hamming",
       "topics": [
-        "control_flow_loops",
+        "control-flow-(loops)",
         "equality",
         "strings"
       ],
@@ -76,7 +76,7 @@
       "difficulty": 1,
       "slug": "raindrops",
       "topics": [
-        "control_flow_conditionals",
+        "control-flow-(conditionals)",
         "filtering",
         "strings"
       ],
@@ -98,7 +98,7 @@
       "difficulty": 3,
       "slug": "acronym",
       "topics": [
-        "regular_expressions",
+        "regular-expressions",
         "strings",
         "transforming"
       ],
@@ -110,7 +110,7 @@
       "difficulty": 2,
       "slug": "pangram",
       "topics": [
-        "control_flow_loops",
+        "control-flow-(loops)",
         "strings"
       ],
       "unlocked_by": "raindrops",
@@ -121,7 +121,7 @@
       "difficulty": 2,
       "slug": "bob",
       "topics": [
-        "control_flow_conditionals",
+        "control-flow-(conditionals)",
         "strings"
       ],
       "unlocked_by": "raindrops",
@@ -133,7 +133,7 @@
       "slug": "triangle",
       "topics": [
         "booleans",
-        "control_flow_conditionals",
+        "control-flow-(conditionals)",
         "logic"
       ],
       "unlocked_by": null,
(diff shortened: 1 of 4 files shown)


creaaa:
Dear tleen,

thanks for the polite review.
Oh, as you say I made `cases_test.go` by hand...
And I added UUID generated `$ configlet uuid`. Is my way wrong?

As you suggested me, I'll fix the first two and then send a PR again. Thank you so much.

Sincerely,
Masa

tleen:
@creaaa you did the UUID right, there is an issue with configlet but you can keep using it for now. `cases_test.go` is part of a whole sequence of using a generator to make Go test data from a [common repository's](https://github.com/exercism/problem-specifications/tree/master/exercises/isbn-verifier) test data file: `canonical-data.json`.

It can get complicated. There is no need to do it as part of this PR. What you should do instead is just fold that test data directly into the test file. We can always add the generator later. It would be odd to have that `cases_test.go` file without a generator.

tleen:
@creaaa this usually works better if you checkout the `nexterism` branch first and then do the changes. This is still carrying forward a bunch of unrelated commits.

config.json
```diff
@@ -1056,4 +1064,4 @@
   ],
   "ignore_pattern": "example(?!.*test)",
   "language": "Go"
-}
+}
```
ferhatelmas: Can we put it back ?

config.json
```diff
@@ -1024,6 +1024,14 @@
       "uuid": "8d5aaf47-7503-4797-85ed-d35a8755cfd4"
     },
     {
+      "difficulty": 2,
+      "slug": "isbn-verifier",
+      "topics": [
+        "arrays"
+      ],
+      "uuid": "6b93edd9-0de4-4c80-6419-68b2f8232b91c3425fa"
+    },
```
ferhatelmas: Since it's not core, we need to set a `unlocked_by`. @tleen your suggestion please ?

exercises/isbn-verifier/isbn_verifier.go
```diff
@@ -0,0 +1,7 @@
+package isbn
+
+const testVersion = 1
```
ferhatelmas: We can safely remove `testVersion` and also in test file and relevant test.

exercises/isbn-verifier/isbn_verifier_test.go
```diff
@@ -0,0 +1,23 @@
...
+	if testVersion != targetTestVersion {
+		t.Fatalf("Found testVersion = %v, want %v", testVersion, targetTestVersion)
+	}
+}
+
+func TestIsVaildISBN(t *testing.T) {
+	for _, test := range testCases {
+		observed := IsVaildISBN(test.isbn)
+		if observed != test.expected {
+			t.Fatalf("IsValidISBN(%s) = %s, want %s (%s)",
```
ferhatelmas: `%t` for booleans ?

exercises/isbn-verifier/example.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func strToAry(isbn string) []int {
+	var result []int
+	for _, char := range isbn {
+		if char == 'X' {
+			result = append(result, 10)
+		} else {
+			s := string(char)
+			i, _ := strconv.Atoi(s)
```
ferhatelmas: There is an assumption here, it returns default (0) on error. 

IMHO, we should handle it to return early or ensure value is zero.

config.json
tleen: `difference-of-squares`?

creaaa:
Dear ferhatelmas, tleen,

Thanks for the polite review. I'm sorry I'm not familiar with OSS process...
I'll fix the files you show me again and send them again.
Moreover, it causes conflict in `config.json`. Would you show me what I should do to solve this?

Thanks.

config.json
ferhatelmas: Thanks!

ferhatelmas:
@creaaa No worries. :heart: That's the way to learn.

Try to rebase your branch (`work`) on top of `nextercism` and in the end there should be only two commits `8fcf3d285a3aef0135e2ab3bda6eb7408fd5511b` and `8916ec7f888d0cdf216f5a2005e44239a4dee024` for review (you can squash them if you want)

creaaa:
Dear community,

Hello, I'm Masa, who opened this PR.
Sorry for delaying to fix this PR.
May I ask something about 2 issues below?

1.
I fixed files related to this PR, you can see them [here](https://github.com/creaaa/go/commit/71b2b9b3f560385a4547312e7dfaa8367ac5e0e6).
In it, I made a new branch `nextercism`, then fixed some files in that branch.
So I'd like to let you review them, but I can't designate this branch using GitHub's UI.



As you can see, I can't select `from` part, while I can `into`.

What I want to do is let you review `creaaa/go:nextercism` branch in this PR, but I can't.
Is that possible to show me how to solve this?

2.
the problem about `config.json`.
I can't avoid mingling unnecessary change due to my inappropriate file handling.
I added a comment  [here](https://github.com/creaaa/go/commit/71b2b9b3f560385a4547312e7dfaa8367ac5e0e6#diff-06b2d3b23dce96e1619d2b53d6c947ec), Could you tell me how to overcome this?

Sorry for any inconvenience.

tleen:
most people find the best way to solve this problem is to close the existing PR. Checkout `nextercism` branch and apply the new exercise there. It will not include these changed files.

tleen:
Closing this in favor of #906.

PR #905, opened 2017-10-19
sublist: add exercise with test generator

ferhatelmas:


exercises/sublist/sublist_test.go
```diff
@@ -0,0 +1,23 @@
+package sublist
+
+import (
+	"reflect"
+	"testing"
+)
+
+func TestSublist(t *testing.T) {
+	for _, tc := range testCases {
+		if actual := Sublist(tc.listOne, tc.listTwo); !reflect.DeepEqual(actual, tc.expected) {
```
tleen: For my curiosity: why reflect.DeepEqual here and not "=="? Does it have something to do with Relation as a type alias?

ferhatelmas: Nope, :eagle: eyes. It's just copy paste from a generic case. Changed it to `!=`.

PR #908, opened 2017-10-23
Normalize the config file to minimize noise later

kytrinyx:
We will be scripting a number of changes to the track configuration.
This runs configlet fmt, which ensures that any further changes can be minimized.

See https://github.com/exercism/meta/issues/95


tleen:
If this is for `nextercism` should we be doing this on the `nextercism` branch?

kytrinyx:
There's a discussion happening in exercism/meta#95 about how this normalization degrades the readability of the configs.

That discussion has some good suggestions for improvements to the serializer.
I'd appreciate it if you would weigh in on that discussion with any suggestions for improving how we handle configs, consistency, and normalization.

Meanwhile, I'm going to close this and wait for that discussion to play out.

I'm sorry about the noise, and any extra work that I've caused you.

tleen:
Will do.

Nothing to apologize for! Script-based normalization ends up being a time-saver and is appreciated! :+1:

This was more about weirdness in our own track.

PR #913, opened 2017-10-25
space-age: add exercise with test generator

ferhatelmas:


exercises/space-age/.meta/gen.go
```diff
@@ -0,0 +1,56 @@
+package main
+
+import (
+	"log"
+	"text/template"
+
+	"../../../gen"
+	space "github.com/exercism/go/exercises/space-age"
+)
```
leenipper: Recommend this change:
```
-       space "github.com/exercism/go/exercises/space-age"
+       space "../../space-age"
```
I think it best to keep the path to the dependent package relative, just as the path to the 'gen' package is relative.

ferhatelmas: Done.

config.json
```diff
@@ -1151,7 +1163,6 @@
     }
   ],
   "foregone": [
-    "space-age",
```
ferhatelmas: Why was it foregone ?

tleen: Unknown, looks like it was a while ago. @soniakeys do you remember why it was ignored in 92cfaeb99ae95e12ab0c1fbc6f673321aad4b1a0?

PR #938, opened 2017-11-01
allergies: add test generator

ferhatelmas:


ferhatelmas:
@tleen thanks for update.

PR #948, opened 2017-11-11
run-length-encoding: create test case generator

leenipper:

Update test program to use generated test case arrays.
Put FAIL and PASS in test result output.

For #605.


leenipper:
Thanks @ferhatelmas

PR #957, opened 2017-11-12
binary-search: add test generator

ferhatelmas:
For #605.

### DIFF
--- a/exercises/binary-search/binary_search_test.go
+++ b/exercises/binary-search/binary_search_test.go
@@ -1,193 +1,52 @@
-// Go has binary search in the standard library.  Let's reimplement
-// sort.SearchInts with the same API as documented in the standard library
-// at http://golang.org/pkg/sort/#Search.  Note that there are some differences
-// with the exercise README.
-//
-// * If there are duplicate values of the key you are searching for, you can't
-// just stop at the first one you find; you must find the first occurrence in
-// the slice.
-//
-// * There is no special "not found" indication.  If the search key is not
-// present, SearchInts returns the index of the first value greater than the
-// search key.  If the key is greater than all values in the slice, SearchInts
-// returns the length of the slice.
-//
-// * You can assume the slice is sorted in increasing order.  There is no need
-// to check that it is sorted.  (That would wreck the performance.)
-//
-// Try it on your own without peeking at the standard library code.
-
 package binarysearch
 
 import (
+	"fmt"
 	"math/rand"
-	"sort"
 	"testing"
 )
 
-var testData = []struct {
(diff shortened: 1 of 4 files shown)


ferhatelmas:
I removed message function since it's not in the canonical data.

For duplicated values, I asked it in the [problem-specifications#1003](https://github.com/exercism/problem-specifications/issues/1003). Let's hold on for consensus.

For missing keys, I propose to return `-1` as in canonical data but we might want to stick to `sort.SearchInts` API. If you agree too, I will update benchmarks. What do you think ?

leenipper:
I think -1 is fine for return value for missing keys for the exercise.

ferhatelmas:
@leenipper this is ready for review.

exercises/binary-search/binary_search_test.go
```diff
@@ -1,193 +1,52 @@
...
 func TestSearchInts(t *testing.T) {
-	for _, test := range testData {
-		if !sort.IntsAreSorted(test.slice) {
-			t.Skip("Invalid test data")
-		}
+	for _, test := range testCases {
 		if x := SearchInts(test.slice, test.key); x != test.x {
-			t.Fatalf("SearchInts(%v, %d) = %d, want %d",
+			t.Fatalf("FAIL: SearchInts(%#v, %d) = %d, want %d",
 				test.slice, test.key, x, test.x)
```
leenipper: Since you have a  test.description, perhaps include it in first line of output, with SearchInts ... on subsequent line?

exercises/binary-search/binary_search_test.go
```diff
@@ -1,193 +1,52 @@
...
-func TestMessage(t *testing.T) {
-	for _, test := range testData {
-		if !sort.IntsAreSorted(test.slice) {
-			t.Skip("Invalid test data")
-		}
-		if res := Message(test.slice, test.key); res != test.ref {
-			t.Fatalf("Message(%v, %d) =\n%q\nwant:\n%q",
-				test.slice, test.key, res, test.ref)
-		}
+		t.Logf("SUCCESS: SearchInts(%#v, %d) = %d", test.slice, test.key, test.x)
```
leenipper: And perhaps let the SUCCESS output simply show test.description.

exercises/binary-search/binary_search_test.go
ferhatelmas: Thanks, updated so.

exercises/binary-search/binary_search_test.go
ferhatelmas: Updated.

PR #958, opened 2017-11-14
collatz-conjecture: add generator

leenipper:
Make use of zero-value of expectError bool.

For #605.

### DIFF
--- a/exercises/collatz-conjecture/.meta/gen.go
+++ b/exercises/collatz-conjecture/.meta/gen.go
@@ -1,0 +1,72 @@
+package main
+
+import (
+	"log"
+	"text/template"
+
+	"../../../gen"
+)
+
+func main() {
+	t, err := template.New("").Parse(tmpl)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


exercises/collatz-conjecture/.meta/gen.go
```diff
@@ -0,0 +1,72 @@
...
+
+// determineExpected examines an .Expected interface{} object and determines
+// whether a test case is valid(bool) and has an answer or expects an error.
+// returning valid and answer.
+func determineExpected(expected interface{}) (bool, int) {
+	ans, ok := expected.(float64)
+	if ok {
+		return ok, int(ans)
+	}
+	return false, 0
```
ferhatelmas: Without if, `return ok, int(ans)` might suffice but as you wish to keep for clarity.

leenipper: I'll keep it.  Thanks @ferhatelmas

PR #959, opened 2017-11-15
grains: add generator

leenipper:
Make use of zero-value of expectError bool.

The test case value for the 'Total()' is retained in TestTotal()
since there wasn't an array of cases in the canonical-data.json
for the 'total' property, and there is only one valid return value.

For #605.

### DIFF
--- a/exercises/grains/.meta/gen.go
+++ b/exercises/grains/.meta/gen.go
@@ -1,0 +1,103 @@
+package main
+
+import (
+	"encoding/json"
+	"log"
+	"strings"
+	"text/template"
+
+	"../../../gen"
+)
+
+func main() {
+	t, err := template.New("").Parse(tmpl)
+	if err != nil {
+		log.Fatal(err)
+	}
+	var j js
+	if err := gen.Gen("grains", &j, t); err != nil {
+		log.Fatal(err)
+	}
+}
+
+// The JSON structure we expect to be able to unmarshal into
+type js struct {
+	Groups []testGroup `json:"Cases"`
+}
+
+type testGroup struct {
+	Description string
+	Cases       []json.RawMessage `property:"RAW"`
+	SquareCases []SquareCase      `property:"square"`
+	// Note: canonical-data.json has a element of "cases"
+	// which includes a test for property 'total', but it is ignored here,
+	// since "expected" is a single known value.
+}
+
+type SquareCase struct {
+	Description string
+	Input       int
+	Expected    interface{}
+}
+
+func (c SquareCase) Valid() bool {
+	valid, _ := determineExpected(c.Expected)
+	return valid
+}
+
+func (c SquareCase) Answer() uint64 {
+	_, answer := determineExpected(c.Expected)
+	return answer
+}
+
+func (c SquareCase) EditedDescription() string {
+	// Go doesn't raise exceptions, so replace the wording in .Description.
+	return strings.Replace(c.Description, "raises an exception", "returns an error", 1)
+}
+
+// determineExpected examines an .Expected interface{} object and determines
+// whether a test case is valid(bool) and has an answer or expects an error.
+// returning valid and answer.
+func determineExpected(expected interface{}) (bool, uint64) {
+	ans, ok := expected.(float64)
+	if ok {
+		if ans == float64(-1) {
+			return false, 0
+		}
+		return ok, uint64(ans)
+	}
+	return false, 0
+}
+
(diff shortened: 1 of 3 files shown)


exercises/grains/.meta/gen.go
```diff
@@ -0,0 +1,103 @@
...
+// determineExpected examines an .Expected interface{} object and determines
+// whether a test case is valid(bool) and has an answer or expects an error.
+// returning valid and answer.
+func determineExpected(expected interface{}) (bool, uint64) {
+	ans, ok := expected.(float64)
+	if ok {
+		if ans == float64(-1) {
+			return false, 0
+		}
+		return ok, uint64(ans)
```
tleen: For my own curiousity, why not return `true` instead of `ok` here? The ok val is a reuse of the ok val for a different expression which is (very slightly) confusing and it has to be true to get here anyway.

exercises/grains/.meta/gen.go
```diff
@@ -0,0 +1,103 @@
...
+// The JSON structure we expect to be able to unmarshal into
+type js struct {
+	Groups []testGroup `json:"Cases"`
+}
+
+type testGroup struct {
+	Description string
+	Cases       []json.RawMessage `property:"RAW"`
+	SquareCases []SquareCase      `property:"square"`
+	// Note: canonical-data.json has a element of "cases"
```
tleen: Have you or @ferhatelmas encountered an other canonical-data files with a cases.cases structure like this one? It seems odd. Wonder if there may be a better way to represent the data in that file we could suggest upstream.

exercises/grains/.meta/gen.go
```diff
@@ -0,0 +1,103 @@
...
+	// since "expected" is a single known value.
+}
+
+type SquareCase struct {
+	Description string
+	Input       int
+	Expected    interface{}
+}
+
+func (c SquareCase) Valid() bool {
```
tleen: Perhaps `HasAnswer` is a more descriptive term for this? It took me a second to realize it was the error check for Answer :smile:. Not having an answer is a valid test case, just that it is looking for an error rather than an answer value? It makes that `if` check in the template below straightforward.

exercises/grains/.meta/gen.go
```diff
@@ -0,0 +1,103 @@
...
+	return valid
+}
+
+func (c SquareCase) Answer() uint64 {
+	_, answer := determineExpected(c.Expected)
+	return answer
+}
+
+func (c SquareCase) EditedDescription() string {
+	// Go doesn't raise exceptions, so replace the wording in .Description.
```
tleen: Do you think this is worth bringing up in problem-specifications? Exception seems to be the narrower term here, while Error is the more general. If anything exception based languages should be rewriting these descriptions if necessary. I've always considered Exceptions to be a sub-term for Errors in generic conversation, but not the other way around.

exercises/grains/.meta/gen.go
leenipper: The are several which have nested cases.cases but only one property value; these are not a problem. Others also have differing property values, which make it a little harder to handle. Two exercises, custom-set and variable-length-quantity have generators that use a PropertyMatch technique. Four others use the RAW special handling with classifyByProperty (allergies, run-length-encoding, clock, and grains). A different structuring in canonical data could help to generate separate test case arrays for different property values. A change would require a retrofit for current generators however.

exercises/grains/.meta/gen.go
leenipper: I agree, HasAnswer is more descriptive.

exercises/grains/.meta/gen.go
leenipper: I was actually surprised to see the exception wording in the canonical data file.  I thought it was very tilted toward languages that use exceptions. Error is a better term to use there.

exercises/grains/.meta/gen.go
leenipper: Yes, agree.

leenipper:
Thanks @tleen.

exercises/grains/.meta/gen.go
ferhatelmas: When I first saw, I was shocked but then checked json schema in problem-specification and it was valid. Thus, even if it's harder to handle, thought that was the decision and tried to live with.

exercises/grains/.meta/gen.go
ferhatelmas: Agreed here.

PR #960, opened 2017-11-15
space-age: do not force rounding of solutions

glikson:
The expected age values are specified (in respective canonical-data.json) with 2-digit precision (e.g., 31.69), while the actual result is a float.
Instead of forcing the solution to perform an arbitrary rounding, the test code verifies the result with certain precision (currently same 2-digit, for compatibility - can be eventually added to canonical-data.json itself).
Aligned with the conclusion of the discussion in exercism/problem-specifications#880

### DIFF
--- a/exercises/space-age/space_age_test.go
+++ b/exercises/space-age/space_age_test.go
@@ -1,12 +1,14 @@
 package space
 
 import (
+	"math"
 	"testing"
 )
 
 func TestAge(t *testing.T) {
+	const precision = 0.01
 	for _, tc := range testCases {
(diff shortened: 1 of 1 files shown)


tleen:
This is funny. I just did the exercise an hour or so ago and was incredibly frustrated about the need to round! Good on you @glikson!

glikson:
@tleen same here - was frustrating enough to actually go and fix it :-)

ferhatelmas:
@glikson @tleen I think it's a good exercise since there is no round function in standard library. :smile: 

Btw, coming in [1.10](https://github.com/golang/go/issues/20100) but not powerful like python's to be very useful in this exercise.

PR #961, opened 2017-11-16
acronym: add a test case for space pre/suffixed strings

dolanor:
Thanks to the comment here: http://exercism.io/submissions/dc3f4dff7b4b4fd8b5f425f2363621f0

### DIFF
--- a/exercises/acronym/cases_test.go
+++ b/exercises/acronym/cases_test.go
@@ -34,4 +34,8 @@ var stringTestCases = []acronymTest{
 		input:    "Complementary metal-oxide semiconductor",
 		expected: "CMOS",
 	},
+	{
+		input:    " Unusable   acronym over the years ",
+		expected: "UAOTY",
+	},
 }


exercises/acronym/cases_test.go
```diff
@@ -34,4 +34,8 @@ var stringTestCases = []acronymTest{
 		input:    "Complementary metal-oxide semiconductor",
 		expected: "CMOS",
 	},
+	{
+		input:    " Unusable   acronym over the years ",
+		expected: "UAOTY",
+	},
```
ferhatelmas: This is a generated test. Its input is [here](https://github.com/exercism/problem-specifications/blob/master/exercises/acronym/canonical-data.json). 

So if it is a good idea to add this test case, it should happen on this parent repo.

tleen:
Yep, our test case generation is a little complex but does make sense in the grand scheme of things. Like @ferhatelmas said, your best bet is to modify the global repository. This way all the Exercism tracks that use these tests will be able to use the new test. Everybody wins! :smile:

It's great to be submitting patches based on exercise feedback. It is the best way to discover edge cases. Thanks for doing it.

dolanor:
Ok, got it! Thank you. I created another pull request on the parent project.
Pretty nice setup!

PR #962, opened 2017-11-16
sieve: add generator

leenipper:

Update example solution to pass test which includes limit as prime.

For #605.


exercises/sieve/.meta/gen.go
```diff
@@ -0,0 +1,48 @@
...
+
+type OneCase struct {
+	Description string
+	Limit       int
+	Expected    []int
+}
+
+// The JSON structure we expect to be able to unmarshal into
+type js struct {
+	Cases []OneCase
```
ferhatelmas: Might inline since we're not adding methods to type, etc.

leenipper: ok.

PR #968, opened 2017-12-04
bob: update generated tests, example and readme

ferhatelmas:


ferhatelmas:
It's interesting that we can't comment directly to review comment, instead to create another comment like this.

:+1: for yelling.

PR #971, opened 2017-12-18
reverse-string: add exercise with generator and property testing

### LINKED ISSUE #950: reverse-string: implement exercise
[Specification](https://github.com/exercism/problem-specifications/tree/master/exercises/reverse-string)

ferhatelmas:
- fixes #950

### DIFF
--- a/exercises/reverse-string/example.go
+++ b/exercises/reverse-string/example.go
@@ -1,0 +1,10 @@
+package reverse
+
+// String reverses given string
+func String(s string) string {
+	runes := []rune(s)
+	for i, j := 0, len(runes)-1; i < j; i, j = i+1, j-1 {
+		runes[i], runes[j] = runes[j], runes[i]
+	}
+	return string(runes)
+}
--- a/exercises/reverse-string/reverse_string_test.go
+++ b/exercises/reverse-string/reverse_string_test.go
@@ -1,0 +1,25 @@
+package reverse
+
+import (
+	"testing"
+	"testing/quick"
+)
+
+func TestReverse(t *testing.T) {
+	for _, testCase := range testCases {
+		if res := String(testCase.input); res != testCase.expected {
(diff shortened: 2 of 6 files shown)


exercises/reverse-string/reverse_string_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"testing"
+	"testing/quick"
+)
+
+func TestReverse(t *testing.T) {
+	for _, testCase := range testCases {
+		if res := String(testCase.input); res != testCase.expected {
+			t.Fatalf("FAIL: %s(%s)\nExpeced: %q\nActual: %q",
```
leenipper: Expeced => Expected

exercises/reverse-string/example.go
```diff
@@ -0,0 +1,10 @@
+package reverse
+
+// String reverses given string
+func String(s string) string {
```
leenipper: would have preferred
```
func Reverse(s string) string {
```
for naming sake.  But since package name is reverse, the call is reverse.String(x) which does read well in itself, perhaps better than reverse.Reverse(x).  Optional.

exercises/reverse-string/reverse_string_test.go
ferhatelmas: Thanks fixed.

exercises/reverse-string/example.go
ferhatelmas: I preferred `String` due package naming and you picked this. Let's wait for @tleen to break even case :wink:

tleen: It gets a little awkward with reverse.Reverse and since you could technically reverse an int or something using the type name as the function name seems ok to me.

ferhatelmas: Thanks, that was similar to the flow of my logic at first so merging.

PR #974, opened 2017-12-20
reverse-string: Implement Benchmark Test

### LINKED ISSUE #973: reverse-string: Implement Benchmark Test
The current version of the tests for the `reverse-string` exercise don't include benchmark tests.

Given the numerous different ways a solution to this exercise could be implemented, I felt like a benchmark test would be appropriate here.

Link to tests: [reverse_string_test.go](https://github.com/exercism/go/blob/master/exercises/reverse-string/reverse_string_test.go)

billglover:
Given the numerous different ways a solution to this exercise could be implemented, I felt like a benchmark test would be appropriate here.

This fixes #973.

### DIFF
--- a/exercises/reverse-string/reverse_string_test.go
+++ b/exercises/reverse-string/reverse_string_test.go
@@ -25,8 +25,8 @@ func TestReverseOfReverse(t *testing.T) {
 }
 
 func BenchmarkReverse(b *testing.B) {
-	for _, test := range testCases {
-		for i := 0; i < b.N; i++ {
+	for i := 0; i < b.N; i++ {
+		for _, test := range testCases {
 			String(test.input)
 		}
 	}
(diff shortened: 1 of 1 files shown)


exercises/reverse-string/reverse_string_test.go
```diff
@@ -23,3 +23,11 @@ func TestReverseOfReverse(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func BenchmarkReverse(b *testing.B) {
+	for _, test := range testCases {
+		for i := 0; i < b.N; i++ {
```
ferhatelmas: I think we should switch order for loops because one of the implementations can be optimized for shorter strings and case might be short then as a result, it might run more and vice versa.

billglover: I hadn't considered that. So essentially, include all the test cases inside the benchmark loop to avoid optimising for a specific implementation. I'll do that now.

PR #977, opened 2017-12-22
reverse-string: rename String to ReverseString to make more obvious

ilmanzo:
renamed "String" function to ReverseString in order to make it more descriptive


ferhatelmas:
I feel I am against it since it reads like `reverse.ReverseString`.

ilmanzo:
I agree with you ; but it feels a little weird for the user who have to implement a function named "String" ... Maybe we can find a better naming for the package

PR #984, opened 2017-12-23
ledger: fix sorting in example

ferhatelmas:


ferhatelmas:
@leenipper Yes, it needs a test. I will update it but maybe in problem specifications since it's lacking canonical data.

PR #1001, opened 2018-01-05
simple-linked-list: add exercise

### LINKED ISSUE #931: simple-linked-list: implement exercise
[Specification](https://github.com/exercism/problem-specifications/tree/master/exercises/simple-linked-list)

ngochai94:
Resolves #931

### DIFF
--- a/exercises/simple-linked-list/example.go
+++ b/exercises/simple-linked-list/example.go
@@ -1,0 +1,74 @@
+package linkedlist
+
+import "errors"
+
+type element struct {
+	data int
+	next *element
+}
+
+type List struct {
+	head *element
+	size int
+}
+
+func New(array []int) *List {
+	var res List
+	for _, elem := range array {
+		res.Push(elem)
+	}
+	return &res
+}
+
+func (list *List) Size() int {
+	return list.size
+}
+
+func (list *List) Push(elem int) {
+	newNode := element{elem, nil}
+	if list.head == nil {
+		list.head = &newNode
+	} else {
+		node := list.head
+		for node.next != nil {
+			node = node.next
+		}
+		node.next = &newNode
+	}
+	list.size++
+}
+
+func (list *List) Pop() (int, error) {
+	var elem int
+	if list.size == 0 {
+		return 0, errors.New("Cannot call pop on empty list")
+	} else if list.size == 1 {
+		elem = list.head.data
+		list.head = nil
+	} else {
+		node := list.head
+		for node.next.next != nil {
+			node = node.next
+		}
+		elem = node.next.data
+		node.next = nil
+	}
+	list.size--
+	return elem, nil
+}
+
+func (list *List) Array() []int {
+	res := []int{}
+	for node := list.head; node != nil; node = node.next {
+		res = append(res, node.data)
+	}
+	return res
+}
+
+func (list *List) Reverse() *List {
+	array, size := list.Array(), list.size
+	for i := 0; i < size/2; i++ {
+		array[i], array[size-1-i] = array[size-1-i], array[i]
+	}
+	return New(array)
+}
--- a/exercises/simple-linked-list/linked_list_test.go
+++ b/exercises/simple-linked-list/linked_list_test.go
@@ -1,0 +1,210 @@
+// API:
+//
+// type Element struct {
+//  data int
+//  next *Element
+// }
+//
+// type List struct {
+//  head *Element
+//  size int
+// }
+//
+// func New([]int) *List
+// func (*List) Size() int
+// func (*List) Push(int)
+// func (*List) Pop() (int, error)
+// func (*List) Array() []int
+// func (*List) Reverse() *List
+
+package linkedlist
+
+import (
+	"reflect"
+	"testing"
+)
+
+var array1To10 = []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}
+
+func TestEmptyListHasSizeZero(t *testing.T) {
+	list := New([]int{})
+	if size := list.Size(); size != 0 {
+		t.Fatalf("Size of empty list: %d, expected: %d", size, 0)
+	}
+	list = New(nil)
+	if size := list.Size(); size != 0 {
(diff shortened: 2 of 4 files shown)


exercises/simple-linked-list/example.go
```diff
@@ -0,0 +1,74 @@
+package linkedlist
+
+import "fmt"
+
+type Element struct {
```
ferhatelmas: do we need to export ?

exercises/simple-linked-list/example.go
```diff
@@ -0,0 +1,74 @@
...
+	next *Element
+}
+
+type List struct {
+	head *Element
+	size int
+}
+
+func New(array []int) *List {
+	res := List{nil, 0}
```
ferhatelmas: defaults are fine so `var res List`

exercises/simple-linked-list/example.go
```diff
@@ -0,0 +1,74 @@
...
+			node = node.next
+		}
+		node.next = &newNode
+	}
+	list.size++
+}
+
+func (list *List) Pop() error {
+	if list.size == 0 {
+		return fmt.Errorf("Cannot call pop on empty list")
```
ferhatelmas: no formatting directive so `errors.New` directly

exercises/simple-linked-list/example.go
```diff
@@ -0,0 +1,74 @@
...
+		node.next = nil
+	}
+	list.size--
+	return nil
+}
+
+func (list *List) Array() []int {
+	res := []int{}
+	node := list.head
+	for node != nil {
```
ferhatelmas: `for node := list.head; node != nil; node = node.next {`

exercises/simple-linked-list/example.go
```diff
@@ -0,0 +1,74 @@
...
+		res = append(res, node.data)
+		node = node.next
+	}
+	return res
+}
+
+func (list *List) Reverse() *List {
+	// not the best algo but it works XD
```
ferhatelmas: let's remove it since we know it's an example.

exercises/simple-linked-list/example.go
```diff
@@ -0,0 +1,74 @@
...
+		node := list.head
+		for node.next != nil {
+			node = node.next
+		}
+		node.next = &newNode
+	}
+	list.size++
+}
+
+func (list *List) Pop() error {
```
ferhatelmas: might be useful to return popped value

exercises/simple-linked-list/linked_list_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+import (
+	"reflect"
+	"testing"
+)
+
+var array1To10 = []int{1, 2, 3, 4, 5, 6, 7, 8, 9, 10}
+
+func TestEmptyListHasSizeZero(t *testing.T) {
+	list := New([]int{})
```
ferhatelmas: how about `New(nil)` too ?

exercises/simple-linked-list/linked_list_test.go
```diff
@@ -0,0 +1,191 @@
...
+	size := list.Size()
+	if size != 0 {
```
ferhatelmas: `if size := list.Size(); size != 0 {`

exercises/simple-linked-list/linked_list_test.go
```diff
@@ -0,0 +1,191 @@
...
+		t.Errorf("Reverse empty list: %v, want empty list", reversed)
+	}
+}
+
+func TestReverseNonEmptyList(t *testing.T) {
+	list := New([]int{1, 2, 3})
+	actual := list.Reverse().Array()
+	expected := []int{3, 2, 1}
+	if !reflect.DeepEqual(actual, expected) {
+		t.Errorf("Reverse non empty list: %v, want %v", actual, expected)
```
ferhatelmas: Since there is one case, it doesn't matter here but we prefer `Fatalf`. Could you please use it ?

exercises/simple-linked-list/example.go
ngochai94: right, good point 👍

exercises/simple-linked-list/example.go
ngochai94: sure, I will change this and update the tests

exercises/simple-linked-list/example.go
ngochai94: right, but I'm wondering why `errors.New` is preferred over `fmt.Errorf`? For me, I always use `fmt.Errorf` in case of later I might need to format the error message and save 1 import

exercises/simple-linked-list/example.go
ngochai94: 👍

exercises/simple-linked-list/linked_list_test.go
ngochai94: should we expect the solution to handle a nil input? In that case, should I add tests for all empty-list tests but with nil input?

exercises/simple-linked-list/linked_list_test.go
ngochai94: 👍

exercises/simple-linked-list/linked_list_test.go
ferhatelmas: I would expect it to behave similar since we didn't supply any element from the point of linked list.

ngochai94:
Thanks @ferhatelmas for reviewing! I have made changes accordingly. Also, I have a question which might not be related to this PR, but I sometimes saw CI fails temporarily such as with `git clone` or fetching `configlet`. Is there a way that we could trigger the CI rebuild?

ferhatelmas:
@ngochai94 push will trigger it. There is a restart button but I'm not sure if you can see it since you don't have write access to repo.

ngochai94:
Thanks @ferhatelmas . Yes, I don't have access to the rebuild button. I'm wondering if we could setup a webhook to trigger the CI rebuild upon a fixed PR comment, such as `Retest this please`

PR #1002, opened 2018-01-05
tree-building: add test cases for duplicate nodes

tbrisker:
Input with duplicate nodes should not be valid even if the nodes themselves are. 
Added one case for a generic duplicate node and another for a duplicate root node, to prevent comparison against zero value (either of ID or of Record) when checking for dupes.


ferhatelmas:
@tbrisker thanks :heart:

PR #1003, opened 2018-01-05
tree-building: rm stub with full solution

ferhatelmas:
not full but it's not a stub


exercises/tree-building/tree_building.go
```diff
@@ -1,107 +0,0 @@
...
-	} else {
-		for i := 0; i < len(n.Children); i++ {
-			err = chk(n.Children[i], m)
-			if err != nil {
-				return
-			}
-		}
-		return
-	}
-}
```
leenipper: I wondered why this stub was there originally, so I checked the README.md, and found this sentence:

> Your job will be to refactor a working but slow and ugly piece of code that implements the tree building logic for highly abstracted records.

That sentence originates in the description.md from problem-specifications.
So, should we simply keep it, or delete it and update the README.md?  The latter would make the Go exercise take a different slant than other language tracks whose focus is a refactor.

ferhatelmas: Yes, let's update readme too.

tleen:
My comment got a little lost as a pending review. Unsure if it was seen. But I am under the impression that the bad code is part of the exercise and refactoring it is the purpose.

leenipper:
Which way are we going with this exercise... shall we keep the stub since it is a re-factor exercise, or proceed to update the readme to remove the refactor focus and remove the stub?  If it's the former, we can close this PR without a merge.

tleen:
I'm of the opinion if we have to completely override [the original README](https://github.com/exercism/problem-specifications/blob/master/exercises/tree-building/description.md) from `problem-specifications` then it should probably be its own independent exercise. So I would have to go with no-merge.

leenipper:
> no-merge.

I'm thinking the same way.

ferhatelmas:
OK :dancer:

PR #1011, opened 2018-01-08
acronym: add benchmarking

strangeman:


exercises/acronym/acronym_test.go
```diff
@@ -12,3 +12,11 @@ func TestAcronym(t *testing.T) {
 		}
 	}
 }
+
+func BenchmarkAcronym(b *testing.B) {
+	for _, test := range stringTestCases {
+		for i := 0; i < b.N; i++ {
```
ferhatelmas: I mean this should be outer.

strangeman: Thank you for the suggestion! I fixed this. But I see this order in many other tests:
https://github.com/exercism/go/blob/master/exercises/anagram/anagram_test.go#L40
https://github.com/exercism/go/blob/master/exercises/accumulate/accumulate_test.go#L40
https://github.com/exercism/go/blob/master/exercises/bob/bob_test.go#L20

ferhatelmas: Then, we need to update them too. Do you want to send a PR for all ?

strangeman: @ferhatelmas yeah, I'll take care of them.

ferhatelmas: Thanks :smile:

PR #1012, opened 2018-01-09
exercises: fix loop ordering in benchmarks

strangeman:
Initial conversation: https://github.com/exercism/go/pull/1011/files/0af38ee4b324bee573ba1debc174cfde5bde6717#r160101420

### DIFF
--- a/exercises/accumulate/accumulate_test.go
+++ b/exercises/accumulate/accumulate_test.go
@@ -37,10 +37,10 @@ func TestAccumulate(t *testing.T) {
 
 func BenchmarkAccumulate(b *testing.B) {
 	b.StopTimer()
-	for _, test := range tests {
+	for i := 0; i < b.N; i++ {
 		b.StartTimer()
 
-		for i := 0; i < b.N; i++ {
+		for _, test := range tests {
 			Accumulate(test.given, test.converter)
 		}
 
--- a/exercises/accumulate/accumulate_test.go
+++ b/exercises/accumulate/accumulate_test.go
@@ -36,14 +36,11 @@ func TestAccumulate(t *testing.T) {
 }
 
 func BenchmarkAccumulate(b *testing.B) {
-	b.StopTimer()
 	for i := 0; i < b.N; i++ {
-		b.StartTimer()
 
 		for _, test := range tests {
 			Accumulate(test.given, test.converter)
 		}
 
-		b.StopTimer()
 	}
 }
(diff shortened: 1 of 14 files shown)


exercises/accumulate/accumulate_test.go
```diff
@@ -37,10 +37,10 @@ func TestAccumulate(t *testing.T) {
 
 func BenchmarkAccumulate(b *testing.B) {
 	b.StopTimer()
-	for _, test := range tests {
+	for i := 0; i < b.N; i++ {
 		b.StartTimer()
 
-		for i := 0; i < b.N; i++ {
+		for _, test := range tests {
```
ngochai94: Since we want to get benchmark for `Accumulate` function call, should `StartTimer` and `StopTimer` surround each function call rather the entire tests?

tleen: I wouldn't think so since the test ranging is consistent and would not add a variable amount of time to each benchmark? These are rough relative measures? I could be incorrect.

tleen: I fairly sure the timer manipulation is used to remove noticeable overhead from a test routine. Like if some heavy initialization or resetting was necessary between the benched function calls.

leenipper: In benchmarks where no "overhead code" is necessary, as in this one and others, it is perfectly fine to omit all the StopTimer and StartTimer calls.such as [prime-factors](https://github.com/exercism/go/blob/master/exercises/prime-factors/prime_factors_test.go#L22-L28). But it is ok as it stands.

exercises/allergies/allergies_test.go
```diff
@@ -42,11 +42,11 @@ func TestAllergicTo(t *testing.T) {
...
-		for _, response := range test.expected {
+	for i := 0; i < b.N; i++ {
+		for _, test := range allergicToTests {
 			b.StartTimer()
 
-			for i := 0; i < b.N; i++ {
+			for _, response := range test.expected {
 				AllergicTo(test.score, response.substance)
 			}
 			b.StopTimer()
```
leenipper: When benchmarking the AllergicTo, I've discovered that its best to omit all the b.StopTimer, b.StartTimer calls; otherwise the test runs for about two minutes on my system. The extra overhead of actually doing the timer manipulation made a huge difference with the loop index changes.  Did you also see that behavior?

strangeman: @leenipper yeah, same problem.

exercises/accumulate/accumulate_test.go
strangeman: I'm not sure about timers in most of the cases. I think we can omit them in almost every test.

ferhatelmas: I am also in the wagon to remove here since there is no big startup cost.

strangeman:
I removed timers for code, where no additional job doing.

PR #1013, opened 2018-01-12
grep: add exercise with generator

### LINKED ISSUE #922: grep: implement exercise
[Specification](https://github.com/exercism/problem-specifications/tree/master/exercises/grep)

leenipper:
Resolves #922

### DIFF
--- a/config.json
+++ b/config.json
@@ -1213,8 +1213,8 @@
       "slug": "grep",
       "topics": [
         "algorithms",
-        "conditionals",
         "arrays",
+        "conditionals",
         "searching",
         "strings"
       ],
--- a/config.json
+++ b/config.json
@@ -1208,6 +1208,20 @@
       "uuid": "6b313720-104a-46c2-8290-4b4af121101f "
     },
     {
+      "core": false,
+      "difficulty": 4,
+      "slug": "grep",
+      "topics": [
+        "algorithms",
+        "conditionals",
+        "arrays",
+        "searching",
+        "strings"
+      ],
+      "unlocked_by": "hamming",
+      "uuid": "ec5dab72-dd96-4dce-8886-aff52a6eb090"
+    },
+    {
       "deprecated": true,
       "slug": "binary",
       "uuid": "bf2dcbc4-266f-4cad-9cd7-76d446360a5d"
--- a/exercises/grep/.meta/gen.go
+++ b/exercises/grep/.meta/gen.go
@@ -13,7 +13,6 @@ func main() {
 		"fileContentData": FileContentData,
 	})
 	t, err := t.Parse(tmpl)
-	//t, err := template.New("").Parse(tmpl)
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/exercises/grep/.meta/gen.go
+++ b/exercises/grep/.meta/gen.go
@@ -1,0 +1,77 @@
+package main
+
+import (
+	"log"
+	"strings"
+	"text/template"
+
+	"../../../gen"
+)
+
+func main() {
+	t := template.New("").Funcs(template.FuncMap{
+		"fileContentData": FileContentData,
+	})
+	t, err := t.Parse(tmpl)
+	//t, err := template.New("").Parse(tmpl)
+	if err != nil {
+		log.Fatal(err)
+	}
+	var j js
+	if err := gen.Gen("grep", &j, t); err != nil {
+		log.Fatal(err)
+	}
+}
+
+// The JSON structure we expect to be able to unmarshal into
+type js struct {
+	Comments []string
+	Cases    []struct {
+		Description string
+		Cases       []struct {
+			Description string
+			Pattern     string
+			Flags       []string
+			Files       []string
+			Expected    []string
+		}
+	}
+}
+
+func FileContentData(comments []string) []string {
+	var start int
+	for i, c := range comments {
+		if strings.Contains(c, "contents are listed below:") {
+			start = i + 1
+			break
+		}
+	}
+	return comments[start:]
+}
+
+// template applied to above data structure generates the Go test cases
+var tmpl = `package grep
+
+{{.Header}}
+
+var fileContentData = []string{ {{range $line := fileContentData .J.Comments}}{{printf "\n%q," $line}}{{end}}
+}
+
+var testCases = []struct {
+	description string
+	pattern     string
+	flags       []string
+	files       []string
+	expected    []string
+}{
+{{range .J.Cases}}
+	{{range .Cases}}{
+		description: {{printf "%q" .Description}},
+		pattern: {{printf "%q" .Pattern}},
+		flags: {{printf "%#v" .Flags}},
+		files: {{printf "%#v" .Files}},
+		expected: {{printf "%#v" .Expected}},
+},
+{{end}}{{end}}
+}
+`
--- a/exercises/grep/.meta/hints.md
+++ b/exercises/grep/.meta/hints.md
@@ -1,0 +1,13 @@
+## Implementation
+
+Define a single Go func, GrepFiles, which accepts a pattern string,
+a slice of flags which are strings, and a slice of filename strings.
+GrepFiles should return a slice of strings of the output for
+the given flags and filenames.
+
+Use the following signature for func GrepFiles:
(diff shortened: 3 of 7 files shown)


ferhatelmas:
@leenipper FYI, will review this tonight

config.json
```diff
@@ -1207,6 +1207,20 @@
...
       "uuid": "6b313720-104a-46c2-8290-4b4af121101f "
     },
+    {
+      "core": false,
+      "difficulty": 4,
+      "slug": "grep",
+      "topics": [
+        "algorithms",
+        "conditionals",
+        "arrays",
```
ferhatelmas: are we sorting these?

exercises/grep/.meta/gen.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"../../../gen"
+)
+
+func main() {
+	t := template.New("").Funcs(template.FuncMap{
+		"fileContentData": FileContentData,
+	})
+	t, err := t.Parse(tmpl)
+	//t, err := template.New("").Parse(tmpl)
```
ferhatelmas: rm

exercises/grep/.meta/hints.md
```diff
@@ -0,0 +1,13 @@
...
+
+Define a single Go func, GrepFiles, which accepts a pattern string,
+a slice of flags which are strings, and a slice of filename strings.
+GrepFiles should return a slice of strings of the output for
+the given flags and filenames.
+
+Use the following signature for func GrepFiles:
+
+```
+func GrepFiles(pattern string, flags []string, files []string) []string {
```
ferhatelmas: `flags, files []string` maybe

exercises/grep/example.go
```diff
@@ -0,0 +1,102 @@
...
+	"strings"
+)
+
+type grepOptions struct {
+	n_line_numbers      bool
+	l_file_only         bool
+	i_any_case          bool
+	v_invert            bool
+	x_match_entire_line bool
+	prefix_filenames    bool
```
ferhatelmas: let's not prefer underscore since it's discouraged by golint

exercises/grep/example.go
```diff
@@ -0,0 +1,102 @@
+package grep
+
+import (
+	"bufio"
+	"os"
+	"strconv"
+	"strings"
+)
+
+type grepOptions struct {
```
ferhatelmas: since we're grep package, might be only options

exercises/grep/example.go
```diff
@@ -0,0 +1,102 @@
...
+	n_line_numbers      bool
+	l_file_only         bool
+	i_any_case          bool
+	v_invert            bool
+	x_match_entire_line bool
+	prefix_filenames    bool
+}
+
+// GrepFiles searches for pattern in the files given, using optional flags.
+func GrepFiles(pattern string, flags []string, files []string) []string {
```
ferhatelmas: `Files` might be a better name since it's `grep.GrepFiles` now

exercises/grep/example.go
```diff
@@ -0,0 +1,102 @@
...
+	output := []string{}
+	options := getOptions(flags, len(files))
+	for _, file := range files {
+		output = append(output, grepOneFile(pattern, file, options)...)
+	}
+	return output
+}
+
+// grepOneFile performs a search for pattern in the given file
+// and returns the proper output according to the options.
```
ferhatelmas: `s/proper/matched/` ?

exercises/grep/grep_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+		b.StartTimer()
+
+		for _, tc := range testCases {
+			GrepFiles(tc.pattern, tc.flags, tc.files)
+		}
+
+		b.StopTimer()
+	}
+
```
ferhatelmas: rm

exercises/grep/.meta/gen.go
leenipper: right.

exercises/grep/.meta/hints.md
leenipper: ok

exercises/grep/example.go
leenipper: ok

exercises/grep/example.go
leenipper: ok

exercises/grep/example.go
leenipper: Files seems not exactly what the function "does".
maybe grep.Search ?

exercises/grep/example.go
leenipper: s/proper output/output/
Since the output depends on the options and not necessarily matched lines.

exercises/grep/grep_test.go
leenipper: Are you saying remove both StartTimer and StopTimer ?  I am thinking to keep an initial StopTimer and immediately follow it with StartTimer to not include the test file creation overhead.

config.json
leenipper: yes.  I overlooked that.

leenipper:
A couple more ideas that I have which I wanted to run by you @ferhatelmas and @tleen which perhaps are more in line with Go practice relating to this exercise:
1. Use the flag package FlagSet type and the Parse function, since one can pass specific string flags[] to it.
2. For the grep output, pass an an io.Writer (the interface type) to Search. Then let the test program use a defined type that collects a slice of output lines as its io.Writer parameter.  The idea being, that with these changes, this little grep package becomes usable for actual work with a very simple main program.

These two ideas would stretch this exercise into learning about Go interface type io.Writer which is really one of the very cool things about Go.
The question - is it worth doing.  And if so, any further suggestions ?

exercises/grep/grep_test.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+		t.Logf("PASS: %s", tc.description)
+	}
+}
+
+func BenchmarkSearch(b *testing.B) {
+	files := createFiles(fileContentData)
+	defer deleteFiles(files)
+
+	b.StopTimer()
```
tleen: Do you think [b.ResetTimer()](https://golang.org/pkg/testing/#B.ResetTimer) would be better here?

leenipper: Oh, yeah. Since I need a stop and a reset, that makes sense.  Will push another commit.

leenipper: Looking at the code in testing/benchmark.go,  using ResetTimer is much better choice with my placement.   I should have had the StopTimer before the call to createFiles, but with the ResetTimer used instead of Start/Stop it will be fine.  Thanks!

tleen: I think involving io.Writer and the std lib flag package would be an interesting Go twist to the project. It may be nice at some point to show that you can build real command line apps with Go. If so this is the obvious exercise to do it. That would probably ramp up the difficulty level because we would be introducing a few new topics all at once. It would have to have a bunch of hints to walk users through the process. We could move it down the track a bit towards the end as a bonus. `[paasio](https://github.com/exercism/go/tree/master/exercises/paasio)` uses io.Writer too...

Overall it seems to be a worthy Go-ification of the exercise. Adds something new and helpful!

leenipper: Ok.  I'll work with it some.  And we'll see how it looks.
I agree with the bump of difficulty level, and more content in hints.md for the exercise "extension".

leenipper:
I suggest letting the grep exercise be first incorporated as it is right now.  After merging, I'll do a PR with [RFC] to consider the flag package and io.Writer use. I've got some potential changes prepared, but I would rather not push them into this PR right now.

exercises/grep/grep_test.go
```diff
@@ -0,0 +1,86 @@
...
+	files := createFiles(fileContentData)
+	defer deleteFiles(files)
+
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		for _, tc := range testCases {
+			Search(tc.pattern, tc.flags, tc.files)
+		}
+	}
+	b.StopTimer()
```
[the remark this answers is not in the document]
leenipper: It is unnecessary.  I'll remove it.

PR #1014, opened 2018-01-12
nucleotide-count: add generator and update example

ilmanzo:
reference issue #605


exercises/nucleotide-count/cases_test.go
```diff
@@ -0,0 +1,39 @@
...
+var testCases = []struct {
+	description   string
+	strand        string
+	expected      map[string]int
+	errorExpected bool
+}{
+	{
+		description: "empty strand",
+		strand:      "",
+		expected:    map[string]int{"A": 0, "C": 0, "G": 0, "T": 0},
```
leenipper: Would it be easy enough to use this expected type ?
> expected      map[byte]int

and then generate this as
> expected:    map[byte]int{'A': 0, 'C': 0, 'G': 0, 'T': 0},

A nucleotide is one single letter after all.  Perhaps too nit-picky, but it seems to match the exercise problem better.
What do you think ?

tleen: Seems reasonable to me. Always a fan of more content in hints.md as well.

ilmanzo:
Hi, I've changed the expected type to map[byte]int and added a hints.md ; the problem in keeping the original nucleotide_count.go is that it's incomplete, and it breaks test builds in Travis CI . Do you have some advice on that ?

exercises/nucleotide-count/example.go
```diff
@@ -5,35 +5,37 @@ import (
...
+func (dna DNA) Counts() (result Histogram, e error) {
 	var total int
-	h := Histogram{}
+	result = make(Histogram)
+
 	for i := range validNucleotides {
 		nucleotide := validNucleotides[i]
-		h[nucleotide], _ = dna.Count(nucleotide)
-		total += h[nucleotide]
+		result[nucleotide], _ = dna.Count(nucleotide)
```
leenipper: Perhaps replace
> result[nucleotide], _ = dna.Count(nucleotide)

with
> if result[nucleotide], e = dna.Count(nucleotide); e != nil {
>    return nil, e
> }

Then var total is unnecessary as well as the later if test using it; it also preserves the specific error from Count.

exercises/nucleotide-count/nucleotide_count.go
```diff
@@ -2,7 +2,7 @@ package dna
 
 // Histogram is a mapping from nucleotide to its count in given DNA.
 // Choose a suitable data type.
-type Histogram
+type Histogram
```
leenipper: Unnecessary white space change ?

exercises/nucleotide-count/example.go
ilmanzo: That would fail the test with invalid nucleotides, like "AGXXACT" in the test suite... Because the algorithm only iterates on valid ones "ACGT"

leenipper: Ah, I see now.

exercises/nucleotide-count/.meta/gen.go
```diff
@@ -0,0 +1,84 @@
...
+// SortedMapString collects key:values for a map in sorted order
+func (c OneCase) SortedMapString() string {
+	strs := make([]string, 0, len(c.Expected))
+	for k, v := range c.Expected {
+		switch t := v.(type) {
+		case float64:
+			strs = append(strs, `'`+k+`': `+strconv.FormatFloat(t, 'f', -1, 64))
+		default:
+		}
+
```
ferhatelmas: rm

leenipper: Yeah,
> default:

That line is unnecessary. Can remove it.

ilmanzo: ok .. I made it a little more useful :) 

in the meantime the problem specification json structure has changed, so I updated also the generator

PR #1018, opened 2018-01-14
book-store: use fixed point output

leenipper:
Update generator for latest canonical-data.json
which uses integer expected values instead of floating point.

Update exercise to use type int for Cost output.
Add hints.md for guidance.
Update README.md.

### DIFF
--- a/exercises/book-store/.meta/hints.md
+++ b/exercises/book-store/.meta/hints.md
@@ -1,0 +1,14 @@
+## Implementation
+
+Define a single Go func, Cost, which calculates the cost
+for a given list of books based on the defined discounts.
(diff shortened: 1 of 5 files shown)


exercises/book-store/.meta/hints.md
```diff
@@ -0,0 +1,14 @@
+## Implementation
+
+Define a single Go func, Cost, which calculates the cost
+for a given list of books based on the defined discounts.
+
+Use the following signature for func Cost:
+
+```
+func Cost(books[]int) int
```
ferhatelmas: `books []int` space might be better

PR #1026, opened 2018-01-19
exercises: update several generators for problem specifications input change

ilmanzo:
fix to generators for the new problem specification input format

### DIFF
--- a/exercises/anagram/cases_test.go
+++ b/exercises/anagram/cases_test.go
@@ -1,8 +1,8 @@
 package anagram
 
 // Source: exercism/problem-specifications
-// Commit: a0f7663 anagram: Update json for new "input" policy (#1036)
-// Problem Specifications Version: 1.1.0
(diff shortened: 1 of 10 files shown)


exercises/anagram/cases_test.go
```diff
@@ -20,23 +20,6 @@ var testCases = []struct {
...
-		expected: []string{
-			"tan"},
-	},
-	{
-		description: "does not detect false positives",
-		subject:     "galea",
-		candidates: []string{
-			"eagle"},
-		expected: []string{},
-	},
```
leenipper: When a generator update causes test cases to drop out, it is probably best to use a separate PR for that one exercise, and note in the commit log that test cases changed.  But not a problem. Just a note for future practice.

ilmanzo: ok! Thanks for pointing it out :)

PR #1042, opened 2018-01-22
parallel-letter-frequency: Add parallel vs concurrent info to README

### LINKED ISSUE #1023: Better local documentation for parallel-letter-freq
Clear up any confusion between the terms "parallel" and Go's use of the term "concurrency". 

See exercism/problem-specifications#1132

tleen:
An attempt to clear up some confusion as to why Go uses the term
"concurrent" where most languages use "parallel".

Resolves #1023

### DIFF
--- a/exercises/parallel-letter-frequency/.meta/hints.md
+++ b/exercises/parallel-letter-frequency/.meta/hints.md
@@ -1,0 +1,14 @@
+## Concurrency vs Parallelism
+
+You may notice that while this exercise is called _Parallel_ letter frequency
+you don't see the term "Parallel" used very often in Go. Gophers prefer to use
+the term **Concurrent** to describe the management of multiple independent
+processes. Although these terms are often used interchangeably Gophers like to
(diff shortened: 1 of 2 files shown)


tleen:
I'm usually in favor of more in the hints than less. Do you think we should go into concurrently in general? Add information for how to get started with the concurrency primitives in Go? IIRC it is the first time a user will have to use these in the nextercism track. It could be very confusing if they are on their own?

leenipper:
> get started with the concurrency primitives

I think I was wanting something like "Go enables concurrency via 'goroutines' which are started with the 'go'  keyword".  And where your hints mentioned 'program' I favored the use of 'goroutine' instead.  See if that will help.

tleen:
:laugh: I think this one is a special case because of how tricky concurrency is as a general concept and how no other language handled it quite like Go.

tleen:
@leenipper @ferhatelmas I added some more info (especially your line @leenipper it is a good one!) and some links to resources. Am I missing any good resources?

PR #1044, opened 2018-01-23
rectangles: add exercise with generator

### LINKED ISSUE #929: rectangles: implement exercise
[Specification](https://github.com/exercism/problem-specifications/tree/master/exercises/rectangles)

ferhatelmas:
fixes #929

### DIFF
--- a/config.json
+++ b/config.json
@@ -1250,6 +1250,20 @@
       "uuid": "ec5dab72-dd96-4dce-8886-aff52a6eb090"
     },
     {
+      "core": false,
+      "difficulty": 4,
+      "slug": "rectangles",
+      "topics": [
+        "arrays",
+        "control-flow (conditionals)",
+        "control-flow (loops)",
+        "matrices",
+        "strings"
+      ],
+      "unlocked_by": "tournament",
+      "uuid": "e7e207dc-bf8b-4bbf-b06e-edd674e09a06"
+    },
+    {
       "deprecated": true,
       "slug": "binary",
       "uuid": "bf2dcbc4-266f-4cad-9cd7-76d446360a5d"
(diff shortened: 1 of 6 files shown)


exercises/rectangles/example.go
```diff
@@ -0,0 +1,48 @@
...
+			prop = diagram[i][colIndex : colIndex+1]
+		} else {
+			prop = diagram[rowIndex][i : i+1]
+		}
+		if !strings.Contains(pattern, prop) {
+			return false
+		}
+	}
+	return true
+}
```
leenipper: I'm thinking that checkSide might be split into two functions, neither of which requires a call with an `ignore` parameter, since it is constant within the function to act like a control variable for path selection. It would make the code longer but perhaps more readable.  What do you think ?  Seems that one case is horizontal and one is vertical side checking, if I am understanding the code accurately.  The pattern might also be local to both the replacement functions as well.

config.json
```diff
@@ -1237,6 +1237,20 @@
...
+      "difficulty": 4,
+      "slug": "rectangles",
+      "topics": [
+        "control-flow (conditionals)",
+        "control-flow (loops)",
+        "arrays",
+        "matrices",
+        "strings"
+      ],
+      "unlocked_by": "matrix",
```
tleen: I would suggest the unlock for this be `tournament`. We have a few similar exercises there all around taking some ASCII representation of a thing and figuring that thing out.

exercises/rectangles/example.go
ferhatelmas: I think this is a great idea since I believe readability is the utmost important aspect of example. Done so.

ferhatelmas:
It seems in the mean time input policy change is applied so updated it, too.

PR #1045, opened 2018-01-24
two-bucket: add exercise with generator

leenipper:


exercises/two-bucket/example.go
```diff
@@ -0,0 +1,169 @@
...
+	sizeBucketTwo,
+	goalAmount int,
+	startBucket string) (goalBucket string, numSteps, otherBucketLevel int) {
+
+	p := problem{
+		capacity: [2]int{sizeBucketOne, sizeBucketTwo},
+		goal:     goalAmount,
+	}
+	var s state
+	if startBucket == "one" {
```
tleen: Think it may make sense to have the start bucket as a package level constant in a type alias? Then we could provide BucketOne/BucketTwo consts. Or should the function allow for the possibility of a bad bucket indicator name? Like if someone passed in startBucket = "three". It's a small thing. Really this LGTM.

leenipper: > start bucket as a package level constant in a type alias?

We can do the constants without needing the type alias, right?  Just use const without an explicit type given. Maybe I'm not completely understanding the suggestion.

IIRC type alias appeared in 1.9.  And we still test  with 1.8.x in travis-ci.

>  possibility of a bad bucket indicator name?

The canonical data didn't contain any test cases with errors expected. So I didn't make the function return an error.  But I agree that if an error can happen, it should be caught & handled. And it is more in keeping with Go philosophy to handle any possible errors.
I can code up extra cases for errors. I'm thinking to also handle errors for bucketSize/goalAmount too (<= 0).  Probably should add a sentence in the hints about handling errors as well as adjust the signature.

tleen: If you think it is a good idea. I like having complex operations return errors, so we should probably have one case for it. I like the idea of a bad bucket name being a possibility but we may be straying too far from the problem-specification.

tleen: Ah I was thinking named types not type aliases!

```go
type Bucket string

const (
	BucketOne Bucket = "one"
	BucketTwo        = "two"
)

func printBucket(b Bucket) {
	fmt.Println("Bucket " + b)
}

func main() {
	fmt.Println("hello world")
	printBucket(BucketOne)
	printBucket(BucketTwo)
	printBucket("three") // still works
	// doesnt work - compile error
	// fourStr := "four"
	// printBucket(fourStr)
}
```

but it wouldn't really stop people from using a direct string :( but it would give us a way to limit the inputs to BucketOne and BucketTwo by providing them. Sort of thinking out loud here. May be too much for an exercise...

leenipper: I didn't see this comment before I merged. Timing was very close!
Are you referring to the findGoal function being changed to return an error ?  It is complex. Will need some work to figure out a unsolvable test case.

tleen: I was thinking of high level errors, like using a non allowed bucket name or some such. Don't want to make the implementation too much of a bear for the users.

exercises/two-bucket/example.go
```diff
@@ -0,0 +1,198 @@
...
+func getPossibleSteps(p problem, s state) (list []step) {
+	for x := emptyOne; x <= pourTwoToOne; x++ {
+		if canPeformStep(p, s, x) {
+			list = append(list, x)
+		}
+	}
+	return list
+}
+
+func canPeformStep(p problem, s state, x step) bool {
```
tleen: Just noticed a missing 'r' in preform.

exercises/two-bucket/example.go
tleen: It is perfectly fine as is I wouldn't worry about it.

PR #1046, opened 2018-01-25
clock: extend exercise with Subtract operation

leenipper:
The problem-specifications canonical-data.json was updated
to define separate subtract property test cases and also indicated
that the exercise should include a Subtract function.

Update generator to recognize "subtract" property separately
and output subtractTests; also update for the new "input" schema.

Update example solution to implement Subtract operation.

Update test program to call Subtract operation
in TestSubtractMinutes and add BenchmarkSubtractMinutes.

### DIFF
--- a/exercises/clock/.meta/gen.go
+++ b/exercises/clock/.meta/gen.go
@@ -29,19 +29,32 @@ type testGroup struct {
 	Description string
 	Cases       []json.RawMessage `property:"RAW"`
 	CreateCases []struct {
-		Description  string
-		Hour, Minute int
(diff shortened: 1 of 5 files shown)


leenipper:
> example_[name]_test.go

I completely missed that!

> reflect the Subtract API change in it

Will do.

leenipper:
@tleen Thanks for catching the example inconsistency.  I learned another cool thing about 'go test' !

tleen:
Yeah! It's interesting isn't it!

PR #1051, opened 2018-01-28
two-bucket: corrections plus bonus

leenipper:
First commit corrects some minor mistakes in test program and example solution.
Second commit extends the exercise with a bonus part that returns an error
when a solution cannot be found for given inputs.


tleen:
@leenipper I have these on my radar just cant get to them till perhaps the weekend.

PR #1067, opened 2018-02-06
Fix test result for luhn test case

mbiagetti:


exercises/luhn/cases_test.go
```diff
@@ -27,7 +27,7 @@ var testCases = []struct {
 	{
 		"a simple valid SIN that becomes invalid if reversed",
 		"59",
-		true,
+		false,
```
leenipper: Note that the test case description does say "simple valid SIN"; for that reason the test case .ok value should remain as true.

The "invalid if reversed" phrase is intended to catch an incorrect implementation.

mbiagetti: Sorry my mistakes

PR #1069, opened 2018-02-06
reverse-string: add UTF-8 multibyte test case

### LINKED ISSUE #1068: reverse-string: improve tests with multibyte character strings?
I don't know if it fits the original problem spirit, but given that Go supports UTF-8 it could be nice to insert a multibyte string example in the test suite.
I think that a simple string like the one used in the official website "Hello, 世界" could be a nice improvement.

leenipper:
Add reverseTestCase type in generator output to be
used in the test program.

Since file cases_test.go is generated from
problem-specifications/exercises/reverse-string/canonical-data.json,
add an additional test case for UTF-8 within the test program.

Resolves #1068

### DIFF
--- a/exercises/reverse-string/.meta/gen.go
+++ b/exercises/reverse-string/.meta/gen.go
@@ -40,11 +40,13 @@ var tmpl = `package reverse
 
 {{.Header}}
 
-var testCases = []struct {
+type reverseTestCase struct {
 	description	string
 	input		string
 	expected	string
-}{ {{range .J.Cases}}
+}
+
+var testCases = []reverseTestCase{ {{range .J.Cases}}
 {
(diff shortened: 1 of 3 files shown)


tleen:
I thought about that but not every language is as thorough as Go with multi-byte characters. I would have to think that putting those entries in the test suite would drastically change the nature of the exercise for some other tracks.

neganp:
I would agree. Having to deal with Unicode would make (for example) the C problems *much* more difficult.

ferhatelmas:
I totally agree with your point but still they can assign much higher difficulty than go.

ferhatelmas:
Done at [problem-specifications#1175](https://github.com/exercism/problem-specifications/issues/1175).

PR #1074, opened 2018-02-13
simple-linked-list: improve benchmark runtime

### LINKED ISSUE #1073: BenchmarkListSize for the Simple Linked List problem in Go hangs for a very long time
BenchmarkListSize for the Simple Linked List problem in Go hangs for a very long time.
The function it tries to benchmark ( list.Size() )is probably too fast (in my solution it just returns the value of list.size). That results in the very large number of iterations that are needed to acurately measure the time, 
but when we also use the b.StopTimer() and b.StartTimer() functions inside the bench loop, the total time needed is much much longer. 

Since the benchmark uses the same array for each list and we only want to benchmark the list.Size() function we might just create a single list before bench loop and use that list without stopping / starting the timer.

leenipper:
Running go test -bench on this exercise is taking
far too much time (several minutes).

Improve the benchmark routines by
doing the list creation outside the loop
and using ResetTimer; this applies to routines
    BenchmarkNewList,
    BenchmarkListSize,
    BenchmarkListToArray, and
    BenchmarkListReverse.

For BenchmarkListPush and BenchmarkListPop,
use an inner loop to measure 100 instances
of the respective Push or Pop operations.
This approach reduces the overhead runtime in relation
to the operations being measured.

Resolves #1073

### DIFF
--- a/exercises/simple-linked-list/linked_list_test.go
+++ b/exercises/simple-linked-list/linked_list_test.go
@@ -165,10 +165,9 @@ func BenchmarkNewList(b *testing.B) {
 }
 
 func BenchmarkListSize(b *testing.B) {
+	list := New(array1To10)
+	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-		b.StopTimer()
-		list := New(array1To10)
-		b.StartTimer()
 		_ = list.Size()
 	}
 }
@@ -176,35 +175,40 @@ func BenchmarkListSize(b *testing.B) {
 func BenchmarkListPush(b *testing.B) {
 	for i := 0; i < b.N; i++ {
 		b.StopTimer()
-		list := New(array1To10)
+		list := New([]int{})
 		b.StartTimer()
-		list.Push(11)
+		for k := 0; k < 100; k++ {
+			list.Push(k)
+		}
 	}
 }
 
 func BenchmarkListPop(b *testing.B) {
 	for i := 0; i < b.N; i++ {
 		b.StopTimer()
-		list := New(array1To10)
+		list := New([]int{})
+		for k := 0; k < 100; k++ {
+			list.Push(k)
+		}
 		b.StartTimer()
-		_, _ = list.Pop()
+		for k := 0; k < 100; k++ {
+			_, _ = list.Pop()
+		}
 	}
 }
 
 func BenchmarkListToArray(b *testing.B) {
+	list := New(array1To10)
+	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-		b.StopTimer()
-		list := New(array1To10)
-		b.StartTimer()
 		_ = list.Array()
 	}
 }
 
 func BenchmarkListReverse(b *testing.B) {
+	list := New(array1To10)
+	b.ResetTimer()
 	for i := 0; i < b.N; i++ {
-		b.StopTimer()
-		list := New(array1To10)
-		b.StartTimer()
 		_ = list.Reverse()
 	}
 }
--- a/exercises/simple-linked-list/linked_list_test.go
+++ b/exercises/simple-linked-list/linked_list_test.go
@@ -177,7 +177,7 @@ func BenchmarkListPush(b *testing.B) {
 		b.StopTimer()
 		list := New([]int{})
 		b.StartTimer()
-		for k := 0; k < 100; k++ {
+		for k := 0; k < 1000; k++ {
 			list.Push(k)
 		}
 	}
@@ -187,11 +187,11 @@ func BenchmarkListPop(b *testing.B) {
 	for i := 0; i < b.N; i++ {
 		b.StopTimer()
 		list := New([]int{})
-		for k := 0; k < 100; k++ {
+		for k := 0; k < 1000; k++ {
 			list.Push(k)
 		}
 		b.StartTimer()
-		for k := 0; k < 100; k++ {
+		for k := 0; k < 1000; k++ {
 			_, _ = list.Pop()
 		}
 	}


krisiasty:
I just run the new version of tests and benchmarks and the BenchmarkListSize is super fast (like less than a second) to finish, the BenchmarkListPush takes about dozen of seconds (which is acceptable), BenchmarkListPop took long time to run (like 2 minutes) and the rest were very quick.

```
 C:\code.go\src\exercism\go\simple-linked-list>go test -bench .
goos: windows
goarch: amd64
pkg: exercism/go/simple-linked-list
BenchmarkNewList-4               3000000               546 ns/op
BenchmarkListSize-4             2000000000               0.63 ns/op
BenchmarkListPush-4               200000              7712 ns/op
BenchmarkListPop-4               3000000               456 ns/op
BenchmarkListToArray-4           2000000               889 ns/op
BenchmarkListReverse-4           1000000              1009 ns/op
PASS
ok      exercism/go/simple-linked-list  157.794s
```

But when I run the Push and Pop benchmarks again the next time Pop finished in a couple of seconds where Push took much much longer (143 seconds). Depending on whether I run the benchmark with -bench or -benchmem flags the one or the other took significantly much more time than the other.
I belive these two benchmarks still need some improvement so they behave in more consistent way.

leenipper:
Thanks for the good feedback and suggestions @krisiasty .  I'll work with it a bit more.

leenipper:
My results were different in the total run time.
```
$ go test -bench .
goos: linux
goarch: amd64
BenchmarkNewList-4               2000000               574 ns/op
BenchmarkListSize-4             2000000000               0.36 ns/op
BenchmarkListPush-4               100000             12522 ns/op
BenchmarkListPop-4                200000              7852 ns/op
BenchmarkListToArray-4           3000000               511 ns/op
BenchmarkListReverse-4           2000000               892 ns/op
PASS
ok      _/home/lee/xgo/exercises/simple-linked-list       18.229s
```
After changing the Pop to use 500 iterations, near the same runtime:
```
$ go test -bench .
goos: linux
goarch: amd64
BenchmarkNewList-4               2000000               559 ns/op
BenchmarkListSize-4             2000000000               0.38 ns/op
BenchmarkListPush-4               200000             12125 ns/op
BenchmarkListPop-4                 10000            162664 ns/op
BenchmarkListToArray-4           5000000               478 ns/op
BenchmarkListReverse-4           2000000              1150 ns/op
PASS
ok      _/home/lee/xgo/exercises/simple-linked-list       18.447s
```
@tleen, could you give us another data point.  I am also using go 1.9.4, but on linux.

krisiasty:
@leenipper I guess you were benchmarking your own solution to the problem, which might be less performant and not as susceptible as other solutions. In your case I see the BenchmarkListPush and BenchmarkListPop takes only 100k and 200k iterations respectively, so the impact of stopping/starting timers is lower comparing to my case with 3M iterations for BenchmarkListPop

krisiasty:
Another set of benchmarks from Go 1.9.4 on Linux running on i7-4790 CPU @ 3.60GHz (4 core / 8 threads). This is with 100 ops as originally proposed in the patch.
``` go test -bench .
goos: linux
goarch: amd64
BenchmarkNewList-8               5000000               301 ns/op
BenchmarkListSize-8             2000000000               0.26 ns/op
BenchmarkListPush-8               500000              3103 ns/op
BenchmarkListPop-8               5000000               352 ns/op
BenchmarkListToArray-8           3000000               471 ns/op
BenchmarkListReverse-8           3000000               477 ns/op
PASS
ok      _/root/exercism.io      94.844s
```

The BenchmarkListPush benchmark run 7.287s and BenchmarkListPop run 81.485s

Another set - Go 1.9.4 on Linux running on server with dual Xeon E5-2660 v3 @ 2.60GHz (20 cores / 40 threads total)
```
 go test -bench .
goos: linux
goarch: amd64
BenchmarkNewList-40              3000000               411 ns/op
BenchmarkListSize-40            2000000000               0.34 ns/op
BenchmarkListPush-40              500000              3948 ns/op
BenchmarkListPop-40              3000000               472 ns/op
BenchmarkListToArray-40          2000000               643 ns/op
BenchmarkListReverse-40          2000000               650 ns/op
PASS
ok      _/root/ex       171.049s
```
Again, there are 3M iterations for BenchmarkListPop and it took 144.935s

I think the general rule should be to keep number of benchmark iterations under 100k range when using StopTimer / StartTimer inside main benchmark loop. 

Same server with 1000 push / pop operations:
```
go test -bench BenchmarkListPush
goos: linux
goarch: amd64
BenchmarkListPush-40               50000             36277 ns/op
PASS
ok      _/root/ex       4.179s

 go test -bench BenchmarkListPop
goos: linux
goarch: amd64
BenchmarkListPop-40       300000              3947 ns/op
PASS
ok      _/root/ex       21.954s

```

On the i7 machine the benchmarks with 1000 ops took 2.360s and 19.368s respectively.

It seems like increasing number of push / pop operations for both most affected benchmarks (BenchmarkListPush and BenchmarkListPop) to 1000 operations works well on wide range of CPUs and problem implementations. But to be sure we need more tests with others people problem solutions.

exercises/simple-linked-list/linked_list_test.go
```diff
@@ -165,46 +165,50 @@ func BenchmarkNewList(b *testing.B) {
...
 }
 
 func BenchmarkListPush(b *testing.B) {
 	for i := 0; i < b.N; i++ {
 		b.StopTimer()
-		list := New(array1To10)
+		list := New([]int{})
 		b.StartTimer()
-		list.Push(11)
+		for k := 0; k < 100; k++ {
```
krisiasty: For some faster solutions to the problem pushing only 100 elements doesn't help much. According to my own tests 1000 should be enough.

exercises/simple-linked-list/linked_list_test.go
```diff
@@ -165,46 +165,50 @@ func BenchmarkNewList(b *testing.B) {
...
+		}
 	}
 }
 
 func BenchmarkListPop(b *testing.B) {
 	for i := 0; i < b.N; i++ {
 		b.StopTimer()
-		list := New(array1To10)
+		list := New([]int{})
+		for k := 0; k < 100; k++ {
```
krisiasty: Same situation as for BenchmarkListPush - 100 push / pop operations is not enough. I suggest 1000.

tleen:
I'm on go1.10rc1 also Linux  w/ i7-4710MQ

```text
goos: linux
goarch: amd64
BenchmarkNewList-8       	 5000000	       348 ns/op
BenchmarkListSize-8      	2000000000	         0.30 ns/op
BenchmarkListPush-8      	  200000	     10729 ns/op
BenchmarkListPop-8       	  200000	      7635 ns/op
BenchmarkListToArray-8   	10000000	       192 ns/op
BenchmarkListReverse-8   	 3000000	       558 ns/op
PASS
```

PR #1079, opened 2018-02-16
added proverb implementation in Go

sunilgopinath:
Closed previous PR and opened a new one after comments

### DIFF
--- a/config.json
+++ b/config.json
@@ -1178,6 +1178,18 @@
       "uuid": "37821140-c0d0-4da8-8f50-47356705a615"
     },
     {
+      "core": false,
+      "difficulty": 4,
+      "slug": "proverb",
+      "topics": [
+        "arrays",
+        "loops",
+        "strings"
+      ],
+      "unlocked_by": "two-fer",
+      "uuid": "b6d05e8f-542b-4789-9cbc-7ba132a5f03d"
+    },
+    {
       "difficulty": 2,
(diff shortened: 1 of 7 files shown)


config.json
```diff
@@ -1177,6 +1177,18 @@
...
+    {
+      "core": false,
+      "difficulty": 4,
+      "slug": "proverb",
+      "topics": [
+        "arrays",
+        "loops",
+        "strings"
+      ],
+      "unlocked_by": "two-fer",
```
tleen: Good job here @sunilgopinath the only suggestion I would have is for the exercise unlock, seems like it falls into the general category of repeat a string with slight positional variations, which sits nicely in this subtree under [twelve-days](../tree/master/exercises/twelve-days):

```text
├─ twelve-days
│  ├─ house
│  ├─ food-chain
│  └─ beer-song
│
```

We can merge this as is and patch that change later. It is related to the next-gen exercism which is a later concern. I imagine we may revisit all the ordering before that anyhow.

sunilgopinath: That makes sense. I can update the ordering on another PR. I used the java repo as an example as I wasn't sure what to put there.

tleen: Yeah it can be tricky and sort of language dependent, your approach to guesstimating is as good as any :smile:.

tleen:
Thanks for doing this, it looks great.

Issue #1086, opened 2018-02-27
crypto-square implementation does not comply with problem description


eraserix:
The problem description of crypto-square expects that:

> Phrases that do not fill perfect rectangles will have n empty spaces. Those spaces should be distributed evenly, added to the end of the last n chunks.

The example in README.md has said spaces. But none of the testcases have it. The problem description should be changed or the implementation and the tests should be fixed.

PR #1087, opened 2018-02-27
crypto-square: add spaces if text does not fill rectangle

eraserix:
The exercise description inside the readme calls for padding with
trailing spaces on the last n chunks if the message does not perfectly
fill the calculated rectangle.  This commit fixes the testcases and the
example implementation.

Fixes: https://github.com/exercism/go/issues/1086

### DIFF
--- a/exercises/crypto-square/example.go
+++ b/exercises/crypto-square/example.go
@@ -18,9 +18,16 @@ func norm(r rune) rune {
 func Encode(pt string) string {
 	pt = strings.Map(norm, pt)
 	numCols := int(math.Ceil(math.Sqrt(float64(len(pt)))))
+	padding := numCols*(numCols-1) - len(pt)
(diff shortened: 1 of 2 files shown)


exercises/crypto-square/example.go
```diff
@@ -18,9 +18,16 @@ func norm(r rune) rune {
...
+	padding := numCols*(numCols-1) - len(pt)
+	if padding < 0 {
+		padding = numCols*numCols - len(pt)
+	}
 	cols := make([]string, numCols)
 	for i, r := range pt {
 		cols[i%numCols] += string(r)
 	}
+	for i := 0; i < padding; i++ {
+		cols[numCols-i-1] += string(' ')
```
leenipper: Perhaps just:
```
cols[numCols-i-1] +=  " "
```

PR #1092, opened 2018-03-13
two-fer: Remove extraneous go files.

waynr:
These files don't look like they serve a purpose since there is already a stub go file that is apparently used for this exercise.


waynr:
Whoops, I didn't notice that at least one example solution was necessary. I'm still pretty sure the other file I removed wasn't necessary thought.

exercises/two-fer/example_two_fer_test.go
```diff
@@ -1,20 +0,0 @@
...
-// compared to the data that comes after the "Output: " comment.
-//
-// Below the result of ShareWith() is passed to standard output
-// using fmt.Println, and this is compared against the expected output.
-// If they are equal, the test passes.
-func ExampleShareWith() {
-	h := ShareWith("")
-	fmt.Println(h)
-	// Output: One for you, one for me.
-}
```
leenipper: This is a testable example.  It was intentional, and not a mistake. See https://blog.golang.org/examples

waynr:
Ah, my bad I guess I kind of glossed over the comments in that file and assumed that this was just a leftover from some previous approach to organizing exercises. Thanks for the explanation!

PR #1096, opened 2018-03-16
Implement yacht for go

zeimusu:
Create an implementation of yacht for the go track.

* Readme, tests and example solution
* metadata and gen.go in .meta directory
* Change config.json to refer to yacht, using self-generated uuid. 
* Change the "poker" configuration to depend on yacht ("I can see this exercise being a precursor to poker since it has similar concepts without requiring the parsing of the cards." -- petertseng)


exercises/yacht/.meta/gen.go
```diff
@@ -0,0 +1,59 @@
...
+type OneCase struct {
+	Description string
+	Input       struct {
+		Dice     []int
+		Category string
+	}
+	Expected int
+}
+
+//func (c OneCase)
```
leenipper:
```
//func (c OneCase)
```
Can remove, just for cleanup sake.

exercises/yacht/cases_test.go
```diff
@@ -0,0 +1,169 @@
+package yacht
+
+// Source: exercism/problem-specifications
+// Commit: 4bb9c29 Update canonical-data.json
```
leenipper: Running the generator makes the // Commit: ... line come out different.  So might as well push a commit with the latest copy to get it in sync.

exercises/yacht/yacht_test.go
```diff
@@ -0,0 +1,25 @@
+package yacht
+
+import (
+	"testing"
+)
+
+func TestCost(t *testing.T) {
```
leenipper: Perhaps TestCost -> TestScore, which is the typical convention.

exercises/yacht/yacht_test.go
```diff
@@ -0,0 +1,25 @@
...
+		if testCase.expected != score {
+			t.Fatalf("FAIL: %s\nScore %v as %s. expected %d, got %d",
+				testCase.description, testCase.dice, testCase.category,
+				testCase.expected, score)
+		}
+		t.Logf("PASS: %s", testCase.description)
+	}
+}
+
+func BenchmarkCost(b *testing.B) {
```
leenipper: And BenchmarkCost -> BenchmarkScore as well.

exercises/yacht/cases_test.go
zeimusu: That's because my changes to problem-specifications were squashed before merging. My branch is now "18 commits ahead and one commit behind". 

What is the git invocation to get my fork realigned with the upstream? (I fixed manually by copying the commit message off of github, that is surely not the Right Way)

zeimusu:
Changes requested by Lee Nipper made.

config.json
```diff
@@ -1072,7 +1087,7 @@
         "sequences",
         "strings"
       ],
-      "unlocked_by": "tournament",
+      "unlocked_by": "yacht",
```
leenipper: This change makes logical sense, but we need to keep this one unlocked by tournament instead of yacht, since yacht is not a core exercise.  All the unlocked_by items must reference only core exercises.

zeimusu: Done, as requested.

exercises/yacht/cases_test.go
leenipper: Locate the older commit in your problem-specifications tree which is  also in the master branch on github.  Then just reset with the hard option using that commit id.

Likely something like:
```
git checkout master
git reset --hard 6a42552060affe0f7745d89d9f719df8f830653b
```
Then do a git fetch followed by git pull.

PR #1101, opened 2018-03-28
Update isogram test cases

ltran:
The longest German isogram is "Heizölrückstoßabdämpfung" (heating oil recoil dampening) with 24 letters.

### DIFF
--- a/exercises/isogram/cases_test.go
+++ b/exercises/isogram/cases_test.go
@@ -54,4 +54,9 @@ var testCases = []struct {
 		input:       "accentor",
 		expected:    false,
 	},
+	{
+		description: "longest German isogram",
+		input:       "Heizölrückstoßabdämpfung",
+		expected:    true,
+	},
 }


ltran:
ya let me add in a failing cases as well. I googled up isogram because of curiosity and found other examples in different languages.

for the usecase you mentioned about the double ss, would the double ss be considered one Rune? 
I tried lowering the ß  but it yield only ß 
https://play.golang.org/p/aBV71Rz7czQ

- [ ] Add failing test case
- [ ] Update readme to be more clear about the expected input

exercises/isogram/cases_test.go
```diff
@@ -54,4 +54,9 @@ var testCases = []struct {
 		input:       "accentor",
 		expected:    false,
 	},
+	{
+		description: "longest German isogram",
+		input:       "Heizölrückstoßabdämpfung",
+		expected:    true,
+	},
```
leenipper: Nice test case.  The cases_test.go file is generated by [isogram/.meta/gen.go](https://github.com/exercism/go/blob/master/exercises/isogram/.meta/gen.go) from [the canonical-data.json file](https://github.com/exercism/problem-specifications/blob/master/exercises/isogram/canonical-data.json) so I'm thinking it may be best to suggest the change in [that repository](https://github.com/exercism/problem-specifications).  If UTF-8 is permitted for the other language tracks it may be acceptable. If not, we could add another test case in  isogram_test.go.

ltran: ah thanks. ill look at that file and check out on eof the other languages to see

ltran:
Ah sorry I got distracted. I'll take a look at how to do it through the recommendations this weekend.

PR #1102, opened 2018-04-01
robot-name: Fix test cases.

waynr:
At first I thought I was such a good programmer when I got my solution working, but then as I looked at others' solutions that couldn't possibly be correct, I realized there is probably something wrong with the test cases. My fixes do the following:

* Add a helper method to `Robot` that lets every test function uniformly check for duplicate names in the same way; but it only performs this check if the name is expected to be unique.
* Adjusts the bounds on the loops in `TestCollision`. The only way to guarantee that the test actually tests for collisions is to exhaustively generate all possible names. The bound is calculated as `26  * 26 * 10 * 10 * 10` which is 676000 names. It's worth noting that any truly random name generator is going to generate an increasing number of collisions as it nears the total name capacity so it's probably worth changing this exercise to limit the time spent randomly searching for an exhaustive list of unique names.
* Adds some code at the bottom of `TestCollision` to validate that the robot name factory detects when the name space is filled. This is determined currently by attempting to recover from a panic; if no panic is detected, then the test should fail at that point. Alternatively, we could change the signature `(r *Robot) Name() string` to `(r *Robot) Name() (string, err)` and check for an error returned from that method instead; which could be done in the `getName` helper method I've added here. I thought I'd see what others think about this before going ahead and doing that.
* I've modified the documentation to reflect that the only way to run the bonus tests is to exclude the normal tests -- which makes sense if you think about it; currently the tests don't dictate any kind of factory factory reset which leaves the programmer with the somewhat unpalatable only recourse of having to use global program state to track which names have been generated already. This means that once we finish running `TestCollision` we can't run any of the other tests so we are out of luck if it happens to run before the non-bonus tests. I haven't been able to find a way to reset global program state in golang's testing library, but if anyone happens to know how to do that I would be happy to change this this PR.


waynr:
I guess there are PR tests on this repo. I wouldn't be surprised if my fixes cause the example implementation to break given that the way the tests were written before they weren't all that effective.

waynr:
Oh yeah, one solution to the global state problem is to have programmers also write a `(r *Robot) ResetFactoryNames()` method, which would allow test cases to properly clean up after themselves.

waynr:
....aaaand one more point about the `ResetFactoryNames` method; this would also allow a test case that checks for randomness by

exercises/robot-name/bonus_test.go
```diff
@@ -5,25 +5,24 @@ package robotname
...
-		}
-		m[n] = true
+	for i := len(seen); i <= maxNames-600000; i++ {
+		_ = New().getName(t, false)
 	}
+
 	// Test that names aren't recycled either.
 	r := New()
-	for i := 0; i < 10000; i++ {
+	for i := len(seen); i < maxNames; i++ {
```
leenipper: With the loop using maxNames to do a full uniqueness check, doing
```
go test -tags bonus -run TestCollisions
```
doesn't complete.  I gave up waiting after a couple minutes.
So I suppose the bonus_example.go algorithm isn't as efficient as it needs to be to complete with these changes. The bonus_example.go used with the test changes must be able to complete and pass the tests.

Have you found an algorithm with a reasonable bound on the runtime when doing a full uniqueness test for the bonus?  Or your suggestion to "limit the time spent randomly searching" could be used.

waynr:
Hey sorry for not responding -- I must have missed the email notification way back when. I'll take a look at this sometime in the next few days (just about to step out for dinner).

exercises/robot-name/bonus_test.go
waynr: FYI I'm about to dig up my solution to see how long it took--if I remember correctly it was fairly reasonable.

waynr:
@leenipper @sebito91 I've added my solution which passes all the fixed tests with a reasonable run time.

waynr:
@sebito91 you're running into the global state problem that I alluded to in one of my earlier comments. You can't just run `go test --tags=bonus` because this will run not only the bonus tests but the default untagged tests also; I'm not sure what determines the order in which these tests are run (maybe alphabetical order by testcase name?) but it seems like the current order has the standard untagged tests run after the bonus collision tests, which (without interface method that we have the user implement that resets the "factory" name database) means we will encounter name a name exhaustion panic when running the untagged tests.

The way to work around this is to specify only the collisions test case:

```

waynr:
I did update the docs to reflect that this now the bonus test needs to be run.

sebito91:
Hmm, thaks for that feedback. I'll take a look at that later today and get back to you. Apologies on the delays!

bitfield:
@sebito91: #1209 presumably affects this, so happy to pick it up if @waynr isn't available.

kytrinyx:
@bitfield Since it's been well over 4 months since we've heard back from @waynr I think having you take over is the right move.

waynr:
My understanding was someone just had to approve the work i already did.
What exactly is left to be done?

As an aside, if you don't want fly by contributors like me to forget about
a PR like this then don't wait 4 months to give feedback.

kytrinyx:
@waynr that's my bad. I misread your comment from August as meaning that you needed to update those tests, and I thought we were waiting for that to happen.

But clearly you did do that. So sorry.

@bitfield Would you mind rebasing this onto master and resolving the conflicts and getting it merged?

waynr:
@kytrinyx no problem, sorry for the somewhat irritable/short tone of my previous responses, I was distractedly emailing while driving. I just rebased and incorporated the recent return signature change for `func (r *Robot) Name() (string, error)` and confirmed that the tests still pass as documented. Having that error available makes it easier to implement the namespace exhaustion test case in the bonus test ;).

bitfield:
@waynr that's awesome.

bitfield:
Perhaps we could dispense entirely with the awkward invocation required to run the bonus test, by just making the bonus test part of the standard tests?

If you take a hard look at it, there isn't a whole lot to this exercise if you're not required to avoid collisions or namespace exhaustion. I always encourage my students to pass the bonus test, and guide them through avoiding the exhaustion problem.

kytrinyx:
@bitfield I don't have an opinion with respect to including or excluding the test by default. If we include the bonus test as part of the exercise proper, should we move it so that it's farther back, or does it make sense where it is?

bitfield:
When you say 'farther back', @kytrinyx , what do you mean? Move the exercise to later in the track, because it's now harder?

kytrinyx:
> When you say 'farther back', @kytrinyx , what do you mean? Move the exercise to later in the track, because it's now harder?

Yeah, sorry about that. That's exactly what I was trying to say!

bitfield:
Yes, it's a tricky one, isn't it? Is this harder than 'Tournament'? There's a lot of machinery to put together in the right way for 'Tournament'. Looking at the exercises further down, they're all pretty demanding (with the possible exception of 'Twelve Days').

On the other hand, it's certainly not easier than 'Parallel Letter Frequency`, because that introduces goroutines and channels. I think on balance it's still in the right place.

PR #1103, opened 2018-04-17
proposal for markdown exercise

ilmanzo:
maybe still need some work; also not very clear how to write the new entry in config.json "unlocked_by" ?

### DIFF
--- a/config.json
+++ b/config.json
@@ -1060,6 +1060,18 @@
       "unlocked_by": "tree-building",
       "uuid": "7aa53a27-4ff3-4891-809f-2af728eb55a0"
     },
+    {
+      "core": false,
+      "difficulty": 4,
+      "slug": "markdown",
+      "topics": [
+        "refactoring",
+        "strings",
+        "text_formatting"
+      ],
+      "unlocked_by": "tree-building",
+      "uuid": "a7469032-c4d0-4e15-aa0c-4e9f6588b4c8"
+    },
     { 
       "core": false,
       "difficulty": 4,
--- a/exercises/markdown/README.md
+++ b/exercises/markdown/README.md
@@ -1,0 +1,29 @@
+# Markdown
+
+Refactor a Markdown parser.
+
+The markdown exercise is a refactoring exercise. There is code that parses a given string with Markdown syntax and returns the associated HTML for that string. Even though this code is confusingly written and hard to follow, somehow it works and all the tests are passing! Your challenge is to re-write this code to make it easier to read and maintain while still making sure that all the tests keep passing.
+
+It would be helpful if you made notes of what you did in your refactoring in comments so reviewers can see that, but it isn't strictly necessary. The most important thing is to make the code better!
+
+## Running the tests
+
+To run the tests run the command `go test` from within the exercise directory.
+
+If the test suite contains benchmarks, you can run these with the `-bench`
+flag:
+
+    go test -bench .
+
+Keep in mind that each reviewer will run benchmarks on a different machine, with
+different specs, so the results from these benchmark tests may vary.
+
+## Further information
+
+For more detailed information about the Go track, including how to get help if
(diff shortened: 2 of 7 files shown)


config.json
```diff
@@ -1060,6 +1060,18 @@
...
+    {
+      "core": false,
+      "difficulty": 4,
+      "slug": "markdown",
+      "topics": [
+        "refactoring",
+        "strings",
+        "text_formatting"
+      ],
+      "unlocked_by": "tree-building",
```
leenipper: In answer to your question about the "unlocked_by" property, it must be another exercise which has "core": true.  For this exercise, "tree-building" does qualify since it is a core exercise.  Generally the unlocked_by exercise would have some intersecting "topics" property. The one common property between tree-building and markdown is "refactoring".  There might be a better core exercise for topics intersection.  See what you think.  The "core" and "unlocked_by" are items in preparation for next revision of exercism user interface which is still in the works.

exercises/markdown/README.md
```diff
@@ -0,0 +1,29 @@
+# Markdown
+
+Refactor a Markdown parser.
+
+The markdown exercise is a refactoring exercise. There is code that parses a given string with Markdown syntax and returns the associated HTML for that string. Even though this code is confusingly written and hard to follow, somehow it works and all the tests are passing! Your challenge is to re-write this code to make it easier to read and maintain while still making sure that all the tests keep passing.
+
```
leenipper: It looks like you might not have used the 'configlet' to generate the README.md. If you need a hint about this, use
```
./bin/fetch-configlet
./bin/configlet --help
./bin/configlet generate --help
```

ilmanzo:
I see.. how about "twelve-days" or "tournament" ?

exercises/markdown/example.go
```diff
@@ -0,0 +1,77 @@
...
+	s := reStrong.ReplaceAll([]byte(markdown), []byte("<strong>$2</strong>"))
+	reEm := regexp.MustCompile("(_)(.*)(_)")
+	s = reEm.ReplaceAll(s, []byte("<em>$2</em>"))
+	//now manage <li> and <hN>
+	var output bytes.Buffer
+	starcount := 0
+	hcount := 0
+	needtoClose := none
+	i := 0
+	for {
```
leenipper: Perhaps
```
for i < len(s)-1 {
```
then the later if / else logic won't be needed.

exercises/markdown/example.go
```diff
@@ -0,0 +1,77 @@
...
+			}
+			if needtoClose == star {
+				output.WriteString("</li>")
+			}
+
+		default:
+			output.WriteByte(s[i])
+
+		}
+		if i < len(s)-1 {
```
leenipper: This if / else sequence can drop after for loop condition change.

exercises/markdown/markdown_test.go
```diff
@@ -0,0 +1,24 @@
...
+	"fmt"
+	"testing"
+)
+
+func TestMarkdown(t *testing.T) {
+	for _, test := range testCases {
+		if html := Render(test.input); html != test.expected {
+			t.Fatalf("FAIL: Render(%q) = %q, want %q.", test.input, html, test.expected)
+		}
+		fmt.Printf("PASS: %s\n", test.description)
```
leenipper: Best to use testing package for all test output;  so:
```
-    fmt.Printf("PASS: %s\n", test.description)
+    t.Logf("PASS: %s\n", test.description)
```

exercises/markdown/example.go
```diff
@@ -0,0 +1,72 @@
...
+			}
+			if needtoClose == star {
+				output.WriteString("</li>")
+			}
+
+		default:
+			output.WriteByte(s[i])
+
+		}
+		i++
```
leenipper: Can drop the i++ here after 'for' loop change.

Issue #1104, opened 2018-05-06
[space-ace]: some edge case with NaN


exklamationmark:
Hi, I noticed that I can pass the test for [space-ace](https://github.com/exercism/go/tree/master/exercises/space-age) with this:

```go
package space

import (
        "math"
)

// Planet represents official planets in the Sol system.
type Planet string

const (
        Earth   Planet = "Earth"
        Mercury Planet = "Mercury"
        Venus   Planet = "Venus"
        Mars    Planet = "Mars"
        Jupiter Planet = "Jupiter"
        Saturn  Planet = "Saturn"
        Uranus  Planet = "Uranus"
        Neptune Planet = "Neptune"
)

// Age provides the age on a particular planet (in Earth years), given the age in seconds.
func Age(ageInSeconds float64, planet Planet) float64 {
        periodInEarthYear, exist := orbitalPeriodsInSecond[planet]
        if !exist {
                return math.NaN()
        }

        // NOTE: periodInEarthYear is assumed to be > 0
        ageInEarthYear := ageInSeconds / periodInEarthYear
        return ageInEarthYear
}

const earthPeriodInSeconds = 31557600.0

var (
        orbitalPeriodsInSecond = map[Planet]float64{
                Earth:   1.0 * earthPeriodInSeconds,
                // Mercury: 0.2408467 * earthPeriodInSeconds,
                // Venus:   0.61519726 * earthPeriodInSeconds,
                // Mars:    1.8808158 * earthPeriodInSeconds,
                // Jupiter: 11.862615 * earthPeriodInSeconds,
                // Saturn:  29.447498 * earthPeriodInSeconds,
                // Uranus:  84.016846 * earthPeriodInSeconds,
                // Neptune: 164.79132 * earthPeriodInSeconds,
        }
)
```

```
$ go test -v
=== RUN   TestAge
--- PASS: TestAge (0.00s)
        space_age_test.go:14: PASS: age on Earth
        space_age_test.go:14: PASS: age on Mercury
        space_age_test.go:14: PASS: age on Venus
        space_age_test.go:14: PASS: age on Mars
        space_age_test.go:14: PASS: age on Jupiter
        space_age_test.go:14: PASS: age on Saturn
        space_age_test.go:14: PASS: age on Uranus
        space_age_test.go:14: PASS: age on Neptune
PASS
ok      exercism/go/space-age   0.005s
```

Do you think it's worthwhile to handle cases like this for the exercise?

******

[...]

PR #1105, opened 2018-05-07
space-age: fix issue #1104

ilmanzo:
added a check for edge case NaN in the test


leenipper:
Thanks @ilmanzo . I think the change is sufficient for issue #1104.

Issue #924, opened 2017-10-30
linked-list: implement exercise


ferhatelmas:
[Specification](https://github.com/exercism/problem-specifications/tree/master/exercises/linked-list)

exklamationmark:
Looks interesting, I would like to work on this.

I have a few questions:

Since Go doesn't have generics/template, what would be better between:
- `NewLinkedList(v int)` --> new doubly linked list with 1 element (`v`)
- `NewLinkedList(v interface{})` -> `v` could be whatever, but no consistent type
- I don't think we want to go too far and have a generator for `NewLinkedListTypenameT(v typenameT)`, do we?

About the methods name:
- Do you think `push_back`, `pop_back`, `push_front`, `pop_front` will be better names (at least to me, since I also work with C++). In particular, `shift` and `unshift` is a bit strange for me.

leenipper:
> I would like to work on this.

Great!

I think either of these would be fine.
```
NewLinkedList(v int) --> new doubly linked list with 1 element (v)
NewLinkedList(v interface{}) -> v could be whatever, but no consistent type
```
I agree that it would be a bit far to use a generator for the exercise.

The push_front and pop_front (or rather PushFront / PopFront with Go conventional naming) seem more self-explanatory than shift and unshift. The exercise README.md can state specifically what the API looks like; usually that is done in a .meta/hints.md file.

exklamationmark:
> The push_front and pop_front (or rather PushFront / PopFront with Go conventional naming) seem more self-explanatory than shift and unshift.

Oops, they should be `PushFront`, `PopFront`, etc.

@leenipper: I added some tests + reference implementation in this [PR](https://github.com/exercism/go/pull/1107/files). Would you mind taking a look first before I fix them according to the CI's results?

PR #1107, opened 2018-05-13
Implement linked-list exercise

exklamationmark:
Issue: https://github.com/exercism/go/issues/924

This PR add an exercise where users implements a doubly linked list. Specifically:

- [x] Add tests for the doubly-linked list.
- [x] Provide a reference implementation that passes the test.
- [ ] Convert the tests to match <exercism.io>'s format.
- [ ] Add README.md and hints

### DIFF
--- a/exercises/linked-list/linked_list.go
+++ b/exercises/linked-list/linked_list.go
@@ -1,0 +1,203 @@
+package linkedlist
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+)
+
+// ListNode is a node in a linked list.
+type ListNode struct {
+	val  int
+	next *ListNode
+	prev *ListNode
+}
+
+// NewListNode constructs a new ListNode with the given value & no next/prev links.
+func NewListNode(v int) *ListNode {
+	return &ListNode{
+		val:  v,
+		next: nil,
+		prev: nil,
+	}
+}
+
+// DoublyLinkedList is a doubly-linked list with Head and Tail.
+type DoublyLinkedList struct {
+	Head *ListNode
+	Tail *ListNode
+}
+
+// NewDoublyLinkedList constructs a doubly linked list from a sequence of integers.
+func NewDoublyLinkedList(vs ...int) *DoublyLinkedList {
+	ll := &DoublyLinkedList{
+		Head: nil,
+		Tail: nil,
+	}
+
+	if len(vs) < 1 {
+		return ll
+	}
+
+	ll.Head = NewListNode(vs[0])
+	ll.Tail = ll.Head
+
+	if len(vs) == 1 {
+		return ll
+	}
+
+	cur := ll.Head
+	for i := 1; i < len(vs); i++ {
+		cur.next = NewListNode(vs[i])
+		cur.next.prev = cur
+		cur = cur.next
+	}
+
+	ll.Tail = cur
+
+	return ll
+}
+
+// Reverse reverses the given linked list in-place.
+func (ll *DoublyLinkedList) Reverse() {
+	if ll.Head == nil || ll.Head.next == nil {
+		return
+	}
+
+	// construct singly-linked list from the back
+	dummy := NewListNode(-1)
+	cur := dummy
+	n := ll.Tail
+	for n != nil {
+		cur.next = n
+
+		cur = cur.next
+		n = n.prev
+	}
+	cur.next = nil // cur will be the new ll.Tail -> set .next = nil
+
+	// add prev -> doubly-linked list
+	prev := dummy.next
+	n = dummy.next.next
+	for n != nil {
+		n.prev = prev
+
+		n = n.next
+		prev = prev.next
+	}
+
+	// update Head & Tail
+	ll.Head, ll.Tail = ll.Tail, ll.Head
+	ll.Head.prev = nil
+}
+
+func (ll *DoublyLinkedList) String() string {
+	buf := bytes.NewBuffer([]byte{'{'})
+
+	for cur := ll.Head; cur != nil; cur = cur.next {
+		buf.WriteString(fmt.Sprintf("%v <-> ", cur.val))
+	}
+
+	buf.WriteByte('}')
+
+	return buf.String()
+}
+
+// DebugString prints the linked list with both node's value, next & prev pointers.
+func (ll *DoublyLinkedList) DebugString() string {
+	buf := bytes.NewBuffer([]byte{'{'})
+	buf.WriteString(fmt.Sprintf("Head= %p; ", ll.Head))
+
+	for cur := ll.Head; cur != nil; cur = cur.next {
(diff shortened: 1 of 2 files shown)


exercises/linked-list/linked_list.go
```diff
@@ -0,0 +1,203 @@
...
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+)
+
+// ListNode is a node in a linked list.
+type ListNode struct {
+	val  int
```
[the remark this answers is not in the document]
exklamationmark: Oh, I forgot to consider that the package could be used elsewhere.

What do you think about exposing all the fields of `ListNode` instead? i.e:
```go
type ListNode struct {
    Val int
    Next *ListNode
    Prev *ListNode
}
```

- package user can call `listNode.Next` and `listNode.Prev` to traverse the list.
    - I assume your `Next()` means `func (ln *ListNode) Next() *ListNode`.
    I have yet to see any benefit in having this getter function vs just exporting the field. Perhaps you can enlighten me? 🙇 
- do we need a `func (ln *ListNode) First() *ListNode`, when user can get it from the `linkedList.Head`?

exercises/linked-list/.meta/description.md
```diff
@@ -0,0 +1,36 @@
...
+* `Head`: the first element in the linked list.
+* `Tail`: the last element in the linked list.
+* `PushBack(v interface{})`: *insert value at back*.
+* `PopBack() (interface{}, error)`: *remove value at back*.
+* `PushFront(v interface{}) `: *remove value at front*.
+* `PopFront() (interface{}, error)`: *insert value at front*.
+* `Reverse()`: reverse the linked list (Head, Tail = Tail, Head).
+
+Instead of not covering error conditions, like calling `PopBack` or `PopFront` on an empty list,
+we will follow Go's idiomatic style and implement the error checks as well.
```
exklamationmark: @leenipper: I have overwritten the default `description.md`, so that the Go implementation is more idiomatic.

`Reverse()` was added into the requirement, as I feel it's where the we touch the  "doubly" aspect of the implementation.

exklamationmark:
@leenipper: I have updated the exercises to match the convention we use.

[The build still fails](https://github.com/exercism/go/pull/1107/files#diff-9180003c44770389866d67870f080c50R17). Do you have a README for on what's checked in the CI build?

exercises/linked-list/linked_list.go
```diff
@@ -0,0 +1,203 @@
...
+		n = n.next
+		prev = prev.next
+	}
+
+	// update Head & Tail
+	ll.Head, ll.Tail = ll.Tail, ll.Head
+	ll.Head.prev = nil
+}
+
+func (ll *DoublyLinkedList) String() string {
```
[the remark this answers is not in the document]
leenipper: > interface is cleaner without it

Agree.

exercises/linked-list/.meta/description.md
leenipper: > significantly touch the "doubly" aspect.

Agree.

An on that "doubly" aspect, if we have First() as suggested, then perhaps also have a Last(), along with a Previous() ?  This assumes exposing less of the struct members for protection of package implemenation.  Otherwise, user can manually traverse. I like exposing less internals.

exklamationmark: >  This assumes exposing less of the struct members for protection of package implemenation. Otherwise, user can manually traverse. I like exposing less internals.

Agree. I was confused because we still return a `*Node` (somehow I associated this with arbitrary content manipulation). It's not the case and package user should not be able to manipulate a node's `next` and `prev`.

So I will change it to:

```go
type Node struct {
    Val interface{}
    prev *Node
    next *Node
}

func (n *Node) First() *Node { ... } // traverse till head
func (n *Node) Last() *Node { ... } // traverse till tail
func (n *Node) Next() *Node { ... }
func (n *Node) Prev() *Node { ... }
```

I will also change the linked list definition to use getter functions, so we are consistent.

```go
type List struct {
    head *Node
    tail *Node
}

func (ll *List) Head() *Node { ... }
func (ll *List) Tail() *Node { ... }
```

exercises/linked-list/linked_list_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+			checkDoublyLinkedList(t, actual, tc.expected)
+		})
+	}
+}
+
+// checkDoublyLinkedList checks that the linked list is constructed correctly.
+func checkDoublyLinkedList(t *testing.T, ll *List, expected []interface{}) {
+	// check that length and elements are correct (scan once from begin -> end)
+	elem, count, idx := ll.head, 0, 0
```
leenipper: Perhaps
```
-    elem, count, idx := ll.head, 0, 0
+    elem, count, idx := ll.First(), 0, 0
```
This gives some lattitude to the implementor instead of requiring an exact duplication of your example struct(s).

exercises/linked-list/linked_list_test.go
```diff
@@ -0,0 +1,135 @@
...
+	for ; elem != nil && idx < len(expected); elem, count, idx = elem.next, count+1, idx+1 {
```
leenipper: And in the same vein,  elem.Next()

exercises/linked-list/linked_list_test.go
```diff
@@ -0,0 +1,135 @@
...
+			t.Errorf("wrong value from %d-th element, expected= %v, got= %v", idx, expected[idx], elem.Val)
+		}
+	}
+	if !(elem == nil && idx == len(expected)) {
+		t.Errorf("expected %d elements, got= %d", len(expected), count)
+	}
+
+	// if elements are the same, we also need to examine the links (next & prev)
+	switch {
+	case ll.head == nil && ll.tail == nil: // empty list
```
leenipper: ll.First()  and ll.Last()

exercises/linked-list/linked_list_test.go
```diff
@@ -0,0 +1,135 @@
...
+		return
+	case ll.head != nil && ll.tail != nil && ll.head.next == nil: // 1 element
+		valid := ll.head == ll.tail &&
```
leenipper: Same here.  And the other places below where the .next and .prev are used.  I won't bother tagging all of them.

exklamationmark:
No problem :) I changed the tests.

exercises/linked-list/README.md
```diff
@@ -0,0 +1,59 @@
...
+If the test suite contains benchmarks, you can run these with the `-bench`
+flag:
+
+    go test -bench .
+
+Keep in mind that each reviewer will run benchmarks on a different machine, with
+different specs, so the results from these benchmark tests may vary.
+
+## Further information
+If you are not familar with linked list, try [this tutorial];dhttps://visualgo.net/en/list).
```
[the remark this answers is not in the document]
exklamationmark: ah, thanks for spotting

PR #1110, opened 2018-05-25
add a test case to the luhn exercise

idealhack:
I'm not sure this is the right way to contribute in test cases, but here it is.

How exercism apply test cases to all languages? I could not find an answer.

### DIFF
--- a/exercises/luhn/cases_test.go
+++ b/exercises/luhn/cases_test.go
@@ -74,4 +74,9 @@ var testCases = []struct {
 		"091",
 		true,
 	},
+	{
+		"strings with non-digits is invalid",
+		":9",
+		false,
+	},
 }


leenipper:
Thanks for asking.

The cases_test.go is generated from the [luhn exercise generator](https://github.com/exercism/go/blob/master/exercises/luhn/.meta/gen.go).

> way to contribute in test cases

This [file](https://github.com/exercism/problem-specifications/blob/master/exercises/luhn/canonical-data.json) within [problem specifications](https://github.com/exercism/problem-specifications) repository is the input to the generator. It would the the one to submit a PR against.  After its acceptance, the generator would pick up the new test case and after running it, a PR to update the generated cases_test.go  is done.

idealhack:
@leenipper Thank you for the explanation! I just saw the documents on website by googling, it seems they are not in any navigations or menus.

- http://exercism.io/contribute/canonical-data
- http://exercism.io/contribute

PR #1121, opened 2018-06-26
hello-world: nit in README, %q instead of %v in test

zellyn:
- s/classical/classic/ in README
- s/%v/%q/ in test
- s/,/;/ in test

### DIFF
--- a/exercises/hello-world/README.md
+++ b/exercises/hello-world/README.md
@@ -1,6 +1,6 @@
 # Hello World
 
-The classical introductory exercise. Just say "Hello, World!".
+The classic introductory exercise. Just say "Hello, World!".
 
 ["Hello, World!"](http://en.wikipedia.org/wiki/%22Hello,_world!%22_program) is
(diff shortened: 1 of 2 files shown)


exercises/hello-world/README.md
```diff
@@ -1,6 +1,6 @@
 # Hello World
 
-The classical introductory exercise. Just say "Hello, World!".
+The classic introductory exercise. Just say "Hello, World!".
```
leenipper: Hmm. For classical vs. classic,  I'm leaning toward classical. I must defer to original [description file](https://github.com/exercism/problem-specifications/blob/master/exercises/hello-world/description.md). If a change were made, that repo is where it would be made. The configlet tool generates README.md.

zellyn: fine by me…

I [looked it up](https://jakubmarian.com/classic-vs-classical-in-english/) but even their description was confused!

PR #1124, opened 2018-06-26
triangle: Fix references to twofer in stub comments

### LINKED ISSUE #1123: Exercise triangle's stub includes comments referencing exercise twofer
Comments should be changed to reference the package triangle and the function KindFromSides.

joswiatek:
Correct comments in stub file to reference the correct package and function name. 

This closes #1123

### DIFF
--- a/exercises/triangle/triangle.go
+++ b/exercises/triangle/triangle.go
@@ -1,7 +1,7 @@
 // This is a "stub" file.  It's a little start on your solution.
 // It's not a complete solution though; you have to write some code.
 
-// Package twofer should have a package comment that summarizes what it's about.
+// Package triangle should have a package comment that summarizes what it's about.
 // https://golang.org/doc/effective_go.html#commentary
 package triangle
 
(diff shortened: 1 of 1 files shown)


joswiatek:
I'm not sure why CI is failing. triangle.go is not passing go fmt because of missing types and values, but those were already missing as part of the exercise.

leenipper:
@joswiatek, I found out the reason the travis-ci failed.  It wasn't your change or any change in the go track itself.  After closer examination of the log, I see that the travis-ci is using go 1.11 beta1, and looks like 1.11 has a small formatting rule change;  as a result the scrabble-score go fmt fails.  We can sit tight for now. No PRs will pass the travis-ci now.  @tleen, do you know the policy for go versions on travis-ci ?  I suspect go 1.11 won't release until later this year. Does the .travis.yml need a change to correctly operate for the track ? Seems that the 1.x magic is picking up 1.11beta1.

leenipper:
Since this change is commentary only, and the travis-ci failed for different reason, it is safe to go ahead with a rebase and merge to complete this one.

PR #1129, opened 2018-07-13
accumulate: Correct the readme

### LINKED ISSUE #1070: accumulate: Readme doesn't match the tests.
The Readme for accumulate uses an example of a numeric list, but the test cases all use strings. The Readme should be rewritten. 

----
Given the collection of strings:

    "the", "cat", "sat", "on", "the", "mat"

And the operation:

`Capitalise` a string (`"cat" => "CAT"`)

Your code should be able to produce the collection of words:

    "THE", "CAT", "SAT", "ON", "THE", "MAT"
----

ooransoy:
Fixes #1070.

### DIFF
--- a/exercises/accumulate/README.md
+++ b/exercises/accumulate/README.md
@@ -5,17 +5,17 @@ operation to perform on each element of the collection, returns a new
 collection containing the result of applying that operation to each element of
 the input collection.
 
-Given the collection of numbers:
+Given the collection of strings:
 
-- 1, 2, 3, 4, 5
+- "cat", "Dog", "b4t", "gO"
 
 And the operation:
 
-- square a number (`x => x * x`)
+- upcase a string
 
-Your code should be able to produce the collection of squares:
+Your code should be able to produce the collection of strings:
 
-- 1, 4, 9, 16, 25
+- "CAT", "DOG", "B4T, "GO"
 
 Check out the test suite to see the expected function signature.
 
--- a/exercises/accumulate/.meta/description.md
+++ b/exercises/accumulate/.meta/description.md
@@ -1,0 +1,24 @@
+Implement the `accumulate` operation, which, given a collection and an
+operation to perform on each element of the collection, returns a new
+collection containing the result of applying that operation to each element of
+the input collection.
+
+Given the collection of strings:
+
+- "cat", "Dog", "b4t", "gO"
+
+And the operation:
+
+- upcase a string
+
+Your code should be able to produce the collection of strings:
+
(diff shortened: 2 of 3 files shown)


exercises/accumulate/README.md
```diff
@@ -29,10 +29,10 @@ Solve this one yourself using other basic tools instead.
 
 To run the tests run the command `go test` from within the exercise directory.
 
-If the test suite contains benchmarks, you can run these with the `--bench` and `--benchmem`
-flags:
+If the test suite contains benchmarks, you can run these with the `-bench`
+flag:
```
leenipper: Did you notice that the differences shown dropped some standard lines for the Go READMEs here? The standard lines come from [here](https://github.com/exercism/go/blob/master/config/exercise_readme.go.tmpl#L11-L14).
And, the README.md files are generated using a tool called configlet, that gathers pieces from standard places, track specific,  and also from the .meta directory (if applicable) under each exercise.  For a few exercises, the base content of the README.md comes from a .meta/description.yml file under the exercise directory. For example, see exercises/yacht; and in those cases, the .meta directory must also include a copy of the metadata.yml from the problem-specifications/exercises/.../metadata.yml (without change).
First, clone the problem-specifications repo, at the same level, directory-wise, as your local exercism/go repo.
Use exercises/accumulate/description.md from that repo to make your adjustments (numbers vs. strings) and put the result in exercises/accumulate/.meta/description.md; copy the metadata.yml for yacht from problem-specifications/exercises/accumulate/metadata.yml into the new accumulate/.meta directory
Then get a up-to-date copy of the configlet tool:
```
./bin/fetch-configlet
```
Then, run configlet:
```
./bin/configlet help generate
./bin/configlet generate --only accumulate
```
The README.md generated for accumulate should show only the specific changes you made for the strings vs. numbers and that text would be in the exercises/accumulate/.meta/description.md.

exercises/accumulate/README.md
```diff
@@ -29,10 +29,10 @@ Solve this one yourself using other basic tools instead.
...
 
-    go test -v --bench . --benchmem
+    go test -bench .
```
leenipper: And here too.  Lost the standard track piece.

bitfield:
Happy to pick this up, @sebito91, if the OP isn't available.

kytrinyx:
@bitfield Since we've not heard back for almost 5 months, I think the right move here is to carry it!

jrake-revelant:
+1 I ran into this problem

Issue #1132, opened 2018-07-19
protein-translation: FromCodon tests force less idiomatic solutions.


hilary:
Please bear with me while I explain, as I'm still quite new to golang.

When solving protein-translation, my inclination was to use an error to signal that `FromCodon` had hit a stop codon (in the same way that `io` uses an error to signal `EOF`. If there were no tests for `FromCodon`, I could do that (or not, depending on my understanding of the language). But because there are tests for `FromCodon` that don’t collect (or ignore) error responses, I’m forced to treat `STOP` like any other response, which produces what feels like far less idiomatic code. (The effect cascades into how I write `FromRNA` as well.)

I asked about this on slack and @kytrinyx said "It would also provides an opportunity to implement something in the style of `os.IsNotExist(err)`." I [looked up the function](https://golang.org/pkg/os/#IsNotExist). Yup! Exactly! I was all set to stretch my wings and fly and ... grounded 😝

sebito91:
I think this is a cool idea! I think the best way around this is to just call out to a helper function of the same name. Usually when I'm creating packages I'll have a publicly exported version of the function, in this case `FromCodon` but just return the value from my helper version that obfustcates the actual content (just like your example doc ref -- https://golang.org/src/os/error.go?s=2154:2185#L65).

In this case, you can create the custom err value yourself and then just do something like this...

```
type customError error
var MyCustomError customError = errors.New("custom error")

func FromCodon(input string) string {
    str, err := fromCodon(input); 
    if err != nil && isMyErr(err) {
        // do something here like return empty or STOP
    }
    return str
}

func fromCodon(input string) (string, error) {
    // do your stuff here and return string, error at the end
}

func isMyErr(err error) bool {
    switch err.(type) {
    case customError:
        return true
    }
    return false
}
```

sebito91:
And I could be completely incorrect too :D

hilary:
It's a cool workaround, but a better solution would be to fix the implementation of the problem in the track.

Hey, golang track maintainers, I'm gonna take a pass at doing just that...

hilary:
Fixed by #1134

PR #1134, opened 2018-07-20
protein-translation: add errors logic

hilary:
Here's my initial pass at a solution for #1132 

Since protein-translation is unlocked by scrabble-score, I thought it was fair game to keep the tests for `FromCodon`, but include the error logic in the tests (and, of course, the example solution). I drew heavily from the golang standard library in designing the solution, but who knows what I may have mangled along the way 😄

### DIFF
--- a/exercises/protein-translation/example.go
+++ b/exercises/protein-translation/example.go
@@ -1,57 +1,71 @@
+// Package protein translates RNA sequences into proteins.
 package protein
 
-func FromCodon(codon string) string {
-	switch codon {
+import "errors"
+
+var (
+	STOP           = errors.New("stop")
+	ErrInvalidBase = errors.New("invalid base")
+)
+
+// IsStopCodon checks whether a codon is a stop codon.
+func IsStopCodon(c string) bool {
+	return c == "UAA" || c == "UAG" || c == "UGA"
+
+}
+
+// FromCodon returns the protein for the given codon.
+// If the codon is a stop codon, it returns STOP.
+// If the codon is unrecognized, it returns invalidBase.
+func FromCodon(c string) (string, error) {
+	if IsStopCodon(c) {
+		return "", STOP
+	}
+
+	switch c {
 	case
 		"AUG":
-		return "Methionine"
+		return "Methionine", nil
 	case
 		"UUU",
 		"UUC":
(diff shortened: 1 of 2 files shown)


sebito91:
Fixes #1132

exercises/protein-translation/protein_translation_test.go
```diff
@@ -42,20 +116,11 @@ var proteinTestCases = []rnaCase{
...
-	}
-}
-
 func TestProtein(t *testing.T) {
 	for _, test := range proteinTestCases {
 		actual := FromRNA(test.input)
 		if !reflect.DeepEqual(actual, test.expected) {
-			t.Errorf("Protein Translation test [%s], expected %q, actual %q", test.input, test.expected, actual)
+			t.Errorf("FAIL: RNA Translation test: %s\nExpected: %q\nActual %q", test.input, test.expected, actual)
 		}
```
leenipper: Perhaps
```
+    t.Logf("PASS: %s", test.input)
```
here as well, to make the 'go test -v' have similar output for TestProtein, as TestCodon does.

exercises/protein-translation/example.go
```diff
@@ -1,57 +1,71 @@
...
-			tempCodon := FromCodon(res)
-			if tempCodon == "STOP" {
-				break
-			}
-			proteins = append(proteins, tempCodon)
-			res = ""
+	bases := []rune(s)
+	for i := 0; i < len(bases); i += 3 {
+		p, err := FromCodon(string(bases[i : i+3]))
+		if err == STOP {
```
sebito91: Only nit I have is that we don't ever handle the `ErrInvalidBase` case here. How would we handle a string like the following from the `rnaCase` test slice?

```
OLD
{"UGGUGUUAUUAAUGGUUU", []string{"Tryptophan", "Cysteine", "Tyrosine"}}

NEW
{"UGGUGUUAUUAAUGGABCUUU"}

        protein_translation_test.go:124: FAIL: RNA Translation test: UGGUGUUAUUAAUGGABCUUU
                Expected: ["Tryptophan" "Cysteine"]
                Actual ["Tryptophan" "Cysteine" "Tyrosine"]
```

PR #1136, opened 2018-07-22
two-fer: add param name to stub function

R4wm:
Not sure if this was intentional, but adding missing parameter name to exercise file.

### DIFF
--- a/exercises/two-fer/two_fer.go
+++ b/exercises/two-fer/two_fer.go
@@ -6,7 +6,7 @@
 package twofer
 
 // ShareWith should have a comment documenting it.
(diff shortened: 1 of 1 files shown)


exercises/two-fer/two_fer.go
```diff
@@ -6,7 +6,7 @@
 package twofer
 
 // ShareWith should have a comment documenting it.
-func ShareWith(string) string {
+func ShareWith(name string) string {
```
leenipper: This stub is intentionally incomplete.  But I think it is fine to add the name to make the func signature complete.

PR #1137, opened 2018-07-23
Collatz conjecture

R4wm:
It feels like the example is giving away a good approach to the exercise..


R4wm:
not sure why still failing CLI, created a more simple PR

PR #1138, opened 2018-07-23
dont give away too much of the exercise / logic

R4wm:
It seems maybe too much good logic is given away with the example.


hilary:
`example.go` is not included when a student downloads a problem.

You can find more information about `example.go` and the other files in the repo in the README.

R4wm:
oh, my mistake, i didn't realize which workspace i was working in. 
THanks

Issue #1143, opened 2018-07-30
Ensure all exercises have benchmarks


hilary:
At present, only some exercises have benchmarks. In Go, benchmarks are considered core tests, therefore all our exercises should have benchmarks.

I can think of five steps (please point out what I've forgotten or a better way to do it) to addressing benchmarks comprehensively

* update the README to include info about benchmarks, including how to write them
* review all the current exercises to come up with a checklist of those that don't have benchmarks
* add benchmarks to all the exercises that don't have them
* add benchmarking as a topic 
* make sure we have exercises that address benchmarking

leenipper:
Go supports benchmarks with only minimal work.  Exercises which address benchmarking is a really good idea.  There are good things to learn in that regard.   Don't any benchmarking focused exercises need to begin with [problem-specifications](https://github.com/exercism/problem-specifications/tree/master/exercises) and be language track independent, at least to start.

hilary:
Thoughts:

* We may have existing problems which address benchmarking, although I can't think of any off the top of my head. I don't think it would be particularly hard to write one.
* I looked at the [topics list](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt), and there isn't a topic for benchmarking (there isn't even a topic for testing 😭).
* As I understand it, a new problem does not *have* to start with a language independent problem-specification. It depends. Many problems that are added to Exercism are adapted from existing problems found elsewhere, in which case, writing a language independent problem-specification is a viable first step. For a new problem, however, I would be wary of writing (and/or requiring) a generalized version before a language specific version exists.

PR #1148, opened 2018-08-01
armstrong-numbers: add benchmarks

hilary:
Part of #1143


hilary:
@sebito91 are we good to go?

PR #1167, opened 2018-09-21
Changes planet type to string

fabioxgn:
This was returning an error: `cannot use "Earth" (type string) as type Planet in field value`. This type Planet just confused me, after changing to string I was able to run the tests and complete the exercise.

### DIFF
--- a/exercises/space-age/cases_test.go
+++ b/exercises/space-age/cases_test.go
@@ -6,7 +6,7 @@ package space
 
 var testCases = []struct {
 	description string
-	planet      Planet
+	planet      string
 	seconds     float64
 	expected    float64
 }{


exercises/space-age/cases_test.go
```diff
@@ -6,7 +6,7 @@ package space
 
 var testCases = []struct {
 	description string
-	planet      Planet
+	planet      string
```
leenipper: The exercise wants you to use a defined type, named Planet, with string as the [underlying type](https://golang.org/ref/spec#Type_definitions).
You may have already understood it by now, but if not, see the [example solution](https://github.com/exercism/go/blob/master/exercises/space-age/example.go#L4).

BTW, cases_test.go is a generated file. No changes are ever made to that file independent of a [generator](https://github.com/exercism/go/blob/master/exercises/space-age/.meta/gen.go) change.

leenipper:
Change unnecessary.

PR #1169, opened 2018-10-03
binary-search: update generator to align with upstream problem-specifications

ilmanzo:
since from [last commit](https://github.com/exercism/problem-specifications/commit/e0ffb0085591467aa946962310d3da6fbbbf9483) the json has changed format, we'd need to manage the 'expected' field both as a value and as an error object. For now I'd keep the old format , adding just an extra 'error' field to the test suite.

### DIFF
--- a/exercises/binary-search/.meta/gen.go
+++ b/exercises/binary-search/.meta/gen.go
@@ -65,19 +65,19 @@ var tmpl = `package binarysearch
 
 {{.Header}}
 
-var testCases = []struct {
+var testCases =	[]struct {
 	description	string
 	slice		[]int
-	key		int
-	x	int
-	err	string
+	key			int
+	x			int
+	err			string
 }{ {{range .J.Cases}}
 {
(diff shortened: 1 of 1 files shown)


exercises/binary-search/.meta/gen.go
```diff
@@ -46,12 +70,14 @@ var testCases = []struct {
...
 	x	int
+	err	string
 }{ {{range .J.Cases}}
 {
 	description:	{{printf "%q"  .Description}},
 	slice:		{{printf "%#v" .Input.Array}},
 	key:		{{printf "%d"  .Input.Value}},
-	x:	{{printf "%d"  .Expected}},
+	x:	{{printf "%d"  .Expected.ValueInt}},
+	err:	{{printf "%q"  .Expected.ValueString}},
```
leenipper: Perhaps align the {{printf ...   using tabs.  The template text string isn't subject to 'go fmt', but it looks better aligned (to me at least).  Maybe also the types for x, and err within the struct above.

PR #1171, opened 2018-10-16
bank-account: Check that the Balance after Close is 0

homelinen:
Noticed a community solution that missed removing the funds from the balance.

Noted that if you get a non-nil error from a method the value is often ignored, but felt is worth checking the balance is 0 after a close.

Maybe over-thinking but you wouldn't expect a closed account to have money left. If used in the real world it would be a big accounting error.

Happy to take any comments and direction if this should be added to some of the other test cases.

### DIFF
--- a/exercises/bank-account/bank_account_test.go
+++ b/exercises/bank-account/bank_account_test.go
@@ -65,6 +65,12 @@ func TestSeqOpenBalanceClose(t *testing.T) {
 		t.Log("Balance still available on closed account.")
 		t.Fatalf("a.Balance() = %d, %t.  Want ok == false", b, ok)
 	}
+			
+	// verify closing balance is 0
+	if b, _ := a.Balance(); b != 0 {
+		t.Log("Balance after close is non-zero.")
+		t.Fatalf("After a.Close() balance is %d and not 0", b)
+	}
 }
 
 func TestSeqOpenDepositClose(t *testing.T) {


exercises/bank-account/bank_account_test.go
```diff
@@ -65,6 +65,12 @@ func TestSeqOpenBalanceClose(t *testing.T) {
 		t.Log("Balance still available on closed account.")
 		t.Fatalf("a.Balance() = %d, %t.  Want ok == false", b, ok)
 	}
+
```
leenipper: The 'go fmt' caused the travis-ci check to fail. Simply run 'go fmt exercises/bank-account/bank_account_test.go' and push a commit for the tiny white-space change.  It should then pass travis-ci check.

kytrinyx:
It's been 6 weeks since we've had any movement on this one. I think it's pretty safe to assume that the original author won't be coming back. @bitfield would you be up for "carrying" this PR? If you're unfamiliar with the term, see https://blog.jessfraz.com/post/the-art-of-closing/ (bullet number 4, but the whole article is great).

bitfield:
No problem @kytrinyx, I'll take care of it.

bitfield:
Hi @homelinen, we really like your patch, but since there hasn’t been a response in a while, we are going to carry this patch and make the edits ourselves. The new PR is #1207.

PR #1176, opened 2018-11-02
Update cases_test.go

lplazas:
Talking forcefully isn't pretty much the same as shouting? Also the phrase ends with an exclamation mark.

### DIFF
--- a/exercises/bob/cases_test.go
+++ b/exercises/bob/cases_test.go
@@ -42,7 +42,7 @@ var testCases = []struct {
 	{
 		"talking forcefully",
 		"Let's go make out behind the gym!",
-		"Whatever.",
+		"Whoa, chill out!",
 	},
 	{
 		"using acronyms in regular speech",


exercises/bob/cases_test.go
```diff
@@ -42,7 +42,7 @@ var testCases = []struct {
 	{
 		"talking forcefully",
 		"Let's go make out behind the gym!",
-		"Whatever.",
+		"Whoa, chill out!",
```
leenipper: > Talking forcefully isn't pretty much the same as shouting?

Thanks for asking.

Although the exclamation mark in the input may make it seem like shouting or yelling, the shouting rule used is words capitalized, which is taken from the [reference](https://pine.fm/LearnToProgram/chap_06.html) in the [README.md](https://github.com/exercism/go/blame/7246816539509b883a98e6477f967887d4ed5d2b/exercises/bob/README.md#L37) and the [input source for the generated test cases](https://github.com/exercism/problem-specifications/blob/master/exercises/bob/canonical-data.json#L54-L59) follows that rule.

PR #1180, opened 2018-11-08
isogram: Add test case for duplicate after non-letter

bitfield:
I had a student submit a solution like this:

```go
	for _, char := range word {
		if !unicode.IsLetter(char) {
			break
		}
```

Obviously they meant `continue` rather than `break`, but the test cases don't catch this, because there is no case where a non-letter character is followed by a duplicated character. This commit adds such a case, which would fail the student's incorrect solution.


exercises/isogram/cases_test.go
```diff
@@ -49,6 +49,11 @@ var testCases = []struct {
 		input:       "thumbscrew-japingly",
 		expected:    true,
 	},
+	{
+		description: "hypothetical word with duplicated character after hyphen",
+		input:       "thumbscrew-jappingly",
+		expected:    false,
+	},
```
leenipper: Looks like an excellent test case to add.  Thanks for the feedback from a student/mentor experience.

bitfield:
Thanks; silly of me, I should have spotted that.

See https://github.com/exercism/problem-specifications/issues/1390

leenipper:
Thanks for following through with problem-specifications!
Looks like a fresh generate (go run .meta/gen.go) followed by a push will get the PR in sync.

PR #1181, opened 2018-11-09
luhn: Add test case for valid number with even length

bitfield:
I had a student submit a solution like this:

```go
		if i == 1 || i%2 == 0 {
			y *= 2
```

The logic for picking 'every second digit from the right' is not correct, but it works with the existing tests, because all the valid numbers have an odd number of digits. This commit adds a test case which would fail this incorrect solution, but pass for correct solutions.

### DIFF
--- a/exercises/luhn/cases_test.go
+++ b/exercises/luhn/cases_test.go
@@ -1,8 +1,8 @@
 package luhn
 
 // Source: exercism/problem-specifications
-// Commit: 3930b0a luhn: add a test case of strings with non-digits (#1246)
-// Problem Specifications Version: 1.2.0
+// Commit: 06b37c1 luhn: Add test case for valid number with even length (fixes #1391)
+// Problem Specifications Version: 1.3.0
 
 var testCases = []struct {
 	description string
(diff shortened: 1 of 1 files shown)


exercises/luhn/cases_test.go
```diff
@@ -39,6 +39,11 @@ var testCases = []struct {
 		"055 444 286",
 		false,
 	},
+	{
+		"valid number with an even number of digits",
+		"095 245 88",
+		true,
+	},
```
leenipper: Thanks for your feedback.  This will be a good additional test case.  As in #1180, it's a generated file.

bitfield:
See https://github.com/exercism/problem-specifications/issues/1391

leenipper:
Thanks again for the p-s update. A generate and a push will make it ready.

kytrinyx:
@bitfield Do you want to have a go at regenerating this one?

bitfield:
Ready for launch, sir!

kytrinyx:
> Ready for launch, sir!

Roger that!

PR #1186, opened 2018-11-17
luhn: add test case with non-digit at the end

tehsphinx:
Problem: I had a student only include the digits before the first invalid character. There was no test that failed his solution.

Solution: Added a test that should return `false` but does not if the student takes only the numbers up to first non-digit.

### DIFF
--- a/exercises/luhn/cases_test.go
+++ b/exercises/luhn/cases_test.go
@@ -1,8 +1,8 @@
 package luhn
 
 // Source: exercism/problem-specifications
-// Commit: 06b37c1 luhn: Add test case for valid number with even length (fixes #1391)
-// Problem Specifications Version: 1.3.0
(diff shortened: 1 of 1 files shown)


kytrinyx:
@tehsphinx This should pass the build if you rebases against the most recent master. Would you mind giving that a whirl?

bitfield:
On behalf of @tehsphinx: opened [problem-specifications #1420](https://github.com/exercism/problem-specifications/pull/1420) to fix this in the canonical data. Once that's merged I'll regenerate the test cases and close this.

kytrinyx:
Awesome, thanks @bitfield! (For reference, same process as https://github.com/exercism/go/pull/1198#issuecomment-449601384 would apply here).

tehsphinx:
Thx @bitfield!

PR #1188, opened 2018-11-18
error-handling: fix for golint

tehsphinx:
In case the student runs `golint` on his solution there is an error in the exercise package. Automatic tools will detect that too and blame the student.


kytrinyx:
@tehsphinx I think if you rebase onto the most recent master, the build should pass. Would you mind kicking the tires on this one?

bitfield:
On behalf of @tehsphinx (who's away for a few days), carried to #1198.

PR #1208, opened 2018-12-22
tree-building: remove refactoring requirement

### LINKED ISSUE #1135: tree-building: refactor vs. rewrite
The README of tree building says the student should refactor the existing code. I imagined that means there is an incremental way of getting from the existing solution to the better solution. Looking at the existing solution I don't see how that is possible. Can someone explain me how this "refactoring" is supposed to work? Can we add some more guidance to the README where to start that process? If there is no incremental process I would suggest changing the description to say "rewrite the Build function" instead of "refactor".

bitfield:
'Tree Building' is a hard enough exercise to implement from scratch, without requiring that the student first understand, and then refactor in place, the deliberately awkward example implementation.

This fixes #1135 by removing the instruction to 'refactor' from the README, and removing the 'bad' example implementation.

It's a great idea to have students learn the basics of refactoring on a relatively simple, straightforward exercise. This isn't it.


bitfield:
Ah, sorry about that; I didn't realise this was generated, because there's no `.meta` directory for this exercise. I'll add one.

bitfield:
So regenerated.

kytrinyx:
> I didn't realise this was generated, because there's no `.meta` directory for this exercise.

Yepp. All the READMEs are generated, and we only typically have a `.meta` if we override something.

PR #1212, opened 2018-12-27
acronym: Rename acronym.nogo to acronym.go

ilmanzo:
The script bin/test-without-stubs renames all the .go files in .nogo; so I guess that the file .nogo has been accidentally committed while the test was running. The right filename is acronym.go


kytrinyx:
Thanks, @ilmanzo

PR #1213, opened 2018-12-27
variable-length-quantity: update generator 

ilmanzo:
upstream problem specifications have been [updated](https://github.com/exercism/problem-specifications/commit/9cdc06462f99064ff7bd4174f9586b93e7b7318f#diff-3030cdac0a2738a37774e6e7ae38ae11) with the addition of an expected 'error' object

### DIFF
--- a/exercises/variable-length-quantity/cases_test.go
+++ b/exercises/variable-length-quantity/cases_test.go
@@ -1,8 +1,8 @@
 package variablelengthquantity
 
 // Source: exercism/problem-specifications
-// Commit: 48dc5e6 variable-length-quantity: Apply new "input" policy
-// Problem Specifications Version: 1.1.0
+// Commit: 191f8086 variable-length-quantity: add mention of 'error' to comment
+// Problem Specifications Version: 1.2.0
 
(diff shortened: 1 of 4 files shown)


exercises/variable-length-quantity/cases_test.go
```diff
@@ -107,46 +107,55 @@ var decodeTestCases = []struct {
 	description string
 	input       []byte
 	output      []uint32 // nil slice indicates error expected.
+	shouldfail  bool
```
kytrinyx: To match the Go idiom, this should be mixed case, e.g. `shouldFail`. However, it would probably be even more idiomatic to have it be named `ok`, which would make the conditional around the assertion:

```go
if !tc.ok {
    // ...
}
```

kytrinyx:
@ilmanzo Would you please make sure that this PR is one commit and does not include the change to the acronym exercise? (If you're unfamiliar with the process, Steve Klabnik has a good hint here: http://blog.steveklabnik.com/posts/2012-11-08-how-to-squash-commits-in-a-github-pull-request)

PR #1224, opened 2019-01-20
word-count: check counting of empty words

tehsphinx:
Ran across this solution that will count empty words on some inputs:

```
package wordcount

import (
	"strings"
	"unicode"
)

// Frequency is the structure of words frequency
type Frequency map[string]int

// WordCount counts the occurrences of each word in a given phrase
func WordCount(input string) Frequency {
	counter := Frequency{}
	phrase := strings.FieldsFunc(input, split)

	for _, word := range phrase {
		word = normalize(word)
		counter[word]++
	}

	return counter
}

func normalize(word string) string {
	word = strings.ToLower(word)

	return strings.TrimFunc(word, func(r rune) bool {
		return !unicode.IsLetter(r) && !unicode.IsNumber(r)
	})
}

func split(r rune) bool {
	return r == ' ' || r == ','
}
```

The added test catches that behaviour.

### DIFF
--- a/exercises/word-count/cases_test.go
+++ b/exercises/word-count/cases_test.go
@@ -35,6 +35,11 @@ var testCases = []struct {
 		Frequency{"one": 1, "three": 1, "two": 1},
 	},
 	{
+		"handles expanded lists with leading space",
+		" one,\n two,\n three",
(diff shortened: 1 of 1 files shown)


tehsphinx:
@leenipper thx, will do

PR #1168, opened 2018-10-01
updating travis to add golanci-lint check

sebito91:
Per the discussion in https://github.com/exercism/go/issues/1024 we decided we wanted to include additional linting as part of the build. Originally we discussed adding gometalinter, but then golangci-lint came along, which is an improvement.

Closes https://github.com/exercism/go/issues/1024

PR #1225, opened 2019-01-23
Updates for golangci-lint

leenipper:
I could have omitted the .travis.yml change until after consensus on the Go code lint "fixes", but I wanted to see if it could operate successfully.

The golangci-lint options used by @sebito91 in #1168 were used for starters.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@
 tmp
 bin/configlet
 bin/configlet.exe
+bin/golangci-lint
+bin/golangci-lint.exe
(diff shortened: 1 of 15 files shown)


leenipper:
The only commit in this PR that I didn't like much was the palindrome-products commentary change to remove the unreferenced bonusData. Is there a better approach? Using a build tag perhaps? Or delete the bonusData and commented out items entirely?

leenipper:
I added commits to bring this one up to date. It now uses golangci-lint v16.0.0.

PR #1227, opened 2019-01-23
bin/test-without-stubs: Removed the redundant slash from the 'file' stub file path variable

ZapAnton:
Fixes the output of the Travis script.

Before:
`moving stub exercises/leap//leap.go`

After:
`moving stub exercises/leap/leap.go`

Also ensures that there will be no unexpected behavior when the `file` variable is used.


kytrinyx:
Nice, thanks!

PR #1228, opened 2019-01-24
leap: Added an argument variable to the IsLeapYear function

ZapAnton:
Out of all the exercises that provide a stub file, `leap` is the only one that does not have an argument variable in the stub function.

This PR tries to address it, by adding an explicit argument variable.

### DIFF
--- a/exercises/leap/leap.go
+++ b/exercises/leap/leap.go
@@ -6,7 +6,7 @@
 package leap
 
 // IsLeapYear should have a comment documenting it.
-func IsLeapYear(int) bool {
+func IsLeapYear(year int) bool {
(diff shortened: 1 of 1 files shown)


kytrinyx:
Thanks for opening this, @ZapAnton 

> This PR tries to address it, by adding an explicit argument variable.

I would prefer not to add the parameter name without a broader discussion, since naming is one of the useful discussions that come out of this exercise.

We're in the process of analyzing the track in order to restructure the exercises and add missing ones where necessary. I think I'd like to hold off on changing function parameters in the stubs until that work is well underway so that we get a feel for the benefits / downsides on naming them explicitly or not.

I'm going to go ahead and close this until then.

Issue #1230, opened 2019-02-04
Suggestion: change name of user-written function in reverse-strings.


sbromberger:
Right now, the reverse-strings exercise does not include a stub. That's not a problem, but a quick glance at the `reverse_string_test.go` shows a source of confusion. There are three functions defined in that file:
1. `TestReverse()` on line 8,
2. `TestReverseOfReverse()` on line 18, and
3. `BenchmarkReverse()` on line 27.

A novice (read: someone who's working on these problems as a user) would therefore assume that the function that s/he needs to write should be named `Reverse()`, since the purpose of the function is to reverse a string, and all the test functions refer to something called `Reverse` in their names.

However, the actual function the user is expected to write is called `String()` (as evidenced by the calls on lines 10, 20, and 30 of the test file).

I would suggest that function that the user is required to create, `String()`, should be changed to `Reverse()`. Not only does this make things consistent with the test file and more intuitive, but it is a more appropriate name for a function that reverses a string, and since we're focusing on programming style as well as substance, perhaps we should encourage the use of meaningful function names.

bitfield:
Agreed, the student's function should be named Reverse(), and that's what the test code should call. PRs welcome!

sbromberger:
Ref: #1231

PR #1231, opened 2019-02-04
reverse-string: rename func String to Reverse

sbromberger:
Renames `String()` to `Reverse()`.

### DIFF
--- a/exercises/reverse-string/example.go
+++ b/exercises/reverse-string/example.go
@@ -1,7 +1,7 @@
 package reverse
 
-// String reverses given string
-func String(s string) string {
+// Reverse reverses given string
+func Reverse(s string) string {
 	runes := []rune(s)
 	for i, j := 0, len(runes)-1; i < j; i, j = i+1, j-1 {
 		runes[i], runes[j] = runes[j], runes[i]
--- a/exercises/reverse-string/reverse_string_test.go
+++ b/exercises/reverse-string/reverse_string_test.go
@@ -7,7 +7,7 @@ import (
 
 func TestReverse(t *testing.T) {
 	for _, testCase := range append(testCases, multiByteCases...) {
-		if res := String(testCase.input); res != testCase.expected {
+		if res := Reverse(testCase.input); res != testCase.expected {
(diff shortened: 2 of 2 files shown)


bitfield:
LGTM! If there's no objection from anyone else I'll merge this shortly.

exercises/reverse-string/example.go
```diff
@@ -1,7 +1,7 @@
 package reverse
```
leenipper: I remember this one.  See https://github.com/exercism/go/pull/971
and the discussion about the func name.
In actual client code (of which there is none for an exercise), you would have reverse.String(x). With the change, a caller would use reverse.Reverse(x). A bit wordy, but I understand the reason for the PR.
@bitfield any new thoughts on this one ?

da-edra:
I think the commit message should be amended to be more descriptive because if you're browsing through the history of the repos a commit saying `implements #1230` doesn't say much.

sbromberger:
Hi,

I'm no longer in a position to support this PR. If someone wants to pick it up, great; otherwise, feel free to close it.

bitfield:
I see @leenipper's point about the `reverse.Reverse` stutter, but I also think this makes a persuasive case:

```go
-		return s == String(String(s))
+		return s == Reverse(Reverse(s))
```

With an improved commit message, I'm 👍.

leenipper:
Ok @bitfield, I'm convinced. I'll change the PR title to:

reverse-string: rename func String to Reverse

Better?

It's a single commit, so I think a Squash and Merge will let the PR title be the commit title. Or is there a better way?  I'm expecting that a Squash and Merge won't alter the original commit author info.

And I would add a "Resolves #1230"

leenipper:
I was mistaken; it looks like a Squash and Merge changes the commit author. I get that impression from the github UI.  I'm not sure there's a way to update the commit subject line while keeping @sbromberger as author, other than a local commit --amend and force push by @sbromberger.  Enlighten me if there is a way.

sbromberger:
I don't care about being an author. It's a simple enough change that someone else can probably just make it. I'd recommend just closing this out and making a new PR.

PR #1232, opened 2019-02-16
largest-series-product: Update cases_test.go

sbromberger:
Correct error messages to conform to expected results.

### DIFF
--- a/exercises/largest-series-product/cases_test.go
+++ b/exercises/largest-series-product/cases_test.go
@@ -20,10 +20,10 @@ var tests = []struct {
 	{"73167176531330624919225119674426574742355349194934", 6, 23520, true, ""},
 	{"0000", 2, 0, true, ""},
 	{"99099", 3, 0, true, ""},
-	{"123", 4, -1, false, "span must be smaller than string length"},
+	{"123", 4, -1, false, "string length must be greater than or equal to span"},
 	{"", 0, 1, true, ""},
 	{"123", 0, 1, true, ""},
-	{"", 1, -1, false, "span must be smaller than string length"},
+	{"", 1, -1, false, "string length must be greater than or equal to span"},
 	{"1234a5", 2, -1, false, "digits input must only contain digits"},
-	{"12345", -1, -1, false, "span must be greater than zero"},
+	{"12345", -1, -1, false, "span must be greater than or equal to zero"},
 }


bitfield:
This looks very helpful, @sbromberger, thank you!

One useful tip for Exercism PRs is it's a good idea to put the exercise name in the PR title, along with a very brief description of what the change does. For example:

> largest-series-product: make error messages more accurate/informative

Since almost all PRs to this repo consist of "updating cases_test.go", a little more information is really helpful to the reviewer.

exercises/largest-series-product/cases_test.go
```diff
@@ -20,10 +20,10 @@ var tests = []struct {
 	{"73167176531330624919225119674426574742355349194934", 6, 23520, true, ""},
 	{"0000", 2, 0, true, ""},
 	{"99099", 3, 0, true, ""},
-	{"123", 4, -1, false, "span must be smaller than string length"},
+	{"123", 4, -1, false, "string length must be greater than or equal to span"},
```
bitfield: This is correct, but it sounds sort of back to front. Would it be better to say:

> span must be smaller than or equal to string length

exercises/largest-series-product/cases_test.go
```diff
@@ -20,10 +20,10 @@ var tests = []struct {
...
 	{"", 0, 1, true, ""},
 	{"123", 0, 1, true, ""},
-	{"", 1, -1, false, "span must be smaller than string length"},
+	{"", 1, -1, false, "string length must be greater than or equal to span"},
```
bitfield: Ditto.

exercises/largest-series-product/cases_test.go
```diff
@@ -20,10 +20,10 @@ var tests = []struct {
...
 	{"1234a5", 2, -1, false, "digits input must only contain digits"},
-	{"12345", -1, -1, false, "span must be greater than zero"},
+	{"12345", -1, -1, false, "span must be greater than or equal to zero"},
```
bitfield: We could maybe phrase this more concisely as:

> span must not be negative

sbromberger:
I'm ok with whatever bikeshedding needs to happen. The problem for me was that the error messages are incorrect as currently specified, and that led to some confusing test results.

exercises/largest-series-product/cases_test.go
sbromberger: "smaller" is a bit of a problem, I think, as it is primarily used as a dimensionality measurement, not a magnitude measurement. I prefer "less than" or "greater than", and the order is interchangeable from my perspective. Thoughts?

exercises/largest-series-product/cases_test.go
sbromberger: I like that more.

leenipper:
Since the cases_test.go is a generated file based on the [problem specifications canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/largest-series-product/canonical-data.json), and the error message text actually originates in the .json, I prefer letting a PR against [problem-specifications](https://github.com/exercism/problem-specifications) make this change. That will help keep the generator maintenance sane, and let error text improvements be shared with other tracks.

PR #1236, opened 2019-03-20
Space Age: Add hint for implementing custom Planet type

BigChief45:
The current test cases use a `Planet` type for specifying a planet in the exercise. However, this `Planet` isn't defined anywhere, and the test cases actually pass a string to the function. When running the tests I get the following error:

```
./cases_test.go:9: undefined: Planet
```

I think this is a mistake? The type should be a string.

### DIFF
--- a/exercises/space-age/.meta/hints.md
+++ b/exercises/space-age/.meta/hints.md
@@ -8,3 +8,7 @@ for your solutions.
 One way to figure out what the function signature(s) you would need is to look
 at the corresponding \*\_test.go file. It will show you what the package level
 functions(s) should be that the test will use to verify the solution.
+
+## Planet Type
+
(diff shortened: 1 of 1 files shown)


bitfield:
Thanks for the report @BigChief45! This is intentional, as the student is supposed to define the `Planet` type in her solution. Maybe we need to make this a bit clearer in the problem description? Although it's a side exercise, it's one a lot of people tackle early on in the track, so some extra help might be good here.

In this case, we could add something to the `hints.md` file. Suggestions welcome!

BigChief45:
@bitfield Understood. I have added the hint as requested.

bitfield:
Yes, this looks good! Since this will likely be the first time a student has implemented a type, it may also be worth hinting that it can simply be a string.

BigChief45:
@bitfield Updated.

bitfield:
Nice work!

PR #1238, opened 2019-04-03
[hamming] Ignore error in test

hpurmann:
Executing `golangci-lint` gave me
```
hamming_test.go:36:12: Error return value of `Distance` is not checked (errcheck)
                        Distance(tc.s1, tc.s2)
                                ^
```
before this change.


bitfield:
Whenever I ignore errors with the blank identifier, I like to add a comment explaining _why_ this is OK. For example:

```go
// ignoring errors and results because we're just timing function execution
_, _ = Distance(tc.s1, tc.s2)
```

hpurmann:
@bitfield Thanks, I added your comment suggestion.

bitfield:
Sweet!

PR #1241, opened 2019-04-22
add meta hint to readme

### LINKED ISSUE #1239: space-age requires a type declaration and maybe shouldn't, or the hint should be exposed.
I did space-age today and I couldn't quite figure out what to do with the type errors I got, because we haven't been defining types yet.

```
./cases_test.go:9:14: undefined: Planet
```

I went and checked out the line at that file, and then thought it may be an error in the test file. I changed the `Planet` type to a `string` and everything worked. I came here to open an issue and found that not only was it intentional, it was talked about in a recent commit to the repo (https://github.com/exercism/go/pull/1236). The issue was opened with the same question I had.

As I am new to exercism, I think this might be a tad too early to make someone define a type. If that's not the case and this is how it is meant to be, the hint in the above mentioned commit is a good one. However, I could not figure out where that hint was supposed to be exposed - I never saw it.

Am I doing something wrong regarding hints and is there a better time to introduce a new type definition requirement?

What do you think @BigChief45 and @bitfield?

bensherman:
There was a hint added to the .meta file in #1236, but the commands to add it to the readme were not run. This PR fixes it.

Fixes #1239.

### DIFF
--- a/exercises/space-age/README.md
+++ b/exercises/space-age/README.md
@@ -28,6 +28,10 @@ One way to figure out what the function signature(s) you would need is to look
 at the corresponding \*\_test.go file. It will show you what the package level
 functions(s) should be that the test will use to verify the solution.
 
+## Planet Type
+
+The test cases make use of a custom `Planet` type that is sent to your function. You will need to implement this custom type yourself. Implementing this new custom type as a string should suffice.
+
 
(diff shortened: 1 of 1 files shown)


bitfield:
Every day, in every way, we're getting better and better!

PR #1242, opened 2019-04-23
isbn-verifier: add benchmark

h311ion:


kytrinyx:
Hi @h311ion, thanks so much for your PR! Was there a discussion about adding this benchmark anywhere? What was the motivation? (I'm not opposed, I just am wondering what led to the PR being opened).

h311ion:
Hi @kytrinyx, there is an a description in task that it may contain benchmarks but it's missing, so i decided to add them

PR #1245, opened 2019-05-06
fix: Update test cases for outdated exercises.

da-edra:
1. Why is this change neccesary?
Because some exercises like isogram allowed you to use a `break` a statement
inside a for loop and having all the tests pass, this can lead to some confusion
between `break` and `continue` statements that are inside a for loop.

With the current test suite one could do something like this:

```
for i, letter := range word {
		letter = unicode.ToLower(letter)
		if !unicode.IsLetter(letter) {
			break
		}
		if strings.ContainsRune(word[i+1:], letter) {
			return false
		}
	}
```

And have all the test pass.

This could be problematic as the `break` statement gives a better benchmark but
it's because of a coincidence with the test cases.

Other exercises had similar problems.

2. How does it address the issue?
The test cases for all of the exercises were updated.

3. What side effects does this change have?
We're up-to-date with the latest test cases! :)

### DIFF
--- a/exercises/isogram/cases_test.go
+++ b/exercises/isogram/cases_test.go
@@ -1,8 +1,8 @@
 package isogram
 
 // Source: exercism/problem-specifications
-// Commit: 7cea153 isogram: add test case with same first and last character (#1308)
-// Problem Specifications Version: 1.6.0
+// Commit: 74869e8 isogram: Add test case for dupe after non-letter (fixes #1390)
+// Problem Specifications Version: 1.7.0
 
 var testCases = []struct {
 	description string
@@ -48,6 +48,11 @@ var testCases = []struct {
 		description: "hypothetical isogrammic word with hyphen",
 		input:       "thumbscrew-japingly",
 		expected:    true,
+	},
+	{
+		description: "hypothetical word with duplicated character following hyphen",
+		input:       "thumbscrew-jappingly",
+		expected:    false,
 	},
 	{
 		description: "isogram with duplicated hyphen",
--- a/exercises/acronym/cases_test.go
+++ b/exercises/acronym/cases_test.go
@@ -1,8 +1,8 @@
 package acronym
 
 // Source: exercism/problem-specifications
-// Commit: 2e46654b Acronym: New test case for apostrophe (#1377)
-// Problem Specifications Version: 1.6.0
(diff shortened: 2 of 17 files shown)


da-edra:
I oringally pointed a [PR with a new test case](https://github.com/exercism/problem-specifications/pull/1517) to the main problems repository but, as @rpottsoh the problem was that the test cases for Go exercises were outdated and didn't cover that specific case.

I updated all of the test cases because I found similar problems in other exercises.

Thanks so much to @rpottsoh for pointing this out.

da-edra:
I'll gradually update the examples of each exercise so the test suite passes on Travis.

exercises/isogram/cases_test.go
```diff
@@ -69,4 +74,9 @@ var testCases = []struct {
 		input:       "angola",
 		expected:    false,
 	},
+	{
+		description: "a non letter before a repeating pattern",
+		input:       "red bat cat",
+		expected:    false,
+	},
```
leenipper: This is a good additional test case.  I see it has not made it into [master of isogram/canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/isogram/canonical-data.json) but is in your [branch](https://github.com/da-edra/problem-specifications/commit/f49969844fe8da2ba5a8a7b24d912ae32ed2ab61#diff-4a2164ce55eed791d3cb1cb39985696a). Perhaps re-open the PR for it there? You may need a less "Go specific" description/log message.

da-edra:
Hey @leenipper , yes, I totally agree with you, we should sync all the test cases to the `canonical-data.json` the `red bat cat` was added by mistake, I'm deleting it so we follow the test cases of all the other exercises.

exercises/isogram/cases_test.go
da-edra: Thank you very much Lee, I decided to delete this test as the `thumbscrew-jappingly` case already covers the  situation were a `break` and a `continue` can cause confusion inside a `for loop`.

da-edra:
Version `1.7.0` is `74869e8`, it's canonical right now. :smile:

exercises/isogram/cases_test.go
leenipper: Looks all good now. Thanks.

PR #1250, opened 2019-06-06
darts: add exercise

leenipper:


exercises/darts/README.md
```diff
@@ -0,0 +1,40 @@
...
+different specs, so the results from these benchmark tests may vary.
+
+## Further information
+
+For more detailed information about the Go track, including how to get help if
+you're having trouble, please visit the exercism.io [Go language page](http://exercism.io/languages/go/resources).
+
+## Source
+
+Inspired by an exercise created by a professor Della Paolera in Argentina
```
bitfield: Closing period?

exercises/darts/cases_test.go
```diff
@@ -0,0 +1,73 @@
...
+		description: "A dart lands outside the target",
+		x:           -9.0,
+		y:           9.0,
+		expected:    0,
+	},
+	{
+		description: "A dart lands just in the border of the target",
+		x:           0.0,
+		y:           10.0,
+		expected:    1,
```
bitfield: I think we need to clarify whether we're using what's called _inward scoring_ (touching a score line earns the higher score) or _outward scoring_ (touching a score line earns the lower score). Both methods are commonly used in shooting sports. Here, the test data suggests inward scoring (if you clip the 1-point line, you get 1 point).

A real dartboard avoids this issue by using lines made of wire; the dart goes one side or the other!

exercises/darts/README.md
leenipper: Its a generated line from configlet app, sourcing exercise metadata.yml.
Surveying all the problem-specifications metadata.yml shows 80% don't have periods in the "source:" property. Oh well.  I will leave it for now.

exercises/darts/cases_test.go
leenipper: Good point.  A real dartboard is different.

I could mention the inward scoring in a hints.md that gets brought into the README.md from configlet run.  Unless maybe this is a case for clarification in the problem-specifications description.md, helping all tracks.

leenipper:
Thanks @bitfield for your review.

leenipper:
@bitfield , I finally revisited this one and pushed some updates.

leenipper:
Improved the test case error message a bit.

PR #1251, opened 2019-06-12
added rune testscase on hamming distance problem

LucaPaterlini:


exercises/hamming/cases_test.go
```diff
@@ -64,4 +64,10 @@ var testCases = []struct {
 		0,
 		true,
 	},
+	{ // check for runes
+		"èà",
+		"èò",
```
leenipper: The cases_test.go is a generated file from the hamming [test case generator](https://github.com/exercism/go/blob/master/exercises/hamming/.meta/gen.go) whose input is the [problem specifications canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/hamming/canonical-data.json).  Because of the generated cases_test.go, I would normally suggest doing a pull request against problem-specifications repo to add the test case in the canonical-data.json.

However, since this test case addition is UTF-8, and not all language tracks can accept UTF-8  input. So, if this test case is to be added, the better approach is to put the new test case inside hamming_test.go; the extra test case can be defined with same struct signature which matches cases_test.go testCases. The for loops in hamming_test.go  can use an append(testCases, extraTestCases...) to easily incorporate the Go specific test case(s).

bitfield:
I'm 💯 behind the goals of this PR: Go's handling of Unicode characters is special, and can lead novices into trouble if they don't understand it properly.

The fundamental issue, for those who haven't been following closely, is that you can process strings as either byte sequences, or rune sequences. Provided you have only ASCII characters, then those both amount to the same thing. But if you have wide Unicode characters, then reading bytes doesn't give you valid characters, it just gives you nonsense bytes, and you lose 'rune sync'; corresponding characters in the two strings no longer line up properly.

> The hamming exercise is an early core exercise, and I'm not sure we should expect understanding of rune and UTF-8 so early in the Go language track.

I've changed my mind on this fairly recently. I used to tell students that in cases like this, where they can _guarantee_ that only ASCII characters will be in the input, then it's fine to read strings by bytes (and index them directly with `[]`). However, in other cases you need to use the other strategy (read runes).

Frankly, two things to learn is one too many, at this stage. My thinking now is that it makes sense to teach just one way, at least at the beginning: the _right_ way. When you're a Go guru, and you can advance all sorts of esoteric, performance-based arguments for using bytes, then fine, you do you. Until that time, my fail-safe teaching strategy is _always treat strings as rune sequences, not byte sequences_.

@leenipper is absolutely right that it's an early exercise, and this is a somewhat arcane thing to understand, especially because few other languages do it this way. But if you're using the Bitfield Method(TM), you'll want to teach runes right away. Otherwise, you can bet someone will say "How come it was fine to use bytes before, but now we're doing runes?"

Maybe the answer is to move it later in the track; it's a trivial exercise in languages where bytes == runes, but not in Go, if you're doing it right.

The question here is not just "should mentors guide students towards a rune-based solution to Hamming?", but "should we actually _enforce_ it with a test case?"

I think we've historically been rather against the idea of enforcing specific implementations with test cases, or even enforcing input validation by testing broken inputs. Also, bear in mind that this is not a generic comparing-random-strings problem. This is specifically a _comparing DNA strands_ problem. The input consists only of As, Cs, Gs, and Ts. So what's an `èà` test case for? It's invalid input, and we explicitly don't test input validation.

Maybe this is an alternate way to achieve more or less the same goals:

1. Leave Hamming as it is for now, but move it a step or two further down the track.
2. Update the mentor notes (if they're not already) to point out the problems with reading strings as bytes, and to suggest that mentors guide students towards always using runes.

Optionally:

3. Introduce another exercise which involves counting or comparing runes, but not involving DNA strands, so that wide characters are now fair play, and add test cases which would fail a byte-oriented solution.

I'd love to hear what @kytrinyx thinks too.

leenipper:
@bitfield , thank you for your excellent points.

I agree that hamming may not be the exercise to emphasize rune awareness.  I also agree that it is best to teach one way, and using rune in a for .. range is best when dealing with string elements.  In whatever exercise is applicable, a hints.md to augment the README.md with Go specific instruction about rune vs. byte will be helpful.

jmrunkle:

PR #1253, opened 2019-06-21
Add a new test case for the isogram exercise

robscotts4rb:
With the current test cases It was possible to stop processing if you detected a symbol character and the tests would all pass, this new test case stops that being a valid way of getting passing test.

### DIFF
--- a/exercises/isogram/cases_test.go
+++ b/exercises/isogram/cases_test.go
@@ -60,6 +60,11 @@ var testCases = []struct {
 		expected:    true,
 	},
 	{
+		description: "isogram with duplicated hyphen and duplicated character",
+		input:       "six-years-old",
+		expected:    false,
+	},
+	{
 		description: "made-up name that is an isogram",
 		input:       "Emily Jung Schwartzkopf",
 		expected:    true,


exercises/isogram/cases_test.go
```diff
@@ -59,6 +59,11 @@ var testCases = []struct {
 		input:       "six-year-old",
 		expected:    true,
 	},
+	{
+		description: "isogram with duplicated hyphen and duplicated character",
+		input:       "six-years-old",
+		expected:    false,
+	},
```
leenipper: This looks like a useful test case addition.

The cases_test.go is a generated file from the isogram test case [generator](https://github.com/exercism/go/blob/master/exercises/isogram/.meta/gen.go) whose input is the [problem specifications canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/isogram/canonical-data.json). Because the cases_test.go is a generated file, I suggest doing a pull request against [problem-specifications repo](https://github.com/exercism/problem-specifications) to add the test case in the canonical-data.json. Then, after the PR is merged there, a PR here using the output of a run of the mentioned generator will get the cases_test.go updated. Updating the canonical-data.json first allows other tracks which use a generator to have their test cases updated as well. And the go track generator maintenance stays "in sync".

jmrunkle:

Issue #1258, opened 2019-07-02
Additional test case for exercise Clock


udhos:
Hi,

I would like to suggest an extra test case for exercise Clock:

```
func TestAddAndCompare(t *testing.T) {
        clock1 := New(15, 45).Add(16)
        clock2 := New(16, 01)
        if !reflect.DeepEqual(clock1, clock2) {
                t.Errorf("clock.New(15,45).Add(16) differs from clock.New(16,01)")
        }
}
```

What do you think?

udhos:
Just for the records, the underlying problem this test case addresses is: current test cases depend on .String(). I have seen a wrong implementation of .Add() and .Subtract() that has been masked by a .String() that fixed the error. That is:

- .Add(): wrong result
- .Subtract(): wrong result
- .Add().String(): correct result
- .Subtract().String(): correct result

However the tests did pass.

PR #1259, opened 2019-07-02
Tests for comparing clock after .Add() and .Subtract().

udhos:
Fixes https://github.com/exercism/go/issues/1258

### DIFF
--- a/exercises/clock/clock_test.go
+++ b/exercises/clock/clock_test.go
@@ -74,6 +74,22 @@ func TestCompareClocks(t *testing.T) {
 	t.Log(len(eqTests), "test cases")
 }
 
+func TestAddAndCompare(t *testing.T) {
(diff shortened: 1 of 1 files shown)


udhos:
I don't think so. This is actually already tested but there is a flaw because the tests depend on the method .String(). See this comment:

https://github.com/exercism/go/issues/1258#issuecomment-508094074

udhos:
We need some tests for .Add() and .Subtract() that don't depend on .String().

leenipper:
@udhos , my comment about the "extra validation" was intended as positive; the tests done for equality after an Add or Subtract (without depending on .String) are useful to insure two clocks representing the same time are equal.  It's a good change.  Sorry for any confusion.

I will merge it now.  I was only waiting to see if there were any more review comments.  Thanks again.

PR #1262, opened 2019-07-20
Eliminate trailing newline in twelve-days

IsaacG:
Eliminate trailing newline in twelve-days

### DIFF
--- a/exercises/twelve-days/example.go
+++ b/exercises/twelve-days/example.go
@@ -50,9 +50,9 @@ func Verse(i int) string {
 }
 
 func Song() string {
-	var song []string
+	var song = make([]string, 12)
 	for i := 1; i <= 12; i++ {
-		song = append(song, Verse(i))
+		song[i-1] = Verse(i)
 	}
 	return strings.Join(song, "\n")
 }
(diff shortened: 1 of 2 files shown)


IsaacG:
String appends are quite as strings are immutable. Each time you do an append, you need to copy the entirety of both strings to a new location. If you got 10 verses, that's 10 copies of the first verse, 9 of the second verse, etc. With an array, you only have two copies of each string - in the array and in the final string. (Well, plus each time the array needs to grow...) The most efficient way to build a string is with a string builder. An array with a preset capacity is pretty close to that.

Additionally, the trailing newline feels like an artifact of the implementation and not something that ought to actually be there. When I call a function to generate a string for something, I tend to not expect that it includes a trailing newline. This allows the caller to control how it's used or printed, in regards to formatting around it.

exercises/twelve-days/example.go
```diff
@@ -49,9 +50,9 @@ func Verse(i int) string {
 }
 
 func Song() string {
-	var song = ""
+	var song []string
```
leenipper: Since the slice capacity is known:
```diff
-       var song []string
+       song := make([]string, 0, 12)
```

PR #1266, opened 2019-08-08
clock - run all add and subtract tests while not depending on correct .String()

udhos:


udhos:
Issue is, saw today another wrong implementation that did pass all the tests only because method .String() fixed the error.

tehsphinx:
Did a quick check. Seems legit.

PR #1268, opened 2019-08-20
twelve days: add benchmark

mcaci:
Hello all,

While working on the "Twelve days" exercise I have created a small benchmark in local with the goal of testing two different approaches:
1) Creating verse by appending strings via "+" operator
2) Creating verse by creating a strings.Builder and appending via the WriteString method.

My idea is that these benchmarks could help those who would like to see the difference in performance for these and potentially other approaches. Would it make sense to you to add them in the test suite?


mcaci:
I'm very happy I could give my contribution, even though a small one :)
Thank you @leenipper !

Issue #1265, opened 2019-08-03
Create stub files for all exercises


kytrinyx:
We have decided to require all file-based tracks to provide stubs for their exercises.

The lack of stub file generates an unnecessary pain point within Exercism, contributing a significant proportion of support requests, making things more complex for our students, and hindering our ability to automatically run test-suites and provide automated analysis of solutions.

We believe that it’s essential to understand error messages, know how to use an IDE, and create files. However, getting this right as you’re just getting used to a language can be a frustrating distraction, as it can often require a lot of knowledge that tends to seep in over time. At the start, it can be challenging to google for all of these details: what file extension to use, what needs to be included, etc. Getting people up to speed with these things are not Exercism’s focus, and we’ve decided that we are better served by removing this source of confusion, letting people get on with actually solving the exercises.

The original discussion for this is at https://github.com/exercism/discussions/issues/238.

Therefore, we’d like this track to provide a stub file for each exercise.

- If this track already provides stub files for all exercises, please close this issue.
- If this track already has an open issue for creating stubs, then my apologies. Please close one as a duplicate.
- Otherwise, please respond to this issue with useful details about what needs to be done to complete this task in this track so that people who are not familiar with the track may easily contribute.

leenipper:
For exercises currently without stubs, I'm thinking that providing a one line stub file which has the correct "package name" line matching the package name in the _test.go file is reasonable.

There are two README.md files (exercises raindrops and space-age) which need the '# No Stub' section replaced/reworded, since the stub file will be there, but with less content than prior exercises in the track.  The .meta/hints.md file is the source of the '# No Stub' section, so edit the .meta/hints.md and re-run configlet to re-generate the README.md for the exercise, as in
```
./bin/fetch-configlet
./bin/configlet generate $PWD --only space-age
``` 

There is no need to provide stub files for the deprecated exercises (binary, counter, hexadecimal, octal and trinary); the config.json has these marked ```"deprecated": true```.

The stub file name must match the exercise directory name, except replace any dash(-) with underline(_) in the stub file name.

Inside each new stub file, place one line with
```package name```
where ```name``` matches the package name in the *_test.go file.

Unless I missed something, that's it.

kytrinyx:
I think this sounds like an excellent plan!

PR #1269, opened 2019-08-21
Add stub files

leenipper:
Adds the missing stub files to address issue #1265 with updates to applicable README.md for two exercises.


bitfield:
One thing which occurs to me is that Parallel Letter Frequency already has a kind of 'stub file' (`frequency.go`), and this is very often the cause of confusion. Students will take the FreqMap type and Frequency function defined there, copy it into their own solution file, and then (to avoid a duplicate definition error) delete the original `frequency.go`. Of course when they submit the solution and a mentor downloads it, the `frequency.go` is still present, meaning the solution doesn't compile. This breaks Exalysis, for example.

Should we make it clearer that students are intended to add their code to `frequency.go` somehow?

bitfield:
That sounds fine; in which case, let's delete the `frequency.go` file from the repo and move the Frequency/FreqMap code into the stub file. They are not actually stubs as such, but the student's code is required to call them.

bitfield:
In that case, should we add them to the example file as well? Or will that simply cause more duplicate declaration errors?

leenipper:
That works. No duplicate declaration errors, since use of example.go is always without the stub file.
And it should solve the problem with students doing the wrong thing.

leenipper:
I simplified the new section in the README.md files.  Sometimes saying less is better.

leenipper:
@bitfield , take another look, time permitting.  Thanks.

leenipper:
@bitfield , ok for merging ?

PR #1271, opened 2019-08-25
two-bucket:  add new complexity bonus test case

LucaPaterlini:
this will force the reader to an optimal solution

### DIFF
--- a/exercises/two-bucket/bonus_test.go
+++ b/exercises/two-bucket/bonus_test.go
@@ -1,4 +1,3 @@
-
 // +build bonus
 
 package twobucket
--- a/exercises/two-bucket/bonus_test.go
+++ b/exercises/two-bucket/bonus_test.go
@@ -13,6 +13,9 @@ var noSolutionCases = []bucketTestCase{
 		"No solution case 2",
 		3, 6, 1, "one", "", 0, 0, true,
 	},
+	{	"Adding Complexity",
+		3333333, 6666667, 1, "one", "one", 19999992, 6666667, false,
+	},
 }
 
 func TestSolve_bonus(t *testing.T) {
--- a/exercises/two-bucket/bonus_test.go
+++ b/exercises/two-bucket/bonus_test.go
@@ -1,3 +1,4 @@
+
 // +build bonus
 
 package twobucket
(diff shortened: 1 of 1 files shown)


exercises/two-bucket/bonus_test.go
```diff
@@ -13,6 +13,9 @@ var noSolutionCases = []bucketTestCase{
 		"No solution case 2",
 		3, 6, 1, "one", "", 0, 0, true,
 	},
+	{	"Adding Complexity",
```
leenipper: travis-ci failed for a "go fmt".  Can be fixed with:
```
-       {       "Adding Complexity",
+       {
+               "Adding Complexity",
```

Also suggest changing var noSolutionCases ... to var bonusCases ...,  since a solution can be found for the new test case.

LucaPaterlini: - checked with go fmt
- added the new variable and the new test case label

bitfield:
Sorry, without any context I've no idea what problem this PR is trying to solve. Is it intended to eliminate a solution which passes the existing test cases, but is pathologically non-performant? If so, what?

LucaPaterlini:
@bitfield  I think I can try to answer then, with the existing set of test cases even using a brute force approach or using a solution with exponential complexity can give a solution in a fraction of a second, the test I have proposed can be solved by a retail computer in around a second in case the solution used have O(mn) complexity where m is the size of the first bucket and n the size of the second bucket

solution proposed ( looks more performant than the example one ) 

O(nm) using dynamic programming

https://exercism.io/tracks/go/exercises/two-bucket/solutions/af03104f53464c14b9e7f97a540f5887

bitfield:
Forgive me for asking the obvious question, but if even a brute-force solution gives the correct answer in a fraction of a second, then what's the problem with a brute-force solution?

LucaPaterlini:
The point is, how can the reader be pushed to optimize the code of a not monitored exercise if any complexity solution make him feel he/or she (likely a male, gender gap its still there) like he/she have solved it?

bitfield:
It sounds like what you're saying is that without additional tests, it's possible to solve this exercise using an algorithm which is of sub-optimal complexity, and thus wouldn't scale to more demanding test cases (this is what I meant by 'pathologically non-performant' in my previous comment). Is that right?

An example from another exercise would be implementing 'Difference of Squares' with loops. Compared to using a formula, the performance is egregiously bad. An analogous test case for that exercise would be very large N inputs.

Assuming I'm understanding what this is about, how does the proposed test case solve this problem? I mean, if it takes a long time to run the tests, but they pass, how does the student know that their solution is unacceptable? And if the tests time out, how does the student know that their inefficient algorithm is the cause?

LucaPaterlini:
answering the last question, we have already set the expectations in students to have their test solved in a few seconds on the average computer, we can add a note in the problem question that if the test execution it's taking more than few seconds they have to think more carefully at the complexity of it.

This expectation comes from the actual habit we have shown them with all the other tests, I thought to suggest complexity checks was an easier battle

bitfield:
The reason I'm asking such hard questions is not because I think it's a bad idea to encourage students towards efficient solutions. I'm completely in favour of that. What I'm not sure about is whether test cases are necessarily the best way to achieve it.

If we had a test that would fail an inefficient solution with a message like this:

```
Looks like your solution would eventually pass this test, but it's not as efficient as it could be. Would you like to explore an alternative solution which scales much more effectively to larger inputs?
```

that would be ideal. Unfortunately, it's really hard to write such tests. The best way I know of to guide students down this road is with mentor feedback. I could give you examples from practically every exercise on the core track of solutions which pass the tests, but are nevertheless far from optimal, and if I could have written test cases which would make them so, I certainly would have!

My suspicion, based on experience, is that if the tests take a long time to run, students will either simply ignore this fact, or say that the tests appear to be broken. So do you think there is a way to detect an inefficient solution _at the test level_ and issue a suitably helpful message?

jmrunkle:

Issue #1270, opened 2019-08-23
New exercise: Tower of Hanoi


udhos:
I would like to contribute an additional exercise for the Go track.

Exercise **Tower of Hanoi** is sketched here:

https://github.com/udhos/exercism-go/tree/master/tower-of-hanoi

If you think it would fit in, please let me know how to contribute it.

Thanks!

leenipper:
Hello @udhos . Thanks for the interest in adding an exercise. And you have a good start on it for the Go track of exercism!

The Tower of Hanoi is a classic programming problem and I think it is worth consideration in Exercism for all language tracks.

New generic exercises enter first via the [problem specifications repo](https://github.com/exercism/problem-specifications).  See this helpful document for [implementing a new exercise](https://github.com/exercism/docs/blob/master/you-can-help/make-up-new-exercises.md).   Refer to this general guide on [Contributing](https://github.com/exercism/problem-specifications/blob/master/CONTRIBUTING.md), particularly the section on [Implementing a Completely New Exercise](https://github.com/exercism/problem-specifications/blob/master/CONTRIBUTING.md#implementing-a-completely-new-exercise).

In the referenced documents, you see the term "slug", which is just the name of the exercise. In this case, the name would be all lower case, tower-of-hanoi.  That sets the name of the new directory in the problem-specifications.  Follow the structure of another exercise in problem-specifications to use as an example, such as [reverse-string](https://github.com/exercism/problem-specifications/tree/master/exercises/reverse-string). So, it's just a new directory within exercises, such as tower-of-hanoi,  and three standard named files within that directory: canonical-data.json,  description.md, and metadata.yml. The canonical-data.json, defines test cases primarily. For tower-of-hanoi, the "property" for each test case is likely just "solve". And the "input" property would perhaps follow what you have in your readme, with nested items such as "disks", "source", "helper", and "target" disk names.  Use your wikipedia.org link as the "source_url" in the metadata.yml.

**I suggest cloning [problem specifications repo](https://github.com/exercism/problem-specifications) and submitting an issue and/or PR there first** to get things rolling.  It's best if new exercise PR in problem-specifications is merged first, since there may be changes which impact the language track exercises implemented.

Feel free to ask questions here, and have fun!

udhos:
Hi @leenipper.

I initially considered to submit the problem for all tracks. However, this issue led me to believe the specification repo is frozen for new exercises:

https://github.com/exercism/problem-specifications/issues/1560

[...]

PR #1272, opened 2019-08-28
New exercise: Tower of Hanoi.

udhos:
See https://github.com/exercism/go/issues/1270

### DIFF
--- a/exercises/tower-of-hanoi/hanoi.go
+++ b/exercises/tower-of-hanoi/hanoi.go
@@ -13,8 +13,11 @@ package towerofhanoi
 // A move is a [2]int value specifying source and destination rods.
 //
 // Example
-// Solve(1, 20, 30, 40) must return list with a single move:
-//  [][2]int{{20, 30}} // move disk from rod 20 to rod 30
+// Solve(1, 'A', 'B', 'C') could return a list with a single move:
+//  [][2]int{{'A', 'B'}} // move disk from rod 'A' to rod 'B'
+//
+// Hint: Use 'go test -v' to see the generated moves.
+//
 //func Solve(disks, from, to, via int) [][2]int {
 //    return nil
 //}
--- a/exercises/tower-of-hanoi/hanoi.go
+++ b/exercises/tower-of-hanoi/hanoi.go
@@ -4,10 +4,10 @@ package towerofhanoi
 // implement the function Solve below.
 
 // Solve returns list of moves for solving the Tower of Hanoi
-// puzzle. 'disks' is the amount of disks. 'from' is source
-// rod, 'aux' is the helper rod, 'to' is the destination rod.
+// puzzle. 'disks' is the amount of disks. 'from' is the source
+// rod, 'to' is the destination rod, 'via' is a helper rod.
 // If Solve is called as Solve(1, "a", "b", "c"), it should
-// return []string{"a -> c"}.
-//func Solve(disks int, from, aux, to string) []string {
+// return []string{"a -> b"}.
+//func Solve(disks int, from, to, via string) []string {
 //	return nil
 //}
--- a/exercises/tower-of-hanoi/hanoi.go
+++ b/exercises/tower-of-hanoi/hanoi.go
@@ -13,8 +13,18 @@ package towerofhanoi
 // A move is a [2]int value specifying source and destination rods.
 //
 // Example
-// Solve(1, 'A', 'B', 'C') could return a list with a single move:
-//  [][2]int{{'A', 'B'}} // move disk from rod 'A' to rod 'B'
+//
+// In order to illustrate the problem, consider the constants below
+// naming the rods. Your implementation is free to define similar
+// constants, but it is not required to do so.
+// const (
+// 	A = 'A'
+// 	B = 'B'
+// 	C = 'C'
+// )
+//
+// Solve(1, A, B, C) could return a list with a single move:
+//  [][2]int{{A, B}} // move disk from rod A to rod B
 //func Solve(disks, from, to, via int) [][2]int {
 //    return nil
 //}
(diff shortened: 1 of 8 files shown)


bitfield:
Would one way to simplify this exercise slightly be to assume that the rods are always called "A", "B", and "C"? Then the solution program no longer needs those inputs, and the only input would be the number of disks.

bitfield:
Yes, I see now, this is a pretty standard thing with Towers of Hanoi solvers. How about calling them `from`, `to`, and `via`, then?

bitfield:
It doesn't seem very Go-like to use strings for the rod names. How about defining constants:

```go
const (
	From = iota
	To
	Via
)
```

udhos:
Pushed a change replacing rods names with numeric IDs.

bitfield:
I'm not sure the numeric IDs are really an improvement over A, B, and C. Given that the key insight in solving this puzzle is swapping around the From, To, and Via rods in recursive calls to Solve, shouldn't we name them `From`, `To`, and `Via` as suggested? If you can come up with better names, by all means do!

udhos:
Well, the func parameters were named 'from', 'to', 'via' already, but their type was string. I thought the point of defining numeric constants was to change their type from string id to numeric id.

So I think I am misunderstanding something.

Can you show the interface you have in mind for the function?

bitfield:
Yes, absolutely: the signature is fine, but I'm suggesting we use the named constants instead of 10, 20, 30, etc. Actually, now that I understand the problem a bit more, I think the names A, B, and C make sense:

```
Your implementation will accept the amount of disks on rod A, then issue a list of moves for solving the puzzle.

Example:

Input:
- number of disks: 2

Output: list of three moves: {A,C}, {A,B}, {C,B} 
```

udhos:
Oh, now I see what you mean: names used in the exercise description.
I am trying to figure how to use the constants in the problem description while not suggesting to the student that those constants would actually help in the implementation, since the recursive solution itself would not benefit from them.
Please see a sketch below.
What do you think?

```
Your implementation will accept the amount of disks on rod A, then issue a list of moves for solving the puzzle.

// Consider the constants below naming the rods, for the sake of the problem description.
// Your implementation is free to define similar constants, but this is not a requirement.
const (
	A = iota
	B
	C
)

Example:

Input:
- number of disks: 2

Output: list of three moves: {A,C}, {A,B}, {C,B}
```

bitfield:
Yes, for the sake of generating readable output, it's maybe better to define, for example:

```go
const A = 'A'
...

```

What do you think?

udhos:
I added pretty printing to tests.
Now we can keep rods names like A,B,C in the generic description, and tests would report similar names.

go test passing:
```
    hanoi_test.go:80: disks=7 move=[A B] before: {[1] [7 6 5 4 3 2] []}
    hanoi_test.go:140: disks=7 move=[A B] after: {[] [7 6 5 4 3 2 1] []}
    hanoi_test.go:31: PASS totalDisks=7
```

go test failing:
```
    hanoi_test.go:80: disks=7 move=[A B] before: {[7 6 5 4 3 2 1] [] []}
    hanoi_test.go:140: disks=7 move=[A B] after: {[7 6 5 4 3 2] [1] []}
    hanoi_test.go:28: FAIL totalDisks=7 rod TO=B has 1 disks, should have 7: [1]
```

In Go we state:

```
// Example
// Solve(1, 'A', 'B', 'C') could return a list with a single move:
//  [][2]int{{'A', 'B'}} // move disk from rod 'A' to rod 'B'
```

udhos:
I am unsure about how much information we should put in the stub file 'hanoi.go'.

Can you guys please have a look at how it looks?

bitfield:
How come you closed this, @udhos?

udhos:
Sorry, I guess I closed it by accident.

leenipper:
> how much information we should put in the stub file 'hanoi.go'

It looks ok I think.  @bitfield ?

exercises/tower-of-hanoi/hanoi.go
```diff
@@ -0,0 +1,30 @@
+package towerofhanoi
+
+// This is an stub file. Remove this comment and
```
bitfield:
```suggestion
// This is a stub file...
```

exercises/tower-of-hanoi/hanoi.go
```diff
@@ -0,0 +1,30 @@
...
+// implement the function Solve below.
+
+// Solve returns list of moves for solving the Tower of Hanoi puzzle.
+//
+// 'disks' is the amount of disks.
+// 'from' is the id for the source rod.
+// 'to' is the id for the destination rod.
+// 'via' is the id for the helper rod.
+//
+// A move is a [2]int value specifying source and destination rods.
```
bitfield: This strikes me as a little weird. I've literally never used an array in Go. Would a struct `{from, to int}` be a bit more idiomatic?

exercises/tower-of-hanoi/hanoi.go
udhos: Added as https://github.com/exercism/go/pull/1272/commits/bc507303b28de3db175928d057303032d4b8c3f2

Unsure whether you suggested the full change or just the grammar issue.

exercises/tower-of-hanoi/hanoi.go
udhos: Please see https://github.com/exercism/go/pull/1272/commits/1741aab3426a00e8173927631eeed43300ab8b13

CoderYihaoWang:
This looks like a very interesting exercise and as a learner I would love to see it on the website. But it seems that it has been ignored since the last modification . Is this abandoned? @leenipper @bitfield @udhos

bitfield:
> Is this abandoned?

So it seems!

jmrunkle:

PR #1285, opened 2019-10-21
space-age: update cases_test.go

odubno:
`Planet` is not defined as a type. Change it to be `string`.

### DIFF
--- a/exercises/space-age/cases_test.go
+++ b/exercises/space-age/cases_test.go
@@ -6,7 +6,7 @@ package space
 
 var testCases = []struct {
 	description string
-	planet      Planet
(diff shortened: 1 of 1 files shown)


olehdubno:
`type Planet string` could get defined inside of `space-age.go`, however it's not immediately clear that we have to do that. Maybe, adding `type Planet string` to `space-age.go`, from the beginning? instead?

odubno:
`type Planet string` could get defined inside of space-age.go, however it's not immediately clear that we have to do that. Maybe, adding `type Planet string` to space-age.go, from the beginning? instead?

PR #1286, opened 2019-10-25
robot-name: change test structure for bonus fix

h311ion:
Bonus test fails on proposed test command - `go test -tags bonus`
It's happens because we are taking all available robot names in bonus test, so base test fails
I have changed test structure - moved common parts to separate file and added exclusion build tag to base tests, so running the command `go test -tags bonus` executes only bonus test.

### DIFF
--- a/exercises/robot-name/common_test.go
+++ b/exercises/robot-name/common_test.go
@@ -1,0 +1,26 @@
+package robotname
+
+import (
+	"regexp"
+	"testing"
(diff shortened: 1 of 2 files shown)


hans-d:
when is this expected to be merged ?

leenipper:
@h311ion , @hans-d, Sorry for the delay before merge. I had waited for a second approval.

PR #1303, opened 2020-01-09
two-bucket: add another unsolvable test case to bonus set

jonmcoe:
The existing two test cases can be addressed by checking that

```
if largerCapacity%smallerCapacity == 0 && goal%smallerCapacity != 0 {
```

but this is incomplete. the addition of this test case requires consideration of the GCD between values as in

```
       bucketsGcd := gcd(bucketOne, bucketTwo)
       if bucketsGcd > gcd(bucketsGcd, goal) {
```
(or an entirely different approach based on checking states within the main step-performing loop)


jmrunkle:

PR #1306, opened 2020-01-14
Add maintainer: tehsphinx

tehsphinx:


config/maintainers.json
```diff
@@ -90,6 +90,16 @@
       "link_url": "https://www.sborza.com/",
       "avatar_url": null,
       "bio": "I'm a Canadian (based in Chicago) golang, C and python developer working on projects within the gaming, finance and data-science worlds respectively. My favorite freetime task is coding challenges and I'm more than happy to help people talk through solutions!"
+    },
+    {
+      "github_username": "tehsphinx",
+      "alumnus": true,
```
iHiD: I think this should be `false` unless there's bad news you've not told me!

leenipper: Yes, anyone intending to be active is not an alum.  Good catch @iHiD .  And @sebito91 I missed this property on yours, so you will want to correct it.

tehsphinx:
@iHiD fixed :)

I guess I did not properly read the translations when looking it up. :D

iHiD:
Phew!!

PR #1309, opened 2020-01-15
small fix to remove alumnus flag from self

sebito91:
Should not have marked this in the beginning, apologies!


tehsphinx:
I think I copied it from yours and had that in my original PR as well. 😄

PR #1310, opened 2020-01-16
tournament: use sort.Slice helper

ferhatelmas:
also, declare some function parameters more idiomatically.


ferhatelmas:
@tehsphinx thanks for review.

PR #1314, opened 2020-01-16
ci: enable a few more linters and fix issues

ferhatelmas:


exercises/list-ops/example.go
```diff
@@ -43,10 +43,9 @@ func (s IntList) Filter(fn predFunc) IntList {
...
 	// for fun, let's do this with recursion
 	var lengthAcc func(IntList, int) int
-	var isEmpty func(IntList) bool
 	// yeah, this is kind of cheaty, but it's hard to determine if a slice is
 	// empty without using len, so walling it off
-	isEmpty = func(l IntList) bool {
+	isEmpty := func(l IntList) bool {
 		return len(l) == 0
 	}
 	lengthAcc = func(lst IntList, acc int) int {
```
sebito91: Do we want to just do the same thing here, define `lengthAcc := func(lst...)`?

ferhatelmas: not possible with recursive functions

sebito91: Ah, fair enough...didn't look into the rest of the example. Should probably do that next time, eh?

exercises/poker/example.go
```diff
@@ -194,7 +194,7 @@ func isStraight(counts []rankCount) bool {
 	}
 	first := counts[0].rank
 	if first == Ace {
-		for i := 1; i < len(counts); i++ {
+		for i := 1; i < len(counts); i++ { //nolint: staticcheck
```
tehsphinx: What is `staticcheck` complaining about here? The start from 1??

ferhatelmas: not updating loop counter, false positive - need to report.

exercises/list-ops/example.go
ferhatelmas: yes, I think it should be simplified to for loop but out of the scope this PR.

ferhatelmas: https://github.com/exercism/go/issues/1316

ferhatelmas:
@tehsphinx @sebito91 thanks for review :+1:

Issue #1304, opened 2020-01-09
Recruiting additional maintainers for Go


kytrinyx:
We're about to start a big push towards version 3 (v3) of Exercism. This is going to be a really exciting step forward for Exercism, with in-browser coding, new Concept Exercises with automated feedback, improved mentoring and much more.

This to be a big community effort, with the work spread out among hundreds of volunteers across Exercism. One key thing is going to be each track having enough maintainers who have the time to manage that community effort. We are therefore putting out a call for new maintainers to bolster our numbers. We're hoping that our existing maintainers will be able to act as mentors to the newer maintainers we add, and take on a parental role in the tracks.

If you are an existing maintainer, could you please reply to this letting us know that you think you'll have time (2-3hrs/week) to help with this over the next 6 months. If you won't have that time, but still want to be a maintainer and just help where you can instead, please tell us that too. If you have come to the end of the road as a maintainer, then we totally understand that and appreciate all your effort, so just let us know.

For anyone new who's interested in becoming a maintainer, thanks for your interest! Being an Exercism maintainer is also a great opportunity to work with some other smart people, learn more about your language of choice, and gain useful skills and experience that are useful for growing your career in the technical leadership direction. Please write a comment below introducing yourself along with your Exercism handle, and telling us why you're interested in becoming a maintainer, and any relevant experience. We will then evaluate every application and contact you using your exercism email address once we have finished the evaluation process.

Thank you!

See also https://github.com/exercism/exercism/issues/5161

sebito91:
I'm an existing maintainer but haven't been holding my weight over the past few months. I'm committing myself to 2-3hrs/week over the next few months!

Prounckk:
Hi, I'd like to be part of this!

tehsphinx:
I'll also like to be part of this!

hans-d:
I'd like to be part of it as well. 
Same handle on exercism, mentoring on go, and known to @tehsphinx as well, and already diving into some of the code involved.

ZoltanOnody:
Hi, I'd like to contribute as well.

I'm a Software Engineer at Exponea, where I'm working on a proprietary database engine for real-time data analyses. I'm working mostly with C++, Go, Python and k8s. My handle on exercism is `ZoltanOnody` and I've been mentoring the Go track in the past.

I'd like to become a Go track maintainer:
- to improve my Go skills (by working on problems outside of my domain),
- to give back to the Go community,
- to work on something useful in my free time (I really like the Exercism's mission),
[...]

PR #1315, opened 2020-01-16
prime-factors: accept order-independent results

### LINKED ISSUE #1295: prime-factors: fails despite valid output
Since the test file use `reflect.DeepEqual`, the slice values must appear in the exact same order to be considered equal. This is poor behaviour because the requirements do not specify ordering. 

Thus `int64[]{2, 3, 2}` fails for the expected output `[]int64{2, 2, 3}`. 

Solution: 
1. Custom implementation of equality that doesn't rely on order.
2. Sort the slice before using `reflect.DeepEqual` in the test harness.

JackMordaunt:
Fix #1295 

Simple sort implementation applied to the list of primes before testing for equality.

### DIFF
--- a/exercises/prime-factors/prime_factors_test.go
+++ b/exercises/prime-factors/prime_factors_test.go
@@ -9,7 +9,8 @@ import (
 func TestPrimeFactors(t *testing.T) {
 	for _, test := range tests {
 		actual := Factors(test.input)
-		sort.Sort(ascending(actual))
+		sort.Slice(actual, ascending(actual))
+		sort.Slice(test.expected, ascending(test.expected))
 		if !reflect.DeepEqual(actual, test.expected) {
 			t.Fatalf("FAIL %s\nFactors(%d) = %#v;\nexpected %#v",
 				test.description, test.input,
@@ -27,8 +28,8 @@ func BenchmarkPrimeFactors(b *testing.B) {
 	}
 }
 
-type ascending []int64
-
-func (l ascending) Len() int             { return len(l) }
-func (l ascending) Swap(ii, jj int)      { l[ii], l[jj] = l[jj], l[ii] }
-func (l ascending) Less(ii, jj int) bool { return l[ii] < l[jj] }
+func ascending(list []int64) func(int, int) bool {
+	return func(ii, jj int) bool {
+		return list[ii] < list[jj]
+	}
+}
--- a/exercises/prime-factors/prime_factors_test.go
+++ b/exercises/prime-factors/prime_factors_test.go
@@ -1,15 +1,15 @@
 package prime
 
-// Return prime factors in increasing order
-
 import (
 	"reflect"
+	"sort"
 	"testing"
 )
 
 func TestPrimeFactors(t *testing.T) {
 	for _, test := range tests {
 		actual := Factors(test.input)
+		sort.Sort(ascending(actual))
 		if !reflect.DeepEqual(actual, test.expected) {
 			t.Fatalf("FAIL %s\nFactors(%d) = %#v;\nexpected %#v",
 				test.description, test.input,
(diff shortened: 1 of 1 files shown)


exercises/prime-factors/prime_factors_test.go
```diff
@@ -1,15 +1,15 @@
...
 import (
 	"reflect"
+	"sort"
 	"testing"
 )
 
 func TestPrimeFactors(t *testing.T) {
 	for _, test := range tests {
 		actual := Factors(test.input)
+		sort.Sort(ascending(actual))
```
ferhatelmas: IMHO, using [sort.Slice](https://golang.org/pkg/sort/#Slice) would make it simpler

ferhatelmas:
> Note: Assumes the test cases are already sorted in ascending order.

let's ensure it too since we dropped order requirement, it would be safer.

exercises/prime-factors/prime_factors_test.go
```diff
@@ -1,15 +1,16 @@
...
 	"reflect"
+	"sort"
 	"testing"
 )
 
 func TestPrimeFactors(t *testing.T) {
 	for _, test := range tests {
 		actual := Factors(test.input)
+		sort.Slice(actual, ascending(actual))
+		sort.Slice(test.expected, ascending(test.expected))
```
tehsphinx: I think this needs to be:

```suggestion
		sort.Slice(test.expected, ascending)
```

Same above.

`func Slice(slice interface{}, less func(i, j int) bool)` expects a function as second argument.

JackMordaunt: Yes, my function returns a closure.

tehsphinx: see also the example on https://golang.org/pkg/sort/#Slice

exercises/prime-factors/prime_factors_test.go
JackMordaunt: I tried in-lining the sorting funcs, but it gets verbose: 

```go
sort.Slice(actual, func(ii, jj int) bool {return actual[ii] < actual[jj]})
sort.Slice(test.expected, func(ii, jj int) bool {return test.expected[ii] < test.expected[jj]})
```

So I opted for a named func. Similar in LOC to the previous sort impl.

exercises/prime-factors/prime_factors_test.go
tehsphinx: right...  my bad. I guess I'm tired :D

exercises/prime-factors/prime_factors_test.go
ferhatelmas: thanks, a func sounds good.

exercises/prime-factors/prime_factors_test.go
```diff
@@ -26,3 +27,9 @@ func BenchmarkPrimeFactors(b *testing.B) {
 		}
 	}
 }
+
+func ascending(list []int64) func(int, int) bool {
+	return func(ii, jj int) bool {
+		return list[ii] < list[jj]
```
ferhatelmas: why not `i` and `j` ?

ferhatelmas:
@JackMordaunt removed note regarding expected from the PR description.

exercises/prime-factors/prime_factors_test.go
JackMordaunt: It's a habit of mine. Easier to identify / search for `ii` than `i`, which appears in many English words.

ferhatelmas: I think it's better, let's change it then good to go. Thanks!

ferhatelmas:
@JackMordaunt awesome contribution, thanks a lot. 

There are a lot to help if you would like to do more. Be sure to read: https://github.com/exercism/go/issues/1304

JackMordaunt:
Thanks! I'll take a look, you might see me some more.

tehsphinx:
@ferhatelmas Are you also taking part in the `v3` initiative? (#1304)

We are just starting to organize ourselves in exercism Slack channel `maintaining-go`.

ferhatelmas:
@tehsphinx yes but I need to read a bit what to do. Probably will do tomorrow. 

If there is something I can, ping me there, will give a look.

ferhatelmas:
@tehsphinx could you shoot an invite for slack ? `elmas.ferhat` at google mail.

tehsphinx:
@ferhatelmas I requested an invitation for you.

PR #1319, opened 2020-01-17
list_ops: get rid of convoluted Length implementation

### LINKED ISSUE #1316: list-ops: simplify example from recursion
to for loop.

context: https://github.com/exercism/go/pull/1314#discussion_r367569813

ferhatelmas:
fixes #1316

### DIFF
--- a/exercises/list-ops/example.go
+++ b/exercises/list-ops/example.go
@@ -40,22 +40,7 @@ func (s IntList) Filter(fn predFunc) IntList {
 }
 
 // Length returns the length of a list
-func (s IntList) Length() int {
-	// for fun, let's do this with recursion
-	var lengthAcc func(IntList, int) int
-	// yeah, this is kind of cheaty, but it's hard to determine if a slice is
-	// empty without using len, so walling it off
-	isEmpty := func(l IntList) bool {
-		return len(l) == 0
-	}
-	lengthAcc = func(lst IntList, acc int) int {
-		if isEmpty(lst) {
-			return acc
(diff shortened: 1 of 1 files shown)


exercises/list-ops/example.go
```diff
@@ -40,22 +40,7 @@ func (s IntList) Filter(fn predFunc) IntList {
...
-	}
-	lengthAcc = func(lst IntList, acc int) int {
-		if isEmpty(lst) {
-			return acc
-		}
-		return lengthAcc(lst[1:], acc+1)
-	}
-	return lengthAcc(s, 0)
-}
+func (s IntList) Length() int { return len(s) }
```
tehsphinx: Isn't the point of this exercise NOT to use already existing functions?

Note: I'm not a big fan of such exercises especially since it's not clear what `existing functions` includes.

ferhatelmas: In general, yes. I get your point.

If somebody wants to learn, they need to implement it by themselves. However, example solution should be idiomatic go. When somebody has a slice, using something else than `len` to find length of this slice will never be seen in the wild. That's why it's fair to make this change. 

That being said, if you have any other suggestions to improve this, I am happy to update.

tehsphinx: I'd probably vote for removing this exercise altogether 😄 

But to answer your question: Changes are fine with me.

ferhatelmas: Same. 

To move on here, merging then.

PR #1321, opened 2020-01-17
Add andres zarta to maintainers.json

andres-zartab:


config/maintainers.json
```diff
@@ -130,6 +130,16 @@
...
+      "show_on_website": true,
+      "name": "Andrés Zarta",
+      "link_text": null,
+      "link_url": null,
+      "avatar_url": null,
+      "bio": "Data Scientist with Operations Research background. Started learning Go to improve my software development skills and ended up loving it for its simplicity and amazing community."
     }
   ]
-}
+}
```
tehsphinx: Can you undo this change?

tehsphinx:
```suggestion
}

```

tehsphinx: Note: files committed to git optimally have an empty line at the end

config/maintainers.json
```diff
@@ -130,6 +130,16 @@
       "link_url": null,
       "avatar_url": null,
       "bio": null
+    },
+    {
+      "github_username": "andres-zartab",
+      "alumnus": true,
```
tehsphinx: This would mean you were a maintainer and have stopped being one.

```suggestion
      "alumnus": false,
```

config/maintainers.json
andres-zartab: Will fix it now!

config/maintainers.json
andres-zartab: Oh, I didn't got that right!

config/maintainers.json
```diff
@@ -130,6 +130,16 @@
...
+      "show_on_website": true,
```
tehsphinx:
```suggestion
      "show_on_website": false,
```

Set this to false for now. It will later be updated once new maintainers grow into more experienced maintainers.

config/maintainers.json
andres-zartab: Is it good enough to just accept your commit suggestion?

config/maintainers.json
tehsphinx: No problem. I did the same 😁

config/maintainers.json
tehsphinx: yes you can do that

andres-zartab:
Thanks for your review and suggestions! :)

PR #1322, opened 2020-01-17
ci: enable gocritic

ferhatelmas:
and fix its issues.

### DIFF
--- a/exercises/atbash-cipher/example.go
+++ b/exercises/atbash-cipher/example.go
@@ -7,12 +7,17 @@ import (
 
 var alphabet = "abcdefghijklmnopqrstuvwxyz"
 
+var (
+	reChunk     = regexp.MustCompile(".{1,5}")
+	reNormalize = regexp.MustCompile("[^a-z0-9]")
+)
+
 func Atbash(s string) string {
 	return chunk(convert(normalize(s)))
 }
 
 func chunk(s string) string {
-	value := regexp.MustCompile(".{1,5}").FindAllString(s, -1)
+	value := reChunk.FindAllString(s, -1)
 	return strings.Join(value, " ")
 }
 
@@ -35,8 +40,7 @@ func convert(s string) string {
 }
 
 func normalize(s string) string {
-	s = strings.ToLower(s)
-	return regexp.MustCompile("[^a-z0-9]").ReplaceAllString(s, "")
+	return reNormalize.ReplaceAllString(strings.ToLower(s), "")
 }
 
 func reverse(s string) string {
--- a/exercises/atbash-cipher/example.go
+++ b/exercises/atbash-cipher/example.go
@@ -12,10 +12,8 @@ func Atbash(s string) string {
 }
 
 func chunk(s string) string {
-	reg, _ := regexp.Compile(".{1,5}")
-	value := reg.FindAllString(s, -1)
-	s = strings.Join(value, " ")
-	return s
+	value := regexp.MustCompile(".{1,5}").FindAllString(s, -1)
+	return strings.Join(value, " ")
 }
 
 func convert(s string) string {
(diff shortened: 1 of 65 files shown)


exercises/atbash-cipher/example.go
```diff
@@ -12,10 +12,8 @@ func Atbash(s string) string {
 }
 
 func chunk(s string) string {
-	reg, _ := regexp.Compile(".{1,5}")
-	value := reg.FindAllString(s, -1)
-	s = strings.Join(value, " ")
-	return s
+	value := regexp.MustCompile(".{1,5}").FindAllString(s, -1)
```
tehsphinx: Maybe in another PR:
`regexp.MustCompile` should be executed only once per regex pattern. Since it's result is basically static I don't see a problem using a global for it.

exercises/atbash-cipher/example.go
```diff
@@ -28,19 +26,17 @@ func convert(s string) string {
...
 	}
 	return result
 }
 
 func normalize(s string) string {
 	s = strings.ToLower(s)
-	reg, _ := regexp.Compile("[^a-z0-9]")
-	s = reg.ReplaceAllString(s, "")
-	return s
+	return regexp.MustCompile("[^a-z0-9]").ReplaceAllString(s, "")
```
tehsphinx: Same here...

exercises/rectangles/example.go
```diff
@@ -1,21 +1,19 @@
 package rectangles
 
-import "strings"
-
 func checkHorizontalSide(diagram []string, start, end, rowIndex int) bool {
-	const pattern = "+-"
 	for i := start; i < end; i++ {
-		if !strings.Contains(pattern, diagram[rowIndex][i:i+1]) {
+		s := diagram[rowIndex][i : i+1]
+		if s != "+" && s != "-" {
```
tehsphinx: Wondering if we should simplify this:

```go
s := diagram[rowIndex][i]
if s != '+' && s != '-' {
```

exercises/rectangles/example.go
```diff
@@ -1,21 +1,19 @@
...
 	}
 	return true
 }
 
 func checkVerticalSide(diagram []string, start, end, colIndex int) bool {
-	const pattern = "+|"
 	for i := start; i < end; i++ {
-		if !strings.Contains(pattern, diagram[i][colIndex:colIndex+1]) {
+		s := diagram[i][colIndex : colIndex+1]
+		if s != "+" && s != "|" {
```
tehsphinx: Same here...

exercises/run-length-encoding/example.go
```diff
@@ -13,12 +13,12 @@ func RunLengthEncode(s string) string {
...
 			} else {
-				count, output = encode(count, i, s, output)
+				count, output = 1, output+encode(count, i, s)
 			}
 		}
 	}
-	if len(s) != 0 {
-		_, output = encode(count, len(s), s, output)
+	if s != "" {
+		_, output = 1, output+encode(count, len(s), s)
```
tehsphinx:
```suggestion
		output += encode(count, len(s), s)
```

exercises/word-count/example.go
```diff
@@ -19,8 +19,7 @@ func WordCount(phrase string) Frequency {
 }
 
 func normalize(phrase string) string {
-	//  Allow for apostrophes in words.
-	r, _ := regexp.Compile(`[^\w|']`)
 	phrase = strings.ToLower(phrase)
-	return r.ReplaceAllLiteralString(phrase, " ")
+	// allow for apostrophes in words
+	return regexp.MustCompile(`[^\w|']`).ReplaceAllLiteralString(phrase, " ")
```
tehsphinx: Another `MustCompile` with static regex pattern better be a global or otherwise only called once.

tehsphinx:
Added some comments on what could be improved. Some changes do not need to be in this PR.

Looks very good!

ferhatelmas:
@tehsphinx thanks for suggestions, they are good, applied.

PR #1323, opened 2020-01-18
Update maintainers.json with user: prounckk

Prounckk:


config/maintainers.json
```diff
@@ -140,6 +140,16 @@
       "link_url": null,
       "avatar_url": null,
       "bio": "Data Scientist with Operations Research background. Started learning Go to improve my software development skills and ended up loving it for its simplicity and amazing community."
+    },
+    {
+      "github_username": "Prounckk",
+      "alumnus": true,
```
tehsphinx: This should be false unless you were a maintainer once but are not any more.

Prounckk: thanks, changed!

PR #1333, opened 2020-02-25
Set online editor config

kntsoriano:
Closes https://github.com/exercism/research_experiment_1/issues/102.

In the issue above, it has been suggested that Go should use tabs for spacing, with a width of 4 spaces. When this is merged, the online editor for Go in the research website should respect this setting.

### DIFF
--- a/config.json
+++ b/config.json
@@ -3,6 +3,10 @@
   "active": true,
   "blurb": "Go is a compiled, open source programming language with a small, consistent syntax, a powerful standard library, and fantastic tooling. It's a great fit for web backends and command-line tools.",
   "ignore_pattern": "example(?!.*test)",
+  "online_editor": {
+    "indent_style": "tab",
+    "indent_size": 4
+  },
   "exercises": [
     {
       "slug": "gigasecond",


kntsoriano:
@tehsphinx Is there anything else I need to do in order to make the tests pass?

tehsphinx:
@kntsoriano I'm not that far into checks on the Go track, that I could help here. Maybe @leenipper can help?

ErikSchierboom:
@kntsoriano @tehsphinx The build failed for the following reason:

```
gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

This just means we hit the rate limit when trying to download the `configlet` tool. I've submitted a fix for that in https://github.com/exercism/configlet/pull/174, but that currently only works for GitHub actions, although Travis support could be added with some effort.

The easiest fix right now is for someone to either merge (there _are_ test passes on two architectures that do pass) or else to restart the Travis build (I can't do that).

iHiD:
Thanks @ErikSchierboom. Merged :)

PR #1334, opened 2020-03-17
Test negative minutes for Add and Subtract methods

jvoll:
The first solution I wrote for this problem was dependent on Add/Subtract only receiving positive numbers. That is not a valid assumption based on the instructions. This adds a test to ensure that case is handled.


tehsphinx:
@jvoll: I think the test cases are automatically created from the https://github.com/exercism/problem-specifications repository. Best create a PR there.

https://github.com/exercism/problem-specifications/blob/master/exercises/clock/canonical-data.json

tehsphinx:
Closing this here...

PR #1339, opened 2020-04-21
Update link to doc about "Named return values as a helpful way to return error information from panic recovery"

b10s:


tehsphinx:
Thx, @b10s!

PR #1341, opened 2020-06-05
Fix name of file

rootulp:
The function `Frequency()` is defined in `parallel_letter_frequency.go` not `frequency.go`.

PR #1342, opened 2020-06-05
Prefer SVG badges to PNGs

rootulp:
because they render better on high resolution displays


rootulp:
^ same failure as #1341 which doesn't appear relevant. This should be safe to merge.

PR #1349, opened 2020-08-02
Use %q to make white space differences visible

suzaku:
Some student complained that the tests failed but he couldn't see any
difference between "expected" and "actual".
After looking into the issue (by changing the %s to %q in this test), I
see that an extra "\n" was appended to the result, which is hard to tell
from the output.

### DIFF
--- a/exercises/tournament/tournament_test.go
+++ b/exercises/tournament/tournament_test.go
@@ -117,7 +117,7 @@ func TestTallyHappy(t *testing.T) {
 				tt.description, err)
 		}
 		if actual != tt.expected {
-			t.Fatalf("Tally for input named %q was expected to return...\n%s\n...but returned...\n%s",
+			t.Fatalf("Tally for input named %q was expected to return...\n\n%q\n\n...but returned...\n\n%q",
 				tt.description, tt.expected, actual)
 		}
 	}


bitfield:
Can you show an example of what the modified output looks like on failure?

suzaku:
It's something like this:

Screen Shot 2020-08-07 at 5 39 19 PM

suzaku:
It's not as readable as the original output, but I can't think of other ways to solve this problem. Do you have any suggestions?

bitfield:
Using `cmp.Diff` is my standard way to do this (using the https://github.com/google/go-cmp library). The output would look something like:

```
--- FAIL: TestSliceSink (0.00s)
    --- FAIL: TestSliceSink/Multiple_lines_pipe (0.00s)
        sinks_test.go:225:   []string{
            -   "testdata/multiple_files/a.txt",
            +   "testdata/multiple_files/1.txt",
                "testdata/multiple_files/2.txt",
                "testdata/multiple_files/3.tar.zip",
              }
```

suzaku:
Is it OK to use `cmp.Diff` in this test?

tehsphinx:
AFAIR the go test runners currently don’t work with any imports outside the standard library…

bitfield:
Why is that? Doesn't it just run `go test` and interpret the output?

tehsphinx:
@bitfield: for security reasons the test runner does not have access to the internet to download any external dependencies. We could preinstall some dependencies if we wanted to, but that should be carefully considered as we have to draw the line somewhere.

bitfield:
Doesn't the Go module proxy more or less solve the security issue? I'm not arguing, just curious. It seems like this kind of diff output would be very valuable for students in particular.

tehsphinx:
This decision was probably not only made for module security. The test code or the code to test could contain malicious code. Letting it download stuff from the internet would not be a good idea, right?

jmrunkle:

PR #1350, opened 2020-08-07
Add 3 benchmarks for the matrix exercise

suzaku:
So that students can readily run benchmarks when they are trying to
improve the performance of their solution.

### DIFF
--- a/exercises/matrix/example.go
+++ b/exercises/matrix/example.go
@@ -11,7 +11,7 @@ import (
 
 type Matrix [][]int
 
-func New(s string) (*Matrix, error) {
+func New(s string) (Matrix, error) {
 	var err error
 	lines := strings.Split(s, "\n")
 	m := make(Matrix, len(lines))
@@ -28,7 +28,7 @@ func New(s string) (*Matrix, error) {
 		}
 		m[i] = row
 	}
-	return &m, nil
+	return m, nil
 }
 
 func (m *Matrix) Set(row, col, val int) (ok bool) {
--- a/exercises/matrix/matrix_test.go
+++ b/exercises/matrix/matrix_test.go
@@ -264,3 +264,47 @@ func TestSet(t *testing.T) {
 		}
 	}
 }
+
+func BenchmarkNew(b *testing.B) {
+	var matrix Matrix
+	for i := 0; i < b.N; i++ {
+		var err error
+		matrix, err = New("1 2 3 10 11\n4 5 6 11 12\n7 8 9 12 13\n 8 7 6 13 14")
+		if err != nil {
+			b.Fatalf("Failed to create the matrix: %v", err)
+		}
+	}
+	if matrix == nil {
+		b.Fatalf("No matrix parsed")
+	}
+}
+
+func BenchmarkRows(b *testing.B) {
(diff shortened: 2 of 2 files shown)


bitfield:
To be honest, while there's nothing wrong with this, I don't really see the need for benchmarks.

In my experience, students are only too ready to benchmark anything and everything. It's sometimes a struggle to persuade them that there are other desirable aspects to code over and above performance. Readability, for example.

suzaku:
The background for this change is a solution in which I want to teach the student about how preallocating space for data structures like `slice` can improve performance. Ideally, I would let him/her try to compare the results of the before/after benchmarks, but I can't find any in this track.

I don't think they would somehow encourage the student's tendency to focus only on performance, because it's not a required step to complete the track. It's not introduced in the instructions. But when we do need to let the student compare the benchmarks, it would be handy.

bitfield:
I'm not objecting to this, just sounding a note of caution. When students see things in the exercise code, they're apt to read too much significance into them. If they see a benchmark, they may assume that the criterion for acceptance is for the code to be as fast as possible.

And it's very hard for them to let go of the idea that the benchmark inputs are representative of real-world inputs, which they aren't. Most real performance problems occur at scale with real user requests and resources in contention, which is difficult to benchmark.

I understand your point about wanting to demonstrate the benefits of a more efficient approach to something. But I often find I can do this by helping the student reason out _why_ it's better, not just prove that it is. "This way allocates 100GiB of stack every time we call the function, whereas this way..."

tehsphinx:
LGTM

bitfield:
I see that you changed the signature of the example `New()` function to return `Matrix` instead of `*Matrix`. One problem with this is that if the student defines `Matrix` as a struct type, this throws compile errors in the tests:

```
# matrix [matrix.test]
./matrix_test.go:156:10: invalid operation: m == nil (mismatched types Matrix and nil)
./matrix_test.go:277:12: invalid operation: matrix == nil (mismatched types Matrix and nil)
```

Structs can't be nil, whereas a slice can. It's actually very common for students to implement Matrix as a struct type, so I think it might be an idea to go back to expecting a pointer. Alternatively, if we require Matrix to be a slice type, then we should probably say that in the instructions, shouldn't we?

suzaku:
> I see that you changed the signature of the example `New()` function to return `Matrix` instead of `*Matrix`. One problem with this is that if the student defines `Matrix` as a struct type, this throws compile errors in the tests:
> 
> ```
> # matrix [matrix.test]
> ./matrix_test.go:156:10: invalid operation: m == nil (mismatched types Matrix and nil)
> ./matrix_test.go:277:12: invalid operation: matrix == nil (mismatched types Matrix and nil)
> ```
> 
> Structs can't be nil, whereas a slice can. It's actually very common for students to implement Matrix as a struct type, so I think it might be an idea to go back to expecting a pointer. Alternatively, if we require Matrix to be a slice type, then we should probably say that in the instructions, shouldn't we?

I agree with you. Although it's better to use `Matrix` instead of `*Matrix` when it's a slice, this is an implementation detail not required in the instructions.

Sorry for the not well-thought change, I'll change it back later.

Issue #1352, opened 2020-08-16
Armstrong Numbers: rename test file


Daveed9:
In Armstrong Numbers, the stub file is named `armstrong_numbers.go` whereas the test file is named `armstrong_test.go`. This is inconviniences students using editors where you can alternate between the stub and test file (ex: `:GoAlternative` with vim-go).

Daveed9:
Fixed!

PR #1353, opened 2020-08-16
Fix names of test files across various exercises

Daveed9:
Fixes issue #1352

Many exercises have test names that do not match the stub files. This makes it inconvenient to switch between the stub and the test file with some text editors (namely vim with vim-go). I first noticed this with armstrong-numbers but I searched for test files that don't have matching stubs and found quite a few examples.


tehsphinx:
Thx, @Daveed9 !!

Issue #1361, opened 2020-10-15
Protein Translation issue


iHiD:
Hello. I was sent this the email below. I've asked for him to send me his code, and I'll add it if he does so, but I thought I'd raise this in case anything jumps out anyway.

---

Dear Exercism team

I think the test suite of the "Protein Translation" side exercise in the
Go track is broken. My code shouldn't succeed but does anyways. Just not
sure where to report / how to fix it. The `if test.errorExected != nil
{` section never comes back true.

I'm not having an account on Github, out of ethical reservations against
that platform, so wouldn't be able to submit an issue there, should that
be the place to do it, but thought I should make you aware nevertheless.

Thanks

GLOFonseca:
Hi, can you give me access to this exercise? I would like to look closer.
My exercism.io io user is the same as GitHub, GLOFonseca.

GLOFonseca:
I've looked at the test and everything looks fine and working. 
But when you have the code from him please let me know!

iHiD:
Here you go:
```
package protein

var codons = map[string]string{
	"AUG": "Methionine",
	"UUU": "Phenylalanine",
	"UUC": "Phenylalanine",
	"UUA": "Leucine",
	"UUG": "Leucine",
	"UCU": "Serine",
	"UCC": "Serine",
	"UCA": "Serine",
	"UCG": "Serine",
	"UAU": "Tyrosine",
	"UAC": "Tyrosine",
	"UGU": "Cysteine",
	"UGC": "Cysteine",
	"UGG": "Tryptophan",
}

var stops = [...]string{"UAA", "UAG", "UGA"}

var ErrStop error
var ErrInvalidBase error

func FromCodon(codon string) (string, error) {

	return codons[codon], nil

}

func FromRNA(rna string) ([]string, error) {

	codon := make([]string, 0)
	lo := 0
	hi := 3

	for x := 1; x <= len(rna)/3; x++ {
		candidate := rna[lo:hi]

		if codons[candidate] == "" {
			break
		}

		codon = append(codon, codons[candidate])
		lo += 3
		hi += 3
	}

	return codon, nil
}
```

He commented: 
> I'm never using ErrStop or ErrInvalidBase, nevertheless it passes the test.

GLOFonseca:
I'll check in the next days.
Thanks!

GLOFonseca:
So I think the problem is with his code.

The test is passing, but his code declares the two errors variables but without value:
```go
var ErrStop error
[...]

PR #1365, opened 2020-10-23
Fix protein_translation_test

GLOFonseca:
This PR creates mock variables for errors to compare with the returned error.

### Problem
Using the variable declared by the actual function may be a problem, because the error variable could be declared with nil value, making every test to pass, when they shouldn't.

### Issue
https://github.com/exercism/go/issues/1361

### DIFF
--- a/exercises/protein-translation/protein_translation_test.go
+++ b/exercises/protein-translation/protein_translation_test.go
@@ -1,6 +1,7 @@
 package protein
 
 import (
+	"errors"
 	"reflect"
 	"testing"
 )
@@ -10,6 +11,9 @@ type codonCase struct {
 	expected      string
 	errorExpected error
 }
+
+var MockErrStop = errors.New("stop")
+var MockErrInvalidBase = errors.New("invalid base")
 
 var codonTestCases = []codonCase{
 	{
@@ -65,22 +69,22 @@ var codonTestCases = []codonCase{
 	{
 		"UAA",
 		"",
-		ErrStop,
+		MockErrStop,
 	},
 	{
 		"UAG",
 		"",
-		ErrStop,
+		MockErrStop,
 	},
 	{
 		"UGA",
 		"",
-		ErrStop,
+		MockErrStop,
 	},
 	{
 		"ABC",
 		"",
-		ErrInvalidBase,
+		MockErrInvalidBase,
 	},
 }
 


bitfield:
Yes, this is definitely a problem, and this looks like a good way to solve it, but I'm wondering if there might be an even neater way: just test that `ErrStop` and `ErrInvalidBase` are not nil! Or did you think of that already?

GLOFonseca:
This could be another solution. But in my opinion doesn't make much sense to use the same variable in the test.

Unless the idea of the exercise is to force the user to create these error variables. Because with this mock variable the user can return the error inside the code, without creating these variables.

If I change to this nil check, where do you thought to do it?

GLOFonseca:
Not related to this specific PR, but why this repo has not the **hacktoberfest** topic? If there is a reason, could you add the label `hacktoberfest-accepted` to this PR?

GLOFonseca:
@bitfield could you review this?

bitfield:
I had a really interesting discussion with an Exercism mentee about exactly this topic: testing that code returns the expected kind of error for given invalid inputs. Testing against sentinel errors (`var ErrStop ...`) is fine, but it means we can't include any data in the error. For `ErrInvalidBase`, for example, we would like to know what the invalid base _was_. That means using a custom error _type_, not a sentinel _value_.

If we are using custom error types, then we need to use `errors.As` to check them, but there's a problem here. Suppose you have a test case like:

```go
type codonCase struct {
	...
	errorExpected error
}
```

If you now call `errors.As` on the error you get from your function, and the `errorExpected` value, it will always be true:

```go
if !errors.As(err, &tc.errorExpected) {
	// it's never not an `error`
```

This makes sense, because the type of `tc.errorExpected` is `error`, and `err` is indeed that type! What we failed to do was check whether it was `ErrInvalidBase`, for example. In order to be able to do that, we need the type of our struct field to be the concrete type:

```go
type invalidBaseCase struct {
	...
	errorExpected ErrInvalidBase
}
```

Therefore, we need different struct types for each test case that expects a distinct error type. In fact, once we have that, we no longer need the `errorExpected` field at all, since every case of this type will expect that specific error type.

It makes sense to me, then, to define `ErrStop` and friends as custom types, and require the student to return a value of that type in these error cases. What do you think?

GLOFonseca:
@bitfield thanks a lot for your review. Sincerely I will need to study and make some tests because I didn't understand everything. But this was my goal in doing this PR, so thanks!

Do you have any references for sentinel values? This is the first point I'm struggling with.

Even with these difficulties, it seems to me that this is a really good solution. After I study and make the changes, I will even think about how to apply to my application tests. Thanks a lot!

bitfield:
This is certainly a good place to start: https://blog.golang.org/go1.13-errors

exercises/protein-translation/protein_translation_test.go
```diff
@@ -11,6 +12,9 @@ type codonCase struct {
 	errorExpected error
 }
 
+var MockErrStop = errors.New("stop")
+var MockErrInvalidBase = errors.New("invalid base")
```
Oliveirakun, line 16: I think the correct approach would be create a custom type like this:
```go
type MockErrInvalidBase struct{}

func (m *MockErrInvalidBase) Error() string {
    return "invalid base"
}


type MockErrStop struct{}

func (m *MockErrStop) Error() string {
    return "stop"
}

```
And then you compare the types to assert the correct type is returned
More info: https://www.digitalocean.com/community/tutorials/creating-custom-errors-in-go

bitfield: Yes, that piece is helpful, but a little out of date: since Go 1.13, we should use `errors.As` (for checking against custom error types) and `errors.Is` (for checking against sentinel errors, but we shouldn't use those: we should use custom error types instead).

GLOFonseca:
@bitfield @Oliveirakun  I've made a new update here. I think I'm close to the best solution for this exercise.

exercises/protein-translation/protein_translation_test.go
```diff
@@ -102,71 +89,115 @@ func TestCodon(t *testing.T) {
...
+		actual, err := FromCodon(test.input)
+		if !errors.As(err, &ErrStop{}) {
+			t.Fatalf("FAIL: Protein translation test: %s\nExpected type error: %T\nActual type error: %T",
+				test.input, ErrStop{}, err)
+		}
+		if actual != test.expected {
+			t.Fatalf("FAIL: Protein translation test: %s\nExpected: %s\nActual: %s",
+				test.input, test.expected, actual)
+		}
+		t.Logf("PASS: Protein translation test: %s", test.input)
```
bitfield, line 116: We don't need to log test passes; log output will only be seen on failure.

exercises/protein-translation/protein_translation_test.go
```diff
@@ -1,98 +1,85 @@
...
+	for _, test := range codonSuccessCases {
 		actual, err := FromCodon(test.input)
-		if test.errorExpected != nil {
-			if test.errorExpected != err {
-				t.Fatalf("FAIL: Protein translation test: %s\nExpected error: %q\nActual error: %q",
-					test.input, test.errorExpected, err)
-			}
-		} else if err != nil {
+		if err != nil {
 			t.Fatalf("FAIL: Protein translation test: %s\nExpected: %s\nGot error: %q",
```
bitfield, line 83: It's easier to read test output if it's all on one line. We can save space here because we already know the package and the test in question; `go test` will output those for us before the failure message. All we need to see is the input and output. So we can write something like:

```suggestion
			t.Fatalf("FromCodon(%q): %v", err)
```

exercises/protein-translation/protein_translation_test.go
```diff
@@ -102,71 +89,115 @@ func TestCodon(t *testing.T) {
...
 		}
 		t.Logf("PASS: RNA translation test: %s", test.input)
 	}
+	for _, test := range proteinInvalidBaseCases {
+		actual, err := FromRNA(test.input)
+		if !errors.As(err, &ErrInvalidBase{}) {
+			t.Fatalf("FAIL: Protein translation test: %s\nExpected type error: %T\nActual type error: %T",
+				test.input, ErrInvalidBase{}, err)
+		}
+		if !reflect.DeepEqual(actual, test.expected) {
```
bitfield, line 169: If the function doesn't return the expected error for an invalid input, that's the end of the test, isn't it? No need to go on and compare the data value against expectations: there are none.

exercises/protein-translation/protein_translation_test.go
```diff
@@ -102,71 +89,115 @@ func TestCodon(t *testing.T) {
...
 }
 
 func BenchmarkCodon(b *testing.B) {
-	for _, test := range codonTestCases {
+	for _, test := range codonSuccessCases {
+		for i := 0; i < b.N; i++ {
+			FromCodon(test.input)
+		}
+	}
+	for _, test := range codonInvalidBaseCases {
```
bitfield, line 182: There's no real value in benchmarking error returns, is there?

bitfield (commented):
This looks fantastic! Not only completely accurate and up to date, but an excellent example for students of how to handle errors. Nice job.

exercises/protein-translation/protein_translation_test.go
GLOFonseca: I agree, but this was already here, I think the idea was to print the tests that had passed when another one failed. But I agree and I will remove it.

exercises/protein-translation/protein_translation_test.go
GLOFonseca: My idea here was to ensure that when an error occurs, the function doesn't return anything besides the error.

exercises/protein-translation/protein_translation_test.go
GLOFonseca: I didn't understand this. As I know, the idea of this benchmarking function is to run all the scenarios without any check of the returns, but checking the time spent to run everything.

GLOFonseca:
@bitfield new update here, and thanks for your review!

exercises/protein-translation/protein_translation_test.go
bitfield: Could you mark all resolved comments as resolved, @GLOFonseca?

exercises/protein-translation/protein_translation_test.go
bitfield: Well, that's a good idea in principle, but in practice, a non-nil error says that the data value should be ignored, and that's what we should do in the test, too.

exercises/protein-translation/protein_translation_test.go
bitfield: Well, when you benchmark code, you're usually interested in its performance with valid data. If the data is invalid, of course we don't expect the function to do anything, so there's nothing to benchmark. So in benchmarks we usually only supply valid inputs.

exercises/protein-translation/protein_translation_test.go
GLOFonseca: That makes sense, we should expect that the code does this.

exercises/protein-translation/protein_translation_test.go
GLOFonseca: Now I understood what was your point. So dividing the scenarios as I did is actually better for the benchmark, and we can only run for the SuccessCases.

GLOFonseca:
@bitfield I've made the last requests

bitfield (approved):
This looks very good! Nice work.

GLOFonseca:
Thanks! @bitfield what is the process now for this PR?

bitfield:
This is a significant update, so I'd like to have someone else cast an eye over it before merging: @tehsphinx?

jmrunkle:

GLOFonseca:
Hello, @jmrunkle we were waiting for @tehsphinx to review it and I totally forgot about it. I will work on it this week and see if this makes sense yet.

PR #1366, opened 2020-10-31
tournament_test: Integrate implicit type assertion to var declaration

MewX:
The unnamed error variable was used to verify the function's return type. However, it's a bit confusing literally. Integrating it back to the variable declaration statement makes it clearer.

### DIFF
--- a/exercises/tournament/tournament_test.go
+++ b/exercises/tournament/tournament_test.go
@@ -127,11 +127,10 @@ func TestTallyError(t *testing.T) {
 	for _, s := range errorTestCases {
 		reader := strings.NewReader(s)
 		var buffer bytes.Buffer
-		err := Tally(reader, &buffer)
+		var err error = Tally(reader, &buffer)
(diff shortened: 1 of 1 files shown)


exercises/tournament/tournament_test.go
```diff
@@ -127,11 +127,10 @@ func TestTallyError(t *testing.T) {
 	for _, s := range errorTestCases {
 		reader := strings.NewReader(s)
 		var buffer bytes.Buffer
-		err := Tally(reader, &buffer)
+		var err error = Tally(reader, &buffer)
```
leenipper, line 130: For consistency, perhaps make same change in func TestTallyHappy.

MewX: Thanks for reviewing! The one in TestTallyHappy might not be necessary as returning non-error type will fail the test anyway.

jmrunkle:

PR #1364, opened 2020-10-21
Move to GitHub Actions

cmccandless:
Resolves #1362

PR #1368, opened 2020-11-13
Fix action

ferhatelmas:
Make action functional #1364 

* fix badge links in readme
* fix a lint case from go1.15
* drop unsupported goos/arch pair
* small simplification where we don't need configuration


leenipper (approved):
It all looks good.

PR #1369, opened 2020-12-19
Changed a rectangles test case to check bottom edge too

raziel2244:
I was able to pass tests without checking the bottom edge of rectangles.

This test case tweak ensures the bottom edge is checked.

### DIFF
--- a/exercises/rectangles/cases_test.go
+++ b/exercises/rectangles/cases_test.go
@@ -91,9 +91,9 @@ var testCases = []struct {
 			"    |",
 			"+-+-+",
 			"| | -",
-			"+-+-+",
+			"+|+-+",
 		},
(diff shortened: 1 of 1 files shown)


tehsphinx:
@ErikSchierboom: Is this the right place to adjust this or are the tests actually coming from somewhere else?

raziel2244:
Ah yes it looks like the test cases are generated in the .meta folder, any idea how to form those strings in the .toml file?

ErikSchierboom:
@tehsphinx The canonical data is defined in the problem-specifications repo: https://github.com/exercism/problem-specifications/blob/master/exercises/rectangles/canonical-data.json

Whether the tests are generated depends on the track. Does Go have a test generator?

Issue #1411, opened 2021-02-05
accumulate: verify input is not modified


rsandtner:
the problem asks for a new collection but the tests do not verify if the input is modified

brugnara:
Hi @rsandtner.

Thank you for writing this issue.

I am trying to replicate the issue you encountered without luck, would you mind to help me, please?

This is what I tried:

```go
func Accumulate(sx []string, fn func(string) string) []string {
	return sx
}
// which breaks the test as expected:
// Accumulate(["first" "letter" "only"], "capitalize"): expected ["First" "Letter" "Only"], actual ["first" "letter" "only"]
// FAIL
```

```go
func Accumulate(sx []string, fn func(string) string) []string {
	ret := make([]string, len(sx))

	return ret
}
// also breaks:
// Accumulate(["echo" "echo" "echo" "echo"], "echo"): expected ["echo" "echo" "echo" "echo"], actual ["" "" "" ""]
// FAIL
```

```go
func Accumulate(sx []string, fn func(string) string) []string {
	return []string{}
}
// and also this breaks:
// Accumulate(["echo" "echo" "echo" "echo"], "echo"): expected ["echo" "echo" "echo" "echo"], actual []
// FAIL
```

Thank you for considering my request.

rsandtner:
hey @brugnara 

if you try

```go
func Accumulate(s []string, m Mapper) []string {
	for i, p := range s {
		s[i] = m(p)
	}

	return s
}
```

all tests pass but the problem clearly asks for a _new_ collection.

brugnara:
Thank you so much for the clarification.

Now I see your point.

My only doubt here is if this exercise is supposed to teach this or not, but I do agree that the `new` should be interpreted and the tests clearly doesn't check this.

ekingery:
We appreciate the report @rsandtner and thanks for the fix, @brugnara!

PR #1413, opened 2021-02-11
check for inplace replacement, fixes 1411

brugnara:
Hi.

This fixes the #1411 issue.

I also refactored the code in order to have less than 80 chars per line.


ekingery (approved):
LGTM

Issue #1351, opened 2020-08-11
hamming: README refers to variations between languages but not the Go way


SaschaMann:
> The Hamming distance is only defined for sequences of equal length, so an attempt to calculate it between sequences of different lengths should not work. **The general handling of this situation (e.g., raising an exception vs returning a special value) may differ between languages.**

It would be nice if this was either removed, or, better yet, if this would describe what I'm expected to do in the Go implementation of the exercise.

PR #1414, opened 2021-02-11
Better suggestions for error handling

brugnara:
Hi.

This should improve the readme, as suggested by #1351


iHiD (changes requested):
(I'll update in a sec with a comment)

exercises/practice/hamming/.docs/instructions.md
```diff
@@ -20,5 +20,12 @@ The Hamming Distance is useful for lots of things in science, not just biology,
...
-exception vs returning a special value) may differ between languages.
+not work, returning an error.
+
+With this exercise, you are introduced to the use of
+[multiple results](https://tour.golang.org/basics/6) feature and here there
+is an example with [errors](https://tour.golang.org/methods/19):
+
+    func Foo(a, b string) (string, error) {
+        return "", errors.New("An error occurred!")
+    }
```
iHiD, line 31: I have three comments. The first is content-based, the second is Exercism-based. Both are to do with the fact that Exercism is in a transitionary window between v2 and v3 right now. So sorry for the extra pain you're feeling as a result of that.

---

The first change is great.

----

Re the extra paragraph, as this is a Practice Exercise, it is expected that the student won't be introduced to any new concepts in it. The content you've added would have been great in v2, but will become redundant in v3. Instead you might want to include a line such as "This exercise is as great place to practice [multiple results] and [errors]" or similar, and link directly to the Exercism Concepts, rather the external concepts. See this doc for more details: https://github.com/exercism/docs/blob/main/anatomy/tracks/internal-linking.md

---

It's worth being aware that this content is pulled from Exercism's [central problem-specifications repository](https://github.com/exercism/problem-specifications). By changing this file here, it will almost certainly get overridden next time that this gets synced to problem-specs. I'm therefore blocking this so that your work doesn't get lost 🙂 

We've recently (last week) added a new file that you can use to do exactly what you're doing here in terms of customising each exercise to your language. See https://github.com/exercism/docs/blob/main/anatomy/tracks/practice-exercises.md#file-docsintroductionappendmd

However, for this to work we need to update the central hamming exercise to the new standard to achieve this. This would be the first time this has been used, so I suspect it would be best if someone worked with you on the problem-specs side of this to get it working end-to-end. Its a great opportunity for us to test it all 🙂 

Can I therefore suggest that @ErikSchierboom works on getting a problem-specifications PR done that allows for this, and then you can update this PR to be the first example of us using this in the wild. We can then showcase this as an examplar for the other tracks on how things should work.

ErikSchierboom: I'll come back to this next week!

ErikSchierboom: I've just [submitted a PR to the exercise's description](https://github.com/exercism/problem-specifications/pull/1784/files). Once that is merged, the problematic bit of text will be removed from the description and this track can use the `instructions.append.md` to append Go-specific instructions. Concretely, what this PR should do is add an additional `.docs/instructions.append.md` file containing the new text:

```markdown
With this exercise, you are introduced to the use of
[multiple results](https://tour.golang.org/basics/6) feature and here there
is an example with [errors](https://tour.golang.org/methods/19):

    func Foo(a, b string) (string, error) {
        return "", errors.New("An error occurred!")
    }
```

brugnara: Thank you so much for all these informations.

I updated the PR accordingly, did I do correctly, @ErikSchierboom?

Please notice this PR is still not ready to merge since links should point to internal pages as iHiD mentioned before, I'll do as soon as I figure them out 👍

ErikSchierboom (approved):
LGTM!

exercises/practice/hamming/.docs/instructions.append.md
```diff
@@ -1,6 +1,14 @@
...
+
+    func Foo(a, b string) (string, error) {
+        return "", errors.New("An error occurred!")
+    }
+
 You may be wondering about the `cases_test.go` file. We explain it in the
 [leap exercise][leap-exercise-readme].
 
-[leap-exercise-readme]: https://github.com/exercism/go/blob/master/exercises/leap/README.md 
+[leap-exercise-readme]: https://github.com/exercism/go/blob/master/exercises/leap/README.md
```
ErikSchierboom, line 14: In the future, this is probably best explained in the track-specific `tests.md` file (see [the spec](https://github.com/exercism/docs/blob/main/anatomy/tracks/shared-files.md#track-specific-files)), but I feel like it's fine for now.

exercises/practice/hamming/.docs/instructions.md
```diff
@@ -20,5 +20,4 @@ The Hamming Distance is useful for lots of things in science, not just biology,
 
 The Hamming distance is only defined for sequences of equal length, so
 an attempt to calculate it between sequences of different lengths should
-not work. The general handling of this situation (e.g., raising an
-exception vs returning a special value) may differ between languages.
+not work, returning an error.
```
ErikSchierboom, line 23: I missed that this PR still added a bit of text. This text should be remove and instead added to the other append document.

```suggestion
not work.
```

ErikSchierboom (approved):

ErikSchierboom (approved):

exercises/practice/hamming/.docs/instructions.md
brugnara: I'm not sure the way I modified those links. Found only a "old" page mentioning the `multiple returns` and I'm also unsure if this is then visible: https://github.com/exercism/go/blob/811db944679d14855027f9ecfd20a8f6da20070a/reference/multiple_returns.md

jmrunkle:

PR #1418, opened 2021-02-13
Matrix - better explanation, as per #1294

### LINKED ISSUE #1294: matrix: test file refers to nonexistent instructions
`matrix_test.go` [states near the top](https://github.com/exercism/go/blob/master/exercises/matrix/matrix_test.go#L4-L5):
> You must implement a constructor and methods Rows() and Cols() as described in the README…

…but there is no such description - or even any clues as to the expected API - [in the README](https://github.com/exercism/go/blob/master/exercises/matrix/README.md). As it is, a user has to derive the expected API from the test failures.

One possibility would be to write the obviously-missing description, but a better one might be to add stubbed-versions of the expected API to `matrix.go`.

brugnara:
This should fix #1294

### DIFF
--- a/exercises/practice/matrix/matrix_test.go
+++ b/exercises/practice/matrix/matrix_test.go
@@ -1,15 +1,17 @@
 // For the Matrix exercise in Go you have to do a few things not mentioned
 // in the README.
 //
-// 1. You must implement a constructor and methods Rows() and Cols() as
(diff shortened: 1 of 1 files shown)


ekingery (approved):
I'm not familiar with this exercise, but the changes look clearer. Let's try it out, and we may take a closer look as this exercise is integrated into v3.

PR #1419, opened 2021-02-15
Basics Concept + Lasagna Concept Exercise

ekingery:
The Basics concept and the Lasagna concept exercise should be working. I have pared down the concepts and corresponding exercises in config.json for now, I think the best workflow will be to add them one by one as they are being readied for launch. I checked in `config.orig.json` for convenience as we migrate exercises. We can remove it any time. 

I attempted to use `git mv` to preserve authorship / history wherever possible when renaming files, although in some cases that was difficult. I rewrote enough of the contents of the Basics concept that I'm comfortable with how things have landed, but I'm definitely open to making adjustments.


ekingery:
Would appreciate a look from any @exercism/reviewers. The build failures will be the next step - for now, the goal is to get the track in a good spot for others to contribute.

config.json
```diff
@@ -18,447 +18,14 @@
...
-          "time",
-          "duration",
-          "select",
-          "channels"
-        ],
-        "status": "wip"
-      },
-      {
-        "slug": "basics",
+        "name": "Gopher's Gorgeous Lasanga",
```
brugnara, line 21: This should be Lasagna.

config.orig.json
```diff
@@ -0,0 +1,2156 @@
...
+          "floating-point"
+        ],
+        "prerequisites": [
+          "numbers",
+          "conditionals"
+        ],
+        "status": "wip"
+      },
+      {
+        "name": "Gopher's Gorgeous Lasanga",
```
brugnara, line 224: Little typo here. Should be Lasagna, I guess.

brugnara (commented):
Massive work!

Just spotted two little typos I'm reporting here.

Hope it's useful.

iHiD (approved):

neenjaw (commented):
For an optimistic merging perspective, this looks really good, great work @ekingery!!!!

catthumbsup

config.json
```diff
@@ -18,447 +18,14 @@
...
-        ],
-        "prerequisites": [
-          "functions",
-          "numbers",
-          "strings",
-          "pointers",
-          "booleans"
-        ],
-        "status": "wip"
+        "status": "active"
```
ErikSchierboom, line 28: I don't know if you already knew, but if the status is set to `actice`, you can also omit the status entirely.

ErikSchierboom (commented):
Nice work!

config.json
ekingery: I just happened  to read that last night, thanks! I'll leave them in for now, until we have some samples with other states in the config file that people can refer to.

PR #1420, opened 2021-02-15
Fix configlet CI

ekingery:


jmrunkle:
Since we already fixed it, so you may want to revert this. Closing the PR in any case.

PR #1424, opened 2021-02-23
fixing some linting errors

brugnara:
Hi.

I tried to fix some linting errors and so here it is a sort of a result. This is a draft because I did something I'm not sure about so I'd like to talk about this.

Almost every exercise in the "practice" will fail the CI because for no one of these the empty function is present so the build will fail accordingly.

I'm not sure if this system is the right choice for this usage, since we should check for linting but not compile error. Just guessing here, don't get me wrong.

I am eager to know your thoughts about this.

### DIFF
--- a/bin/test-without-stubs
+++ b/bin/test-without-stubs
@@ -19,6 +19,9 @@ for dir in exercises/*/; do
   fi
 done
 
+# This raises many problems since it's an empty file
+mv exercises/concept/lasagna/lasagna_test.go exercises/concept/lasagna/lasagna_test.notgo
+
 root=$(pwd)
 
 cleanup () {
@@ -29,7 +32,10 @@ cleanup () {
 
   # could use `rename` here, but it's not uniform among platforms.
   # some renames take two arguments while others take a single regex.
-  for file in exercises/*/*.notgo; do
+  # for file in exercises/*/*.notgo; do # not working, neither with **
+
+  find . -name '*.notgo' | while read -r file; do
+    echo "moving: $file"
     dir=$(dirname $file)
     base=$(basename $file .notgo)
     mv "$dir/$base.notgo" "$dir/$base.go"
--- a/exercises/concept/maps/maps_test.go
+++ b/exercises/concept/maps/maps_test.go
@@ -66,12 +66,11 @@ func TestAddItem(t *testing.T) {
 			true,
 		},
 	}
-	units := Units()
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			bill := NewBill()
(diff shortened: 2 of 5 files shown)


brugnara:
As per saying, the CI now fails because `practices` are not fixed, yet:

image

bin/test-without-stubs
```diff
@@ -19,6 +19,9 @@ for dir in exercises/*/; do
   fi
 done
 
+# This raises many problems since it's an empty file
+mv exercises/concept/lasagna/lasagna_test.go exercises/concept/lasagna/lasagna_test.notgo
```
tehsphinx, line 23: can we just make it non-empty? It just needs a package name, then it should not create problems, right?

bin/test-without-stubs
```diff
@@ -29,7 +32,10 @@ cleanup () {
 
   # could use `rename` here, but it's not uniform among platforms.
   # some renames take two arguments while others take a single regex.
-  for file in exercises/*/*.notgo; do
+  # for file in exercises/*/*.notgo; do # not working, neither with **
```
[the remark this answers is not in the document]
tehsphinx: But it works here on github?

exercises/concept/maps/maps_test.go
```diff
@@ -66,12 +66,11 @@ func TestAddItem(t *testing.T) {
 			true,
 		},
 	}
-	units := Units()
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			bill := NewBill()
 			for _, item := range tt.entry {
-				ok := AddItem(bill, units, item.name, item.unit)
+				ok := AddItem(bill, item.name, item.unit)
```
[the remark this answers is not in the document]
tehsphinx: The exercise seems to be inconsistent. In the `.meta/exemplar.go` file there are `units` passed to the function to retrieve data from.

bin/test-without-stubs
tehsphinx: apart from the fact that tests are needed...

brugnara: The problem within the lasagna (but also for every practice test) is the tests are calling methods not declared yet in the .go counterpart. These packages are impossible to build without adding empty methods for every exercise.

As per saying, the lasagna concepts, should teach to write methods, so this one cannot be fixed in any way. This will never build.

brugnara:
@tehsphinx no, the `for` doesn't work on github:

image

But with this fix, yes:

image

bin/test-without-stubs
tehsphinx: ok, got it.

jmrunkle:

PR #1430, opened 2021-03-20
Fix broken leap link

codedge:
The link to the leap readme led to a 404.

### DIFF
--- a/exercises/practice/hamming/.docs/instructions.append.md
+++ b/exercises/practice/hamming/.docs/instructions.append.md
@@ -3,4 +3,4 @@
 You may be wondering about the `cases_test.go` file. We explain it in the
 [leap exercise][leap-exercise-readme].
 
-[leap-exercise-readme]: https://github.com/exercism/go/blob/master/exercises/leap/README.md 
+[leap-exercise-readme]: https://github.com/exercism/go/blob/main/exercises/practice/leap/.docs/instructions.md


tehsphinx (approved):

tehsphinx:
@ErikSchierboom: how could we deal with those failing linters?

For example there is this test:

```go

func TestGetMonth(t *testing.T) {
	tests := map[string]struct {
		arg, want int
	}{
		"GetMonth(Jan)": {arg: Jan, want: 1},
		"GetMonth(Mar)": {arg: Mar, want: 3},
		"GetMonth(Oct)": {arg: Oct, want: 10},
	}
	for name, tc := range tests {
		t.Run(name, func(t *testing.T) {
			if got := GetMonth(tc.arg); got != tc.want {
				t.Errorf("GetMonth(%v) = %v, want %v", tc.arg, got, tc.want)
			}
		})
	}
}
```

The constants `Jan`, `Mar` and `Oct` are not defined because the student should define them. How are other tracks solving this?

ErikSchierboom:
> @ErikSchierboom: how could we deal with those failing linters?

I think most tracks don't actually lint the stub implementation so they might not have this issue. It is definitely good to lint as much as possible, but in this case you might want to exclude that file.

tehsphinx:
@ErikSchierboom: can we get this merged anyway. The linters are not really a problem of this PR.

PR #1431, opened 2021-03-23
Update matrix_test.go

jecarr:
Hi,

I was doing this exercise and had to tweak the test file. I am still learning Go; I'd be much appreciative of explanations to any errors on my part with this pull request:

1. function BenchmarkNew()
   - Change `var matrix Matrix` to `var matrix *Matrix`
   - Due to syntax error 
`>go test`
`# matrix [matrix.test]`
`.\matrix_test.go:279:12: invalid operation: matrix == nil (mismatched types Matrix and nil)`
`FAIL    matrix [build failed]`
   - I did further reading to learn how to set a struct instance to nil and [learnt I can't](https://stackoverflow.com/questions/30131052)
2. TestRows() and TestCols() final if statements
    - `r[0][0]++` and `c[0][0]++` occurs
    - `if !reflect.DeepEqual(m.Rows(), test.rows)` then throws an error
    - As does `if !reflect.DeepEqual(m.Cols(), test.cols)`
    - If the increment (`r[0][0]++`) takes place, isn't it that we want DeepEqual() to return false?
    - Should it not be that if DeepEqual() returns true, then the increment affected the outcome of Rows() and Cols() hence a failed test?

Apologies if I have misunderstood something! Thanks


tehsphinx (commented):
Sounds legit to me although I'd not commit the comments behind the code. Maybe before the block that is responsible for doing that?

Maybe someone else more familiar with the matrix exercise can check this as well.

bitfield:
I think this conflates two different issues.

The first, which breaks the tests if the student defines Matrix as a struct, was introduced in #1350. I pointed out the problem at the time, but the PR was merged anyway. So if that was correct, then _this_ PR must be wrong, mustn't it? We don't want to keep flip-flopping this issue back and forth from PR to PR.

For what it's worth, I think it was less confusing for students when they were able to define Matrix as a struct. (I think it's better to define it as a slice, but that's a discussion we can have in mentoring, rather than having the tests fail.)

The other issue (wrong sense of `reflect.DeepEqual` is puzzling. I agree that it looks like this change should be made, but how long has it been wrong, and when was it first broken?

dittrichlucas (approved):

jmrunkle:
OK, I have puzzled over this and I believe I have convinced myself that the current implementation of the `reflect.DeepEqual` tests is correct.

Here is how those tests work:

```
r := m.Rows()
```

https://github.com/exercism/go/blob/main/exercises/practice/matrix/matrix_test.go#L174

`Rows()` should return a defensive copy of the rows (rather than a direct reference to the internal representation). So when we modify `r` and then call `m.Rows()` again, we are verifying that it did not modify the internal representation of the Matrix:

```
r[0][0]++
if !reflect.DeepEqual(m.Rows(), test.rows) {
  t.Fatalf("Matrix.Rows() returned slice based on Matrix " +
           "representation.  Want independent copy of element data.")
}
```

https://github.com/exercism/go/blob/main/exercises/practice/matrix/matrix_test.go#L184-L188

This **should** use `!` because we want to enter that codeblock if a fresh defensive copy of the rows no longer matches test.rows.

---

Since we have a separate PR (#1475) for the other issue, I am closing this one since I believe the `reflect.DeepEquals` checks are correct.

PR #1458, opened 2021-06-24
Move exercise's after.md to concept's about.md

angelikatyborska:
This migration was already done for all exercises but this one. Somehow one `after.md` stayed behind.


ekingery (approved):
Good catch, thanks @angelikatyborska!

PR #1460, opened 2021-06-24
Mark some hints as TODO; always use lists

angelikatyborska:
`hints.md` requirements: https://github.com/exercism/docs/blob/main/building/tracks/concept-exercises.md#file-docshintsmd

All hints must be lists.

With the conditionals and slices hints, there were single lines that were not in lists, but otherwise lists were used. Is this because that was meant to be like a comment to the document's author, and not a hint for the student?

There are some exercises where the hints are free-form documents or missing. I added TODO comments ther because I lack the skills to write proper hints for go.

@ErikSchierboom I have questions about two things that aren't clear to me from the docs:
1. Can hints be multilevel lists? (If no, I think there's one more file to be fixed)
2. If there is a paragraph next to the hint list, will it simply be ignored and the list rendered as hints, or will it completely break rendering any hints for that heading?


ekingery (approved):
Recursive Hint: Fix the hints ;-)

Thanks again!

ErikSchierboom:
> @ErikSchierboom I have questions about two things that aren't clear to me from the docs:

> Can hints be multilevel lists? (If no, I think there's one more file to be fixed)

Currently that's not supported.

> If there is a paragraph next to the hint list, will it simply be ignored and the list rendered as hints, or will it completely break rendering any hints for that heading?

It will break rendering at the moment.

@iHiD Do we want to support either of the above options?

iHiD:
> Can hints be multilevel lists?
> Currently that's not supported.

This might actually work. I don't **think** it would break the "splitting" code. I'm just not sure if the HTML would render correctly. However, you wrote this Erik, so you have more knowledge than me 🙂 

> If there is a paragraph next to the hint list, will it simply be ignored and the list rendered as hints, or will it completely break rendering any hints for that heading?

I don't hugely want to change/enhance this right now, so I think this option is out.

ErikSchierboom:
> This might actually work. I don't think it would break the "splitting" code. I'm just not sure if the HTML would render correctly. However, you wrote this Erik, so you have more knowledge than me 🙂

I just checked and it looks like it is not working. I'd have to dig a little deeper to see what I can do.

angelikatyborska:
Since nested lists aren't supported, I tried to flatten that one multilevel list that was in hints.

PR #1461, opened 2021-06-24
Luhn: A new test case with odd number of digits

fahadhk:
A new positive test case with odd number of digits and non-zero first digit.
I have seen a solution which passes all the tests for Luhn containing odd number of digits and never reads the first digit.
It happens because in all such cases the first digit is zero .


ekingery (approved):
Thank you for catching and fixing this!

PR #1475, opened 2021-07-28
Fix matrix test variable type

antklim:
In the matrix exercise, the function New is expected to return a tuple with the types `*Matrix, error`.
However, in the `BenchmarkNew` test, the matrix variable defined a variable of type `Matrix`. It causes compilation errors due to type mismatch. 

This PR proposes a test code fix. It is also related to other open PR #1431.

Thanks.

### DIFF
--- a/exercises/practice/matrix/.meta/example.go
+++ b/exercises/practice/matrix/.meta/example.go
@@ -11,7 +11,7 @@ import (
 
 type Matrix [][]int
 
-func New(s string) (Matrix, error) {
+func New(s string) (*Matrix, error) {
 	var err error
 	lines := strings.Split(s, "\n")
 	m := make(Matrix, len(lines))
@@ -28,7 +28,7 @@ func New(s string) (Matrix, error) {
 		}
 		m[i] = row
 	}
-	return m, nil
+	return &m, nil
 }
 
 func (m *Matrix) Set(row, col, val int) (ok bool) {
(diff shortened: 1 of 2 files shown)


dittrichlucas (approved):

ekingery:
Unfortunately I don't have time to dive into the original issue in #1431. I was going to approve this change, however I think we would need to also update the `.meta/example.go` file. Ideally someone could take a fresh look at this exercise and make a call on the issues discussed in #1431, make sure it conforms to the new V3 specs, etc.

jmrunkle:
@antklim - can you pull in the changes from main and then rebase so that we can get a clean run of the CI tests? I want to see if this is actually passing. If it is, I am in favor of approving this so that students can actually choose between different solutions and discuss that in mentoring.

I could also be persuaded that we should actually get rid of benchmark tests completely from our test suites and simply allow students to do their own benchmarking if they would like to. Benchmarking is something that seems outside of the normal purview of exercism v3's MO.

antklim:
@jmrunkle I just rebased from main and resolved conflicts.

jmrunkle:
Yeah, it is what I was afraid of, this does not pass CI:

```
practice/matrix/matrix_test.go:277:17: cannot use New("1 2 3 10 11\n4 5 6 11 12\n7 8 9 12 13\n 8 7 6 13 14") (value of type Matrix) as *Matrix value in assignment (typecheck)
		matrix, err = New("1 2 3 10 11\n4 5 6 11 12\n7 8 9 12 13\n 8 7 6 13 14")
```

jmrunkle:
So the stub actually make no assertion about what it returns and the exemplar returns (Matrix, error): https://github.com/exercism/go/blob/main/exercises/practice/matrix/.meta/example.go#L14

antklim:
@jmrunkle I made changes in stub signature. I hope it will fix the build.

jmrunkle (approved):

PR #1476, opened 2021-08-11
wordy add check of using default 0 minus number

zhujinhe:
Some solutions use 0 as the default value. The added test case can reject this scenario.


ekingery (approved):
Hello. I don't have the time to verify the fix at the moment, but the code seems reasonable. Thanks and sorry for the delay in merging!

PR #1478, opened 2021-08-17
Fix analyns infiltration description

siebenschlaefer:
The comment of the function `CanFreePrisoner` talks about "the prisoner [...] and the other 3 characters". But there are only two other characters.

I also took the opportunity to add missing articles to make the comments more consistent ("the prisoner", "the archer", and "the knight").


ekingery (approved):
Thanks!

PR #1491, opened 2021-09-01
fix fmt.printf example at strings indtroduction

bazuara:
Fixed a fit.printf typo inside the example at the end of strings introduction. Provided example didn't work, now it does according to expected commented output


ekingery (approved):
Good find, thanks!

PR #1493, opened 2021-09-01
Fix party robot

siebenschlaefer:
None of the tests were executed or even compiled because `go.mod` did used the wrong module name.
I fixed the module name, tests, and test descriptions.


oanaOM (approved):

oanaOM:
nice work !

PR #1496, opened 2021-09-02
party-robot go.mod using wrong module

goatthinking:
`module string` is incorrect and thus `go test` is not running tests properly. 
Switched to `module partyrobot` based off `packack partyrobot` in party_robot.go


goatthinking:
found this being fixed in https://github.com/exercism/go/pull/1493 closing this one out.

PR #1500, opened 2021-09-04
 zero_zilch_nada: remove ambiguous reference to empty

sudomateo:
In Go, there can be a difference between a type's zero value and its
empty value. For example, consider the following code snippet:

```go
func main() {
	var s1 []string
	fmt.Printf("%#v\n", s1)
	if s1 == nil {
		fmt.Println("s1 is equal to nil")
	}

	s2 := []string{}
	fmt.Printf("%#v\n", s2)
	if s2 == nil {
		fmt.Println("s2 is equal to nil")
	}
}
```

The variable `s1` will be set to its zero value (`nil`) whereas the
variable `s2` will be set to a slice of strings with zero elements. This
behavior makes the term "empty" ambiguous and may confuse students as they
continue learning about Go.

In addition to code updates, the exercise text was updated to introduce
the `var` keyword as well as various verbiage updates.

Updated the documentation for the nil and zero-values concepts to better
reflect the new `zero_zilch_nada` exercise content. Long term, I'd
probably recommend merging the two concepts into one but for now keeping
their content the same ensures that students will learn the relevant
content regardless of which concept they click on.


concepts/nil/about.md
```diff
@@ -1,30 +1,67 @@
...
+  fmt.Println(myBool)
+}
+```
+
+```go
+func main() {
+  var myMap map[int]int
+  fmt.Println(myMap)
+}
+```
```
ErikSchierboom, line 39: I would probably leaving this out, as this document is about `nil` and not about zero values. If you'd like to link between concepts, I'd suggest using the Concept Widget as described in [this document](https://exercism.org/docs/building/markdown/widgets).

concepts/nil/about.md
```diff
@@ -1,30 +1,67 @@
-# Zero Values
+# Introduction
 
-Unlike some other languages, Go has no empty/null/undefined state for uninitialized variables. Every type has a [zero value][zero_values] that uninitialized variables will hold upon declaration.
+Go does not have a concept of undefined for variable values. Variables declared without an explicit initial value default to the [zero value][zero_values] for their respective type.
 
-These default values are called the zero values for their respective types:
+The zero value for primative types such as booleans, numeric types, and strings are `false`, `0`, and `""`, respectively.
+
+The identifier `nil`, meaning zero, is the zero value for more complex types such as pointers, interfaces, channels, maps, slices, and functions.
```
ErikSchierboom, line 7: Maybe consider adding backticks around the complex types and using their actual type name?

concepts/nil/about.md
```diff
@@ -1,30 +1,67 @@
-# Zero Values
+# Introduction
 
-Unlike some other languages, Go has no empty/null/undefined state for uninitialized variables. Every type has a [zero value][zero_values] that uninitialized variables will hold upon declaration.
+Go does not have a concept of undefined for variable values. Variables declared without an explicit initial value default to the [zero value][zero_values] for their respective type.
```
iHiD, line 3:
```suggestion
Go does not have a concept of empty/null/undefined for variable values. Variables declared without an explicit initial value default to the [zero value][zero_values] for their respective type.
```

I'd like to keep this bit in - it's useful for people from coming from other languages to understand there is no `null` (as `null` and `undefined` are quite different things)

concepts/nil/introduction.md
```diff
@@ -1,19 +1,67 @@
 # Introduction
```
ErikSchierboom, line 1: It looks like this file is just a copy of the `about.md` document, which is not exactly how it should be (see [the docs](https://exercism.org/docs/building/tracks/concept-exercises)). In short: the introduction should be a minimal explanation of the concept.

concepts/nil/introduction.md
```diff
@@ -1,19 +1,67 @@
...
+
+```go
+func main() {
+  var mySlice []int
+
+  if mySlice != nil {
+    fmt.Println(mySlice)
+  }
+}
+```
```
ErikSchierboom, line 65: If I look at the exemplar implementation, it looks like students don't need to know about this to succesfully finish the exercise. If true, could you omit this example? You can read more on this here: https://exercism.org/docs/building/tracks/concept-exercises

concepts/nil/introduction.md
```diff
@@ -1,19 +1,67 @@
...
 
-The identifier `nil`, meaning zero, is the zero value for the more complex types in Go.
+[zero_values]: https://golang.org/ref/spec#The_zero_value
```
ErikSchierboom, line 67: Introduction documents should generally not link to external resources. Could you remove the link?

exercises/concept/zero-zilch-nada/.docs/introduction.md
```diff
@@ -1,23 +1,67 @@
...
+
+```go
+func main() {
+  var mySlice []int
+
+  if mySlice != nil {
+    fmt.Println(mySlice)
+  }
+}
+```
```
ErikSchierboom, line 65: If I look at the exemplar implementation, it looks like students don't need to know about this to succesfully finish the exercise. If true, could you omit this example? You can read more on this here: https://exercism.org/docs/building/tracks/concept-exercises

exercises/concept/zero-zilch-nada/.docs/instructions.md
```diff
@@ -1,40 +1,41 @@
 # Instructions
 
-In this exercise, we will write some simple functions that return zero values for various Go types. Some types
-you might not have seen before. They will be introduced in detail in later exercises.
+In this exercise, we will write some functions that return the zero value for
+Go's types. You may not have seen some of the types yet. They will be
+introduced in detail in later exercises.
```
ErikSchierboom, line 5: I wonder if it would be possible to have this exercise depend on those other exercises. For example, one could teach booleans first in another exercise without that exercise having to depend on the zero concept.

concepts/zero-values/introduction.md
```diff
@@ -1,25 +1,67 @@
...
+
+```go
+func main() {
+  var mySlice []int
+
+  if mySlice != nil {
+    fmt.Println(mySlice)
+  }
+}
+```
```
ErikSchierboom, line 65: If I look at the exemplar implementation, it looks like students don't need to know about this to succesfully finish the exercise. If true, could you omit this example? You can read more on this here: https://exercism.org/docs/building/tracks/concept-exercises

concepts/zero-values/.meta/config.json
```diff
@@ -1,5 +1,5 @@
 {
-  "blurb": "TODO: add blurb",
+  "blurb": "Learn about zero values.",
   "authors": [],
```
ErikSchierboom, line 3: Maybe add yourself as the author?

concepts/zero-values/introduction.md
```diff
@@ -1,25 +1,67 @@
...
 
-The identifier `nil`, meaning zero, is the zero value for the more complex types in Go.
+[zero_values]: https://golang.org/ref/spec#The_zero_value
```
ErikSchierboom, line 67: Introduction documents should generally not link to external resources. Could you remove the link?

concepts/zero-values/about.md
```diff
@@ -1,30 +1,67 @@
 # Zero Values
 
-Unlike some other languages, Go has no empty/null/undefined state for uninitialized variables. Every type has a [zero value][zero_values] that uninitialized variables will hold upon declaration.
+Go does not have a concept of undefined for variable values. Variables declared without an explicit initial value default to the [zero value][zero_values] for their respective type.
 
-These default values are called the zero values for their respective types:
+The zero value for primative types such as booleans, numeric types, and strings are `false`, `0`, and `""`, respectively.
+
+The identifier `nil`, meaning zero, is the zero value for more complex types such as pointers, interfaces, channels, maps, slices, and functions.
```
ErikSchierboom, line 7: I would argue that this document does not need to describe `nil` values (as that is a different concept). If you'd like to link between concepts, I'd suggest using the Concept Widget as described in this document.

ErikSchierboom (approved):
Thanks a lot for working on this! I've left some comments to help explain what is expected in the various documents.

I don't know if you're interested, but this exercise could use a little love. In its current state it is, well, rather matter-of-factly. You just return a single value from the functions and there is no real story/theme around the functionality to give the students something fun to work with. Would you be interested in trying to add a bit more flavor to this exercise and make it more fun for students? If so, let me know, and I can help you with that.

exercises/concept/zero-zilch-nada/.docs/introduction.md
sudomateo: Omitted the example!

concepts/zero-values/.meta/config.json
sudomateo: Done!

concepts/zero-values/introduction.md
sudomateo: Omitted the example!

concepts/zero-values/about.md
sudomateo: Removed the `nil` values and added a Concept Widget.

concepts/zero-values/introduction.md
sudomateo: Done!

sudomateo:
Thank you for the feedback! I reworked the `zero-values` concept per your feedback. I also updated the `zero-zilch-nada` exercise to be a bit more involved for the student.

I'll do the same for the `nil` concept once I get more time to work on this. For now I'm leaving conversations open around the `nil` concept.

ErikSchierboom:
> I also updated the zero-zilch-nada exercise to be a bit more involved for the student.

Maybe I'm missing something, but the instructions of the exercise look to be relatively unchanged. With more involved, I meant giving the exercise a little bit of story/theme, to have the tasks make sense for the student (and not just a list of fill-in methods). Note that this story/theme doesn't have to be very involved, it can be relatively simple. For some inspiration, I'd look at existing exercises like https://github.com/exercism/elixir/blob/main/exercises/concept/dna-encoding/.docs/instructions.md or https://github.com/exercism/elixir/blob/main/exercises/concept/high-school-sweetheart/.docs/instructions.md

junedev:
@sudomateo As a story I can recommend the amusement park story: https://github.com/exercism/javascript/blob/main/exercises/concept/amusement-park/.docs/instructions.md

Write some introduction sentences about working for the amusement park IT and being in charge of data quality of the visitor data etc.

Then you can set up the tasks the following way:

- First Task: let the student write a NewVisitor function that creates and returns a visitor struct given some data as parameters. Describe in the task descriptions which fields the struct should have, e.g. name as string, age as int, address as map (street, city), etc. This task does not practice anything yet really, it is more for the setup.
- Then the following tasks can be things like "Determine whether the visitor provided a name" where you pass a visitor struct as parameter and the function e.g. "nameIsSet" returns true or false depending on the value of that field.
- I would see the above as the minimum to make this a bit more story-like. If you want to go further you can also mix it up with something like this task where you provide some map and the student needs to return some text: https://github.com/exercism/javascript/blob/main/exercises/concept/amusement-park/.docs/instructions.md#4-improve-the-ticket-status-response

I don't think it is necessary that the exercise covers all the types. Concentrate on a few tasks to cover the most common types. That's enough for the student to understand the idea.

If you go ahead with these types of tasks, don't forget to add structs as a prerequisite.

Erik says it is ok to change the slug and folder name to "amusement-park".

ErikSchierboom:
> I don't think it is necessary that the exercise covers all the types. Concentrate on a few tasks to cover the most common types. That's enough for the student to understand the idea.

100% this ☝️

sudomateo:
Thank you for the feedback @ErikSchierboom and @junedev. I've been a bit busy in and out of work the past week. I'll try to take a look at this in the next few days. I like the idea of making the exercise a bit more story-like. When I first opened this pull request, I was under the impression I should leave the exercise alone. It's nice to see that I can help shape the exercise to be more engaging.

jmrunkle:
FYI: CI should be fixed at head if you do not mind syncing your fork with main and then rebasing this branch.

sudomateo:
I'm going modify the `zero-zilch-nada` exercise and rename it to `census`. This `census` exercise will ask the user to create new residents with their necessary information, check whether or not required information is missing from a given resident, allow the counting of valid residents with all required information present, and allow residents to "move out" by setting their fields to their zero value. It's similar to the amusement park exercise with a bit more focus on zero values. I'll update this pull request once I finish the exercise. Apologies for the delay, last minute travel plans.

ErikSchierboom:
> I'm going modify the zero-zilch-nada exercise and rename it to census. This census exercise will ask the user to create new residents with their necessary information, check whether or not required information is missing from a given resident, allow the counting of valid residents with all required information present, and allow residents to "move out" by setting their fields to their zero value. It's similar to the amusement park exercise with a bit more focus on zero values. I'll update this pull request once I finish the exercise. Apologies for the delay, last minute travel plans.

That sounds excellent.

concepts/nil/about.md
sudomateo: Done!

concepts/nil/introduction.md
sudomateo: Updated!

concepts/nil/introduction.md
sudomateo: Done!

config.json
```diff
@@ -228,8 +228,8 @@
         "status": "beta"
       },
       {
```
sudomateo, line 230: I only changed the `name` and `slug` for now. Perhaps I should also change the `prerequisites` as well?

exercises/concept/census/.docs/instructions.md
```diff
@@ -0,0 +1,98 @@
+# Instructions
```
ErikSchierboom, line 1: I really, really like this story and their tasks! Well done 👍

ErikSchierboom (commented):
This looks great! I'll let one of the Go maintainers review the Go side of things.

Thanks a lot!

ErikSchierboom:
There is a merge conflict.

exercises/concept/census/.docs/hints.md
```diff
@@ -0,0 +1,20 @@
+# Hints
+
+## 1. Create a new resident
+
+- Revisit the [structs concept][concept-structs] for information on how to create structs.
```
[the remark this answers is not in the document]
sudomateo: Switched to widgets!

exercises/concept/census/.docs/instructions.md
```diff
@@ -0,0 +1,98 @@
...
+
+It is your job to prepare the city's IT system for an upcoming census. Specifically, you are responsible for the program that will injest the data from the census workers.
```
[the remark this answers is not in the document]
sudomateo: Used the word `process` instead.

exercises/concept/census/.docs/instructions.md
sudomateo: Thank you kindly!

concepts/nil/about.md
```diff
@@ -1,30 +1,47 @@
-# Zero Values
+# Introduction
```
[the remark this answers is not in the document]
sudomateo: You're right. Both `nil` and `zero-values` are quite identical. Currently, they both point to the same exercise. Perhaps for now we leave the content as is and defer the work of differentiating their content until someone can update both the `nil` content and create a specific exercise for it? What do you think?

config.json
```diff
@@ -234,6 +234,14 @@
...
         "status": "beta"
       },
+      {
+        "name": "The Farm",
+        "slug": "the-farm",
+        "uuid": "43e12a9e-ed8e-4552-a889-d86e284b21c0",
+        "concepts": [],
+        "prerequisites": [],
+        "status": "wip"
+      },
```
ErikSchierboom: Why did you add this entry? Was it to fix CI?

exercises/concept/census/.docs/hints.md
```diff
@@ -2,19 +2,27 @@
 
 ## 1. Create a new resident
 
-- Revisit the [structs concept][concept-structs] for information on how to create structs.
+- Revisit the structs concept for information on how to create structs.
+
+[concept:go/structs]()
```
ErikSchierboom, line 7: For now, concept widgets render inline:

```suggestion
- Revisit the [concept:go/structs]() concept for information on how to create structs.
```

exercises/concept/census/.docs/hints.md
```diff
@@ -2,19 +2,27 @@
...
 
 ## 2. Validate resident information
 
 - The zero value for a string is `""`.
 - The zero value for a map is `nil`.
-- Revisit the [maps concept][concept-maps] for information on how to determine if a map key exists.
+- Revisit the maps concept for information on how to determine if a map key exists.
+
+[concept:go/maps]()
```
ErikSchierboom, line 15:
```suggestion
- Revisit the [concept:go/maps]() concept for information on how to determine if a map key exists.
```

exercises/concept/census/.docs/hints.md
```diff
@@ -2,19 +2,27 @@
...
 
 ## 3. Delete resident information
 
 - A resident is considered deleted when all of their fields are set to their zero value.
-- Revisit the [zero-values concept][concept-zero-values] for the zero value for each Go type.
+- Revisit the zero-values concept for the zero value for each Go type.
+
+[concept:go/zero-values]()
```
ErikSchierboom, line 22:
```suggestion
- Revisit the [concept:go/zero-values]() concept for the zero value for each Go type.
```

exercises/concept/census/.docs/hints.md
```diff
@@ -2,19 +2,27 @@
...
 
 ## 4. Count the residents
 
-- Revisit the [range-iteration concept][concept-range-iteration] for information on how to range over a slice.
+- Revisit the range-iteration concept for information on how to range over a slice.
+
+[concept:go/range-iteration]()
```
ErikSchierboom, line 28:
```suggestion
- Revisit the [concept:go/range-iteration]() concept for information on how to range over a slice.
```

config.json
sudomateo: I did not add this. I merged `main` into this branch. This entry is already on `main` [here](https://github.com/exercism/go/blob/main/config.json#L237-L244).

concepts/nil/about.md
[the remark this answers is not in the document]
junedev: I agree with @sudomateo this is hard to separate out in this context. What about having this exercise just teach one concept `zero-values` and have all the content we really need in there?

Then we don't have the duplication and we can still decide later on to add separate concept here or in the new "pointers" exercise (https://github.com/exercism/go/issues/1643).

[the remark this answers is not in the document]
ErikSchierboom: Sounds good.

exercises/concept/census/.docs/hints.md
```diff
@@ -0,0 +1,20 @@
+# Hints
+
+## 1. Create a new resident
+
+- Revisit the [concept:go/structs]() concept for information on how to create structs.
```
junedev, line 5: @ErikSchierboom Does it make sense to use widgets inside of a line of text? I thought they render as rather big boxes once they are actually implemented. I would have expected this to be a normal relative link to e.g. `/tracks/go/concepts/structs`.

exercises/concept/census/.docs/instructions.md
```diff
@@ -0,0 +1,96 @@
...
+## 2. Validate required information
+
+Residents may be reluctant to provide personal data to census workers. In those cases it's necessary to determine if the resident provided the required information to be counted in the census.
+
+In order to be counted, a resident must provide a non-zero value for their name and an address that contains a non-zero value for the `street` key. All other information, such as the resident's age, is optional. Implement the `HasRequiredInfo` method that returns a boolean indicating if the resident has provided the required information.
+
+```go
+name := "Matthew Sanabria"
+age := 0
+address := make(map[string]string, 0)
```
junedev, line 37: `make` should be only called with 1 argument in case of map

exercises/concept/census/.docs/instructions.md
```diff
@@ -0,0 +1,96 @@
...
+- The name of the resident.
+- The age of the resident.
+- The address of the resident.
+
+The function should return a pointer to a `Resident` struct that holds this information.
+
+```go
+name := "Matthew Sanabria"
+age := 29
+address := map[string]string{"street": "Main St.",}
```
junedev, line 22: no comma at end if the bracket is not on a new line

exercises/concept/census/census_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+				residents = append(residents, resident)
+			}
+
+			if got := census.Count(residents); got != test.want {
+				t.Errorf("Count() = %d, want %d", got, test.want)
+			}
+		})
+	}
+}
```
junedev, line 143: A separate test function for each function/method would give much better test output for the student. Not sure the rules of the Go track also allow the current version. I asked on Slack.

junedev:
> Nice! @junedev could you look at the Go side of things?

Done

exercises/concept/census/census_test.go
ekingery: Thanks for pointing this out, @junedev. As written, the entire test will be returned for every failure. In order to parse out the individual subtests properly, this test would need to be refactored to loop through the test cases immediately after they are defined. Here is the spec: https://github.com/exercism/go-test-runner#subtests

exercises/concept/census/.docs/instructions.md
sudomateo: In this case, yep!

exercises/concept/census/.docs/hints.md
sudomateo: I'm happy to implement either way here.

exercises/concept/census/census_test.go
sudomateo: Updated to use multiple test functions.

concepts/nil/about.md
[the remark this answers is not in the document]
sudomateo: @junedev Does this mean we should remove the `nil` concept entirely and move its unique information into the `zero-value` concept?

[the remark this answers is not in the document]
junedev: Exactly!

concepts/zero-values/about.md
```diff
@@ -1,30 +1,57 @@
...
-The predeclared identifier `nil`, meaning zero, is the zero value for the more complex types in Go. Comparing a type whose zero value is not `nil` to `nil`, like string, is an error:
+| boolean   | `false`    |
+| numeric   | `0`        |
+| string    | `""`       |
+| pointer   | `nil`      |
+| function  | `nil`      |
+| interface | `nil`      |
+| slice     | `nil`      |
+| channel   | `nil`      |
+| map       | `nil`      |
```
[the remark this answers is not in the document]
junedev: Sorry, I got confused with the nil concept. Can you double check some short info on structs is also available in the introduction of the concept and exercise. I didn't see it there.

concepts/zero-values/about.md
```diff
@@ -1,30 +1,57 @@
...
+## Zero Value Construction
+
+The `var` keyword can be used to construct any type to its zero value:
+
+```go
+func main() {
+  var myBool bool
+  fmt.Println("Zero value boolean:", myBool)
+}
+```
```
junedev, line 32: You can remove the main function from those examples if you want. But in any case this is missing a line below that shows the actual result, it is usually written like this on the Go track:  
`// Output: Zero value ....`

Sorry for all the nit picks, in general you did great work and I really appreciate your help!

concepts/nil/about.md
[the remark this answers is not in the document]
sudomateo: Removed the `nil` concept and updated the `zero-value` concept with its information.

concepts/zero-values/about.md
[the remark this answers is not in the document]
sudomateo: Done!

exercises/concept/census/.docs/hints.md
sudomateo: I used normal relative links to the relevant concept.

concepts/zero-values/about.md
sudomateo: Never a nit pick! Thank you for the feedback.

junedev (approved):

junedev:
@sudomateo The past thing that remains is getting the CI to pass. 

For the prerequisites you need to write "string-formatting" instead of "strings" for now. (It will be changed later when "strings" becomes available. Additionally you need to remove "pointers" for now as it does not exists yet. You can leave a comment here https://github.com/exercism/go/issues/1643 that it needs to be added again later.

junedev:
As for the other CI failures, can you try to use "package census" instead of "package census_test" for the test file? This is how it was done in the other concept exercises. (Not sure this will fix the issue though.)

ekingery (approved):
Rubber stamping based on the other reviews / fixes and the tests passing.

sudomateo:
Fixed the merge conflict.

ekingery:
> Fixed the merge conflict.

By the way, thanks for your work on this exercise, @sudomateo! Looks great.

sudomateo:
> > Fixed the merge conflict.
> 
> By the way, thanks for your work on this exercise, @sudomateo! Looks great.

Anytime! I'm happy to help!

ErikSchierboom:
Yeah, great work @sudomateo!

PR #1498, opened 2021-09-03
Corrects one minor typo

renan-eccel:
I got wandering if the third character was the dog or Annalyn itself 😁

PR #1502, opened 2021-09-04
annalyns-infiltration: Fix CanFreePrisoner comment

louis-cl:
Very silly change, but sounds small enough to practice collaborating :smile: 
Now it matches with introduction text.


louis-cl:
Duplicated https://github.com/exercism/go/pull/1498

PR #1509, opened 2021-09-05
Remove comments prereq until that exercise is live

iHiD:


iHiD:
@ekingery @oanaOM I'm just pushing this to get things tidied but please could you see whether the prereqs can be added back in with the correct labels etc?

Issue #1503, opened 2021-09-04
cars_assemble setup failed


lnds:
Fail in web editor and locally shows this error:

    cars_assemble_test.go:4:2: package numbers is not in GOROOT (/usr/local/opt/go/libexec/src/numbers)
    FAIL    cars [setup failed]

louis-cl:
Should be fixed by https://github.com/exercism/go/commit/56bb8a80292b84d91793f852404c8197fdc01469

PR #1510, opened 2021-09-05
cars-assemble: Fix tests

louis-cl:
Fixes https://github.com/exercism/go/issues/1503
Test were expecting solution to be in package numbers.
Also I changed the tests to accept results within a margin of test case value(1e-9)


louis-cl:
Already solved by https://github.com/exercism/go/pull/1519
Opening new PR for the float comparison (that failed on my machine)

PR #1512, opened 2021-09-06
getmonth takes a string as input

LucaPaterlini:


ekingery:
@LucaPaterlini I think you may be confusing the `untyped numeric constants` as strings?

LucaPaterlini:
Could be, which library shall the user imports them from? ( Thinking face)

ekingery:
> Could be, which library shall the user imports them from? ( Thinking face)

I think the idea is that you have to define them. See the `.meta/examplar.go` file for a cheat sheet ;-)

PR #1514, opened 2021-09-06
TestGetMonth string as input

LucaPaterlini:


ekingery:
@LucaPaterlini I think you may be confusing the untyped numeric constants as strings? Same with #1512

PR #1516, opened 2021-09-06
removed phantom numbers package

LucaPaterlini:


ekingery:
@LucaPaterlini I'm sorry but #1519 had caught my attention last week. I am just seeing this now. It is the correct fix, but has already been merged into `main`. Apologies!

PR #1517, opened 2021-09-06
Updated parametes to be consistent with the tests

LucaPaterlini:


LucaPaterlini:
need to be amended because the unit should be int not string as well

jmrunkle:
Thanks for this PR. It also requires a docs update which I already took care of in #1538.

PR #1519, opened 2021-09-06
Fixing Test / Main Packaging

geoffreywiseman:
Test seems to expect the 'system under test' to be in a numbers package, but it's in a `cars` package, same as the test -- when working on this locally, the simplest thing seemed to be to remove the import and drop the qualified method names, which was in keeping with the way `savings-account` was written. There are other paths, but this is simple.

### DIFF
--- a/exercises/concept/cars-assemble/cars_assemble_test.go
+++ b/exercises/concept/cars-assemble/cars_assemble_test.go
@@ -1,7 +1,6 @@
 package cars
 
 import (
-	"numbers"
 	"testing"
 )
 
@@ -40,7 +39,7 @@ func TestCalculateProductionRatePerHour(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got := numbers.CalculateProductionRatePerHour(tt.speed)
+			got := CalculateProductionRatePerHour(tt.speed)
 			if got != tt.want {
 				t.Errorf(
 					"CalculateProductionRatePerHour(%d) = %f, want %f",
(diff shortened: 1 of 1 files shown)


ekingery (approved):
Thanks @geoffreywiseman for your contribution! The change looks good to me.

bitfield:
May I add a comment to this? It's conventional, if not entirely universal, to put Go tests in a different package from the one being tested. This is partly to enforce the principle that we should only test our public API, and partly so that if your package-qualified names don't make sense, you'll spot it right away because you have to use them in the tests.

So if I were working on this, I might put the test into `package cars_test`, and have it import `cars` and reference `cars.Calculate...`

ekingery:
> May I add a comment to this? It's conventional, if not entirely universal, to put Go tests in a different package from the one being tested. This is partly to enforce the principle that we should only test our public API, and partly so that if your package-qualified names don't make sense, you'll spot it right away because you have to use them in the tests.
> 
> So if I were working on this, I might put the test into `package cars_test`, and have it import `cars` and reference `cars.Calculate...`

Thanks for the comment, @bitfield! I didn't see it initially. I do agree with you. As you are probably aware, this would represent a change in the majority of Go exercises. I wouldn't want to start this with one-off fixes because I think the consistency is more valuable than the changes in the small number we'd likely catch and fix. Unfortunately we don't have an active lead for the Go track who could help or advise on the bigger initiative of converting all of the exercises. Unfortunately at the moment we're struggling to get the CI to pass!

geoffreywiseman:
Yeah, I am not opposed to applying that pattern, but agree that it should be consistent. My main goal was to fix the failure, secondary goal was consistency.

bitfield:
> this would represent a change in the majority of Go exercises

Yes, I'm absolutely fine with that. Let's do it!

> agree that it should be consistent

Then we _all_ agree!

PR #1521, opened 2021-09-08
Correct typo in annalyns_infiltration exercise

jessealter:
There are only three characters in total (not counting the player and the player's dog, who do not sleep), so if the prisoner is awake then there are only two characters remaining (the knight and the archer).


andrerfcsantos:
A PR for this was already opened: https://github.com/exercism/go/pull/1478

jessealter:
Duplicate of #1478

PR #1523, opened 2021-09-09
gross-store: Fixing failing UTC

jitcompile:
Updated the unit test case call for AddItem as the function only accepts 3 arguments and not 4
Updated the unit test case call for RemoveItem as the function only accepts 3 arguments and not 4
Removed unnecessary call to store units

### DIFF
--- a/exercises/concept/gross-store/gross_store_test.go
+++ b/exercises/concept/gross-store/gross_store_test.go
@@ -66,12 +66,12 @@ func TestAddItem(t *testing.T) {
 			true,
 		},
 	}
-	units := Units()
+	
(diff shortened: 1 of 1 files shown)


geoffreywiseman (approved):
This change looks ok to me -- it's the PR I would have made.

ekingery:
I like the simplicity here, but I decided to merge #1538 as it appeared to be a more comprehensive solution, and included the docs updates. @jitcompile, thanks for your contributions. If you still feel that the current version in `main` needs adjusted, feel free to merge it and re-open / submit. Alternatively, you can reach me on slack and we can discuss there.

PR #1530, opened 2021-09-09
cars-assemble: Compare floats with threshold in test

louis-cl:
Split from https://github.com/exercism/go/pull/1510
On my machine tests failed because of a very small float difference, this is doing a more robust float comparison


ekingery (approved):
Hmm, I'm a little hesitant on this, but I don't have a better solution. The tests still worked for me, so let's merge it.

PR #1538, opened 2021-09-12
Add missing parameters to stub

### LINKED ISSUE #1527: Gross Store wrong arguments number or wrong signature of AddItem and RemoveItem functions
In "Gross Store" exercise the function `AddItem` has the following signature:

```go
func AddItem(bill map[string]int, item string, unit string) bool {}
```

3 arguments.

In test file it is called:

```go
ok := AddItem(bill, units, item.name, item.unit)
```

4 arguments.

The same is for `RemoveItem` function.

What should be changed? The test or the source?

### LINKED ISSUE #1537: gross-store function signatures for AddItem and RemoveItem are missing "units map[string]int" parameter
The tests require the units parameter:

* https://github.com/exercism/go/blob/main/exercises/concept/gross-store/gross_store_test.go#L74
* https://github.com/exercism/go/blob/main/exercises/concept/gross-store/gross_store_test.go#L155

The exemplar has the correct parameters:

* https://github.com/exercism/go/blob/main/exercises/concept/gross-store/.meta/exemplar.go#L21
* https://github.com/exercism/go/blob/main/exercises/concept/gross-store/.meta/exemplar.go#L32

But the stub given to students does not:

* https://github.com/exercism/go/blob/main/exercises/concept/gross-store/gross_store.go#L14
* https://github.com/exercism/go/blob/main/exercises/concept/gross-store/gross_store.go#L19

This is particularly confusing and frustrating for students using the browser editor because they cannot see the tests to know why they are failing to compile correctly.

jmrunkle:
Fixes #1527

### DIFF
--- a/exercises/concept/gross-store/.docs/instructions.md
+++ b/exercises/concept/gross-store/.docs/instructions.md
@@ -18,12 +18,12 @@ In order to use the measurement, you need to store the measurement in your progr
 ```go
 units := Units()
 fmt.Println(units)
-// Output: map[...]
+// Output: map[...] with entries like ("dozen": 12)
 ```
 
 ## 2. Create a new customer bill
 
-You need to implement a function that create a new bill for the customer.
+You need to implement a function that create a new (empty) bill for the customer.
 
 ```go
 bill := NewBill()
@@ -35,14 +35,15 @@ fmt.Println(bill)
 
 To implement this, you'll need to:
 
-- Check whether the given unit of measurement is correct
-- Add the item to the customer bill, indexed by the item name. You probably also need a variable to represent the customer bill, you are expected to use `map`
+- Return `false` if the given `unit` is not in the `units` map.
+- Add the item to the customer `bill`, indexed by the item name and return `true`.
 
 ```go
 bill := NewBill()
-ok := AddItem(bill, "carrot", "dozen")
+units := Units()
+ok := AddItem(bill, units, "carrot", "dozen")
 fmt.Println(ok)
-// Output: true or false
+// Output: true (since dozen is a valid unit)
 ```
 
 > Note that the returned value is type of `bool`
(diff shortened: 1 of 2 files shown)


geoffreywiseman:
If exercism were to use this path, the instructions should also be updated, since they show invocations without the units.

I still prefer the other path, mostly for aesthetic reasons, but I recognize that either is a valid fix.

jmrunkle:
> If exercism were to use this path, the instructions should also be updated, since they show invocations without the units.

Updated the instructions which were pretty bad / not explicit enough even if we used only 3 params.

jmrunkle:
Apparently gocritic complains if you do not combine the two parameters with the same type:

```
exercises/concept/gross-store/gross_store.go:14:1: paramTypeCombine: func(bill map[string]int, units map[string]int, item string, unit string) bool could be replaced with func(bill, units map[string]int, item, unit string) bool (gocritic)
func AddItem(bill map[string]int, units map[string]int, item string, unit string) bool {
^
exercises/concept/gross-store/gross_store.go:19:1: paramTypeCombine: func(bill map[string]int, units map[string]int, item string, unit string) bool could be replaced with func(bill, units map[string]int, item, unit string) bool (gocritic)
func RemoveItem(bill map[string]int, units map[string]int, item string, unit string) bool {
```

I'll do that even though I find it less clear personally.

jmrunkle:
NOTE: All other test failures are pre-existing issues. There's actually one more set for gross-store, but I did not change the tests (so also pre-existing):

```
exercises/concept/gross-store/gross_store_test.go:14:2: SA4006: this value of `tests` is never used (staticcheck)
	tests := []struct {
	^
exercises/concept/gross-store/gross_store_test.go:42:2: SA4006: this value of `tests` is never used (staticcheck)
	tests := []struct {
	^
exercises/concept/gross-store/gross_store_test.go:100:2: SA4006: this value of `tests` is never used (staticcheck)
	tests := []struct {
	^
```

They sure look used to me...

ekingery (approved):
I don't fully understand the nuances here and can't commit the time to test thoroughly, but the code looks good. Thanks @jmrunkle!

PR #1541, opened 2021-09-12
Comment/wording update for CanFreePrisoner test

dstockto:
I believe there are only 3 people total in reference, the prisoner, the knight and the archer. I think the comment should read that if the prisoner is awake and the other *2* characters are asleep ...


dstockto:
Looks like there have been several pull requests that do this. Closing this one.

Issue #1495, opened 2021-09-02
Syllabus -> Constants = blocked by "Comments" but "Comments" not visible.


goatthinking:
While going through the Syllabus I completed the following concepts:
* Basics
* Nil
* Zero Values
* Booleans
* Strings

My next concept would be "Constants" which shows as Locked w/ message: `Learn Basics, Comments, and Strings to unlock this concept.`

However, there is no "Comments" available to do. While looking at things quick I think it should be related to this in some way, but could be very wrong.
https://github.com/exercism/go/blob/2f58e6d0aaf6c379d02741c59d3b8ba297f7d3c3/config.json#L71-L82

Good work on the V3 so far! It looks amazing. Thank you to everyone who has put in time to accomplish this.

SterverSVK:
+1 to priority

louis-cl:
At a quick glance, it is missing tests
https://github.com/exercism/go/blob/a167c7853aed6b62bc3117e798f8cc51755a4bde/exercises/concept/weather-forecast/weather_test.go#L1

pwierer:
Same here. +1

oanaOM:
Thanks everyone for reporting this! sorry for the inconvenience.  Will work on fixing this today.

iHiD:
I've hotfixed this here: https://github.com/exercism/go/pull/1509

Hopefully @oanaOM will be able to come up with a more solid fix later :)

micuffaro:
If it's only about adding proper tests to `go/exercises/concept/weather-forecast/weather_test.go`, I can probably take a stab at it.
Would that be alright @oanaOM , @iHiD ?
Anything else besides the tests that I need to be aware of?

ekingery:
The code looks good (and the tests pass with the exemplar). Thanks and nice work using `ast`, @micuffaro! I am seeing three items to cover before or just after merge:

1. In the first test, do we want to support >= 4, or will the AST parsing definitely catch and count multi-line comments as 1? 
2. Can you please add a `go.mod` file? 
```
module weather

go 1.13
```
3. We'll want to re-enable comments after merge, undoing [L92](https://github.com/exercism/go/pull/1509/files#diff-587cb980af76fdc7e52369fd0b9d926dff266976b6f8ac631e358fecc49ff8cfL92) in #1509

micuffaro:
Thanks @ekingery.

> In the first test, do we want to support >= 4, or will the AST parsing definitely catch and count multi-line comments as 1?

[...]

PR #1543, opened 2021-09-12
Add tests for weather concept exercise

micuffaro:
addresses #1495


ekingery (changes requested):
Code looks good! I added comments in #1495 regarding a go.mod file and a question about the first test which expects exactly 4 comments.

ekingery (approved):

Issue #1485, opened 2021-08-24
Fix CI


ErikSchierboom:
This repository's `main` branch currently does not pass all CI checks.
This makes detecting issues in PRs harder and might result in a broken track, and so is an important thing to fix.

The following CI workflows are currently failing:

https://github.com/exercism/go/runs/3382441936
https://github.com/exercism/go/runs/3382441888
https://github.com/exercism/go/runs/3382441808
https://github.com/exercism/go/runs/3382441775
https://github.com/exercism/go/runs/3382441694
https://github.com/exercism/go/runs/3382441690

## Tracking

https://github.com/exercism/v3-launch/issues/55

jmrunkle:
I will make sure this issue is taken care of. #1534 seems like a good start. I will make sure all of the CI issues are addressed before I close this. Still working on getting write access to this repo.

jmrunkle:
I believe there are still a few remaining issues with the CI.

Here are the errors from configlet:

```
The Concept Exercise `chessboard` has `range-iteration` in its `concepts`, which is not a `slug` in the top-level `concepts` array:
/home/runner/work/go/go/config.json

The `authors` array is empty:
/home/runner/work/go/go/concepts/anonymous-functions/.meta/config.json

The `authors` array is empty:
/home/runner/work/go/go/concepts/application-timing/.meta/config.json

The `authors` array is empty:
/home/runner/work/go/go/concepts/bytes/.meta/config.json

The `authors` array is empty:
/home/runner/work/go/go/concepts/channels/.meta/config.json

The `authors` array is empty:
/home/runner/work/go/go/concepts/conditionals/.meta/config.json

The `authors` array is empty:
/home/runner/work/go/go/concepts/context/.meta/config.json

The `authors` array is empty:
/home/runner/work/go/go/concepts/defer/.meta/config.json

The `authors` array is empty:
/home/runner/work/go/go/concepts/duration/.meta/config.json

The `authors` array is empty:
/home/runner/work/go/go/concepts/empty-interface/.meta/config.json

The `authors` array is empty:
/home/runner/work/go/go/concepts/floating-point/.meta/config.json

The `authors` array is empty:
/home/runner/work/go/go/concepts/functions/.meta/config.json

The `authors` array is empty:
/home/runner/work/go/go/concepts/goroutines/.meta/config.json

The `authors` array is empty:
/home/runner/work/go/go/concepts/interfaces/.meta/config.json

The `authors` array is empty:
[...]

PR #1544, opened 2021-09-12
Fix configlet lint issues

jmrunkle:
Addresses the configlet lint issues from #1485

Namely: one missing slug and a bunch of empty author entries.


ekingery (approved):
I think Ideally the true author would be dug out of the commit history (assuming these are more than stub exercises), but I am approving because I think giving credit to @ErikSchierboom is better than having a broken build.

jmrunkle:
I did. The concepts are technically orthogonal to the exercises and the commits for the stuff in the concepts comes from Erik.

PR #1547, opened 2021-09-12
Fix test-without-stubs to work with new structure

jmrunkle:
The former structure assumed the exercises were all directly under the exercises directory, but that broke in v3 once we added sub-directories "concept" and "practice"

I also found the whole thing a bit wonky since it was doing the work directly in the directories where the actual files that we want to upload are happening. This made it very tricky to test locally. Instead I just opted to move everything over to a temporary directory (at the cost of the errors now complaining about the temporary directory).

This also just fixes the issues uncovered by this now that we can properly identify them. There were a number of them.

This is one more part to fixing #1485


jmrunkle:
NOTE: the configlet lint will be fixed by #1544

ekingery (approved):
Thanks so much for your efforts here, and nice shell scripting! The changes look good overall. You may run into some conflicts with a few of these (like gross-store which might have just had a similar fix merged). I also know there was some debate about the matrix test (I just commented on that in [#1475](https://github.com/exercism/go/pull/1475#issuecomment-917657443)). That said, I think getting the build fixed is the highest priority and we can revisit any of the specifics as-needed in a future issue/PR.

jmrunkle:
> I also know there was some debate about the matrix test

I think we are in agreement here that getting this working is the most important issue. Personally, I see no good reason to tie two exercises together like that, but it is a debate for another time.

I think I merged everything into this PR that was merged into main so I think we are good to go.

PR #1568, opened 2021-09-14
annalyns-infiltration: remove const and ; from instructions

### LINKED ISSUE #1552: annalyns-infiltration: instructions should not use `const`
The [instructions.md file](https://github.com/exercism/go/blob/main/exercises/concept/annalyns-infiltration/.docs/instructions.md#1-check-if-a-fast-attack-can-be-made) of the `annalyns-infiltration` exercise uses `const` in the code samples, but the exercise's [only prerequisite](https://github.com/exercism/go/blob/main/config.json#L55) is `basics`, which only teaches [about variables](https://github.com/exercism/go/blob/main/concepts/basics/about.md#variables), and not constants.

The easy fix is to just use `var` in the instructions (note that what syntax can be used is determined by the exercise's (transitive) prerequisites).

### LINKED ISSUE #1554: annalyns-infiltration: don't use unneeded semicolons
The [instructions.md file](https://github.com/exercism/go/blob/main/exercises/concept/annalyns-infiltration/.docs/instructions.md#1-check-if-a-fast-attack-can-be-made) of the `annalyns-infiltration` exercise use a semicolon which is unneeded (as far as my basic Go knowledge goes):

```go
const knightIsAwake = false;
const archerIsAwake = true;
const prisonerIsAwake = false;
fmt.Println(CanSpy(knightIsAwake, archerIsAwake, prisonerIsAwake))
// Output: true
```

The easy fix is to remove these semicolons:

```go
const knightIsAwake = false
const archerIsAwake = true
const prisonerIsAwake = false
fmt.Println(CanSpy(knightIsAwake, archerIsAwake, prisonerIsAwake))
// Output: true
```

andrerfcsantos:
Fixes #1552
Fixes #1554

### DIFF
--- a/exercises/concept/annalyns-infiltration/.docs/instructions.md
+++ b/exercises/concept/annalyns-infiltration/.docs/instructions.md
@@ -18,7 +18,7 @@ You have four tasks: to implement the logic for determining if the above actions
 Define the `CanFastAttack()` function that takes a boolean value that indicates if the knight is awake. This function returns `true` if a fast attack can be made based on the state of the knight. Otherwise, returns `false`:
 
 ```go
-const knightIsAwake = true;
+var knightIsAwake = true
 fmt.Println(CanFastAttack(knightIsAwake))
 // Output: false
 ```
@@ -28,9 +28,9 @@ fmt.Println(CanFastAttack(knightIsAwake))
 Define the `CanSpy()` function that takes three boolean values, indicating if the knight, archer and the prisoner, respectively, are awake. The function returns `true` if the group can be spied upon, based on the state of the three characters. Otherwise, returns `false`:
 
 ```go
-const knightIsAwake = false;
-const archerIsAwake = true;
-const prisonerIsAwake = false;
+var knightIsAwake = false
+var archerIsAwake = true
+var prisonerIsAwake = false
 fmt.Println(CanSpy(knightIsAwake, archerIsAwake, prisonerIsAwake))
 // Output: true
 ```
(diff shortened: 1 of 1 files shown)


exercises/concept/annalyns-infiltration/.docs/instructions.md
```diff
@@ -18,7 +18,7 @@ You have four tasks: to implement the logic for determining if the above actions
 Define the `CanFastAttack()` function that takes a boolean value that indicates if the knight is awake. This function returns `true` if a fast attack can be made based on the state of the knight. Otherwise, returns `false`:
 
 ```go
-const knightIsAwake = true;
+var knightIsAwake = true
```
jmrunkle, line 21: I feel like the := syntax is more commonly used for variable declarations. Should we steer students towards that?

andrerfcsantos: On the other hand, using the full var declaration is more "context agnostic" as it can be used outside any function. I'm just thinking about the use case where a student might want to copy/paste this and might copy/paste the code to be outside any function - the full var declaration will work.

But I don't really have a strong opinion about this, so if you feel the change to `:=` is better, I'll happily make it.

jmrunkle: No strong preference on my part, especially if there is a reasonable pedagogical reason for using this one.

jmrunkle (approved):

PR #1569, opened 2021-09-14
Need for speed misleading comments

### LINKED ISSUE #1536: need-for-speed comments for the Drive function do not match tests
The comment says:

> If there is not enough battry to drive on more time, the car will not move **but use the leftover battery**.

The bold part of the text is wrong because the test for this asserts that the leftover battery is not used: https://github.com/exercism/go/blob/main/exercises/concept/need-for-speed/need_for_speed_test.go#L111

I think the incorrect comment should be removed / corrected so that students are not confused by this.

voroskoi:
Hi,

Those comments are misleading as tests do not expect the battery to be drained.

(Sorry, I could not squash those two commits on github webui)

Fixes #1536


jmrunkle (approved):
Thanks for fixing this

Issue #1553, opened 2021-09-14
annalyns-infiltration: introduction should use `true` or `false` in code example


ErikSchierboom:
The [introduction.md document(https://github.com/exercism/go/blob/main/exercises/concept/annalyns-infiltration/.docs/introduction.md#introduction) of `annalyns-infiltration` has the following syntax example:

```go
var a bool
```

As `true` and `false` are also new keywords that are introduced, I think it would make more sense to also use one of these keywords in the example. For example:

```go
speeding := true
```

PR #1570, opened 2021-09-14
annalyn-infiltration: add uses of 'true' and 'false' to examples

andrerfcsantos:
Fixes: #1553

### DIFF
--- a/exercises/concept/annalyns-infiltration/.docs/introduction.md
+++ b/exercises/concept/annalyns-infiltration/.docs/introduction.md
@@ -4,7 +4,9 @@ Booleans in Go are represented by the predeclared boolean type `bool`, which val
 It's a defined type.
 
 ```go
-var a bool
+var closed bool    // boolean variable 'closed' implicitly initialized with 'false'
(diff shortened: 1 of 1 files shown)


exercises/concept/annalyns-infiltration/.docs/introduction.md
```diff
@@ -4,7 +4,9 @@ Booleans in Go are represented by the predeclared boolean type `bool`, which val
 It's a defined type.
 
 ```go
-var a bool
+var a bool         // boolean variable 'a' implicitly initialized with 'false'
+speeding := true   // boolean variable 'speeding' initialized with 'true'
+hasError := false  // boolean variable 'hasError' initialized with 'false'
```
jmrunkle, line 9: Just noting a general inconsistency with the variables. Either rename `a` to something semantically meaningful on the lines above or else have these two be `b` and `c` as we would expect.

andrerfcsantos: Fixed. Ended up changing the name of `a` to `closed`.

dittrichlucas (approved):

Issue #1561, opened 2021-09-14
numbers: cleanup introduction.md document


ErikSchierboom:
The [introduction.md file](https://github.com/exercism/go/blob/main/concepts/numbers/introduction.md) of the `numbers` concept contains a TODO on the document needing to be cleaned up.

Looking at the contents, the only things that need to be done are:

- Remove the level 2 headings
- Remove the TODO

Issue #1562, opened 2021-09-14
numbers: about document is fairly empty


ErikSchierboom:
The [about.md document](https://github.com/exercism/go/blob/main/concepts/numbers/about.md) of the `numbers` concept seems to be missing most of its content when compared to its [introduction.md document](https://github.com/exercism/go/blob/main/concepts/numbers/introduction.md). In general, the `about.md` document should contain _all_ the information from the `introduction.md`, but it can also add links to external resources, more in-depth discussion of the topic, etc. The https://exercism.org/docs/building/tracks/concepts page describes these documents in detail.

For now, the minimal fix would be to copy the content from the `introduction.md` file to the `about.md` file (minus the things mentioned here: https://github.com/exercism/go/issues/1561) and add any missing links. The existing link in the `about.md` file is probably good to have.

andrerfcsantos:
Submitted the quick fix for now. But I think this concept can be improved further. Made my suggestions in a separate issue: https://github.com/exercism/go/issues/1574

PR #1572, opened 2021-09-14
numbers concept: cleanup introduction.md

andrerfcsantos:
Fixes: #1561

### DIFF
--- a/concepts/numbers/introduction.md
+++ b/concepts/numbers/introduction.md
@@ -1,11 +1,5 @@
 # Introduction
 
-TODO: the content below is copied from the exercise introduction and probably needs rewriting to a proper concept introduction
-
-## numbers
-
-## conditionals
-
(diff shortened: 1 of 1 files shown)


concepts/numbers/introduction.md
```diff
@@ -1,11 +1,5 @@
 # Introduction
 
-TODO: the content below is copied from the exercise introduction and probably needs rewriting to a proper concept introduction
```
jmrunkle, line 3: This TODO is not really addressed. There are still mentions of the exercise in the section below, but this material is supposed to be exercise agnostic.

andrerfcsantos: I opened a separate issue for this #1562 - I agree with you, but I think this change is out of the scope of this PR as it requires thinking about `about.md` too.

jmrunkle (approved):

PR #1573, opened 2021-09-14
numbers concept: fix empty about.md

andrerfcsantos:
Fixes: #1562

### DIFF
--- a/concepts/numbers/about.md
+++ b/concepts/numbers/about.md
@@ -1,8 +1,50 @@
 # About
 
-## Type Conversion
+Go contains basic numeric types that can represent sets of either integer or
+floating-point values. There a different types depending on the size of value
+you require and the architecture of the computer where the application is
+running (e.g. 32-bit and 64-bit).
+
+For the sake of this exercise you will only be dealing with:
+
+- `int`: e.g. `0`, `255`, `2147483647`. A signed integer that is at least 32
+  bits in size (value range of: -2147483648 through 2147483647). But this will
+  depend on the systems architecture. Most modern computers are 64 bit,
+  therefore `int` will be 64 bits in size (value rate of:
+  -9223372036854775808 through 9223372036854775807).
+
+- `float64`: e.g. `0.0`, `3.14`. Contains the set of all 64-bit floating-point
+  numbers.
+
+For a full list of the available numeric types and more detail see the
+following resources:
+
+- [Go builtin type declarations][go-builtins]
+- [Digital Ocean - Understanding Data Types in Go][do-understanding-types]
+- [Arden Labs - Understanding Type in Go][arden-understanding-types]
+
+Go supports the standard set of arithmetic operators of `+`, `-`, `*`, `/`
+and `%` (remainder not modulo).
+
+In Go, assignment of a value between different types requires explicit
+conversion. For example, to convert an `int` to a `float64` you would need to
+do the following:
+
+```go
+var x int = 42
+f := float64(x)
+
+fmt.Printf("x is of type: %s\n", reflect.TypeOf(x))
+// Output: x is of type: int
+
+fmt.Printf("f is of type: %s\n", reflect.TypeOf(f))
+// Output: f is of type: float64
+```
 
 For more information on Type Conversion please see
 [A Tour of Go: Type Conversions][type-conversion]
 
 [type-conversion]: https://tour.golang.org/basics/13
+[go-builtins]: https://github.com/golang/go/blob/master/src/builtin/builtin.go
+[do-understanding-types]: https://www.digitalocean.com/community/tutorials/understanding-data-types-in-go
+[arden-understanding-types]: https://www.ardanlabs.com/blog/2013/07/understanding-type-in-go.html


concepts/numbers/about.md
```diff
@@ -1,8 +1,50 @@
 # About
 
-## Type Conversion
+Go contains basic numeric types that can represent sets of either integer or
+floating-point values. There a different types depending on the size of value
+you require and the architecture of the computer where the application is
+running (e.g. 32-bit and 64-bit).
+
+For the sake of this exercise you will only be dealing with:
```
jmrunkle, line 8: These concepts are not necessarily viewed within the context of a specific exercise, so please remove any mention of a particular exercise.

concepts/numbers/about.md
```diff
@@ -1,8 +1,50 @@
...
+
+- `int`: e.g. `0`, `255`, `2147483647`. A signed integer that is at least 32
+  bits in size (value range of: -2147483648 through 2147483647). But this will
+  depend on the systems architecture. Most modern computers are 64 bit,
+  therefore `int` will be 64 bits in size (value rate of:
+  -9223372036854775808 through 9223372036854775807).
+
```
jmrunkle, line 15: This seems to be missing an explanation of the other int types: https://tour.golang.org/basics/11

In particular, that you can get an explicit 64-bit version using int64 and uint64.

concepts/numbers/about.md
```diff
@@ -1,8 +1,50 @@
 # About
 
-## Type Conversion
+Go contains basic numeric types that can represent sets of either integer or
```
jmrunkle, line 3: Feel free to disregard if you disagree, but I advocate putting line breaks at sentence breaks in this documentation. It line wraps both in the UI and in the reviewing UI, but breaking at sentences means that your changes to individual sentences end up having less of a blast radius.

The way it is currently broken down means that you have to reformat the whole paragraph whenever you edit one sentence (causing a larger blast radius and making it slightly harder to review).

As I said, feel free to disregard if you prefer this style though.

concepts/numbers/about.md
```diff
@@ -1,8 +1,50 @@
...
+```go
+var x int = 42
+f := float64(x)
+
+fmt.Printf("x is of type: %s\n", reflect.TypeOf(x))
+// Output: x is of type: int
+
+fmt.Printf("f is of type: %s\n", reflect.TypeOf(f))
+// Output: f is of type: float64
+```
```
jmrunkle, line 43: We should probably explicitly mention what happens with division by zero.

jmrunkle (changes requested):
Thanks for doing this!

andrerfcsantos:
While I agree with the suggested changes, I think they belong in a separate PR, as some of these changes would probably need to be reflected into `instroduction.md`. In fact, I opened [an issue](https://github.com/exercism/go/issues/1574) together with this PR to improve this description further - your points can be added there.

The point of this PR is just to quick fix  #1562 and provide a basis for further improvement.

ErikSchierboom (approved):

concepts/numbers/about.md
ErikSchierboom: I like the above suggestion.

concepts/numbers/about.md
ErikSchierboom: That depends. What _does_ happen? If what happens involves a yet unknown concept (which is likely), then it shouldn't be mentioned here but instead in the documentation of that concept.

concepts/numbers/about.md
ErikSchierboom: I would argue not to mention the other integer types. The goal of this exercise is very simple: teach students how to use the two most-used numeric types: `int` and `float64`. There is so much more that can be taught about integers and floats that there should be separate exercises to teach that. For example, C# has the [hyper-optimized-telemetry exercise](https://github.com/exercism/csharp/tree/main/exercises/concept/hyper-optimized-telemetry) to teach about all the integer types, and the [interest-is-interesting exercise](https://github.com/exercism/csharp/tree/main/exercises/concept/interest-is-interesting) to teach about all the floating point types.

ErikSchierboom (approved):

ErikSchierboom:
>The point of this PR is just to quick fix #1562 and provide a basis for further improvement.

@jmrunkle I agree with this statement by @andrerfcsantos. I'd prefer to get this merged and then do changes in a follow-up PR.

Note to both: I've commented on some on @jmrunkle's comments, which can help understand why this exercise was designed as it is.

concepts/numbers/about.md
jmrunkle: I mean it is nothing you can react to, so you have to code against it. If you write a literal 0 for the division it will not compile. If a variable may be zero it causes a panic.

ErikSchierboom: I'd be fine with mentioning briefly what happens, but I would refrain from using terminology that the student won't yet understand. Maybe something like: "division with a literal zero won't compile and division with a variable that contains zero will fail" or something like that.

jmrunkle (approved):

PR #1606, opened 2021-09-15
Fix the typo in the test output function name

vissree:
The output of 'CanFreePrisoner' tests has the wrong function name mentioned. Correcting it so that the test outputs are not confusing.


jmrunkle (approved):
Thanks!

Issue #1607, opened 2021-09-16
Benchmark tests should not run on the website


jmrunkle:
This requires changes both in the test code as well as in the actual test runner, but the simplest option is to [skip tests](https://pkg.go.dev/testing#hdr-Skipping) using the short flag.

I will be creating another issue in the go test runner repo to make the relevant change for the test runner itself so that it sets the `-short` flag when running `go test`.

By using the short flag, CLI students that simply run `go test` will automatically see the benchmarks, but students in the UI will not be weighed down by them.

ErikSchierboom:
Closed by https://github.com/exercism/go-test-runner/issues/37

PR #1608, opened 2021-09-16
Skip benchmark tests in short mode

jmrunkle:
Part of #1607 - this prepares the way for us to use short mode in the test runner to skip these tests.


jmrunkle:
NOTE: I basically produced this using sed and then used sed to reformat. I think it is correct, but feel free to look more closely. The fact that the tests still pass CI gives me pretty good confidence that this is correct.

ekingery (approved):
Pretty nifty, looks good. Thanks!

PR #1627, opened 2021-09-16
Have card-tricks only use int type

### LINKED ISSUE #1600: card-tricks: replace uint8 with int
The `card-tricks` exercise uses `uint8` but the prerequisites haven't unlocked this concept. The uses of `uint8` should be replaced with `int`.

jmrunkle:
Fixes #1600 

None of the concepts teach about `uint8` and there is no real reason to confuse the students so we will just use `int` type everywhere in this exercise.


ekingery (approved):
As discussed in slack, :shipit: 

Thanks!

iHiD:
This closes https://github.com/exercism/go/issues/1600 btw.

ErikSchierboom:
Thanks!

PR #1635, opened 2021-09-17
Mild tweaks to CI tests

jmrunkle:
1. use /bin/bash
2. only check for the expected exemplar files (example.go and exemplar.go)
3. specify the backup name so that the sed commands work with Mac


bin/test-without-stubs
```diff
@@ -25,18 +26,24 @@ for parentdir in concept practice; do
...
+      exit 1
+    fi
     mod="${dir}go.mod"
     if [ ! -f $mod ]; then
-      echo "Creating a basic go module at $mod, like this:"
+      echo "Please add a basic go module at $mod, like this:"
       cat << EOD
 module $stub
 
 go 1.14
```
ekingery, line 43:
```suggestion
go 1.15
```

bin/test-without-stubs
```diff
@@ -68,7 +73,7 @@ trap cleanup EXIT INT TERM
 # This is the last command in the file,
 # so the exit status of this script is the exit status of go test.
 # If this ever changes, remember to preserve the exit status accordingly.
-# Otherwise Travis may falsely report a test as passed when it has failed.
+# Otherwise CI may falsely report a test as passed when it has failed.
```
ekingery, line 76: 🎉

ekingery (approved):

PR #1647, opened 2021-09-17
Additional test case for practice/luhn that often fails

moshegood:
I've mentored this particular problem for a while, and this is a commonly seen failing test case.


jmrunkle:
This exercise comes from the problem-specifications and contains [canonical data](https://github.com/exercism/problem-specifications/blob/main/exercises/luhn/canonical-data.json). If you think more needs to be tested here, I think you should first create an issue over there, get them to agree to update the canonical data for this exercise and then add a test case for this. Once that is done, feel free to re-open this PR.

Issue #1625, opened 2021-09-16
party-robot: change taught concepts


ErikSchierboom:
The `party-robot` exercise currently is designed to teach the `strings` concept, but this will be done by the Poetry Door Club policy exercise that is to be added. As such, the goal of this exercise changes to: teach the `string-formatting` concept (and the [`package-import` concept](https://github.com/exercism/go/issues/1586).

The `strings` concept should be renamed to `string-formatting`: https://github.com/exercism/go/blob/main/config.json#L64

bobtfish:
This is done!

junedev:
@bobtfish I think there is a part missing here. The the exercise introduction and the concept introduction and about file still contain a section that explain strings in general. That section should be removed as this part will be covered by another exercise.

You can paste the text as a comment into this issue https://github.com/exercism/go/issues/1613 so the person creating the string exercise/concept can use it.

Sorry that I did not notice this in the review earlier but it was a bit hard to understand which PR was supposed to cover which issue. You can help with that by writing "resolves #xyz" (where xyz is the issue number) into the PR description. GitHub will then automatically create a link. See https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword

PR #1648, opened 2021-09-17
Change party-robot to teach the string-formatting concept (#1625)

bobtfish:
I've changed all existing exeercises to reference  rather than string so that
the config stays consistent, we'll want to update this again when we have the Poetry Door Club policy exercise.


config.json
```diff
@@ -258,7 +258,7 @@
         "prerequisites": [],
```
junedev, line 258: All the "topics" fields below should not be changed. Topics are not used anymore by the website but they serve as orientation when setting "prerequisites" and "practices" keys. The "topic" of those exercises is still "strings", not "string-formatting".

config.json
```diff
@@ -1825,8 +1825,8 @@
       "uuid": "23e1c40a-300c-4a44-bc32-f3440a057216"
     },
     {
-      "name": "Strings",
-      "slug": "strings",
+      "name": "String formatting",
```
junedev, line 1828: Since this is the title that shows for example in the concept box in the tree, it should be "String Formatting".
```suggestion
      "name": "String Formatting",
```

junedev (changes requested):

junedev (commented):
The CI says there is one more exercise ("Elons Toys") where strings needs to be changed to string-formatting for now in the config.json.

bobtfish:
> The CI says there is one more exercise ("Elons Toys") where strings needs to be changed to string-formatting for now in the config.json.

Yeah, I can't see where this is at all...

https://github.com/bobtfish/go/blob/party-robot-changeconcept/config.json#L167-L178

What am I missing @junedev ? 🙃

junedev:
@bobtfish This was merged into the main branch yesterday: https://github.com/exercism/go/blob/4e6ef8292315a04a51775ed7d72dfa4d9b70ef52/config.json#L176

You need to merge the main branch, then you should be able to see "strings" there and then you can update it.

bobtfish:
> @bobtfish This was merged into the main branch yesterday:
> 
> https://github.com/exercism/go/blob/4e6ef8292315a04a51775ed7d72dfa4d9b70ef52/config.json#L176
> 
> You need to merge the main branch, then you should be able to see "strings" there and then you can update it.

AHA! Thank you, that's what I was missing <3

junedev (approved):

Issue #1613, opened 2021-09-16
[Sept Sprint] New Exercise for Strings


iHiD:
This is part of our September Spring to get the Go Track in shape. Please read the [Overview issue](https://github.com/exercism/go/issues/1610) for more details. **This exercise is unclaimed. Please comment below if you'd like to work on it.**

---

**Before building this exercise, please read up on [the documentation](https://exercism.org/docs/building/tracks/concept-exercises).**

Copy/paste/modify from the [corresponding Javascript exercise](https://github.com/exercism/javascript/blob/main/exercises/concept/poetry-club-door-policy/). This exercise will be teaching the `strings` concept instead of having `logs-logs-logs` teach it. The latter will be modified to teach about `runes`. 

You can reuse lots of the text from here: https://github.com/exercism/go/blob/e41496d3ae8e03b2bf17ea0686c3fe78f1206ba9/exercises/concept/party-robot/.docs/introduction.md (That text is getting removed)

- The new exercise should give a basic introduction to strings (double quotes, escape sequences, backticks, `+`). 
-  Additionally this exercise should introduce the strings package. This can be done as part of the `strings` concept or as a separate concept `strings-package`. In any case this content should show how to access things from the package and show one usage example and a table with commonly used functions from the package. The concept about page should link to the full list of functions from the docs.
- It must  **not** use runes, slices, mutation, UTF-8 encoding or string formatting. This might require changes to the exercise.

The new exercise will have one prerequisite: `basics`.

For reference, the Swift track has also [ported the JS exercise](https://github.com/exercism/swift/tree/main/exercises/concept/poetry-club).

junedev:
@ErikSchierboom I just realized the poetry club door policy exercise relies on syntax like `str[0]`. In Go that does not return a string and we didn't want to go into the details of runes, type conversion etc at this point because this can be the second learning exercise the student does on the Go track. Not sure how to "fix" that problem yet. Maybe we need to chat about it on Monday.

bobtfish:
Text removed from the `string-formatting` concept in issue #1625 to be used in this concept / exercise is below:

# About

A `string` in Go is an immutable sequence of bytes. Strings may contain arbitrary bytes but usually they contain human-readable text.
Text strings are conventionally interpreted as UTF-8 encoded sequence of Unicode code points (runes) which will be explained in a future exercise.
A `string` value can be written as a string literal, which is a sequence of bytes enclosed in double quotes:

```go
s := "Hello World!"
```

[...]

PR #1652, opened 2021-09-18
Remove text generally about strings (and not fmt) from party-robot.

### LINKED ISSUE #1625: party-robot: change taught concepts

bobtfish:
This exercise is becoming just about string formatting.
The text removed from here will be re-added in the strings concept being introduced in issue #1613

Closes #1625

### DIFF
--- a/concepts/string-formatting/about.md
+++ b/concepts/string-formatting/about.md
@@ -1,29 +1,5 @@
 # About
 
-A `string` in Go is an immutable sequence of bytes. Strings may contain arbitrary bytes but usually they contain human-readable text.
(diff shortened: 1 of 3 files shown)


junedev (changes requested):
The change also needs to be applied to the exercise introduction file.

junedev (approved):

Issue #1590, opened 2021-09-15
need-for-speed: consider not pre-defining the structs


ErikSchierboom:
The whole purpose of the `need-for-speed` exercise is to teach students how to define and work with `struct` types. The [stub file](https://github.com/exercism/go/blob/main/exercises/concept/need-for-speed/need_for_speed.go#L4) already defines the struct types though, which means that the student won't have to do this themselves. This is not wrong per se, but is likely a sub-optimal learning experience, as we prefer to have new syntax be written from scratch by students.

Tracks usually "solve" this by adding a comment in the source code with a TODO requesting the student to define a certain type of syntax. See [this example](https://github.com/exercism/fsharp/blob/main/exercises/concept/bandwagoner/Bandwagoner.fs#L3).

PR #1656, opened 2021-09-19
fix: remove struct and add TODO

jaimem88:
Fixes https://github.com/exercism/go/issues/1590

### DIFF
--- a/exercises/concept/need-for-speed/.docs/instructions.md
+++ b/exercises/concept/need-for-speed/.docs/instructions.md
@@ -1,8 +1,10 @@
 # Instructions
 
-In this exercise you'll be organizing races between various types of remote controlled cars. Each car has its own speed and battery drain characteristics.
+In this exercise you'll be organizing races between various types of remote controlled cars.
+Each car has its own speed and battery drain characteristics.
 
-Cars start with full (100%) batteries. Each time you drive the car using the remote control, it covers the car's speed in meters and decreases the remaining battery percentage by its battery drain.
+Cars start with full (100%) batteries. Each time you drive the car using the remote control,
+it covers the car's speed in meters and decreases the remaining battery percentage by its battery drain.
 
 If a car's battery is below its battery drain percentage, you can't drive the car anymore.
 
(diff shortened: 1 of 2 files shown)


jaimem88:
@iHiD @ErikSchierboom ready for review 😄

junedev (changes requested):
Now that the stub is changed, the instructions need to be changed as well so that the student knows what fields to include in the struct. Currently e.g. distance is not mentioned at all (and it looks like even the example output in task 1 is incorrect).

ErikSchierboom:
@jaimem88 I'll let @junedev review, she actually knows Go and I don't :)

jaimem88:
Thank you both! I'll get back to this in a couple of days 🙇

exercises/concept/need-for-speed/.docs/instructions.md
```diff
@@ -1,22 +1,32 @@
 # Instructions
 
-In this exercise you'll be organizing races between various types of remote controlled cars. Each car has its own speed and battery drain characteristics.
+In this exercise you'll be organizing races between various types of remote controlled cars.
+Each car has its own speed and battery drain characteristics.
```
jaimem88, line 4: not sure if there are specific guidelines for text, but it helps to read by splitting into multiple lines. I can revert if prefered :)

exercises/concept/need-for-speed/need_for_speed.go
```diff
@@ -1,27 +1,16 @@
...
 	panic("Please implement the NewCar function")
 }
 
-
-// Track implements a race track.
-type Track struct{
-	distance int
-}
-
+// TODO: define the 'Track' type struct
```
jaimem88, line 10: Removed the `type Track struct` part, it's also defined in the instructions 😄

exercises/concept/need-for-speed/need_for_speed.go
```diff
@@ -1,27 +1,16 @@
...
 
 // NewTrack created a new track
-func NewTrack(distance int) Track{
+func NewTrack(distance int) Track {
```
jaimem88, line 13: I run goimports locally so it took care of this

jaimem88:
> Thanks for taking care of this! Now that the stub is changed, the instructions need to be changed as well so that the student knows what fields to include in the struct. Currently e.g. distance is not mentioned at all (and it looks like even the example output in task 1 is incorrect). Would be great if you could include this in your PR so the content remains consistent.

@junedev I've updated the docs a bit, let me know what you think!

junedev (approved):

exercises/concept/need-for-speed/.docs/instructions.md
ErikSchierboom: This is totally fine!

ErikSchierboom:
Thanks again!

PR #1658, opened 2021-09-19
blackjack: move FirstTurn function to tests

### LINKED ISSUE #1558: blackjack: move `FirstTurn` function from stub to tests file
The [stub file](https://github.com/exercism/go/blob/main/exercises/concept/blackjack/blackjack.go#L24) of the `blackjack` exercise contains a `FirstTurn` function which the comments state should not be edited. This is problematic for two reasons:

1. Students might not notice the comment and edit this function, leading to invalid implementations
2. Students might notice the comment and edit the function, purposefully getting invalid implementations to pass the tests

For the above reasons, this function should really be moved into the [test file](https://github.com/exercism/go/blob/main/exercises/concept/blackjack/blackjack_test.go).

andrerfcsantos:
Fixes #1558


andrerfcsantos:
Some CI tests are failing but I believe this is some sort of problem with the CI - it's saying the function is declared twice, but the function was just moved to another file. Caching issue maybe?

jmrunkle (approved):

PR #1659, opened 2021-09-19
blackjack: streamline introduction

### LINKED ISSUE #1556: blackjack: streamline introduction
Firstly, the [introduction.md file](https://github.com/exercism/go/blob/main/exercises/concept/blackjack/.docs/introduction.md#introduction) of the `blackjack` exercise states that a `switch` allows for more "advanced scenarios" than an `if` statement, but it doesn't clearly state _what_ those advanced scenarios are. There are some hints in the comments in the syntax example, but it would likely be easier for the student if the text itself would explain how a `switch` is different from an `if` statement and in what ways.

Secondly, some explicit mention of the `case` and `default` keywords is probably also useful. 

Thirdly, the syntax example uses method invocation syntax (`strings.HasPrefix(value, "val")`), but I don't think students have been taught this concept looking at [the prerequisites](https://github.com/exercism/go/blob/main/config.json#L103). Looking at [the exemplar implementation](https://github.com/exercism/go/blob/main/exercises/concept/blackjack/.meta/exemplar.go#L52), students just need to know that a `switch` can have arbitrary boolean expressions in which for example numbers are compared. I would recommend changing the syntax example to something similar to the exemplar implementation.

andrerfcsantos:
Fixes #1556

Summary of the changes:

 * I revamped the introduction of the concept and the introduction of the exercise as per the suggestions in #1556
 * I also made the `about.md` of the concept to be more comprehensive
 * More friendly link names for the concept

### DIFF
--- a/concepts/conditionals-switch/.meta/config.json
+++ b/concepts/conditionals-switch/.meta/config.json
@@ -1,5 +1,8 @@
 {
   "blurb": "Switches control the flow of a program based on the value of a variable or expression.",
-  "authors": ["tehsphinx"],
+  "authors": [
+    "tehsphinx",
+    "andrerfcsantos"
+  ],
   "contributors": []
 }
--- a/exercises/concept/blackjack/.docs/introduction.md
+++ b/exercises/concept/blackjack/.docs/introduction.md
@@ -1,25 +1,33 @@
 # Introduction
 
-Instead of using an `if` statement for conditional logic, Go also provides a `switch` statement for more advanced scenarios.
+Like other languages, Go also provides a `switch` statement. Switch statements are a shorter way to write long `if ... else if` statements. To make a switch, we start by using the keyword `switch` followed by a value or expression. We then declare each one of the conditions with the `case` keyword. We can also declare a `default` case, that will run when none of the previous `case` conditions matched:
 
 ```go
-var value string
+operatingSystem := "windows"
 
-// switch statement on variable content
-switch value {
-case "val1":
-    // conditional code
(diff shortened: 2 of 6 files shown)


concepts/conditionals-switch/.meta/config.json
```diff
@@ -1,5 +1,7 @@
 {
   "blurb": "Switches control the flow of a program based on the value of a variable or expression.",
   "authors": ["tehsphinx"],
-  "contributors": []
+  "contributors": [
+    "andrerfcsantos"
```
junedev, line 5: You did so much work for the concept, you should put yourself as author.

exercises/concept/blackjack/.docs/introduction.md
```diff
@@ -1,25 +1,38 @@
...
+case age > 20 && age < 30:
+    // do something if age is between 20 and 30
+case age == 10:
+    // do something if age is equal to 10
+default:
+    // do something else for every other case
 }
 ```
+
+To learn more about this topic, check [Go by Example: Switch][go_by_example_switch] or [Tour of Go: Switch][switch_statement].
```
junedev, line 35: Please remove this line and the link in line 3 from the exercise introduction. The introduction is "self-sufficient" without them as it should be and we don't really want the student to wander off when they are just about to work on the exercise.

In the concept docs, the links are ok.

junedev (approved):
Great work! Left 2 minor comments.

andrerfcsantos:
Thanks for the review @junedev.

Applied the suggested changes.

Issue #1621, opened 2021-09-16
[Sept Sprint] Gross Store


iHiD:
This is part of our September Spring to get the Go Track in shape. Please read the [Overview issue](https://github.com/exercism/go/issues/1610) for more details. **This exercise is claimed by @MiroslavGatsanoga .**

---

- [ ] https://github.com/exercism/go/issues/1593
- [ ] https://github.com/exercism/go/issues/1592
- [x] https://github.com/exercism/go/issues/1591

MiroslavGatsanoga:
I'd like to work on this one :)

Issue #1619, opened 2021-09-16
card-tricks: introduce multiple-return-values concept


ErikSchierboom:
This is part of our September Spring to get the Go Track in shape. Please read the [Overview issue](https://github.com/exercism/go/issues/1610) for more details. **This exercise is claimed by @jmrunkle.**

---

The `card-tricks` exercise uses multiple return values, but it does not have this as a taught concept.
This concept also includes teaching students how to do `if x, y := z; !y `. (i.e. checking for a value and then having a condition.
Note: don't teach about discard values (`_`).

- [ ] Add `multiple-return-values` as an official concept
  - The exercise introduction should introduce multiple return values
  - Add `multiple-return-values` to the `concepts` property of the `card-tricks` exercise's entry in the `config.json` file
  - Add an entry to the `concepts` array for the `multiple-return-values` concept
  - Create a `concepts/multiple-return-values` directory with its [required files](https://exercism.org/docs/building/tracks/concepts)

moshegood:
To be clear: Multiple return values should be its own concept connected to functions. It should not live inside the slices concept.

MiroslavGatsanoga:
Once the `multiple-return-values` concept is created, it needs to be added to Gross Store exercise.

junedev:
@moshegood What Erik means is that the exercise "Card Tricks" should teach two concepts, `slices` and `multiple-return-values`. 

I agree with you that the new exercise about functions would be a better fit but the problem @ErikSchierboom saw is that a lot exercises already need multiple return values before the student would get to functions in the concept tree. This is why he suggested to add this as a separate concept here.

ErikSchierboom:
Indeed. It is definitely possible for `multiple-return-values` to be explained in a new `functions` exercise, but we would have to do careful analysis to see if the dependencies make sense. For now, the easiest fix is to explain `multiple-return-values` alongside `slices` in the "Card Tricks" exercise and then we can always move this concept to a different exercise later on.

jmrunkle:
I am taking care of this issue. Not sure if the first comment on this issue should be updated to reflect that.

PR #1660, opened 2021-09-19
[Sept Sprint] Gross Store improvements

MiroslavGatsanoga:
Fixes https://github.com/exercism/go/issues/1621

### DIFF
--- a/exercises/concept/gross-store/.docs/introduction.md
+++ b/exercises/concept/gross-store/.docs/introduction.md
@@ -8,11 +8,7 @@ Syntactically, `map` looks like this:
 map[KeyType]ElementType
 ```
 
-_`KeyType` must be any [comparable type][gospec-comparable], while `ElementType` can be any valid type in go, which means you can store anything from primitive variable to a slice._
-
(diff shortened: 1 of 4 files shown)


exercises/concept/gross-store/.docs/introduction.md
```diff
@@ -8,12 +8,8 @@ Syntactically, `map` looks like this:
 map[KeyType]ElementType
 ```
 
-_`KeyType` must be any [comparable type][gospec-comparable], while `ElementType` can be any valid type in go, which means you can store anything from primitive variable to a slice._
```
junedev, line 11: Could you please apply the same changes you did for this introduction file to the maps concept `introduction.md`.

junedev:
@MiroslavGatsanoga Here is how you can fix the CI error for now: Instead of `strings` use `string-formatting`. We will correct this later once the strings exercise is in place. `multiple-return-values` does not exist, so you need to delete it. Instead leave a comment in this issue https://github.com/exercism/go/issues/1619 that it needs to be added back to the config for gross store once the concept was created.

junedev (approved):

PR #1663, opened 2021-09-19
Update the type conversion introduction

### LINKED ISSUE #1576: cars-assemble: update type conversion introduction
The [introduction.md document](https://github.com/exercism/go/blob/main/exercises/concept/cars-assemble/.docs/introduction.md) of the `cars-assemble` has the following section on type conversion:

````
In Go, assignment of a value between different types requires explicit
conversion. For example, to convert an `int` to a `float64` you would need to
do the following:

```go
var x int = 42
f := float64(x)

fmt.Printf("x is of type: %s\n", reflect.TypeOf(x))
// Output: x is of type: int

fmt.Printf("f is of type: %s\n", reflect.TypeOf(f))
// Output: f is of type: float64
```
````

There are two improvements to be made:

1. The `reflect.TypeOf` calls are using syntax that the student has never seen (it's not in the exercise's prerequisites). As such, it needs to be removed from the code example. One way to replace this is to just have comments describing the behavior
2. The introductory text does not really mention that converting between types is done via a function call which name is the same as the type to convert to. Making this explicit will help the student.

jmrunkle:
Fixes #1576

### DIFF
--- a/exercises/concept/cars-assemble/.docs/introduction.md
+++ b/exercises/concept/cars-assemble/.docs/introduction.md
@@ -24,17 +24,12 @@ Go supports the standard set of arithmetic operators of `+`, `-`, `*`, `/` and `
 ## Type Conversion
 
 In Go, assignment of a value between different types requires explicit conversion.
+Converting between types is done via a function with the name of the type to convert to.
 For example, to convert an `int` to a `float64` you would need to do the following:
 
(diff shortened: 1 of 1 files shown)


ekingery (approved):
Accomplishes the objectives in #1576, thanks!

PR #1664, opened 2021-09-20
Cars assemble enhance test output

vissree:
Add function name to the panic message and fix the function name typo in the test failure error message.

### DIFF
--- a/exercises/concept/cars-assemble/cars_assemble.go
+++ b/exercises/concept/cars-assemble/cars_assemble.go
@@ -3,13 +3,13 @@ package cars
 // CalculateProductionRatePerHour for the assembly line, taking into account
 // its success rate
 func CalculateProductionRatePerHour(speed int) float64 {
-	panic("not implemented")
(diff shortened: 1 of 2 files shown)


jmrunkle:
I am not convinced we want to go this route (ie. improving TDD). The concept exercises are supposed to be trivially easy and not really built around TDD. This is why (for example) we do not show the tests in the website editor for concept exercises.

@ekingery - do you have any thoughts?

ekingery:
> I am not convinced we want to go this route (ie. improving TDD). The concept exercises are supposed to be trivially easy and not really built around TDD. This is why (for example) we do not show the tests in the website editor for concept exercises.
> 
> @ekingery - do you have any thoughts?

We do want to fix the test, so that commit should be approved. As for the test names being in the panic, I agree that `panic("not implemented")` is sufficient. That said, I don't necessarily have an issue with hardcoding the test name into the panic argument.

junedev (approved):

PR #1676, opened 2021-09-22
Add a `multiple-return-values` concept to `card-tricks`

### LINKED ISSUE #1619: card-tricks: introduce multiple-return-values concept

jmrunkle:
Fixes #1619


bitfield:
This looks great! One interesting thing I've found from experience is that the metasyntactic variable `Foo` confuses people. You and I know what it means, because it's part of hacker culture, but of course the people we're talking to here don't have that background—they're in the process of acquiring it!

Accordingly, I try to avoid using `foo` (and `bar`) in examples, instead looking for some more concrete example (`User`, or something the reader can imagine being part of the problem domain). Do you have any suggestions for this one?

jmrunkle:
> This looks great! One interesting thing I've found from experience is that the metasyntactic variable `Foo` confuses people. You and I know what it means, because it's part of hacker culture, but of course the people we're talking to here don't have that background—they're in the process of acquiring it!
> 
> Accordingly, I try to avoid using `foo` (and `bar`) in examples, instead looking for some more concrete example (`User`, or something the reader can imagine being part of the problem domain). Do you have any suggestions for this one?

For sure - since we are in the `card-tricks` exercise, I can use `Card` instead of `Foo`.

ekingery (approved):
Nice!

PR #1682, opened 2021-09-22
Some fixes for the concept tree

junedev:
Now that we have some more concepts, we can set the correct perquisites for the exercise that teaches (advanced) "constants". This will move it further down in the tree. Also by removing "constants" as prerequisite for "cars assemple" (since we now cover it in basics), the concepts in cars-assemble will take the correct place in the concept tree.

I also fixed the prereqs for the `time` exercise and the currently wip errors exercise.

### DIFF
--- a/config.json
+++ b/config.json
@@ -89,9 +89,8 @@
           "constants"
         ],
         "prerequisites": [
-          "basics",
-          "comments",
-          "string-formatting"
+          "numbers",
+          "type-definitions"
         ],
         "status": "beta"
       },
@@ -118,9 +117,9 @@
           "time"
         ],
         "prerequisites": [
-          "basics",
           "string-formatting",
-          "numbers"
+          "numbers",
+          "multiple-return-values"
         ],
         "status": "beta"
       },
@@ -134,7 +133,7 @@
           "conditionals-if"
         ],
         "prerequisites": [
-          "constants"
+          "basics"
         ],
         "status": "beta"
       },
@@ -146,7 +145,6 @@
           "maps"
         ],
         "prerequisites": [
-          "basics",
           "conditionals-if",
           "numbers",
           "string-formatting",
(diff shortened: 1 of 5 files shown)


junedev:
@ErikSchierboom You can probably best review this.

jmrunkle:
This is a bit hard to review because the changes are somewhat context-less.

@junedev - Mind updating your initial comment on this PR to say why each change makes sense?

For example:

`savings-account`
- Remove `basics` because it relies on other concepts that rely on `basics`
- Remove `comments` because ...
- Remove `string-formatting` because ...
- Add `numbers` because ...

ErikSchierboom (approved):

junedev:
@jmrunkle Sorry, didn't have time for this yesterday.  Here you go:

**savings account - constants**
  - removed "basics" because we only use it as prerequisite for the very first exercises that don't have any other prerequisites as to not add is for all the exercises
  - removed "strings" (currently called string-formatting) because it is already a prerequisite to reach "type-defintions" (once we have the proper strings exercise in place) and there is nothing besides `"abc"` about strings in the exercise
  - removed comments because it does not have any more requirements on comments than all the other exercises that dont have comments as prerequisite
  - added "type-defintions" because the exercise uses `type Abc string`
  -  added numbers (although not strictly necessary because you cant reach type defs without it) because the exercise contains ints and floats and you need to understand numbers to understand numeric constants
 
**Booking up for Beauty - time**
  - removed basics (same reason as above)
  - added "multiple return values" because the time functions all return multiple return values

**Cars Assemble**
We added a basic introduction to constants to the basics exercise. the current constants exercise introduces advanced constants ("enum workarounds"). So I replaced constants with basics as prerequisites since we only need the what is covered in basics for the exercise.

**Gross Store - maps**
  - removed basics (same reason as mentioned for savings account)
 
**Deep Thought**
  - removed constants (same reason as mentioned for cars assemble above)
  - added the actual things that are needed in the exercise: string formatting  and multiple return values

jmrunkle:
Thanks June!

Issue #1611, opened 2021-09-16
[Sept Sprint] Cars Assemble


iHiD:
This is part of our September Spring to get the Go Track in shape. Please read the [Overview issue](https://github.com/exercism/go/issues/1610) for more details. **This exercise is claimed by @jmrunkle.**

---

## Cars Assemble

- [x] Add `arithmetic-operators` as a concept
  - Intro should have a table with the standard operators +,-,*,/,%, 
  - Intro should explain what happens for integer division (e.g. `5 / 2 = 2`)
  - Intro should introduce shorthand assignments `+=` etc.
  - Add `arithmetic-operators` to the `concepts` property of `cars-assembly` entry in the `config.json` file
  - Add an entry to the `concepts` array for the `arithmetic-operators` concept
  - Create a `concepts/arithmetic-operators` directory with its [required files](https://exercism.org/docs/building/tracks/concepts)
- [x] Add `comparison` as a concept
  - Add the `comparison` concept to the `cars assemble` exercise's `concepts` array in the config.json
  - Add an entry for the `comparison` concept to the `concepts` array in the config.json
  - Create a `concepts/comparison` directory with its [required files](https://exercism.org/docs/building/tracks/concepts) (check [this JS example](https://github.com/exercism/javascript/blob/main/concepts/comparison/introduction.md) for inspiration)
- [x] https://github.com/exercism/go/issues/1542
- [x] https://github.com/exercism/go/issues/1576
- [x] https://github.com/exercism/go/issues/1575
- [x] https://github.com/exercism/go/issues/1644
- [x] https://github.com/exercism/go/issues/1675
- [x] #1710

jmrunkle:
Claiming this one since it looks like low-hanging fruit.

iHiD:
Great. There's a question about whether this should also teach comparison. @junedev/@erikschierboom are discussing and will follow up on that on Tues :)

ErikSchierboom:
@jmrunkle https://github.com/exercism/go/issues/1675 Has been added to the list of tasks.

andrerfcsantos:
Since the comparison concept was previously planned to one of the exercises I was looking into, I already started making it and can take care of that task.

ErikSchierboom:
Perfect!

junedev:
@jmrunkle That was a copy and paste error. I will fix it.

[...]

PR #1686, opened 2021-09-23
Add `arithmetic-operators` as a concept for `cars-assemble`

jmrunkle:
Part of #1611

### DIFF
--- a/concepts/arithmetic-operators/about.md
+++ b/concepts/arithmetic-operators/about.md
@@ -48,3 +48,6 @@ a++ // same as a += 1, a == 11
 b := 10
 b-- // same as b -= 1, b == 9
 ```
+
+NOTE: these are statements and cannot be used as expressions (ie. they do not return a value).
+Also, only the postfix notation is allowed (ie. no `++a` or `--a`).
--- a/concepts/arithmetic-operators/about.md
+++ b/concepts/arithmetic-operators/about.md
@@ -1,0 +1,50 @@
+# About Arithmetic Operators
+
+## Operators
+
+Go supports many standard arithmetic operators:
+
+| Operator | Example        |
+| -------- | -------------- |
+| `+`      | `4 + 6 == 10`  |
+| `-`      | `15 - 10 == 5` |
+| `*`      | `2 * 3 == 6`   |
+| `/`      | `13 / 3 == 4`  |
+| `%`      | `13 % 3 == 1`  |
+
+For integer division, the remainder is dropped (eg. `5 / 2 == 2`).
+
+## Shorthand Assignments
+
+These can be used in shorthand assignments to update and assign a variable using the operator:
+
+```go
+a := 1
+a += 2 // same as a = a + 2 == 3
+
+b := 2
+b -= 2 // same as b = b - 2 == 0
+
+c := 4
+c *= 2 // same as c = c * 2 == 8
+
+d := 8
+d /= 2 // same as d = d / 2 == 4
+
+e := 16
+e %= 2 // same as e = e % 2 == 0
+```
+
+## Increment and Decrement
+
+There are also two special statements: increment (`++`) and decrement (`--`).
+They modify the value of a variable by increasing (or decreasing) the value by 1.
+For example:
+
+```go
+a := 10
+a++ // same as a += 1, a == 11
+
+b := 10
+b-- // same as b -= 1, b == 9
+```
--- a/concepts/arithmetic-operators/introduction.md
+++ b/concepts/arithmetic-operators/introduction.md
@@ -1,0 +1,50 @@
+# Introduction to Arithmetic Operators
+
+## Operators
+
+Go supports many standard arithmetic operators:
+
+| Operator | Example        |
+| -------- | -------------- |
+| `+`      | `4 + 6 == 10`  |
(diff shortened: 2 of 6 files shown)


concepts/arithmetic-operators/introduction.md
```diff
@@ -0,0 +1,50 @@
...
+They modify the value of a variable by increasing (or decreasing) the value by 1.
+For example:
+
+```go
+a := 10
+a++ // same as a += 1, a == 11
+
+b := 10
+b-- // same as b -= 1, b == 9
+```
```
ErikSchierboom, line 50: The exercise doesn't require the student to know this to solve the exercise. Could you remove (they're great for the `about.md`)?

```suggestion
```

exercises/concept/cars-assemble/.docs/introduction.md
```diff
@@ -15,6 +15,55 @@ For the sake of this exercise you will only be dealing with:
...
+For example:
+
+```go
+a := 10
+a++ // same as a += 1, a == 11
+
+b := 10
+b-- // same as b -= 1, b == 9
+```
+
```
ErikSchierboom, line 66: The exercise doesn't require the student to know this to solve the exercise. Could you remove (they're great for the `about.md`)?

```suggestion
```

ErikSchierboom (commented):
Awesome. Some minor nits.

concepts/arithmetic-operators/introduction.md
junedev: @ErikSchierboom I would recommend keeping a short version (1 example each) in the introduction and then maybe later adapt the exercise so they can we used. If we put it in about the student will not see them by default and they are needed in the following exercises. Also the Exercism docs say it is allowed to include introduction content that is necessary for the concept itself, that's how I would count this info.

ErikSchierboom: I'm fine with that, but I would make it a minimal version then. The current version is quite exhaustive (which is fine for the about.md, but less so for the introduction.md).

concepts/arithmetic-operators/about.md
```diff
@@ -0,0 +1,50 @@
...
+d /= 2 // same as d = d / 2 == 4
+
+e := 16
+e %= 2 // same as e = e % 2 == 0
+```
+
+## Increment and Decrement
+
+There are also two special statements: increment (`++`) and decrement (`--`).
+They modify the value of a variable by increasing (or decreasing) the value by 1.
```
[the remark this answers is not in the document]
junedev: Can here to make a comment about this ... but @ErikSchierboom was faster.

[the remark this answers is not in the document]
jmrunkle: That is actually why I specifically used the word "statements" on line 40. I can be more explicit though.

jmrunkle:
> Awesome. Some minor nits.

Addressed the comments (unless I misunderstood a comment). PTAL

exercises/concept/cars-assemble/.docs/introduction.md
```diff
@@ -15,6 +15,20 @@ For the sake of this exercise you will only be dealing with:
 
 Go supports the standard set of arithmetic operators of `+`, `-`, `*`, `/` and `%` (remainder not modulo).
 
+## Arithmetic Operators
```
junedev, line 18: The sentence on line 16 seems obsolete now.

exercises/concept/cars-assemble/.docs/introduction.md
```diff
@@ -15,6 +15,20 @@ For the sake of this exercise you will only be dealing with:
...
+| Operator | Example        |
+| -------- | -------------- |
+| `+`      | `4 + 6 == 10`  |
+| `-`      | `15 - 10 == 5` |
+| `*`      | `2 * 3 == 6`   |
+| `/`      | `13 / 3 == 4`  |
+| `%`      | `13 % 3 == 1`  |
+
+For integer division, the remainder is dropped (eg. `5 / 2 == 2`).
+
```
junedev, line 31: As said before this is the only file the student really sees so in the spirit of "introducing the concept" I would add one sentence like this even if it currently not part of the exercise:
"Go also has shorthand assignment operators for the operators above, e.g. `a += 5` is short for `a = a + 5`, and it support the increment and decrement operators `++` and `--`."

junedev (approved):
Two minor comments.

jmrunkle:
> Two minor comments.

Done

concepts/arithmetic-operators/introduction.md
```diff
@@ -0,0 +1,35 @@
+# Introduction to Arithmetic Operators
```
junedev, line 1: @ErikSchierboom Is it a problem if the h1 header is more than just the word "Introduction"? (same question for about)

jmrunkle: NOTE: if this matters, this is not the only one that would have a problem. Also, it is not mentioned in the contributing guides so it really must be in there if this is a thing that matters.

ErikSchierboom: @junedev @jmrunkle It doesn't matter. We strip the h1 heading

PR #1689, opened 2021-09-23
Add `comparison` concept

andrerfcsantos:
Part of #1611

### DIFF
--- a/concepts/comparison/about.md
+++ b/concepts/comparison/about.md
@@ -1,0 +1,53 @@
+# About
+
+In Go, we can compare values using the equality and ordering operators.
+We have 6 of these operators:
+
+| Comparison        | Operator  |
+| ------------------| --------- |
+| equal             | `==`      |
+| not equal         | `!=`      |
+| less              | `<`       |
+| less or equal     | `<=`      |
+| greater           | `>`       |
+| greater or equal  | `>=`      |
+
+The result of a comparison is always a boolean value:
+
+```go
+a := 3
+
+a != 4 // true
+a > 5  // false
+```
+
+## Equality operators
+
+The equality operators `==` and `!=` check whether a value is equal to another or not, respectively.
+We can use the equality operators on the majority of types in Go. Here are some common exmaples:
+
+```go
+2 == 3 // false, integer comparison
+
+2.1 != 2.2 // true, float comparison 
+
+"hello" == "hello" // true, string comparison
+
+'a' != 'b' // true, rune/byte comparison
+```
+
+## Ordering operators
+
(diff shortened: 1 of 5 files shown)


concepts/comparison/about.md
```diff
@@ -0,0 +1,49 @@
+# About
+
+In Go, we can compare values using the equality and ordering operators. We have 6 of these operators:
+
+| Comparison        | Operator  |
+| ------------------| --------- |
+| Equal             | `==`      |
+| Not Equal         | `!=`      |
```
jmrunkle, line 8:
```suggestion
| equal             | `==`      |
| not equal         | `!=`      |
```

Consistent capitazilation.

concepts/comparison/about.md
```diff
@@ -0,0 +1,49 @@
+# About
+
+In Go, we can compare values using the equality and ordering operators. We have 6 of these operators:
```
jmrunkle, line 3:
```suggestion
In Go, we can compare values using the equality and ordering operators.
We have 6 of these operators:
```

nit: one sentence per line

concepts/comparison/about.md
```diff
@@ -0,0 +1,49 @@
...
+```go
+a := 3
+
+a != 4 // true
+a > 5  // false
+```
+
+## Equality operators
+
+The equality operators `==` and `!=` check whether a value is equal to another or not, respectively. We can use the equality operators on the majority of types in Go. Here are some common exmaples:
```
jmrunkle, line 25:
```suggestion
The equality operators `==` and `!=` check whether a value is equal to another or not, respectively.
We can use the equality operators on the majority of types in Go. Here are some common exmaples:
```

concepts/comparison/about.md
```diff
@@ -0,0 +1,49 @@
...
+2.1 != 2.2 // true, float comparison 
+
+"hello" == "hello" // true, string comparison
+
+'a' != 'b' // true, rune/byte comparison
+```
+
+## Ordering operators
+
+Ordering operators check if one value is greater than (`>`), greater or equal to (`>=`), less than (`<`) and less or equal to (`<=`) to another value. This kind of comparison is available for numbers, bytes, runes and strings. When comparing strings, the dictionary order (also known as lexicographic order) is followed.
```
jmrunkle, line 39:
```suggestion
Ordering operators check if one value is greater than (`>`), greater or equal to (`>=`), less than (`<`) and less or equal to (`<=`) to another value.
This kind of comparison is available for numbers, bytes, runes and strings.
When comparing strings, the dictionary order (also known as lexicographic order) is followed.
```

jmrunkle (approved):
Mostly looks good, just some very nitpicky comments. Thanks!

andrerfcsantos:
Thanks for the review!

Suggestions applied to both files.

PR #1692, opened 2021-09-24
Add tasks to `the-farm` introduction

### LINKED ISSUE #1639: the-farm: add tasks to introduction
The [instructions.md document](https://github.com/exercism/go/blob/main/exercises/concept/the-farm/.docs/instructions.md) of the `the-farm` exercise does not conform to the standard. The first part is fine, but the second part should be formatted differently and have one or more tasks. See [these instructions](https://github.com/exercism/go/blob/main/exercises/concept/lasagna/.docs/instructions.md) for an example.

jmrunkle:
Also
- changed the tasks slightly to not only use a specific error, but also to return a specific error.
- change the test structure so that it uses the current style of table-driven tests

Fixes #1639

### DIFF
--- a/exercises/concept/the-farm/.docs/hints.md
+++ b/exercises/concept/the-farm/.docs/hints.md
@@ -6,10 +6,14 @@
 * use `errors.Is(error, error) bool` to check if it is an `ErrWeight` error
 * for any other error, return that back to the caller, with no computed value for division
 
-## 2. Prevent division by zero
+## 2. Return an error for negative fodder
 
 * use `errors.New(string)` to return a custom error for this case
 
-## 3. Return a `SillyNephew` error for a negative number of cows
+## 3. Prevent division by zero
+
+* use `errors.New(string)` to return a custom error for this case
+
+## 4. Return a `SillyNephew` error for a negative number of cows
 
 * return the custom `SillyNephew` error for this case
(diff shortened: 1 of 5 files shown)


bitfield:
One suggestion: instead of using sentinel errors (`ErrWeight`, `SillyNephew`), what do you think about using custom error types instead? The advantage of this would be that you can convey information about the invalid input within the error:

```go
return SillyNephew{fmt.Errorf("negative cows: %d", cows)}
```

The type of the error can be checked at the other end with `errors.As` (instead of `errors.Is` for sentinel errors).

jmrunkle:
> One suggestion: instead of using sentinel errors (`ErrWeight`, `SillyNephew`), what do you think about using custom error types instead? The advantage of this would be that you can convey information about the invalid input within the error:
> 
> ```go
> return SillyNephew{fmt.Errorf("negative cows: %d", cows)}
> ```
> 
> The type of the error can be checked at the other end with `errors.As` (instead of `errors.Is` for sentinel errors).

I like the idea, and there is another issue for teaching about interfaces in this exercise. I could just take care of both in this PR, but I was planning to do it separately since adding the `interfaces` concept will also involve adding a bunch of documentation / metadata for the new concept. Mind if I punt on that @bitfield?

bitfield:
Absolutely. Keep up the fine work!

junedev:
regarding the error wrapping (not relevant for this specific PR): I think it is important to also know how not to unwrap, so I would use this exercise to show the old equal and type assertion way of error checking and then use the "deep thought" exercise later to explain error wrapping and errors.Is and errors.As. Just a thought, feel free to ignore.

junedev (approved):

PR #1693, opened 2021-09-24
Update the string formatting examples to fix bug and be clearer

### LINKED ISSUE #1691: string-formatting: Example might be wrong
From the bug report dump on Slack

---

I think the example code is incorrect:

f := 4.3242
fmt.Printf("%.4", f)
// Output: 4.32

If the desired output is 4.32 given 4.3242, and you want rounding, then I think this is correct:

f := 4.3242
fmt.Printf("%.2f", f)
// Output: 4.32

For that matter, I think "f" is a confusing variable name given the Printf formatting operator "f". I would suggest "number".

---

INCORRECT EXAMPLE IN FLOATING POINT PRECISION
In your example under String track for Go, the example you gave for implementing floating point precision is this: 
f := 4.3242
fmt.Printf("%.4", f)
// Output: 4.32

The above won't work because the formatting verb (f) is not stated immediately after the precision figure. Also, the correct precision to arrive at that output will be 2 and not 4. It should be:

  f := 4.3242
fmt.Printf("%.2f", f)
// Output: 4.32

bobtfish:
Fixes #1691

### DIFF
--- a/concepts/string-formatting/about.md
+++ b/concepts/string-formatting/about.md
@@ -6,8 +6,8 @@ Go provides an in-built package called `fmt` (format package) which offers a var
 In Go floating point values are conveniently printed with Printf's verbs: `%g` (compact representation), `%e` (exponent) or `%f` (non exponent). All three verbs allow the field's width and numeric position to be controlled.
 
 ```go
-f := 4.3242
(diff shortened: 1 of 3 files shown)


jmrunkle (approved):
Thanks

PR #1697, opened 2021-09-25
variable-length-quantity: add test for bytes not fitting 32 bits

louis-cl:
I expected this to be covered but it was not.
Instructions make the student assume that values fit in uint32 but I think it's good to detect the situation.
Similar to how invalid VLQ sequences are.


jmrunkle:
These tests come from problem-specifications repo for variable-length-quantity and (in general) we do not want to vary too much from that repo. If you think this case should be tested, please create an issue PR over in that repo to get agreement that this should be checked before making a PR here: https://github.com/exercism/problem-specifications

PR #1698, opened 2021-09-26
doc: change the expected message to the correct function

vanildosouto:


junedev (approved):
Good catch, thanks for the fix!

Issue #1644, opened 2021-09-17
cars-assemble: explain `if expression; condition { }` syntax


ErikSchierboom:
The `cars-assemble` exercise explains how to do `if` statements in Go. One very common pattern in Go is to add a short statement before the condition in an if statement, see https://tour.golang.org/flowcontrol/6 for an example.

The `cars-assemble` exercise should teach this. At the very least, this should be added to the concept's and exercise's `introduction.md` document. Ideally, we'd modify the exercise such that the student should be using this pattern. This would involve:

- Add a new task or updating an existing one
- If a new function is used, updating the stub
- Updating the exemplar implementation
- Updating the exercise's introduction
- Updating the concept's introduction

PR #1699, opened 2021-09-26
Fix the introduction to the conditionals-if concept.

bobtfish:
Now covers the correct material for the related exercise.
No longer mentions switch which is covered elsewhere

### DIFF
--- a/concepts/conditionals-if/about.md
+++ b/concepts/conditionals-if/about.md
@@ -13,12 +13,14 @@ if value == "val" {
 In scenarios involving more than one case many `if` statements can be chained together using the `else if` and `else` statements.
 
 ```go
-if value == "val" {
-    // conditional code
-} else if value == "val2" {
-    // conditional code
+var number int
+
+if number > 0 {
+    return "positive"
+} else if number < 0 {
+    return "negative"
 } else {
-    // default code
+    return "zero"
 }
 ```
 
--- a/concepts/conditionals-if/about.md
+++ b/concepts/conditionals-if/about.md
@@ -6,7 +6,7 @@ Conditionals in Go are similar to conditionals in other languages. The underlyin
 var value string
 
 if value == "val" {
-    // conditional code
+    return "was val"
 }
 ```
 
@@ -14,13 +14,14 @@ In scenarios involving more than one case many `if` statements can be chained to
 
 ```go
 var number int
+result := "This number is "
 
 if number > 0 {
-    return "positive"
+    result += "positive"
 } else if number < 0 {
-    return "negative"
+    result += "negative"
 } else {
-    return "zero"
+    result += "zero"
 }
 ```
 
--- a/concepts/conditionals-if/introduction.md
+++ b/concepts/conditionals-if/introduction.md
@@ -9,9 +9,18 @@ if value == "val" {
     return "was val"
 }
 
+```
+
+In scenarios involving more than one case many `if` statements can be chained together using the `else if` and `else` statements.
+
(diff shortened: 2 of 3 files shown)


concepts/conditionals-if/introduction.md
```diff
@@ -1,57 +1,17 @@
...
 ```go
 var value string
 
 if value == "val" {
-    // conditional code
+    return "was val"
 }
-```
-
-In scenarios involving more than one case many `if` statements can be chained together using the `else if` and `else` statements.
```
junedev, line 23: Seeing an example for `else if` and `else` is important to understand this concept so I would recommend to not remove this part.

Maybe instead of giving 2 if-only examples you can have one if-only and one "full" version.

bobtfish:
Pretty sure that the test fail here is a flake:

```
level=error msg="Running error: context loading failed: failed to load packages: timed out to load packages: context deadline exceeded"
level=error msg="Timeout exceeded: try increasing it by passing --timeout option"
```

jmrunkle (approved):
Looks good. I think we should probably mention the fact that we can have an initial statement as part of the if statement. For example:

```go
if var := someFunction(); var == nil {
  ...
}
```

I just do not know where else we could introduce that idea. Anyway, it does not have to be part of this change so I'm merging this one.

junedev (commented):
Thanks for the fix.

If you have not done so already could you please check whether any of the changes also need to be reflected in the about.md file or in the concept introduction.

junedev:
@jmrunkle FYI We have a separate ticket for that part: https://github.com/exercism/go/issues/1644

jmrunkle:
> FYI We have a separate ticket for that part: #1644

Thanks - I knew I remembered seeing it somewhere. That issue is also assigned to me, so after we merge this, I can wrap it up.

jmrunkle (approved):

jmrunkle:
I want to give @junedev a chance to look at it again before we merge. June, feel free to merge if the changes look good to you.

concepts/conditionals-if/about.md
```diff
@@ -13,12 +13,14 @@ if value == "val" {
...
+var number int
+
+if number > 0 {
+    return "positive"
+} else if number < 0 {
+    return "negative"
 } else {
-    // default code
+    return "zero"
 }
```
junedev, line 24: In this example all the else statements are obsolete because of the returns.  
There are a couple of options to fix this that come to mind:

**Option 1**
Use fmt.Println instead of return. Drawback is the student might not know this function but it is pretty self explanatory.

**Option 2**
Show a less concrete example. Drawback is that it is less concrete. 
```go
if number > 0 {
    // Code that should be executed for positive numbers
} else if number < 0 {
    // Code that ....
} else {
    // Code that ...
}
```

**Option 3**
Change the example so returning is not the best option anymore. Make the example more complex.
```go
var number int
result := "This number is "

if number > 0 {
    result += "positive"
} else if number < 0 {
    result +="negative"
} else {
    result += "zero"
}
```

@jmrunkle Any preference?

junedev (commented):
If I see this correctly you fixed the concept introduction and about but the exercise introduction still shows the old version in the "conditionals" section (you might want to rename that section so "If Statements" while you are there.

junedev (approved):

concepts/conditionals-if/about.md
jmrunkle: I trust your judgement and it looks like it has been merged now. :D

PR #1705, opened 2021-09-27
Fill in `type-conversion` concept docs

### LINKED ISSUE #1702: `sorting-room`: fill out `type-conversion` concept documentation
Everything inside concepts/type-conversion is basically empty right now.

jmrunkle:
Fixes #1702

### DIFF
--- a/concepts/type-conversion/about.md
+++ b/concepts/type-conversion/about.md
@@ -16,3 +16,28 @@ For example, to convert an `int` to a `float64` you would need to do the followi
 var x int = 42 // x has type int
 f := float64(x) // f has type float64 (ie. 42.0)
 ```
+
+## Converting between primitive types and strings
+
+There is a `strconv` package for converting between primitive types (like `int`) and `string`.
+
+```go
+import "strconv"
+
+var intString string = "42"
+var i int = strconv.Atoi(intString)
+
+var number int = 12
+var s string = strconv.Itoa(number)
+```
+
+Here are some of the other common conversion methods:
+
+| Method      | Purpose                                       |
+| ----------- | --------------------------------------------- |
+| ParseBool   | Convert `string` to `bool`                    |
+| FormatBool  | Convert `bool` to `string`                    |
+| ParseFloat  | Convert `string` to `float`                   |
+| FormatFloat | Convert `float` to `string`                   |
+| ParseInt    | Convert `string` to `int` (if base is not 10) |
+| FormatInt   | Convert `int` to `string` (if base is not 10) |
--- a/concepts/type-conversion/about.md
+++ b/concepts/type-conversion/about.md
@@ -1,3 +1,19 @@
-# About
+# About Type Conversion
 
-TODO: add information on type-conversion concept
+In Go, assignment of a value between different types requires explicit conversion.
+Converting between types is done via a function with the name of the type to convert to.
+A value `v` is converted to type `T` using `T(v)`.
+For example, to convert an `int` to a `float64` you would need to do the following:
+
+```go
+var x int = 42 // x has type int
+f := float64(x) // f has type float64 (ie. 42.0)
+```
+
+Assignment between items of different type requires an explicit conversion.
+For example, this is a compiler error:
+
+```go
+var x int = 42
+var f float64 = x // requires float64(x) to compile
+```
--- a/concepts/type-conversion/about.md
+++ b/concepts/type-conversion/about.md
@@ -1,20 +1,28 @@
 # About Type Conversion
 
-In Go, assignment of a value between different types requires explicit conversion.
-For example, this is a compiler error:
+Go requires explicit conversion between different types.
+For example, this causes a compiler error:
 
 ```go
 var x int = 42
 var f float64 = x // compiler error x is an int not a float64
 ```
 
-Converting between types is done via a function with the name of the type to convert to.
+Converting between types (also know as **type casting**) is done via a function with the name of the type to convert to.
 A value `v` is converted to type `T` using `T(v)`.
 For example, to convert an `int` to a `float64` you would need to do the following:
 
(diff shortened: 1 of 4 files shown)


concepts/type-conversion/about.md
```diff
@@ -1,3 +1,18 @@
-# About
+# About Type Conversion
```
jmrunkle, line 1: I need a second opinion: should we mention about converting to/from string specifically using the strconv package?

```go
var fortyTwo string = "42"
v, _ := strconv.Atoi(fortyTwo) // convert string to int
var i int = 42
str := strconv.Itoa(i) // convert int to string
```

junedev: I would include the strconv package in this concept for now. It can then still be removed later if we want to make a separate exercise for it.

jmrunkle: Done

jmrunkle:
@junedev - PTAL?

concepts/type-conversion/about.md
```diff
@@ -1,3 +1,43 @@
...
+For example, to convert an `int` to a `float64` you would need to do the following:
+
+```go
+var x int = 42 // x has type int
+f := float64(x) // f has type float64 (ie. 42.0)
+```
+
+## Converting between primitive types and strings
+
+There is a `strconv` package for converting between primitive types (like `int`) and `string`.
```
junedev, line 22: As introduction to `strconv` I would use the following pitfall so the student is warned about that
```go
var num int = 65
str := string(num)
// str is now "A" not "65"
```
Probably not needed for the introduction but definitely good content for the about imo.

concepts/type-conversion/about.md
```diff
@@ -1,3 +1,43 @@
...
+Here are some of the other common conversion methods:
+
+| Method      | Purpose                                       |
+| ----------- | --------------------------------------------- |
+| ParseBool   | Convert `string` to `bool`                    |
+| FormatBool  | Convert `bool` to `string`                    |
+| ParseFloat  | Convert `string` to `float`                   |
+| FormatFloat | Convert `float` to `string`                   |
+| ParseInt    | Convert `string` to `int` (if base is not 10) |
+| FormatInt   | Convert `int` to `string` (if base is not 10) |
```
junedev, line 43: I would remove the "if base is not 10" from the lines above. Those functions work fine also for base 10 and tbh I prefer to use them over the alternatives because they are more readable if you don't already know "Atoi" from some other languages.

concepts/type-conversion/about.md
```diff
@@ -1,3 +1,43 @@
...
+
+Converting between types is done via a function with the name of the type to convert to.
+A value `v` is converted to type `T` using `T(v)`.
+For example, to convert an `int` to a `float64` you would need to do the following:
+
+```go
+var x int = 42 // x has type int
+f := float64(x) // f has type float64 (ie. 42.0)
+```
+
```
junedev, line 19: I would add a note that this conversion syntax also applies for custom types/non primitive types, e.g.
```go
type List []string
stringArray := []string{"a","b"}
list := List(stringArray)
```
Maybe you can think of a better example.

concepts/type-conversion/introduction.md
```diff
@@ -1,3 +1,25 @@
-# Introduction
+# Introduction to Type Conversion
 
-TODO: add introduction for type-conversion concept
+In Go, assignment of a value between different types requires explicit conversion.
+Converting between types is done via a function with the name of the type to convert to.
+A value `v` is converted to type `T` using `T(v)`.
```
junedev, line 5: This line is nice for the about file but could probably be omitted since there is the text above and the example below in the introduction already.
```suggestion
```

concepts/type-conversion/about.md
```diff
@@ -1,3 +1,43 @@
-# About
+# About Type Conversion
 
-TODO: add information on type-conversion concept
+In Go, assignment of a value between different types requires explicit conversion.
```
junedev, line 3: I find this sentence confusing because type conversion is not only needed in the context of an assignment but also for comparison for example, or when providing an argument for a function.
```go
var i int8 = 12
var j int32 = 13
fmt.Println(i == j)
// invalid operation: i == j (mismatched types int8 and int32)
```
Here a possible alternative:
```suggestion
Go does not implicitly convert values from one type to another. If you want to use a value of one type in a context where another type is required, you need to explicitly convert the value.
```
(Needs to be fixed in the all the places if you decide to change something here.

concepts/type-conversion/about.md
```diff
@@ -1,3 +1,43 @@
...
+For example, this is a compiler error:
```
junedev, line 4:
```suggestion
For example, this causes a compiler error:
```

concepts/type-conversion/about.md
```diff
@@ -1,3 +1,43 @@
...
+
```
junedev, line 5: General comment: you might want to mention "type casting" as a term somewhere, e.g. like "this is also called ..." so the student knows what that means even if there previous language did not use the term.

junedev (approved):
Another look? I didn't even take a first look, I just answered your question. :wink: 
Well, anyway. Here is my review now. All comments are minor/optional so pick what you like.

junedev:
You should add yourself as author of the concept in the config of the concept (Erik can be removed, he is just a "placeholder").

jmrunkle:
> You should add yourself as author of the concept in the config of the concept (Erik can be removed, he is just a "placeholder").

I believe that I did that already: https://github.com/exercism/go/blob/678740210a17fc908b193d7abdd0a1b0775e3c0a/concepts/type-conversion/.meta/config.json

junedev:
> believe that I did that already
Sorry, my local branch was not up to date.

Issue #1704, opened 2021-09-27
`sorting-room`: teach about type-switch syntax


jmrunkle:
Teach about this syntax:

```go
switch v := i.(type) {
case T:
    // here v has type T
case S:
    // here v has type S
default:
    // no match; here v has the same type as i
}
```

junedev:
@jmrunkle If you want you can also include this as part of the type assertion concept. Many tracks said that things are a bit easier if the concepts are not cut too small.

jmrunkle:
> If you want you can also include this as part of the type assertion concept. Many tracks said that things are a bit easier if the concepts are not cut too small.

Sounds good. I will do that since it simplifies things.

jmrunkle:
This is done between the exercise itself and the `type-assertion` concept

PR #1706, opened 2021-09-27
Fill in documentation for `type-assertions`

### LINKED ISSUE #1703: `sorting-room`: fill out `type-assertion` concept documentation
Everything inside concepts/type-assertion is basically empty right now.

jmrunkle:
Fixes #1703
Partially addresses #1704

### DIFF
--- a/concepts/type-assertion/about.md
+++ b/concepts/type-assertion/about.md
@@ -13,11 +13,11 @@ var input interface{} = 12
 str := input.(string) // panic at runtime since input is not a string!
 ```
 
-We can test whether an interface value holds a specific concrete type by returning two values: the underlying value and a boolean value that reports whether the assertion succeeded.
+We can test whether an interface value holds a specific concrete type by making use of both return values of the type assertion: the underlying value and a boolean value that reports whether the assertion succeeded.
 For example:
 
 ```go
-str, ok := input.(string) // no panic if input is a string
+str, ok := input.(string) // no panic if input is not a string
 ```
 
 If `input` holds a `string`, then `str` will be the underlying value and `ok` will be true.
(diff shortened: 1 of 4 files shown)


concepts/type-assertion/about.md
```diff
@@ -1,3 +1,34 @@
-# About
+# About Type Assertions
```
junedev, line 1: You wanted to also include the type switch in this concept if I understood correctly.

concepts/type-assertion/about.md
```diff
@@ -1,3 +1,34 @@
...
+If we assign the result of this statement to a single value, we assert that the interface variable holds a value of the concrete type.
+As a result, it may panic at runtime if the interface value does not have the specified concrete type.
+For example:
+
+```go
+var input interface{} = 12
+str := input.(string) // panic at runtime since input is not a string!
+```
+
+We can test whether an interface value holds a specific concrete type by returning two values: the underlying value and a boolean value that reports whether the assertion succeeded.
```
junedev, line 16:
```suggestion
We can test whether an interface value holds a specific concrete type by making use of both return values of the type assertion: the underlying value and a boolean value that reports whether the assertion succeeded.
```

concepts/type-assertion/about.md
```diff
@@ -1,3 +1,34 @@
...
+For example:
+
+```go
+str, ok := input.(string) // no panic if input is a string
```
junedev, line 20: I think you meant this:
```suggestion
str, ok := input.(string) // no panic if input is not a string
```

jmrunkle:
Thanks @junedev - addressed your initial round of comments there. Not sure if we can be any more succinct than that about the type-switch without removing it from the intro completely, but I am open to suggestions.

junedev:
I think it's fine as is. Can always be changed later someone feels the need for that.

PR #1711, opened 2021-09-27
Cleanup the introduction for the booleans concept

bobtfish:
Should just be about booleans, not about conditionals, switch statements etc).

### DIFF
--- a/concepts/booleans/introduction.md
+++ b/concepts/booleans/introduction.md
@@ -1,49 +1,18 @@
 # Introduction
 
-Go supports the three logical operators `&&` (AND), `||` (OR), and `!` (NOT).
+Booleans in Go are represented by the `bool` type, which values can be either `true` or `false`.
 
-Conditionals in Go are similar to conditionals in other languages. The underlying type of any conditional operation is the `bool` type, which can have the value of `true` or `false`. Conditionals are often used as flow control mechanisms to check for various conditions. For checking a particular case an `if` statement can be used, which executes its code if the underlying condition is `true` like this:
+Go supports three boolean operators: `!` (NOT), `&&` (AND), and `||` (OR).
 
 ```go
-var value string
-
-if value == "val" {
-    // conditional code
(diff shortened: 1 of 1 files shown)


concepts/booleans/introduction.md
```diff
@@ -1,49 +1,18 @@
 # Introduction
 
-Go supports the three logical operators `&&` (AND), `||` (OR), and `!` (NOT).
+Booleans in Go are represented by the `bool` type, which values can be either `true` or `false`.
```
jmrunkle, line 3:
```suggestion
Booleans in Go are represented by the `bool` type.
A `bool` is either `true` or `false`.
```

Split the sentence since there are two distinct statements rather than joining them with a comma.

concepts/booleans/introduction.md
```diff
@@ -1,49 +1,18 @@
...
-if value == "val" {
-    // conditional code
-}
+true || false // Output: true
+true && false // Output: false
+!true // Output: false
 ```
 
-In scenarios involving more than one case many `if` statements can be chained together using the `else if` and `else` statements.
+The three boolean operators each have a different _operator precedence_. As a consequence, they are evaluated in this order: `!` first, `&&` second, and finally `||`. If you want to 'escape' these rules, you can enclose a boolean expression in parentheses (`()`), as the parentheses have an even higher operator precedence.
```
jmrunkle, line 13:
```suggestion
The three boolean operators each have a different _operator precedence_.
As a consequence, they are evaluated in this order: `!` first, `&&` second, and finally `||`.
If you want to force a different ordering, you can enclose a boolean expression in parentheses (ie. `()`), as the parentheses have an even higher operator precedence.
```

Use one sentence per line and a couple other trivial changes.

jmrunkle (changes requested):
Just a couple small nitpicks, thanks!

jmrunkle (approved):
LGTM

PR #1712, opened 2021-09-27
Fix slices concept about to be superset of introduction.

bobtfish:
There's a lot of good info in the about page for the concept, but it didn't cover any of the information from the introduction.
This fixes that by copying the introduciton into the about, making it a superset rather than a completely different doc

### DIFF
--- a/concepts/slices/about.md
+++ b/concepts/slices/about.md
@@ -3,7 +3,9 @@
 Slices in Go are similar to lists or arrays in other languages.
 They hold a number of elements of a specific type (or interface).
 
-Slices in Go are based on arrays. Arrays have a fixed size. A slice, on the other hand, is a dynamically-sized, flexible view into the elements of an array.
+Slices in Go are based on arrays.
+Arrays have a fixed size.
+A slice, on the other hand, is a dynamically-sized, flexible view into the elements of an array.
 
 A slice is written like `[]T` with `T` being the type of the elements in the slice:
 
--- a/concepts/slices/about.md
+++ b/concepts/slices/about.md
@@ -14,7 +14,7 @@ var empty []int                 // an empty slice
 withData := []int{0,1,2,3,4,5}  // a slice pre-filled with some data
 ```
 
-You can get/set an element at a given zero-based index using square-bracket notation:
+You can get or set an element at a given zero-based index using square-bracket notation:
 
 ```go
 withData[1] = 5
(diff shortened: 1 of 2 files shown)


concepts/slices/about.md
```diff
@@ -1,5 +1,31 @@
 # About
 
+Slices in Go are similar to lists or arrays in other languages. They hold a number of elements of a specific type (or interface).
```
jmrunkle, line 3:
```suggestion
Slices in Go are similar to lists or arrays in other languages.
They hold a number of elements of a specific type (or interface).
```

Ideally one sentence per line.

concepts/slices/about.md
```diff
@@ -1,5 +1,31 @@
...
+
+Slices in Go are based on arrays. Arrays have a fixed size. A slice, on the other hand, is a dynamically-sized, flexible view into the elements of an array.
```
jmrunkle, line 5:
```suggestion
Slices in Go are based on arrays.
Arrays have a fixed size.
A slice, on the other hand, is a dynamically-sized, flexible view into the elements of an array.
```

concepts/slices/about.md
```diff
@@ -1,5 +1,31 @@
...
+
+A slice is written like `[]T` with `T` being the type of the elements in the slice:
+
+```go
+var empty []int                 // an empty slice
+withData := []int{0,1,2,3,4,5}  // a slice pre-filled with some data
+```
+
+You can get/set an element at a given zero-based index using square-bracket notation:
```
jmrunkle, line 14:
```suggestion
You can get or set an element at a given zero-based index using square-bracket notation:
```

I think we should prefer being more explicit about this rather than implying they are interchangeable or something like that.

concepts/slices/about.md
```diff
@@ -1,5 +1,31 @@
...
+```
+
+You can create a new slice from an existing slice by getting a range of elements, once again using square-bracket notation, but specifying both a starting (inclusive) and ending (exclusive) index.
+If you don't specify a starting index, it defaults to 0.
+If you don't specify an ending index, it defaults to the length of the slice.
+
+```go
+newSlice := withData[2:4] // newSlice == []int{2,3}
+```
+
```
jmrunkle, line 28: We should probably mention what happens if start (eg. `[:2]`) or end (eg. `[2:]`) is missing. And perhaps even what happens if **both** are missing (eg. `[:]`).

jmrunkle (changes requested):
Mostly good, just a few nitpicks.

Issue #1643, opened 2021-09-17
[Sept Sprint] New exercise teaching pointers


ErikSchierboom:
This is part of our September Spring to get the Go Track in shape. Please read the [Overview issue](https://github.com/exercism/go/issues/1610) for more details. **This exercise is unclaimed. To claim it, please comment below.**

---

**Before building this exercise (or exercises), please read up on [the documentation](https://exercism.org/docs/building/tracks/concept-exercises).**

We need an exercise for the `pointers` and `nil` concepts. These concepts go well together so it makes sense to teach them in the same exercise.

The exercise should teach:

- What pointers are
- Pointer types for primitives
- Which types are already pointers
- Explain that go has no pointer arithmetic
- Referencing and dereferencing syntax
- `nil` as zero value for the pointer
- pitfalls when working with pointers (always check whether they are nil before dereferencing)

Keep in mind that the concept of pointers might be hard for people coming from languages that don't have explicit pointer types.

junedev:
We have noticed in the exercise teaching `zero-values` that it might not make sense to create a separate concept for `nil`. You can also just include the info in the `pointers` concept instead and have this exercise just teach `pointers`.

ErikSchierboom:
> have this exercise just teach `pointers`.

Sounds like that might be better actually, seeing how people often struggle with pointers.

andrerfcsantos:
Just to be clear, this actually requires the implementation of a concept + exercise, right?

If so, what would be the requirements for the concept?

junedev:
Yes, exactly. The introduction file of the exercise should cover the list that you find in the issue description. Then you can use that same content as introduction file of the concept. You can also copy the content into about.md of the concept and then think about whether you would like to add more detailed content/examples there etc. Also if some item on the issue description list is not needed in the exercise and not needed to understand the concept in general, then it should be moved to that about.md file of the concept.

Besides introduction and about there is not much needed for a concept, you can find the general documentation on Concepts here https://exercism.org/docs/building/tracks/concepts. 

TLDR If you focus on the exercise and its introduction, you have all the minimal things you need for the concept as well.

ErikSchierboom:
@andrerfcsantos I'm in the process of adding a new exercise, so you can check my PR to give you an idea of what it looks like: https://github.com/exercism/go/pull/1680 Note that it's perfectly fine to create the exercise in a series of smaller, incremental PRs if you'd like that.

andrerfcsantos:
[...]

PR #1720, opened 2021-09-27
Functions concept

### LINKED ISSUE #1642: [Sept Sprint] New exercise teaching functions
This is part of our September Spring to get the Go Track in shape. Please read the [Overview issue](https://github.com/exercism/go/issues/1610) for more details. **This exercise is unclaimed. To claim it, please comment below.**

---

**Before building this exercise (or exercises), please read up on [the documentation](https://exercism.org/docs/building/tracks/concept-exercises).**

We need an exercise for the `functions` concept, which should teach about:

- Arguments (incl. syntax for arguments of the same type)
- Return values
- Named return values
- Naked return
- pass by value but the value can be a pointer
- Can also include: Anonymous functions/callbacks

It should **not** teach (these will be covered later):

- Variadic functions (that should be introduced with rest-and-spread)

It is worth to look if it is possible to port this exercise from the JS [lasagna master exercise](https://github.com/exercism/javascript/tree/main/exercises/concept/lasagna-master/.docs).

bobtfish:
This is my first attempt at something as ambitious as a full exercise and a concept, so any/all feedback and tips very gratefully received.

Specific things that I'm unsure about / have questions on:
  * Fitting this concept/exercise into the existing hierarchy - I've done this somewhat, but if I add dependencies for everything that's relevant then I create cycles. Feedback welcomes from someone who has a more holistic view of the progression of concepts they think is right for the Go track.
  * The concept introduction and exercise introduction are nearly identical. Is this OK/appropriate? (Should I just be using .docs/introduction.md.tpl if so?)
  * .meta/design.md - this is my first attempt at one of these, I've read a few others but it's largely cargo cult.
  * Should my exercise include `forked_from`? It's not clear to me from the docs what the contents of this field would be? (I.e. is it just same track exercises, or can you reference cross tracks etc)

Fixes #1642

### DIFF
--- a/concepts/functions/about.md
+++ b/concepts/functions/about.md
@@ -34,7 +34,7 @@ func Hello(name string) string {
   return "Hello " + name
 }
 
-func HelloAndGoodbye(name string) (string. string) {
+func HelloAndGoodbye(name string) (string, string) {
   return "Hello " + name, "Goodbye " + name
 }
 ```
--- a/concepts/functions/introduction.md
+++ b/concepts/functions/introduction.md
@@ -1,46 +1,83 @@
 # Introduction
 
-TODO: the content below is copied from the exercise introduction and probably needs rewriting to a proper concept introduction
-
-## packages
-
-## functions
-
-## variables
-
-Go is a statically-typed language, which means that everything has a type at compile-time. Assigning a value to a name is referred to as defining a variable. A variable can be defined either by explicitly specifying its type, or by assigning a value to have the Go compiler infer its type based on the assigned value.
+A function allows you to group code into a reusable unit.
+It consists of the `func` keyword, the name of the function, and a comma-separated list of zero or more parameters and types in round brackets.
+All parameters must be explicitly typed, there is no type inference for parameters.
+There are no default values for parameters, all function parameters are required.
 
 ```go
-var explicit int // Explicitly typed
-implicit := 10   // Implicitly typed
+import "fmt"
+
+// No parameters
+func PrintHello() {
+    fmt.Println("Hello")
+}
+
+// One parameter
+func PrintHelloName(name string) {
+  fmt.Println("Hello " + name)
+}
 ```
 
-The value of a variable can be assigned using the `:=` and updated using the `=`. Once defined, a variable's type can never change.
+## Return Values
+
+The function arguments are followed by zero or more return values which must also be explicitly typed.
+Single return values are left bare, multiple return values are wrapped in parenthesis.
+Values are returned to the calling code from functions using the `return` keyword.
+There can be multiple `return` statements in a function.
+The execution of the function ends as soon as it hits one of those `return`s.
+If multiple values are to be returned from a function, they are comma seperated.
 
 ```go
-count := 1 // Assign initial value
-count = 2  // Update to new value
-
-// Compiler error when assigning different type
-// count = false
-```
-
-A function can have zero or more parameters. All parameters must be explicitly typed, there is no type inference for parameters. A function can also have multiple return values which must also be explicitly typed. Values are returned from functions using the `return` keyword. To allow a function to be called by code in other packages, the name of the function must start with a capital letter.
-
-```go
-package greeting
-
-// Hello is a public function
 func Hello(name string) string {
-    return hello(name)
+  return "Hello " + name
 }
 
-// hello is a private function
-func hello(name string) string {
-    return "Hello " + name
+func HelloAndGoodbye(name string) (string. string) {
+  return "Hello " + name, "Goodbye " + name
 }
 ```
 
-Invoking a function is done by specifying the function name and passing arguments for each of the function's parameters.
+## Invoking Functions
 
-Go supports two types of comments. Single line comments are preceded by `//` and multiline comments are inserted between `/*` and `*/`.
+Invoking a function is done by specifying the function name and passing arguments for each of the function's parameters in parenthesis.
+
+```go
+import "fmt"
+
+// No parameters, no return value
+func PrintHello() {
+    fmt.Println("Hello")
+}
+// Called like this:
+PrintHello()
+
+// One parameter, one return value
+func Hello(name string) string {
+  return "Hello " + name
+}
+// Called like this:
+greeting := Hello("Dave")
+
+// Multiple parameters, multiple return values
+func SumAndMultiply(a, b int) (int, int) {
+    return a+b, a*b
+}
+// Called like this:
+aplusb, atimesb := SumAndMultiply(a, b)
+```
+
+## Named Return Values and Naked Return
+
+As well as parameters, return values can optionally be named.
+If named return values are used, a `return` statement without arguments will return those values, this is known as a 'naked' return.
+
+```go
+func SumAndMultiplyThenMinus(a, b, c int) (sum, mult int) {
+    sum, mult = a+b, a*b
+    sum -= c
+    mult -= c
+    return
+}
+```
+
--- a/concepts/functions/introduction.md
+++ b/concepts/functions/introduction.md
@@ -33,7 +33,7 @@ func Hello(name string) string {
   return "Hello " + name
 }
 
-func HelloAndGoodbye(name string) (string. string) {
+func HelloAndGoodbye(name string) (string, string) {
   return "Hello " + name, "Goodbye " + name
 }
 ```
--- a/exercises/concept/lasagna-master/.docs/instructions.md
+++ b/exercises/concept/lasagna-master/.docs/instructions.md
@@ -1,0 +1,80 @@
+# Instructions
+
+In this exercise you are going to write some more code related to preparing and cooking your brilliant lasagna from your favorite cookbook.
+
+You have four tasks.
+The first one is related to the cooking itself, the other three are about the perfect preparation.
+
+## 1. Estimate the preparation time
+
+For the next lasagna that you will prepare, you want to make sure you have enough time reserved so you can enjoy the cooking.
+You already made a plan which layers your lasagna will have.
+Now you want to estimate how long the preparation will take based on that.
+
+Implement a function `preparationTime` that accepts an array of layers as a `[]string` and the average preparation time per layer in minutes as an `int`.
+The function should return the estimate for the total preparation time based on the number of layers as an `int`.
+Go has no default values for functions.
+If the average preperation time is passed as `0` (the default initial value for an `int`), then the default value of `2` should be used.
+
+```go
+layers := []string{"sauce", "noodles", "sauce", "meat", "mozzarella", "noodles"}
+preparationTime(layers, 3)
+// => 18
+preparationTime(layers, 0)
+// => 12
+```
+
+## 2. Compute the amounts of noodles and sauce needed
+
+Besides reserving the time, you also want to make sure you have enough sauce and noodles to cook the lasagna of your dreams.
+For each noodle layer in your lasagna, you will need 50 grams of noodles.
+For each sauce layer in your lasagna, you will need 0.2 liters of sauce.
+
+Define the function `quantities` that takes an array of layers as parameter as a `[]string`.
+The function will then determine the quantity of noodles and sauce needed to make your meal.
+The result should be returned as two values of `noodles` as an `int` and `sauce` as a `float64`.
+
+```go
+quantities([]string{"sauce", "noodles", "sauce", "meat", "mozzarella", "noodles"})
+// => 100, 0.4
+```
+
+## 3. Add the secret ingredient
+
+A while ago you visited a friend and ate lasagna there.
+It was amazing and had something special to it.
+The friend sent you the list of ingredients and told you the last item on the list is the "secret ingredient" that made the meal so special.
+Now you want to add that secret ingredient to your recipe as well.
+
+Write a function `addSecretIngredient` that accepts two arrays of ingredients of type `[]string` as parameters.
+The first parameter is the list your friend sent you, the second is the ingredient list for your own recipe.
+The function should generate a new slice and add the last item from your friends list to the end of your list.
+Neither argument should not be modified.
+
+```go
+friendsList := []string{"noodles", "sauce", "mozzarella", "kampot pepper"}
+myList := []string{"noodles", "meat", "sauce", "mozzarella"}
+
+addSecretIngredient(friendsList, myList)
+// => []string{"noodles", "meat", "sauce", "mozzarella", "kampot pepper"}
+```
+
+## 4. Scale the recipe
+
+The amounts listed in your cookbook only yield enough lasagna for two portions.
+Since you want to cook for more people next time, you want to calculate the amounts for different numbers of portions.
+
+Implement a function `ScaleRecipe` that takes two parameters.
+
+- A slice of float64 amounts needed for 2 portions.
+- The number of portions you want to cook.
+
+The function should return a slice of float64s with the amounts needed for the desired number of portions.
+You want to keep the original recipe though.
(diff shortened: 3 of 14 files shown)


exercises/concept/lasagna-master/.meta/config.json
```diff
@@ -0,0 +1,12 @@
...
+  "blurb": "Dive deeper into Go functions while preparing to cook the perfect lasagna.",
+  "authors": ["bobtfish"],
+  "contributors": [],
+  "files": {
+    "solution": ["lasagna-master.go"],
+    "test": ["lasagna-master_test.go"],
+    "exemplar": [".meta/exemplar.go"]
+  },
+  "source": "Inspired by the Lasagna Master exercise in the Javascript track",
+  "source_url": "https://github.com/exercism/javascript/blob/main/exercises/concept/lasagna-master/.docs/instructions.md"
```
ErikSchierboom, line 11: You don't need the source if you specify the forked_from field (which is appropriate here)

```suggestion
  "forked_from": ["javascript/lasagna-master"]
```

exercises/concept/lasagna-master/.meta/exemplar.go
```diff
@@ -0,0 +1,35 @@
+package lasagna
+
+// Determines whether the lasagna is done based on the
+func cookingStatus(time int) string {
```
ErikSchierboom, line 4: Should these functions start with a capital letter? I've seen that used in other exercises.

exercises/concept/lasagna-master/lasagna-master.go
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * Implement the functions needed to solve the exercise here.
+ * Here an example of the syntax as reminder:
+ *
+ * function yourFunction(param string, otherParam int) string {
+ *   ...
+ *   return "foo"
+ * }
+ */
```
ErikSchierboom, line 11: I'd recommend doing the same here as in the [lasagna-exercise](https://github.com/exercism/go/blob/main/exercises/concept/lasagna/lasagna.go#L5), where there are TODOs for individual functions.

```suggestion
// TODO: define the 'CookingStatus()' function
```

concepts/functions/about.md
```diff
@@ -1,40 +1,167 @@
...
+    *x = *y
+    *y = temp
+}
+
+var a int = 100
+var b int = 200
+swap(&a, &b)
+// a = 200, b = 100
+```
+
```
ErikSchierboom, line 116: There will be a separate exercise on pointers: https://github.com/exercism/go/issues/1643

```suggestion
```

concepts/functions/about.md
```diff
@@ -1,40 +1,167 @@
...
-A function can have zero or more parameters. All parameters must be explicitly typed, there is no type inference for parameters. A function can also have multiple return values which must also be explicitly typed. Values are returned from functions using the [`return` keyword][return]. To allow a function, or a method to be [called by code in other packages][public-vs-private], the name of the function must start with a capital letter.
+## Return Values
+
+The function arguments are followed by zero or more return values which must also be explicitly typed.
+Single return values are left bare, multiple return values are wrapped in parenthesis.
+Values are returned to the calling code from functions using the [`return` keyword][return].
+There can be multiple `return` statements in a function.
+The execution of the function ends as soon as it hits one of those `return`s.
+If multiple values are to be returned from a function, they are comma seperated.
+More information about idiomatic use of [multiple return values][concept-multiple-return-values] can be found in the linked concept.
```
ErikSchierboom, line 31: Nicely explained!

concepts/functions/about.md
```diff
@@ -1,40 +1,167 @@
 # About
 
-Functions in Go are considered [first class citizens][first-class-functions] making them very powerful. They can be used as:
+A function allows you to group code into a reusable unit.
+It consists of the `func` keyword, the name of the function, and a comma-separated list of zero or more parameters and types in round brackets.
+Other possibilities like [methods][concept-methods] and anonymous functions will be covered in other concepts.
```
ErikSchierboom, line 5: I'm not a fan of referring to other concepts like this upfront. Maybe change these to be [concept widget links](https://exercism.org/docs/building/markdown/widgets) and move this text somewhat later in this document?

concepts/functions/about.md
```diff
@@ -1,40 +1,167 @@
...
+```
 
-Invoking a function from another package is done by specifying the package name as well as the function name.
+A similar syntax can be used to unpack a slice into a function call:
 
 ```go
-phrase := greeting.Hello("John")
+numbers := []int{1, 2, 4}
+result := sum(...numbers)
+// result => 7
```
ErikSchierboom, line 156: Variadic functions are out of scope for this exercise (see the [issue description](https://github.com/exercism/go/issues/1642#issue-999186434)). Could you remove this section?

concepts/functions/about.md
```diff
@@ -1,40 +1,167 @@
...
+
+var a int = 100
+var b int = 200
+swap(&a, &b)
+// a = 200, b = 100
+```
+
+## Public and Private Functions
+
+To allow a function to be called by [code in other packages][public-vs-private], the name of the function must start with a capital letter.
```
ErikSchierboom, line 119: @junedev Do we want to include public/private here?

concepts/functions/introduction.md
```diff
@@ -1,46 +1,113 @@
...
+    *x = *y
+    *y = temp
+}
 
-Go supports two types of comments. Single line comments are preceded by `//` and multiline comments are inserted between `/*` and `*/`.
+var a int = 100
+var b int = 200
+swap(&a, &b)
+// a = 200, b = 100
+```
```
ErikSchierboom, line 113: This section can be removed as this exercise should not cover pointers (nor is it needed by the student to solve the exercise).

concepts/functions/introduction.md
```diff
@@ -1,46 +1,113 @@
...
-// Hello is a public function
-func Hello(name string) string {
-    return hello(name)
+```go
+func SumAndMultiplyThenMinus(a, b, c int) (sum, mult int) {
+    sum, mult = a+b, a*b
+    sum -= c
+    mult -= c
+    return
 }
```
ErikSchierboom, line 81: This example uses multiple returns, which is something that is introduced in the `card-tricks` exercise. The dependencies here are tricky.

ErikSchierboom (commented):
Thank you so much for doing this! I've roughly reviewed this, but decided to hold off on some reviewing more to first solve the dependency cycles.

> Fitting this concept/exercise into the existing hierarchy - I've done this somewhat, but if I add dependencies for everything that's relevant then I create cycles. Feedback welcomes from someone who has a more holistic view of the progression of concepts they think is right for the Go track.

@junedev could you perhaps take a look at how you envision the concept hierarchy? One immediate issue I could see was that this exercise depends on slices, but that the exercise teaching slices also teaches multiple returns. We could consider moving the multiple returns to this exercise? It's also important to note that we should decide if we want this to be an early exercise or not (as in: which exercises should depend on this)?

> The concept introduction and exercise introduction are nearly identical. Is this OK/appropriate? (Should I just be using .docs/introduction.md.tpl if so?)

This is perfectly ok. At some point the .tpl files will get good support, but we have only basic support for it now so I'd ignore it.

> .meta/design.md - this is my first attempt at one of these, I've read a few others but it's largely cargo cult.

It's perfectly fine.

> Should my exercise include forked_from? It's not clear to me from the docs what the contents of this field would be? (I.e. is it just same track exercises, or can you reference cross tracks etc)

Forked from is exclusively for cross-track references. So in this case it should be:

`"forked_from": ["javascript/lasagna-master"]

concepts/functions/introduction.md
bobtfish: Damn, I knew I'd end up leaving something extraneous in the introduction, thanks for spotting!

concepts/functions/introduction.md
bobtfish: The other option is to return a `struct` type - the `structs` concept only depends on `numbers` and `conditonals-if` so this doesn't seem unreasonable...

bobtfish:
> Thank you so much for doing this! I've roughly reviewed this, but decided to hold off on some reviewing more to first solve the dependency cycles.

Sounds good - thanks for the rough cut that was really helpful.

Agree that we should sort the high level dependency issues and decide where this fits and what it can/can't depend on before we get into the minutiae.

concepts/functions/about.md
junedev: No, that's not necessary. It is taught in the packages concept.

junedev:
Disclaimer: I did not read the PR content yet, I am mostly responding to the question regarding the concept tree.

The functions concept should include explaining multiple return values, not as a separate concept just as part of the content. This is where it belongs. Content can probably be copied. Thus multiple return values can be safely used as part of the exercise here.

This exercise should have slices and structs as prerequisites. When you modify the scaleRecipe exercise to use a slice of structs instead of a map that should be enough to make it work. (Make sure not to use `range`, it not available yet. Just use a normal for loop.)

We can then later go ahead and remove the multiple return values concept from slices. The exercise would just need a minimal change, there is no conceptual dependency (unlike for maps, there you need multiple return values). Until then I don't think anyone will think this is bad track because we repeat a couple of lines of concept content. Everything that depended on multiple return values will then depend on functions later.

The general hierarchy on the end would be that the student learns the basic types and conditionals, that unlocks slices and structs, which unlocks functions which unlocks maps.

With regards to pointers I would go with the "learn the details later approach". Strictly speaking you would need pointers to understand how the complex  types are passed to the function. However currently pointers has maps as prerequisite and I would like to keep it a bit further down the tree. So instead of going into the details here I would just talk about how the passing works for "reference types" and primitives.

Does that help/ make sense?

junedev:
I had a quick look at the exemplar file:
* I dont think the first task is needed here, it was about "not passing an argument" which does not exist in Go
* The second task should be changed to practice Gos version of a default parameter (an if statement). If the second argument is 0, use 2 as default.

concepts/functions/about.md
ErikSchierboom: @bobtfish Could you remove the public/private section here (and in any other places in this PR)?

ErikSchierboom:
@bobtfish Does @junedev's explanation help sort the dependencies issue?

concepts/functions/about.md
bobtfish: This is not currently taught or mentioned anywhere in the packages concept or exercise that I can see, unless I'm missing it.

Does adding this need a ticket?

concepts/functions/introduction.md
bobtfish: Per @junedev we're OK with multiple returns here, and will eventually remove them from `card-tricks` - resolving this.

bobtfish:
> @bobtfish Does @junedev's explanation help sort the dependencies issue?

Yes! Thanks @junedev - I'll work on getting this into shape in the next day or two and let you know when it's ready for review.

junedev:
Yes please, packages should teach how to import AND export things.
-------- Ursprüngliche Nachricht --------Von: Tomas Doran ***@***.***> Datum: 30.09.21  11:15  (GMT+01:00) An: exercism/go ***@***.***> Cc: June ***@***.***>, Mention ***@***.***> Betreff: Re: [exercism/go] Functions concept (#1720) 
@bobtfish commented on this pull request.



In concepts/functions/about.md:
> +## Public and Private Functions
+
+To allow a function to be called by [code in other packages][public-vs-private], the name of the function must start with a capital letter.

This is not currently taught or mentioned anywhere in the packages concept or exercise that I can see, unless I'm missing it.
Does adding this need a ticket?

concepts/functions/about.md
junedev: Yes please, packages should teach how to import AND export things.

ErikSchierboom:
@junedev I've opened an issue: https://github.com/exercism/go/issues/1728

junedev:
@bobtfish Another thought regarding scaleRecipe. Instead of making it a slice of structs you could also just have a slice of floats an say "each item represents the amount of some ingredient ...". The task is about not modifying the argument so a slice will do to make this point.

junedev:
@bobtfish Setting this to draft for now. Mark as "ready for review" again whenever you are ready.

bobtfish:
> Make sure not to use `range`, it not available yet. Just use a normal for loop.

I'm unsure about this - it's totally possible to do this, but the exercise is a lot nicer to solve if you can use `range`, and there isn't anything in the dependencies at the moment which stops this from depending on `range-iteration`.

Why do we imagine that this has to be before we cover `range-iteration`?

junedev:
> Why do we imagine that this has to be before we cover range-iteration?

Range iteration need to show how to iterate over maps so maps need to be known to the student, maps need multiple return values from here so we can't move maps before functions.

bobtfish:
> > Why do we imagine that this has to be before we cover range-iteration?
> 
> Range iteration need to show how to iterate over maps so maps need to be known to the student, maps need multiple return values from here so we can't move maps before functions.

Aha, got it, thanks for the explanation. I'll take `range-iteration` out later tonight.

exercises/concept/lasagna-master/.docs/instructions.md
```diff
@@ -0,0 +1,80 @@
...
+## 1. Estimate the preparation time
+
+For the next lasagna that you will prepare, you want to make sure you have enough time reserved so you can enjoy the cooking.
+You already made a plan which layers your lasagna will have.
+Now you want to estimate how long the preparation will take based on that.
+
+Implement a function `preparationTime` that accepts an array of layers as a `[]string` and the average preparation time per layer in minutes as an `int`.
+The function should return the estimate for the total preparation time based on the number of layers as an `int`.
+Go has no default values for functions.
+If the average preperation time is passed as `0` (the default value for an `int`), then the default value of `2` should be used.
```
junedev, line 17:
```suggestion
If the average preperation time is passed as `0` (the default initial value for an `int`), then the default value of `2` should be used.
```

exercises/concept/lasagna-master/.meta/design.md
```diff
@@ -0,0 +1,44 @@
...
+
+## Prerequisites
+
+- `strings` are needed in the exercise
+- `numbers` are needed in the exercise
+- `comparison` is needed in the exercise
+- `conditionals-if` is needed in the exercise
+- `range-iteration` is needed in the exercise
+- `slices` are needed in the exercise
+- `structs` are needed in the exercise
```
junedev, line 37: I think you removed this requirement.
```suggestion
```

exercises/concept/lasagna-master/.meta/exemplar.go
```diff
@@ -0,0 +1,39 @@
+package lasagna
+
+// Estimates the preparation time based on the number of layers.
```
junedev, line 3: All function comments need to conform to the Go spec (see concept comments).

junedev (commented):
Well done overall! I really like the examples you used in the concepts.

Some more small things need to be fixed before this can be merged.

bobtfish:
> Well done overall! I really like the examples you used in the concepts.

Thanks!

> Some more small things need to be fixed before this can be merged and the merge conflict needs to be resolved.

All fixed up. I wasn't 100% sure what you were suggesting around the `PrintHelloName` example, but I have had a go at amending it.

I've also rebased and squashed as it was getting unwieldy - no point committing adding then removing then re-adding the same thing , and that also allowed me to fix the merge conflict so it'll merge cleanly without a load of back-and-forth merges to make future history exploration easy. (I'm not sure if there's a house style/recommendation about merges/rebases for PRs in progress)

junedev (approved):

PR #1722, opened 2021-09-27
card-tricks: Allow nil empty slices when an empty slice is expected

### LINKED ISSUE #1721: card-tricks: tests are unnecessarily restrictive for nil empty slices
In the `card-tricks` exercise, students are expected to create a slice of a given size, filled with the same value in all positions. A solution that can accomplish this and it is idiomatic is this:

```go
func PrefilledSlice(value, length int) []int {
    var s []int

    for i := 0; i < length; i++ {
        s = append(s, value)
    }

    return s
}
```

However this solution fails the current tests, and the reason is a bit subtle. Some tests pass 0 as the length in the second argument. With this solution and with a length=0, this function will return a **nil** slice with length of 0. But the tests are expecting actually a **non-nil** slice of length 0. If the student actually makes this solution, the error they get is also criptic:

```
=== RUN   TestPrefilledSlice/Length_zero
    card_tricks_test.go:191: PrefilledSlice(value:3, length:0) = [], want []
=== RUN   TestPrefilledSlice/Negative_length
    card_tricks_test.go:191: PrefilledSlice(value:3, length:-3) = [], want []
--- FAIL: TestPrefilledSlice (0.01s)
```

The tests say that the function returned `[]`, but the tests wanted `[]`, which is obviously confusing. The reason for this is that a nil slice and a non-nil empty slice both are represented as `[]`.

Right now, the only way to make this test pass is to create an non-nil empty slice explicitly with `make` or with `[]int{}`:

```go
func PrefilledSlice(value, length int) []int {
    s := make([]int, 0, length)

    for i := 0; i < length; i++ {
        s = append(s, value)
    }

    return s
}
```

However, the `var s []int` way of creating a slice should also be allowed. In fact, even [the hints for this task](https://github.com/exercism/go/blob/main/exercises/concept/card-tricks/.docs/hints.md) encourage that:

> There are several ways to create a new slice. The simplest is `var s []int`. More options you get with the `make` function.

The problem is that if the student actually tries to create a slice with `var s []int` the tests may fail, like in the example above.

andrerfcsantos:
Fixes #1721

### DIFF
--- a/exercises/concept/card-tricks/card_tricks_test.go
+++ b/exercises/concept/card-tricks/card_tricks_test.go
@@ -1,7 +1,6 @@
 package cards
 
 import (
-	"reflect"
 	"testing"
 )
 
@@ -134,7 +133,7 @@ func TestSetItem(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if got := SetItem(tt.args.slice, tt.args.index, tt.args.value); !reflect.DeepEqual(got, tt.want) {
+			if got := SetItem(tt.args.slice, tt.args.index, tt.args.value); !slicesEqual(got, tt.want) {
 				t.Errorf("SetItem(slice:%v, index:%v, value:%v) = %v, want %v",
 					tt.args.slice, tt.args.index, tt.args.value, got, tt.want)
 			}
@@ -187,7 +186,7 @@ func TestPrefilledSlice(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if got := PrefilledSlice(tt.args.value, tt.args.length); !reflect.DeepEqual(got, tt.want) {
+			if got := PrefilledSlice(tt.args.value, tt.args.length); !slicesEqual(got, tt.want) {
 				t.Errorf("PrefilledSlice(value:%v, length:%v) = %v, want %v", tt.args.value, tt.args.length, got, tt.want)
 			}
 		})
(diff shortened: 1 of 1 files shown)


junedev (approved):
Not sure everyone agrees but besides that this fixes the issue I like that it does not use "reflect".

exercises/concept/card-tricks/card_tricks_test.go
```diff
@@ -247,13 +246,32 @@ func TestRemoveItem(t *testing.T) {
...
+	}
+
+	size := len(a)
+	for i := 0; i < size; i++ {
+		if a[i] != b[i] {
+			return false
+		}
+	}
+
+	return true
```
jmrunkle, line 272: nit: this is precisely the sort of code that is easy to have off-by-one errors on and would probably need to be tested on its own to be truly confident. Personally, I would trust this more if this part where delegated to reflect.DeepEquals since you already handled the `nil` case before getting here.

jmrunkle (approved):

exercises/concept/card-tricks/card_tricks_test.go
andrerfcsantos: A middle ground between making this code less susceptible to off-by-one errors while still not using the `reflect` package would be to do something like:

```go
for i := range a {
    if a[i] != b[i] {
        return false
    }
}
```

I can do a follow-up PR if you think it's worth it. Personally I also would like to not use the `reflect` package. In Exercism we kind of encourage students to look at the tests (specially in practice exercises), and introducing them to the reflect package via the tests seems overkill. Not only because we are showing them a package they should think twice before using, but also because we can write a simple function ourselves. By not using the reflect package we also avoid all the runtime checks for types that the reflect package brings.

Issue #1645, opened 2021-09-17
[Sept Sprint] New exercise teaching `type-conversion`


ErikSchierboom:
This is part of our September Spring to get the Go Track in shape. Please read the [Overview issue](https://github.com/exercism/go/issues/1610) for more details. **This exercise is claimed by @jmrunkle.**

---

**Before building this exercise (or exercises), please read up on [the documentation](https://exercism.org/docs/building/tracks/concept-exercises).**

We need an exercise for the `type-casting` and `type-assertions` concepts. Since there is sometimes confusion around those two things, it makes sense to teach both in the same exercise to demonstrate how they are related.

If it makes sense, this exercise could also teach about the `type-switch` concept.

The exercise should teach:

- How to convert from one type to another
- How to assert a type
- How to assert an interface type

This exercise will have at least `interfaces` as a prerequisite. 

---

Remaining work:

- [x] #1702
- [x] #1703
- [x] #1704

ErikSchierboom:
See https://github.com/exercism/go/blob/ed630183cccef21791ca21da0ecaafc5b52f9118/exercises/concept/cars-assemble/.docs/introduction.md#type-conversion for the original type conversion introduction text.

jmrunkle:
This is blocked on https://github.com/exercism/go/issues/1638. Until then we cannot add an exercise that requires the `interfaces` concept. We can add it as `status: "wip"`, but configlet will complain if you try to enable the new concept exercise since its prereq is not taught by some exercise.

jmrunkle:
I would be happy to take this one, but I am having trouble coming up with a good story for this one. I looked at the canonical stories list (https://exercism.org/docs/building/tracks/stories), but nothing jumped out at me.

The best I could come up with is you have some kind of sorting room which has to handle anything that is put in there. Some things have predefined results (eg. "This is a number" or "This is a string"), but a few things need to be unpacked and have their contents inspected (eg. "This is a box containing a number 4"). Anything unexpected will be marked "Return to sender".

If no one has a better story I can get rolling with that.

I'll probably start with a PR just to fix the actual concept documentation for `type-conversion` and `type-assertion` which is nothing but TODOs right now. I can also add the `type-switch` concept which seems pretty relevant here.

ErikSchierboom:
I think that story could work fine. Maybe you could make it Santa themed and have the boxes be presents or something like that?

jmrunkle:
[...]

PR #1723, opened 2021-09-28
`sorting-room`: Implement exercise to teach about `type-conversion` and `type-assertion`

jmrunkle:
Part of #1645

This is not complete yet since the introduction should be filled out from details in the concepts so leaving it as "wip", but I wanted to start getting feedback.

### DIFF
--- a/config.json
+++ b/config.json
@@ -233,7 +233,7 @@
           "interfaces",
           "string-formatting"
         ],
-        "status": "active"
+        "status": "wip"
       },
       {
         "name": "Chessboard",
--- a/config.json
+++ b/config.json
@@ -223,6 +223,19 @@
         "status": "beta"
       },
       {
+        "name": "Sorting Room",
+        "slug": "sorting-room",
+        "uuid": "411f4f1e-7ee2-4f11-955d-8af86da69c2a",
+        "concepts": [
+          "type-assertion",
+          "type-conversion"
+        ],
+        "prerequisites": [
+          "interfaces"
+        ],
+        "status": "wip"
+      },
+      {
         "name": "Chessboard",
         "slug": "chessboard",
         "uuid": "b5f1c789-adf6-487c-95b0-c6bce23d711b",
@@ -1817,6 +1830,11 @@
       "uuid": "a960da78-c940-4725-8069-929f42564cdb"
     },
     {
+      "name": "Interfaces",
+      "slug": "interfaces",
+      "uuid": "a492060c-11e7-448e-bdbd-6a907cec9493"
+    },
+    {
       "name": "Iteration",
       "slug": "iteration",
       "uuid": "e18ca160-bbb8-4cd0-882a-0e67f1e60b47"
@@ -1885,6 +1903,11 @@
       "name": "Time",
       "slug": "time",
       "uuid": "4673000c-7822-4252-88c5-723f47d6ad06"
+    },
+    {
+      "name": "Type Assertion",
+      "slug": "type-assertion",
+      "uuid": "601911e6-ecb8-4e37-bc11-395aaa2be3ce"
     },
     {
       "name": "Type Conversion",
--- a/config.json
+++ b/config.json
@@ -131,7 +131,6 @@
         "concepts": [
           "arithmetic-operators",
           "numbers",
-          "type-conversion",
           "conditionals-if",
           "comparison"
         ],
@@ -231,9 +230,10 @@
           "type-conversion"
         ],
         "prerequisites": [
-          "interfaces"
+          "interfaces",
+          "string-formatting"
         ],
-        "status": "wip"
+        "status": "active"
       },
       {
         "name": "Chessboard",
--- a/exercises/concept/sorting-room/.docs/instructions.md
+++ b/exercises/concept/sorting-room/.docs/instructions.md
@@ -1,0 +1,34 @@
+# Instructions
+
+Jen is working in the sorting room in a large factory.
+The sorting room needs to process anything that comes into it by categorizing it with a label.
+She needs a program to help her with the sorting.
+
+Most primitive values should get straight-forward labels.
+For numbers, she wants strings saying `"This is the number 2.0"` (if the number was 2).
+Jen wants the same output for integers and floats.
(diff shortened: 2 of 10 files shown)


junedev:
FYI I merged the concept MRs so their content is now available to be included here.

exercises/concept/sorting-room/.meta/exemplar.go
```diff
@@ -0,0 +1,55 @@
+package sorting
```
junedev, line 1: In case you are open to more general feedback on the exercise itself, here it is:
Of course the exercise works as is but currently the two boolean tasks don't practice anything that was taught in the concepts. Additionally I thought whether it would be possible to modify the exercise so that the student actually needs to use `x, ok := T(y)` and `strconv`.

A possible way to address both things would be to make the exercise only about number sorting and to remove the boolean parts from the existing content. Now that we "made some space" you could add another task about a very special type of number where the student needs to check for this number type and use strconv to extract the number. Something along the lines of this:
```go
type FancyNumber struct {
	n string
}

func (i FancyNumber) Value() string {
	return i.n
}

type FancyNumberBox interface {
	Value() string
}

func ExtractFancyNumber(nb FancyNumberBox) int {
	fancyNum, ok := nb.(FancyNumber)
	if !ok {
		return 0
	}

	num, _ := strconv.Atoi(fancyNum.Value())
	return num
}
```

exercises/concept/sorting-room/.meta/exemplar.go
```diff
@@ -0,0 +1,55 @@
...
+// DescribeBooleanBox should return a string describing the BooleanBox.
+func DescribeBooleanBox(bb BooleanBox) string {
+	if bb.Boolean() {
+		return "This box contains true"
+	}
+	return "This box contains false"
+}
+
+// DescribeAnything should return a string describing whatever it contains.
+func DescribeAnything(i interface{}) string {
```
junedev, line 40: The student does not know about `interface{}` yet since we have no exercise that introduces it. The plan was to combine it with introducing JSON encoding but it might be a while until that exercise will be available. To fix maybe we could add a short bit on this in introduction. E.g. something like "A common use case for type assertion is to find the actual type for values that are specified as empty interface `interface{}` which can potentially hold values of all types."

jmrunkle: As mentioned in #1645:

> This exercise will have at least `interfaces` as a prerequisite.

I would expect the `interfaces` concept / exercise to introduce the empty interface. Is that not a reasonable assumption? Coincidentally, I also have the issue for teaching `interfaces` in `the-farm` so I could certainly make sure it is mentioned there (even though it will not specifically be used there).

junedev: In my head `interfaces` (with actual methods) and the empty interface were separate concepts but of course your version also works.

bitfield:
I dare say you've already thought about this, but it's worth mentioning that [generics](https://bitfieldconsulting.com/golang/generics) will be shipping soon, perhaps as soon as December. We'll need to fit that into our syllabus along with interfaces, won't we? I don't know if you already have an exercise planned for this.

junedev:
@bitfield I haven't looked at the final proposal that made it through yet. Do you see an issue with just adding a separate exercise on generics somewhat further down the concept tree?

config.json
```diff
@@ -222,6 +222,19 @@
         ],
         "status": "beta"
       },
+      {
+        "name": "Sorting Room",
+        "slug": "sorting-room",
+        "uuid": "411f4f1e-7ee2-4f11-955d-8af86da69c2a",
+        "concepts": [
+          "type-assertion",
+          "type-conversion"
```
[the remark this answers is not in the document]
jmrunkle: FYI: It does check, but it ignores exercises marked "wip" so that you can stage changes like this:

> The "exercises.concept[].concepts" values must not be in any other concept exercise's "concepts" property
> https://github.com/exercism/configlet/issues/249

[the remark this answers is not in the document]
junedev: Thanks for clarifying!

exercises/concept/sorting-room/.meta/exemplar.go
jmrunkle: Yep, that sounds good. Went with this option.

jmrunkle:
Ready for another look whenever you are ready @junedev

bitfield:
My blog post summarises roughly what students will need to know. Yes, we need to tackle interfaces before generics—you could think of generics as further loosening the constraints on what kind of things a function can accept. Instead of just being constrained by method sets, for example `io.Reader`, they can now be constrained by type lists (all integer types, for example) or by attributes like `comparable` or `constraints.Ordered`. So I'm sure we could come up with a similar exercise along those lines.

exercises/concept/sorting-room/.docs/instructions.md
```diff
@@ -0,0 +1,78 @@
+# Instructions
+
+Jen is working in the sorting room in a large factory.
+The sorting room needs to process anything that comes into it by categorizing it with a label.
+She needs a program to help her with the sorting.
```
junedev, line 5:
```suggestion
Jen is responsible for things that were pre-categorized as numbers and needs a program to help her with the sorting.
```

exercises/concept/sorting-room/.meta/config.json
```diff
@@ -0,0 +1,17 @@
+{
+  "blurb": "Learn about type casting and type assertions in the sorting room.",
+  "authors": [
+    "jmrunkle"
+  ],
```
junedev, line 5: Maybe I ("junedev") could get a contributor slot here for helping with exercise structure/adding the fancy number part?

junedev (approved):

junedev:
Exercise looks good now! Left 2 more tiny comments, feel free to merge whenever you see fit.

exercises/concept/sorting-room/.meta/config.json
jmrunkle: Oh for sure, let me do that super quick. Thanks!

PR #1724, opened 2021-09-28
Add pointer concept

### LINKED ISSUE #1643: [Sept Sprint] New exercise teaching pointers

andrerfcsantos:
Fixes #1643


concepts/pointers/about.md
```diff
@@ -1,3 +1,155 @@
...
+
+
+A note of caution however: always check if a pointer is not `nil` before dereferencing. Dereferencing a `nil` pointer will make the program panic at runtime!
+
+### Pointers to structs
+
+So far we've only seen pointers to primitive values. We can also create pointers for structs:
+
+```go
+type Point struct {
```
sudomateo, line 105: I'd personally recommend using another name for this struct type as the `Point` name is too similar to the pointer concept and may confuse readers. Perhaps a `Person` struct instead?

andrerfcsantos: That's a good point. After writing it down I also thought about that. I'll change it :)

sudomateo: Sorry to butt in! I was just poking around at the open pull requests. Appreciate your work here!

ErikSchierboom (commented):
I'm really impressed with this PR! You did an excellent job. I've left some minor comments.

I don't really know anything about Go, so I'll leave it to others to review the Go parts of the code.

concepts/pointers/introduction.md
```diff
@@ -1,3 +1,149 @@
...
+
+## Variables and Memory
+
+Let's say we have a regular integer variable `a`:
+
+```go
+var a int
+```
+
+When we declare a variable, Go has to find a place in memory to place its value. Conviniently, when we need to fetch the value stored in that piece of memory, we can just refer to it by the variable name.
```
iHiD, line 13:
```suggestion
When we declare a variable, Go has to find a place in memory to store its value. This is largely abstracted from us — when we need to fetch the value stored in that piece of memory, we can just refer to it by the variable name.
```

concepts/pointers/introduction.md
```diff
@@ -1,3 +1,149 @@
...
+
+For instance, when we do:
+
+```go
+a + 2
+```
+
+We are effectively fetching the value stored in the memory associated with the variable `a` and adding 2 to it.
```
iHiD, line 21:
```suggestion
For instance, when we write `a + 2`, we are effectively fetching the value stored in the memory associated with the variable `a` and adding 2 to it.
```

iHiD: I don't think this one needs to be split over multiple lines.

concepts/pointers/introduction.md
```diff
@@ -1,3 +1,149 @@
...
+
+Similarly, when we need to change the value in the piece of memory of `a`, we can use the variable name to do an assignment:
+
+```go
+a = 3
+```
+
+The piece of memory that `a` is associated with now will have the value `3`.
```
iHiD, line 29:
```suggestion
The piece of memory that is associated with `a` will now will now be storing the value `3`.
```

concepts/pointers/introduction.md
```diff
@@ -1,3 +1,149 @@
...
+
+## Pointers
+
+While variables allow us to refer to values in memory, sometimes it's useful to know the memory address of the value that the variable refers. Pointers hold the memory addresses of those values. A pointer declaration looks like this:
```
iHiD, line 33:
```suggestion
While variables allow us to refer to values in memory, sometimes it's useful to know the **memory address** to which the variable is pointing. **Pointers** hold the memory addresses of those values. A pointer declaration looks like this:
```

I've just tweaked this so that we show why they're called pointers (I always find if you can make the naming seem "natural" when teaching something, it sinks in better)

concepts/pointers/introduction.md
```diff
@@ -1,3 +1,149 @@
...
+
+```go
+var p *int // 'p' can hold the memory address of an integer
+```
+
+Here we declaring a variable `p` of type `*int`. The type `*int` means "pointer to int". In other words, we are saying that `p` is a pointer to int, which means that `p` will hold the memory address of an integer. The zero value of pointers is `nil`. A `nil` pointer holds no memory address.
```
iHiD, line 39:
```suggestion
Here we declaring a variable `p` of type `*int`. The type `*int` means "pointer to int". In other words, we are saying that `p` is a pointer to int, which means that `p` will hold the memory address of an integer. The zero value of pointers is `nil` because a `nil` pointer holds no memory address.
```

iHiD: I felt those two sentences should be linked so it feels a little less like two seperate facts.

concepts/pointers/introduction.md
```diff
@@ -1,3 +1,149 @@
...
+
+### Getting a pointer to a variable
+
+The operator `&` gives the memory address of its operand. This means that to get the memory address of the value of a variable `a`, we can do `&a`. `&a` will create a pointer to the value of `a`:
```
iHiD, line 43: I found this a little confusing, mainly because I'm not used to the word "operand". 

```suggestion
To find the memory address of the value of a variable, we can use the `&` operator. 
For example, if we want to find and store the member address of variable `a` in the pointer `p`, we can do the following:
```

concepts/pointers/introduction.md
```diff
@@ -1,3 +1,149 @@
...
+
+```go
+var a int
+a = 2
+
+var p *int
+p = &a // the variable 'p' wil hold the memory address of 'a'
```
iHiD, line 50:
```suggestion
p = &a // the variable 'p' will hold the memory address of 'a'
```

concepts/pointers/introduction.md
```diff
@@ -1,3 +1,149 @@
...
+### Accessing the value via a pointer (dereferencing)
+
+When we have a pointer, we might want to know the value that is in the memory address of the pointer. We can do this using the `*` operator:
+
+```go
+var a int 
+a = 2
+
+var p *int 
+p = &a // the variable 'p' wil hold the memory address of 'a'
```
iHiD, line 62:
```suggestion
p = &a // the variable 'p' will hold the memory address of 'a'
```

concepts/pointers/introduction.md
```diff
@@ -1,3 +1,149 @@
...
+var a int
+a = 2
+
+var p *int
+p = &a // the variable 'p' wil hold the memory address of 'a'
+```
+
+### Accessing the value via a pointer (dereferencing)
+
+When we have a pointer, we might want to know the value that is in the memory address of the pointer. We can do this using the `*` operator:
```
iHiD, line 55:
```suggestion
When we have a pointer, we might want to know the value store in the memory address to which it points. We can do this using the `*` operator:
```

concepts/pointers/introduction.md
```diff
@@ -1,3 +1,149 @@
...
+pa = &a          // 'pa' now refers to the memory address of 'a'  
+*pa = *pa + 2    // increment by 2 the value at memory address 'pa'
+
+fmt.Println(a)   // Output: 4
+                 // 'a' will have the new value that was changed via the pointer!
+```
+
+Assigning to `*pa` will change the value stored at the memory address `pa` holds. Since `pa` holds the memory address of `a`, by assigning to `*pa` we are effectively changing the value of `a`! 
+
+A note of caution however: always check if a pointer is not `nil` before dereferencing. Dereferencing a `nil` pointer will make the program panic at runtime!
```
iHiD, line 86:
```suggestion
A note of caution however: always check if a pointer is not `nil` before dereferencing. Dereferencing a `nil` pointer will make the program panic at runtime!
```

Could we use a different word to "panic" here? Maybe "crash". I'm not sure someone coming from a different language (e.g. Ruby) would be familiar with the word, and so I have images in my mind of someone running around panicking, rather than the more technical usage :)

concepts/pointers/introduction.md
```diff
@@ -1,3 +1,149 @@
...
+}
+
+var peter Person
+peter = Person{Name: "Peter", Age: 22}
+
+var p *Person
+p = &peter
+```
+
+We could've also created a new `Person` and immediately get a pointer to it:
```
iHiD, line 105:
```suggestion
We could have also created a new `Person` and immediately stored a pointer to it:
```

concepts/pointers/introduction.md
andrerfcsantos: I was largely inspired by [Tour of Go: Pointers](https://tour.golang.org/moretypes/1) that also calls it operand.

But I agree the suggestion makes it a little more clear!

concepts/pointers/introduction.md
andrerfcsantos: In Go, a crash (in the sense of unhandled exception) is called a panic, that was the main reason for the use of panic there.

But I it doesn't hurt to make the introduction more inclusive and use a more general term.

Thanks for the suggestion.

concepts/pointers/introduction.md
```diff
@@ -1,3 +1,149 @@
-# Introduction
+# About
+
+Like many other languages, Go has pointers. A pointer hold the memory address of a value.
```
iHiD, line 3: I think we need to give people a reason for why they're learning this before they work through such a complex topic. I've copied and shortened the bit at the bottom of document here. 

```suggestion
Like many other languages, Go has pointers. 
If you're new to pointers, they can feel a little mysterious but once you get used to them, they're quite straight-forward.
They're a crucial part of Go, so take some time to really understand them.

Before digging into the details, it's worth understanding the use of pointers. Pointers are a way to share memory with other parts of our program, which is useful for two major reasons:
1. When we have large amounts of data, making copies to pass between functions is very inefficient. 
  By passing the memory location of where the data is stored instead, we can dramatically reduce the resource-footprint of our programs.
2. By passing pointers between functions, we can access and modify the computer's memory directly, meaning that any changes made by one function are immediately visible to other parts of the program when the function ends.
```

concepts/pointers/introduction.md
```diff
@@ -1,3 +1,149 @@
...
+fmt.Println(ages)
+// Output: map[Peter:22]
+// The changes the function 'addPeterAge' made to the map are visible after the function ends!
+```
+
+## Why are pointers useful?
+
+Pointers are a way to share memory with other parts of our program. When you pass a pointer as an argument to a function, you are sharing the memory address of a value with that function. The function can then use that memory address to access and modify the value. All the modifications that the function makes to that value through the pointer are then visible to other parts of the program when the function ends. Thus, pointers can be a way to share memory.
+
+Another situation where sharing memory can be useful is to avoid copies of large amounts of data. If we have a piece of data that takes a lot of space in memory, it might not be practical to copy that data to every place we need it. In that situation, we might consider using a pointer and pass the pointer around instead. That way we avoid copies of large amounts of data, but we allow several parts of the program to access it via the pointer.
```
iHiD, line 148: We can probably remove this from here now, as we have a summary version at the top. However, it'd be good to have this slightly more detailed version stay in the about.md

iHiD:
I've reviewed the `introduction.md` - It's really great 🙂

andrerfcsantos:
Thanks @iHiD and @ErikSchierboom for the extensive review!

These are great suggestions, really appreciate it. At a quick glance, I agree with all of them. I'll go through them in detail and submit a new patch.

exercises/concept/election-day/election_day_test.go
```diff
@@ -0,0 +1,191 @@
...
+	"testing"
+)
+
+// ElectionResult represents an election result
+type ElectionResult struct {
+	// Name of the candidate
+	Name string
+	// Number of votes the candidate had
+	Votes int
+}
```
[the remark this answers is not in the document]
andrerfcsantos: I like that suggestion a lot. I completely missed the `files.editor` key when reading through the docs.

Thanks for pointing that out.

iHiD:
@andrerfcsantos I don't think I've ever seen Erik as impressed with a first-pass of a totally new exercise before 👏

exercises/concept/election-day/.meta/exemplar.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// DisplayResult creates a message with the result to be displayed
+func DisplayResult(result *ElectionResult) string {
+	return fmt.Sprintf("%s (%d)", result.Name, result.Votes)
+}
+
+// DecrementVotesOfCandidate decrements by one the vote count of a candidate in a map
+func DecrementVotesOfCandidate(results map[string]int, candidate string) {
+	results[candidate]--
```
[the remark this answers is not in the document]
andrerfcsantos: Added.

exercises/concept/election-day/.meta/exemplar.go
```diff
@@ -0,0 +1,40 @@
...
+func NewElectionResult(candidateName string, votes int) *ElectionResult {
+	return &ElectionResult{
+		Name:  candidateName,
+		Votes: votes,
+	}
+}
+
+// DisplayResult creates a message with the result to be displayed
+func DisplayResult(result *ElectionResult) string {
+	return fmt.Sprintf("%s (%d)", result.Name, result.Votes)
```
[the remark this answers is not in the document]
andrerfcsantos: Added.

exercises/concept/election-day/.meta/exemplar.go
```diff
@@ -0,0 +1,40 @@
+package electionday
+
+import "fmt"
```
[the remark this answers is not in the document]
andrerfcsantos: Added.

concepts/pointers/introduction.md
andrerfcsantos: Fixed.

concepts/pointers/introduction.md
andrerfcsantos: I like this change a lot!

I talked a bit with some people to know their troubles learning pointers before doing this exercise. One of the things they told me is that they had trouble learning pointers because they were not seeing why they were useful. That's why I made this little section, but I agree 100% that having this upfront is even better.

Change made.

andrerfcsantos:
Changes applied. Ready for further review!

iHiD:
Great! I probably won't have time to review again before GoBridge launches, so if you're happy with the changes you've applied, I am too :)

I guess we just need a review on the Go side of things and then we can merge :)

exercises/concept/election-day/.docs/instructions.md
```diff
@@ -0,0 +1,124 @@
+# Instructions
+
+A local school near you has a very active student's association.
```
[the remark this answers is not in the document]
andrerfcsantos: Just checked and you are right. Fixed!

ErikSchierboom (approved):

ErikSchierboom:
@iHiD Yep! Let's wait for a proper Go review, but I'll give it my approval to show the other reviewers I like this PR!

bobahop:
Great work!

A couple of things.

For pointer to a slice, if the slice is appended to in a function it was passed into, the change will not be seen in the original, since the length and maybe capacity  will change, but the pointer is only for the address. So in that way a slice does not behave the same as a map. I only mention because they are grouped together in the explanation, and it mentions that map changes will be seen in the original.

The other issue is maybe to caveat against something like

`*p = &a`

I want to say I read somewhere that it's a common mistake for newer programmers to use the dereference operator when reassigning a pointer to a new address.

concepts/pointers/about.md
```diff
@@ -1,3 +1,154 @@
...
+pa = &a          // 'pa' now refers to the memory address of 'a'  
+*pa = *pa + 2    // increment by 2 the value at memory address 'pa'
+
+fmt.Println(a)   // Output: 4
+                 // 'a' will have the new value that was changed via the pointer!
+```
+
+Assigning to `*pa` will change the value stored at the memory address `pa` holds. Since `pa` holds the memory address of `a`, by assigning to `*pa` we are effectively changing the value of `a`! 
+
+A note of caution however: always check if a pointer is not `nil` before dereferencing. Dereferencing a `nil` pointer will make the program crash at runtime!
```
junedev, line 88: I think it is very important to show the actual error message here so people can recognize it later when it happens to them.
```suggestion
A note of caution however: always check if a pointer is not `nil` before dereferencing. Dereferencing a `nil` pointer will make the program crash at runtime!

```go
var p *int // p is nil initially
fmt.Println(*p)
// panic: runtime error: invalid memory address or nil pointer dereference
```

concepts/pointers/about.md
```diff
@@ -1,3 +1,154 @@
...
+If we create a map and call this function, the changes the function made to the map persist after the function ended. This is a similar behavior we get if we were using a pointer, but note how on this example we are not using any referencing/dereferencing or any of the pointer syntax:
+
+```go
+ages := make(map[string]int)
+addPeterAge(ages)
+fmt.Println(ages)
+// Output: map[Peter:22]
+// The changes the function 'addPeterAge' made to the map are visible after the function ends!
+```
+
```
junedev, line 145: Suggestion how to address the slices issue mentioned here https://github.com/exercism/go/pull/1724#issuecomment-932351399
```suggestion

The same applies when changing an existing item in a slice. However, actions that return a new slice like `append` will **not** modify the slice outside of the function.
```

concepts/pointers/about.md
```diff
@@ -1,3 +1,154 @@
...
+## Pointer arithmetic
+
+Unlike other languages, pointer arithmetic is not allowed in Go. This means that snippets like the following are invalid:
```
junedev, line 148: Suggesting to add an explanation of that pointer arithmetic actually is. Also rephrased the first bit a little.
```suggestion
Unlike other languages, Go does not support pointer arithmetic. This means you cannot go from a memory address of one value to the address of another by incrementing the pointer value. So snippets like the following are invalid:
```

concepts/pointers/introduction.md
```diff
@@ -1,3 +1,144 @@
-# Introduction
+# About
```
junedev, line 1:
```suggestion
# Introduction
```

exercises/concept/election-day/.docs/instructions.md
```diff
@@ -0,0 +1,124 @@
+# Instructions
+
+A local school near you has a very active students' association.
+The student's association is managed by a president and once every 2 years,
+elections are run to elect a new president.
+
+In this year's election, it was decided that a new digital system to
+count the votes was needed. The school needs your help building this new system.
+## 1. Create a vote counter
```
junedev, line 9:
```suggestion
count the votes was needed. The school needs your help building this new system.

## 1. Create a vote counter
```

exercises/concept/election-day/.meta/design.md
```diff
@@ -0,0 +1,32 @@
...
+
+- Know how to create a pointer from a variable
+- Know how to dereference a pointer
+- Know the pitfalls of dereferencing a nil pointer
+- Know how to work with pointers to structs
+- Know which pointer are already pointers
+
+## Out of scope
+
+- Unsafe pointers
```
junedev, line 17:
```suggestion
- Unsafe pointers
- Pointer receivers for methods (this is parts of the `methods` concept)
```

junedev (approved):
Great work! Really well done! :clap:  
Left some minor suggestions, if you like them they probably need to be applied in all the places.

concepts/pointers/about.md
moshegood: This is actually a common edge case in go.
```
func addOne(in []int) []int {
	return append(in, 1)
}

func main() {
	slice := []int{1, 2, 3, 4}
	slice = append(slice, 5)
	slice2 := append(slice, 6)
	fmt.Println(addOne(slice)) // prints 1,2,3,4,5,1
	fmt.Println(slice2)        // also prints 1,2,3,4,5,1
}
```

concepts/pointers/about.md
```diff
@@ -1,3 +1,160 @@
...
 
-TODO: add information on pointers concept
+Like many other languages, Go has pointers. 
+If you're new to pointers, they can feel a little mysterious but once you get used to them, they're quite straight-forward.
+They're a crucial part of Go, so take some time to really understand them.
+
+Before digging into the details, it's worth understanding the use of pointers. Pointers are a way to share memory with other parts of our program, which is useful for two major reasons:
+1. When we have large amounts of data, making copies to pass between functions is very inefficient. 
+  By passing the memory location of where the data is stored instead, we can dramatically reduce the resource-footprint of our programs.
+2. By passing pointers between functions, we can access and modify the computer's memory directly, meaning that any changes made by one function are immediately visible to other parts of the program when the function ends.
```
moshegood, line 10: s/computer's memory/single copy of the data/

concepts/pointers/about.md
```diff
@@ -1,3 +1,160 @@
...
+p = &Person{Name: "Peter", Age: 22}
+
+fmt.Println(p.Name) // Output: "Peter" 
+                    // Go automatically dereferences 'person' to allow
+                    // access to the 'Name' field
+```
+
+## Slices and maps are already pointers
+
+Slices and maps are special types because they already have pointers in their implementation. This means that more often that not, we don't need to create pointers for these types to share the memory address for their values. Imagine we have a function that adds a key to a map:
```
moshegood, line 132: This is actually a potential edge case:
```
func addOne(in []int) []int {
	return append(in, 1)
}

func main() {
	slice := []int{1, 2, 3, 4}
	slice = append(slice, 5)
	slice2 := append(slice, 6)
	fmt.Println(addOne(slice)) // prints 1,2,3,4,5,1
	fmt.Println(slice2)        // also prints 1,2,3,4,5,1
}
```

concepts/pointers/about.md
```diff
@@ -1,3 +1,160 @@
...
+
+```go
+ages := make(map[string]int)
+addPeterAge(ages)
+fmt.Println(ages)
+// Output: map[Peter:22]
+// The changes the function 'addPeterAge' made to the map are visible after the function ends!
+```
+
+The same applies when changing an existing item in a slice. However, actions that return a new slice like `append` will **not** modify the slice outside of the function.
```
moshegood, line 151: This is not necessarily true:
```
func addOne(in []int) []int {
	return append(in, 1)
}

func main() {
	slice := []int{1, 2, 3, 4}
	slice = append(slice, 5)
	slice2 := append(slice, 6)
	fmt.Println(addOne(slice)) // prints 1,2,3,4,5,1
	fmt.Println(slice2)        // also prints 1,2,3,4,5,1
}
```

andrerfcsantos: Maybe we can say "might not modify the slice" and link to a resource that explains this in more detail like [Arrays, slices (and strings): The mechanics of 'append'](https://go.dev/blog/slices).

andrerfcsantos:
Made the suggested changes.

In addition I also made other improvements:

* In the pointer arithmetic example I declared the variables using the long form instead of the shorthand syntax. This is to 1) make the types of each variable more explicit 2) be more consistent with the rest of the examples
* I modified the map example slightly to increment a value on an existing map instead of adding a new key. This is just to be closer to what the student will be required to do in the exercise.

PR #1725, opened 2021-09-29
`cars-assemble`: add a task to implement `SuccessRate` and add hints

### LINKED ISSUE #1710: `cars-assemble`: add implementation of `successRate()` as a task
Hi,
just did the exercise and found the instructions to be a bit confusing (but maybe that's just me?)

In the Instructions, it says:
```
Note that the calculation of the success rate has been provided already.
```
To me this implies that the function `sucessRate()` has already been implemented - which is not the case. Seems to be a leftover of #1662

Suggestions:
- remove the above mentioned sentence
- Implemenation of `successRate()` should be added as the first task, since the calculation requires it. Also move the function to the first place of the stubs.

Thoughts?

jmrunkle:
Fixes #1710

### DIFF
--- a/exercises/concept/cars-assemble/.docs/hints.md
+++ b/exercises/concept/cars-assemble/.docs/hints.md
@@ -1,28 +1,14 @@
 # Hints
 
-TODO: write hints
+## 1. Calculate the success rate
 
-## Arithmetic Operators in Go
+- Use if conditionals to return the values in the table.
 
-Below shows the set of arithmetic operators available in Go in use:
+## 2. Calculate the production rate per hour
 
-```go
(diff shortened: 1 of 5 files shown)


junedev (commented):
To make this exercise consistent with the structure of other exercises, I would recommend to move successRate to the top of the stub and exemplar, make it exported and add a test case that the student can get passing once the task is completed.

jmrunkle:
> To make this exercise consistent with the structure of other exercises, I would recommend to move successRate to the top of the stub and exemplar, make it exported and add a test case that the student can get passing once the task is completed.

All done. Care to take another look?

junedev (approved):
Thanks for fixing the other files as well!

PR #1729, opened 2021-09-30
Format msg in booking up for beauty

vanildosouto:


jmrunkle (approved):
The need_for_speed_test.go changes should probably have been in a different PR, but it's a clear improvement so I'm not going to force the issue.

Issue #1638, opened 2021-09-17
the-farm: introduce interfaces concept


ErikSchierboom:
`The Farm` should unlock `interfaces`. This will involve:

- Updating the exercise's instructions.md file
- Updating the exercise's introduction.md
- Updating the concept's introduction.md
- Updating the concept's about.md file
- Add `interfaces` to the `concepts` for the `the-farm` exercise in the `config.json`

junedev:
@jmrunkle Could you give me an update whether I should unassign you or not?

jmrunkle:
Sorry, still recovering from new baby. I'm nearly done with this so if you do not mind being a little patient, I can finish it.

junedev:
@jmrunkle I know how that feels :blue_heart: and I would really like that you don't have this issue in your neck anymore. So if you don't mind, @andrerfcsantos offered to finish this up. If you could push whatever you have to some branch/ create a draft PR, he can pick it up from there.

andrerfcsantos:
@jmrunkle I don't mind waiting a bit more, but like June said, I'll happily finish this so you can have more time with your family!

andrerfcsantos:
@jmrunkle Can you give us an update on this?

jmrunkle:
I just turned my in-progress code into a draft PR: #1976

I have lost a bit of context on where exactly I had left this, but I will try to make it complete today.

jmrunkle:
Looks like the main thing that was missing was to copy the intro material about interfaces into the intro for the exercise. Let me patch that in and then send the PR out for review.

junedev:
@jmrunkle To avoid further delays (students keep asking for this), I would suggest @andrerfcsantos can directly push small fixes to your branch if anything comes up instead of writing review comments and waiting for them to be resolved. This way it can be merged quickly. If you object, let us know.

jmrunkle:
Sounds fine to me. I don't mind either way

PR #1737, opened 2021-10-01
Fill out the `interfaces` concept

jmrunkle:
Part of #1638

### DIFF
--- a/concepts/interfaces/about.md
+++ b/concepts/interfaces/about.md
@@ -1,3 +1,53 @@
-# About
+# About Interfaces
 
-TODO: add information on interfaces concept
+An **interface** is effectively a set of method signatures.
+Any value that defines those methods "implements" the interface type (implicitly).
+Here is what an interface definition might look like:
+
+```go
+type InterfaceName interface {
+    MethodOne() MethodOneReturnType
+    ...
+}
+```
+
+## The `error` interface
+
+For example, here is the built-in `error` interface:
+
+```go
+type error interface {
+    Error() string
+}
+```
+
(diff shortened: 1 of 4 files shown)


concepts/interfaces/about.md
```diff
@@ -1,3 +1,53 @@
...
+Any value that defines those methods "implements" the interface type (implicitly).
+Here is what an interface definition might look like:
+
+```go
+type InterfaceName interface {
+    MethodOne() MethodOneReturnType
+    ...
+}
+```
+
```
junedev, line 13: I would have liked two more pieces of info in this section.
1. that a type can implement multiple interfaces
2. what are interfaces good for/ when are they used

(inspired by https://golang.org/doc/effective_go#interfaces)

Here some possible text but feel free to write your own if you don't like it.
```suggestion

Since interfaces are implemented implicitly, a type can easily implement multiple different interfaces.

If all function/method parameters and return values would need to be of one specific concrete (ie. non-interface) type, that would strongly limit their re-usability.
By using interface types for the parameters, only the behavior that is needed for the function to do its job is defined.
Now the function can be used with parameters of different concrete types, as long as they implement the correct interface.
With this, you can for example achieve dependency injection in Go.
A similar logic applies when using interface types as return values as we will see in the example below.

```

junedev (approved):
Sorry for the wait. Didn't have time to properly sit down and work on this earlier.

I only commented on the about file. Please judge for yourself which of the changes you apply there also need to be transferred to the introduction.

In general its definitely "approvable" so I am doing this now. You can apply what you think is helpful from the comments and then merge.

PR #1738, opened 2021-10-01
add new hacktoberfest label

junedev:
This is on hold until I can get confirmation that using only this label without opting in the whole repo actually works.

Update: DJ confirmed this works. By providing the `hacktoberfest-accepted` label we can mark indivial PRs so they count for Hacktoberfest without setting hacktoberfest as topic for the repo. The label cannot be applied if its not in the appends (it will be removed) so that is why this change is needed.

### DIFF
--- a/.appends/.github/labels.yml
+++ b/.appends/.github/labels.yml
@@ -51,6 +51,10 @@
   description: ""
   color: "cc8744"
 
+- name: "hacktoberfest-accepted"
+  description: "Makes an individual PR count towards the hacktoberfest goal"
+  color: "a6692d"
+
 - name: "help wanted"
   description: ""
   color: "5319e7"


iHiD:
@SaschaMann @ErikSchierboom Should we add this to org-wide-files?

junedev:
> Should we add this to org-wide-files?

My 2 cents on this: I thought that would cause issues because some tracks already have that label and we can't prefix it with `x:` because then it will probably not count for Hacktoberfest.

iHiD:
> I thought that would cause issues because some tracks already have that label 

In those cases, it should probably just be a no-op.

SaschaMann:
I think duplicates in the label file won't break anything but I'll check

SaschaMann:
Seems to work fine: https://github.com/SaschaMann/julia-1/runs/3767391419?check_suite_focus=true (search for `hacktoberfest-accepted`)

The colour and description will be overwritten/changed but I doubt that's an issue.

ErikSchierboom:
@SaschaMann It won't remove the old label, right?

SaschaMann:
> label: {Name:hacktoberfest-accepted Description:Opt-in to hacktoberfest Color:4a4a4a} not changed on SaschaMann/julia-1
>label {Name:hacktoberfest-accepted Description:Opt-in to hacktoberfest :) Color:5b5b5b} updated on: SaschaMann/julia-1

I think it doesn't.

junedev:
Closing this PR since this will be handled via the org wide files then.

Issue #1694, opened 2021-09-24
Debug docs don't mention "fmt" import


junedev:
The debug docs https://github.com/exercism/go/blob/ab614ba795c5cec9a28411f422a578cae5236f97/exercises/shared/.docs/debug.md explain which method to use but they don't that you have to import "fmt" and how to do that. So if a student wants to debug something in a very early exercise, just writing `fmt.Println` like explained in the docs will not work for them.

People reported issues with fmt.Println via the Editor Bug Report feature and that might be one reason for someone having trouble.

PR #1742, opened 2021-10-01
Mention `import "fmt"` in debug.md

angelikatyborska:
Closes https://github.com/exercism/go/issues/1694

Is this enough? It would have been enough to help me out. I hit this problem too as a total Golang noob. I didn't know how to write imports so first I wrote `import fmt` and hit a weird error about "expected string, got newline" 😁 

I think for people that wrote software before it will be obvious that the imports go at the top of the file usually, but the `fmt.Println` calls can go anywhere, right? I don't think it's necessary to spell out explicitly.

### DIFF
--- a/exercises/shared/.docs/debug.md
+++ b/exercises/shared/.docs/debug.md
@@ -3,6 +3,7 @@
 When a test fails, a message is displayed describing what went wrong and for which input. You can also use the fact that [console output][fmt-println] will be shown too. You can write to the console using:
 
 ```go
+import "fmt"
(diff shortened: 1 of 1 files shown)


jmrunkle (approved):

jmrunkle:
> Is this enough? It would have been enough to help me out. I hit this problem too as a total Golang noob. I didn't know how to write imports so first I wrote `import fmt` and hit a weird error about "expected string, got newline" 😁

I think this is good enough. Thanks!

PR #1744, opened 2021-10-01
Add out-of-bounds to `card-tricks` instructions

### LINKED ISSUE #1743: card-tricks: instructions should mention what to do if indices are out-of-bounds
ynfle:
> The instructions don't mention what to do with out of bound values (negative and >= len(slice)).

This applies to most of the exercises.

jmrunkle:
Fixes #1743

### DIFF
--- a/exercises/concept/card-tricks/.docs/instructions.md
+++ b/exercises/concept/card-tricks/.docs/instructions.md
@@ -9,8 +9,21 @@ To make things a bit easier she only uses the cards 1 to 10.
 Return the card at position `index` from the given stack.
 
 ```go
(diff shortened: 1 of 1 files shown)


junedev (approved):
I don't think the second example for the out of bounds case is necessary because the text is rather self-explanatory but I am fine with merging this as is as well.

jmrunkle:
> I don't think the second example for the out of bounds case is necessary because the text is rather self-explanatory but I am fine with merging this as is as well. I leave it up to you.

Frankly, pretty much all of the examples should be pretty self-explanatory for a native English speaker, but I think the examples remove all possible doubt for anyone that may struggle with English.

PR #1752, opened 2021-10-02
cars-assemble: explain and use the `if expression; condition { }` syntax

### LINKED ISSUE #1644: cars-assemble: explain `if expression; condition { }` syntax

jmrunkle:
Fixes #1644

### DIFF
--- a/concepts/conditionals-if/about.md
+++ b/concepts/conditionals-if/about.md
@@ -43,7 +43,7 @@ function getVal(connected bool) int {
 }
 ```
 
-If statements can also use a short initialization statement that can be used to initialize one or more variables for the if statement.
+If statements can also include a short initialization statement that can be used to initialize one or more variables for the if statement.
 For example:
 
 ```go
--- a/concepts/conditionals-if/about.md
+++ b/concepts/conditionals-if/about.md
@@ -1,6 +1,9 @@
 # About
 
-Conditionals in Go are similar to conditionals in other languages. The underlying type of any conditional operation is the `bool` type, which can have the value of `true` or `false`. Conditionals are often used as flow control mechanisms to check for various conditions. For checking a particular case an [`if` statement][if_statement] can be used, which executes its code if the underlying condition is `true` like this:
+Conditionals in Go are similar to conditionals in other languages.
+The underlying type of any conditional operation is the `bool` type, which can have the value of `true` or `false`.
+Conditionals are often used as flow control mechanisms to check for various conditions.
+For checking a particular case an [`if` statement][if_statement] can be used, which executes its code if the underlying condition is `true` like this:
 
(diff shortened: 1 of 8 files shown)


concepts/conditionals-if/about.md
```diff
@@ -40,7 +43,24 @@ function getVal(connected bool) int {
...
+For example:
+
+```go
+num := 7
+if v := 2 * num; v > 10 {
+    fmt.Println(v)
+} else {
+    fmt.Println(num)
+}
+// Output: 7
```
junedev, line 56: Usually these are always used with functions so I am contemplating whether to make this look a bit more like it would look in real Go code even if the example is a little more abstract then, e.g. like this:
```suggestion
num := 8
if ok := isEven(num); ok {
    // ...
}
```

I also removed the else part and the code inside as they make it harder to see what the example is actually about.

concepts/conditionals-if/about.md
```diff
@@ -40,7 +43,24 @@ function getVal(connected bool) int {
 }
 ```
 
-Coming from other languages one may be tempted to try to use one-line conditionals. Go does not support ternary operators or one-line conditionals. This is a purposeful design decision to limit the amount of possibilities, making code simpler and easier to read.
+If statements can also use a short initialization statement that can be used to initialize one or more variables for the if statement.
```
junedev, line 46:
```suggestion
If statements can also include a short initialization statement that can be used to initialize one or more variables for the if statement.
```

junedev (approved):
Great that you also added a task on this!

I left two minor comments. If you apply any of those, they need to be applied in all the places.

concepts/conditionals-if/about.md
jmrunkle: So, the problem I had with going this route is that it would beg the question of why use the initialization at all? For example, that one could be re-written as:

```go
if isEven(num) {
  // ...
}
```

I contrived the example that I wrote specifically to avoid that problem (since you want to print out the value returned). I suppose I could have the example I wrote invoke a function called `double`, but that means you have to infer what that function does in order to understand which branch it chose. In the end, I think removing the need for a function at all is simpler for the student.

PR #1754, opened 2021-10-02
Add a hint to `cars-assemble` task 2 about casting

angelikatyborska:
This hint is based on my own struggles with this exercise. I am not 100% sure if what I wrote is true, it's just my interpretation of the error that running the code produced.


junedev (approved):
If the introduction / arithmetic operator concept does not mention anything about the types of the operands, then please open another issue saying this needs to be clarified.

PR #1756, opened 2021-10-02
lasagna: Reorder instructions

SaschaMann:
The first task requires defining a constant, not a function, so this matches that order.

### DIFF
--- a/exercises/concept/lasagna/.docs/introduction.md
+++ b/exercises/concept/lasagna/.docs/introduction.md
@@ -13,35 +13,6 @@ All source files in a directory must share the same package name.
 package lasagna
 ```
 
-## Functions
-
-Go functions accept zero or more parameters.
-Parameters must be explicitly typed, there is no type inference.
-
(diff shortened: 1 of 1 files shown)


junedev (commented):
I think generally it makes more sense to introduce variables before functions. Could you apply the same change to the concept docs (introduction and about)?

junedev:
@bobtfish Is it ok to merge this or will that mess with the lasagna branch you are working on and you rather want to make the changes yourself there?

SaschaMann:
> Could you apply the same change to the concept docs (introduction and about)?

Sure, I'll wait for @bobtfish's response first

bobtfish:
I don’t mind! Please feel free to fix everywhere @SaschaMann and we can merge this, or I’m happy to take care of it in my branch 🙃

junedev:
@bobtfish Then I decide you will take care of this of this in your PR. I will leave a note in the draft PR.

PR #1762, opened 2021-10-02
Fix instructions for this problem to suggest returning an error.

### LINKED ISSUE #1351: hamming: README refers to variations between languages but not the Go way

bobtfish:
Fixes #1351

### DIFF
--- a/exercises/practice/hamming/.docs/instructions.md
+++ b/exercises/practice/hamming/.docs/instructions.md
@@ -18,7 +18,4 @@ The Hamming Distance is useful for lots of things in science, not just biology, 
 
 # Implementation notes
 
-The Hamming distance is only defined for sequences of equal length, so
-an attempt to calculate it between sequences of different lengths should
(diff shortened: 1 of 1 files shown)


junedev:
This change means we would diverge from the problem spec. I am leaving this to someone else to decide this is the way to go here.

jmrunkle:
> This change means we would diverge from the problem spec. I am leaving this to someone else to decide this is the way to go here.

Unless configlet starts complaining about this, I think we can have slightly divergent wording that is more specific to the actual way the exercise is written.

jmrunkle (approved):

Issue #1728, opened 2021-09-30
party-robot: update introduction


ErikSchierboom:
The `party-robot` exercise unlocks the `packages` concept, but its `introduction.md` document does not explain what packages are (it does show how they can be used for the `fmt` package).

Firstly, the `introduction.md` document of the `party-robot` exercise should be updated to better explain what packages are. There is already content for this in the `introduction.md` document of the `packages` concept.

Secondly, the `packages` concept should mention how public/private visibility works. See https://github.com/exercism/go/pull/1720 for sample text for this. The public/private visibility explanation should be added to the exercise's `introduction.md` and the concept's `introduction.md` document.

bobtfish:
I'm handling this

PR #1765, opened 2021-10-02
Update party robot introduction

bobtfish:
Fixes issue #1728

cc @junedev

### DIFF
--- a/concepts/packages/about.md
+++ b/concepts/packages/about.md
@@ -52,3 +52,18 @@ errors.New("Connection not established")
 // using the errors package located at github.com/pkg/errors by alias
 errs.New("Connection not established")
 ```
+
+Similar to program languages like Java that use access modifiers such as `public` and `private`, Go determines if an item is exported and unexported through how it is declared. 
+To allow an item to be called by code in other packages (i.e. [export an identifier](https://golang.org/ref/spec#Exported_identifiers) , the name must start with a capital letter:
+
+```go
+package greeting
+// Hello is a public function (callable from other packages)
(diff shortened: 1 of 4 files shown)


concepts/packages/about.md
```diff
@@ -52,3 +52,19 @@ errors.New("Connection not established")
...
+package greeting
+// Hello is a public function (callable from other packages)
+func Hello(name string) string {
+    return "Hello " + name
+}
+// hello is a private function (not callable from other packages)
+func hello(name string) string {
+    return "Hello " + name
+}
+```
```
junedev, line 70: I think it would be good also add a sentence that the same applies for struct field names, I dont think that is obvious from the above

bobtfish: Good call for the `about.md`, but should this make it to the `introduction` for the concept or exercise, as it isn't needed to solve them?

junedev: The introduction files are not only about doing the exercise but also understanding the concept itself. This is a case for something like that.

In about I would write something and show an example. In the introduction files I would just write one sentence about it.

junedev (approved):

junedev:
@bobtfish Small info regarding your PR description text: "fixes issue" does not create a link between the issue and the PR so that the issue is automatically closed then the PR is merged. You can find a list of the keywords Github understands here https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword.

PR #1766, opened 2021-10-02
Cleanups for Lasagna exercise and basics concept from feedback.

### LINKED ISSUE #1745: Lasagna (basics) feedback
Hi! I'm going through the very first learning exercise on the track as a student who has never written a single line of Golang before. I managed to get through it on my own without cheating, so that's great :) but I still have some suggestions for improvements.

Below might be a lot text, but the kind of changes that I am suggesting would be actually tiny, like adding an extra word here and there, or a single sentence.

This sounds mindblowing to me. How can a single function have multiple return values? I have never seen this before. Or am I misinterpreting things and does this mean only that the type definition of the return value allows something like a union type? Anyway, my problem here is that there is no follow up to that statement. No examples of a function with "multiple return values", whatever that means in practice. Is it even relevant to the exercise? If yes, maybe a code example would be nice to have. If not, maybe it doesn't need to say anything? I have no idea. I was just confused by this sentence.

## 2. Lowercase names vs uppercase names

In the example code, a variable's name is lowercase while a constant's name is uppercase. Is this a requirement? Is it just a convention? Is it just like with function names, that both variable names and constant names can be either lowercase or uppercase and that changes whether they are public or not? I wish there was at least one sentence for each explaining the name casing, like there is for functions.

## 3. Variable names - camelCase or not?

I don't see any hints in the introduction or in the code itself as to whether I'm supposed to give variable names using camelCase or snake_case, or something else. I do have examples that constants and functions are named in PascalCase, and a variable name is supposed to be lowercase, so probably camelCase, but a specific mention of this would be nice for a total Golang beginner.

## 4. Types? I don't know them... 

So, I need to define the functions from zero, there are no function stubs. I don't know any types in Golang beyond `string` and `int` that are in the code examples in the introduction.

When I defined the function without arguments as a start, I got this compilation error:

```
# lasagna [lasagna.test]
./lasagna_test.go:39:31: too many arguments in call to RemainingOvenTime
	have (number)
	want ()
./lasagna_test.go:39:31: RemainingOvenTime(25) used as value
./lasagna_test.go:63:29: too many arguments in call to PreparationTime
	have (int)
	want ()
./lasagna_test.go:63:29: PreparationTime(tt.layers) used as value
./lasagna_test.go:88:25: too many arguments in call to ElapsedTime
	have (int, int)
	want ()
```

From this compilation error, I was convinced that `number` is the correct type of the argument in `RemainingOvenTime`, but that was wrong. They're all supposed to be `int`s. Could the instructions maybe say that? "...function that takes the actual minutes the lasagna has been in the oven (`int`)"? 

Why is this compilation error saying `have (number)` and not `have (int)` for `RemainingOvenTime`?

---

That's all my thoughts! I hope it's not too much. I can probably answer a lot of those questions myself by googling, but I wanted to share this because I think a lot of Golang beginners might have similar doubts.

### LINKED ISSUE #1763: lasagna: error unclear when not specifying a return type
```go
package lasagna

// TODO: define the 'OvenTime' constant
const OvenTime = 40

// TODO: define the 'RemainingOvenTime()' function
func RemainingOvenTime(t int) {
    return OvenTime - t
}
```

results in the following error

```go
# lasagna [lasagna.test]
./lasagna.go:8:5: too many arguments to return
	have (int)
	want ()
./lasagna_test.go:39:31: RemainingOvenTime(25) used as value
```

It's not clear that this is a "Go-level error" rather than a task-level error. The error can be interpreted as "You're returning and int but the exercise expects `()`".

Clearing this up with a custom error message or some additional info might make it clearer for people coming from languages where return types aren't given in the signature. It's mentioned in the docs but it's easy to overlook and this error doesn't point one towards the solution, imo.

bobtfish:
Fixes #1745 
Fixes #1763

### DIFF
--- a/exercises/concept/lasagna/.docs/hints.md
+++ b/exercises/concept/lasagna/.docs/hints.md
@@ -4,9 +4,19 @@
 
 - An [integer value][integers] can be defined as one or more consecutive digits.
 
+If you see a `panic:` error when running the tests, this is as you haven't implemented one of the functions (it should say which one) or you have left the boilerplate in place, you need to remove the `panic(...)` line from the supplied code.
+
 ## 1. Define the expected oven time in minutes
 
 - You need to define a [constant][constants] and assign it the expected oven time in minutes.
+
+If you see an `undefined: OvenTime` error then double check that you have the constant defined.
+
+If you see an `invalid operation: got != tt.expected (mismatched types float64 and int)` error then you've put a decimal point into the `OvenTime` - remove the decimal so that it'll be an integer.
+
+If you see a `syntax error: non-declaration statement outside function body` error then this is as you forgot the `const` keyword.
+
+If you see an `syntax error: unexpected :=, expecting =` error then this is as you are trying to assign to the constant like a variable; contants are assigned using `=` not `:=`.
 
 ## 2. Calculate the remaining oven time in minutes
 
--- a/exercises/concept/lasagna/.docs/hints.md
+++ b/exercises/concept/lasagna/.docs/hints.md
@@ -4,7 +4,8 @@
 
 - An [integer value][integers] can be defined as one or more consecutive digits.
 
-If you see a `panic:` error when running the tests, this is as you haven't implemented one of the functions (it should say which one) or you have left the boilerplate in place, you need to remove the `panic(...)` line from the supplied code.
+If you see a `panic:` error when running the tests, this is because you have not implemented one of the functions (it should say which one) or you have left the boilerplate in place.
+You need to remove the `panic(...)` line from the supplied code and replace it with a real implementation.
 
 ## 1. Define the expected oven time in minutes
 
@@ -12,11 +13,12 @@ If you see a `panic:` error when running the tests, this is as you haven't imple
 
 If you see an `undefined: OvenTime` error then double check that you have the constant defined.
 
-If you see an `invalid operation: got != tt.expected (mismatched types float64 and int)` error then you've put a decimal point into the `OvenTime` - remove the decimal so that it'll be an integer.
+If you see an `invalid operation: got != tt.expected (mismatched types float64 and int)` error then you have likely put a decimal point into the `OvenTime` causing Go to infer the type as a floating point number.
+Remove the decimal and the type will be inferred as an `int`.
 
-If you see a `syntax error: non-declaration statement outside function body` error then this is as you forgot the `const` keyword.
+If you see a `syntax error: non-declaration statement outside function body` error then it is likely that you forgot the `const` keyword.
 
-If you see an `syntax error: unexpected :=, expecting =` error then this is as you are trying to assign to the constant like a variable; contants are assigned using `=` not `:=`.
+If you see a `syntax error: unexpected :=, expecting =` error then you are likely trying to assign the constant using `:=` like a variable; constants are assigned using `=` not `:=`.
 
 ## 2. Calculate the remaining oven time in minutes
 
--- a/exercises/concept/lasagna/.docs/hints.md
+++ b/exercises/concept/lasagna/.docs/hints.md
@@ -3,22 +3,15 @@
 ## General
 
 - An [integer value][integers] can be defined as one or more consecutive digits.
-
-If you see a `panic:` error when running the tests, this is because you have not implemented one of the functions (it should say which one) or you have left the boilerplate in place.
-You need to remove the `panic(...)` line from the supplied code and replace it with a real implementation.
+- If you see a `panic:` error when running the tests, this is because you have not implemented one of the functions (it should say which one) or you have left the boilerplate in place. You need to remove the `panic(...)` line from the supplied code and replace it with a real implementation.
 
 ## 1. Define the expected oven time in minutes
 
 - You need to define a [constant][constants] and assign it the expected oven time in minutes.
-
-If you see an `undefined: OvenTime` error then double check that you have the constant defined.
-
-If you see an `invalid operation: got != tt.expected (mismatched types float64 and int)` error then you have likely put a decimal point into the `OvenTime` causing Go to infer the type as a floating point number.
-Remove the decimal and the type will be inferred as an `int`.
-
-If you see a `syntax error: non-declaration statement outside function body` error then it is likely that you forgot the `const` keyword.
-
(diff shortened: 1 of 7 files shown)


junedev:
This should include reordering the sections in the docs for this so that variables are explained before functions as suggested by @SaschaMann in #1756.

bobtfish:
@SaschaMann @angelikatyborska - this PR should address your feedback about the basics concept/exercise.

If you have time to have a look and let me know what you think then any further feedback that you have would be very welcome!

exercises/concept/lasagna/.docs/hints.md
```diff
@@ -4,10 +4,20 @@
 
 - An [integer value][integers] can be defined as one or more consecutive digits.
 
+If you see a `panic:` error when running the tests, this is as you haven't implemented one of the functions (it should say which one) or you have left the boilerplate in place, you need to remove the `panic(...)` line from the supplied code.
```
jmrunkle, line 7:
```suggestion
If you see a `panic:` error when running the tests, this is because you have not implemented one of the functions (it should say which one) or you have left the boilerplate in place.
You need to remove the `panic(...)` line from the supplied code and replace it with a real implementation.
```

exercises/concept/lasagna/.docs/hints.md
```diff
@@ -4,10 +4,20 @@
...
+
 ## 1. Define the expected oven time in minutes
 
 - You need to define a [constant][constants] and assign it the expected oven time in minutes.
 
+If you see an `undefined: OvenTime` error then double check that you have the constant defined.
+
+If you see an `invalid operation: got != tt.expected (mismatched types float64 and int)` error then you've put a decimal point into the `OvenTime` - remove the decimal so that it'll be an integer.
```
jmrunkle, line 15:
```suggestion
If you see an `invalid operation: got != tt.expected (mismatched types float64 and int)` error then you have likely put a decimal point into the `OvenTime` causing Go to infer the type as a floating point number.
Remove the decimal and the type will be inferred as an `int`.
```

exercises/concept/lasagna/.docs/hints.md
```diff
@@ -4,10 +4,20 @@
...
+
+If you see a `syntax error: non-declaration statement outside function body` error then this is as you forgot the `const` keyword.
```
jmrunkle, line 17:
```suggestion
If you see a `syntax error: non-declaration statement outside function body` error then it is likely that you forgot the `const` keyword.
```

exercises/concept/lasagna/.docs/hints.md
```diff
@@ -4,10 +4,20 @@
...
+
+If you see an `syntax error: unexpected :=, expecting =` error then this is as you are trying to assign to the constant like a variable; contants are assigned using `=` not `:=`.
```
jmrunkle, line 19:
```suggestion
If you see a `syntax error: unexpected :=, expecting =` error then you are likely trying to assign the constant using `:=` like a variable; constants are assigned using `=` not `:=`.
```

exercises/concept/lasagna/.docs/instructions.md
```diff
@@ -18,24 +18,37 @@ OvenTime
...
+}
+
 RemainingOvenTime(30)
 // Output: 10
 ```
 
 ## 3. Calculate the preparation time in minutes
 
-Define the `PreparationTime()` function that takes the number of layers you added to the lasagna as a parameter and returns how many minutes you spent preparing the lasagna, assuming each layer takes you 2 minutes to prepare.
+Define the `PreparationTime(layers int) int` function that takes the number of layers you added to the lasagna as a parameter and returns how many minutes you spent preparing the lasagna, assuming each layer takes you 2 minutes to prepare.
```
jmrunkle, line 31: This seems likely to get out of sync if we ever update the method signature for any reason.

Just remove any type information and leave the function name only.

```suggestion
Define the `PreparationTime` function that takes the number of layers you added to the lasagna as a parameter and returns how many minutes you spent preparing the lasagna, assuming each layer takes you 2 minutes to prepare.
```

exercises/concept/lasagna/.docs/instructions.md
```diff
@@ -18,24 +18,37 @@ OvenTime
...
+}
+
 PreparationTime(2)
 // Output: 4
 ```
 
 ## 4. Calculate the elapsed working time in minutes
 
-Define the `ElapsedTime()` function that takes two parameters: the first parameter is the number of layers you added to the lasagna, and the second parameter is the number of minutes the lasagna has been in the oven. The function should return how many minutes in total you've worked on cooking the lasagna, which is the sum of the preparation time in minutes, and the time in minutes the lasagna has spent in the oven at the moment.
+Define the `ElapsedTime(numberOfLayers, actualMinutesInOven int) int` function that takes two parameters: the first parameter is the number of layers you added to the lasagna, and the second parameter is the number of minutes the lasagna has been in the oven.
```
jmrunkle, line 44:
```suggestion
Define the `ElapsedTime` function that takes two parameters: the first parameter is the number of layers you added to the lasagna, and the second parameter is the number of minutes the lasagna has been in the oven.
```

jmrunkle (changes requested):
Mostly looks good, just a few nitpicky comments if you do not mind addressing those.

exercises/concept/lasagna/.docs/hints.md
```diff
@@ -4,10 +4,20 @@
 
 - An [integer value][integers] can be defined as one or more consecutive digits.
 
+If you see a `panic:` error when running the tests, this is as you haven't implemented one of the functions (it should say which one) or you have left the boilerplate in place, you need to remove the `panic(...)` line from the supplied code.
+
 ## 1. Define the expected oven time in minutes
 
 - You need to define a [constant][constants] and assign it the expected oven time in minutes.
 
+If you see an `undefined: OvenTime` error then double check that you have the constant defined.
```
SaschaMann, line 13: I wonder if these should be hints?

SaschaMann (commented):
Looks good!

exercises/concept/lasagna/.docs/hints.md
bobtfish: Good point, the spec says these should be bullet points!

jmrunkle (approved):

PR #1771, opened 2021-10-03
Sync concepts/functions about.md and intro.md

IsaacG:


junedev:
These parts were deliberately left out from the introduction. It is supposed to be shorter that the about file and should not contain links if possible so the student does not wander off before actually starting the respective exercise.

Docs for reference: https://exercism.org/docs/building/tracks/concepts

PR #1772, opened 2021-10-03
Grammar: replace comma with an idea connector (concepts/functions)

IsaacG:


junedev (commented):
Thanks for reviewing the texts! 

Could you make sure all your fixes are applied consistently in those three files (if the respective text is in there):
- concept introduction.md
- concept about.md
- exercise introduction.md

That would very helpful.

IsaacG:
Done. I noticed `concept about` makes "return value" a link and the rest do not have a link for that. Is that supposed to be synced or is that intentionally different?

junedev:
It's different on purpose. Introduction should be rather short and not invite the student to wander off to other pages. In contrast about is shown after the exercise was completed for future reference so there it is ok to have links to invite exploring.

junedev (approved):

PR #1779, opened 2021-10-03
Grammar and word smithing: maps and gross-store

IsaacG:


concepts/maps/about.md
```diff
@@ -46,24 +46,26 @@ or
 >
 > Therefore it's important to initialize a map before using it
 
-Here are some operations that you can use with map
+Here are some operations that you can do with a `map`
```
junedev, line 49: Could you please remove the backticks around map everywhere?
The text does not refer to the keyword `map` (where we would use the backticks) but to the English word/general concept "map". That they are both the same word is just coincidence. You wouldn't write "you can do that with a `slice` for example".

junedev (commented):
Thanks a lot for already checking all the relevant files here. Just one small comment.

concepts/maps/about.md
IsaacG: Sorry about that. Fixed!

junedev (approved):

concepts/maps/about.md
junedev: Thanks for the quick fix!

IsaacG: Thanks for the quick approvals :D

PR #1784, opened 2021-10-04
Fix Go code formatting in codeblock (concepts/methods)

IsaacG:


junedev (commented):
Would be great if you can apply the fix(es) to the about.md and the exercise introduction as well.

IsaacG:
> Would be great if you can apply the fix(es) to the about.md and the exercise introduction as well.

This issue does not exist in those files. One does not have this code. The other has the "fixed" version.

junedev:
? I can see the issue with the missing space here https://github.com/exercism/go/blob/main/concepts/methods/about.md for example

IsaacG:
Sorry about that. I'm not sure why I couldn't see that. I did a `grep -rlF '){'` to pick up all occurrences this time.

junedev (approved):

PR #1785, opened 2021-10-04
Fix census.go package doc to pass go lint

IsaacG:


exercises/concept/census/census.go
```diff
@@ -1,4 +1,4 @@
-// Packages census simulates a system used to collect census data.
+// Package census simulates a system used to collect census data.
```
junedev, line 1: Could you apply the same fix for the .meta/exemplar.go file please?

IsaacG: Done! Sorry for missing that.

junedev (approved):

PR #1788, opened 2021-10-04
fixes #1776

### LINKED ISSUE #1776: Party Robot does not test the table number
It is possible to pass the tests with a hardcoded table number.

The instructions are confusing as to how to get from a number to a table 1B. I tried hardcoding it and it passed.

sharno:
change the 3rd task from converting to hex to adding leading zeroes
add more hints about formatting verbs

### DIFF
--- a/exercises/concept/party-robot/.docs/instructions.md
+++ b/exercises/concept/party-robot/.docs/instructions.md
@@ -27,14 +27,12 @@ HappyBirthday("Frank", 58)
 
 ## 3. Give directions
 
-Implement the `AssignTable` function to give directions. The robot provides the table number in hex (uppercase)
-due to a misalignment of its creator and the rest of the world. Fortunately the precision on the distance
-was limited to 1 digit.
+Implement the `AssignTable` function to give directions. The robot provides the table number in a 3 digits format. If the number is less than 3 digits it gets extra leading zeroes to become 3 digits (eg. 3 becomes 003). The robot mentions too the distance of the table with but only with a precision that's limited to 1 digit.
 
 ```go
 AssignTable("Christiane", 27, "Frank", "on the left", 23.7834298)
 // Output:
 // Welcome to my party, Christiane!
-// You have been assigned to table 1B. Your table is on the left, exactly 23.8 meters from here.
+// You have been assigned to table 027. Your table is on the left, exactly 23.8 meters from here.
 // You will be sitting next to Frank.
 ```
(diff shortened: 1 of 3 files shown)


exercises/concept/party-robot/.docs/instructions.md
```diff
@@ -27,14 +27,12 @@ HappyBirthday("Frank", 58)
 
 ## 3. Give directions
 
-Implement the `AssignTable` function to give directions. The robot provides the table number in hex (uppercase)
-due to a misalignment of its creator and the rest of the world. Fortunately the precision on the distance
-was limited to 1 digit.
+Implement the `AssignTable` function to give directions. The robot provides the table number in a 3 digits format. If the number is less than 3 digits it gets extra leading zeroes to become 3 digits (eg. 3 becomes 003). The robot mentions too the distance of the table with but only with a precision that's limited to 1 digit.
```
junedev, line 30:
```suggestion
Implement the `AssignTable` function to give directions. The robot provides the table number in a 3 digits format. If the number is less than 3 digits it gets extra leading zeroes to become 3 digits (eg. 3 becomes 003). The robot also mentions the distance of the table. Fortunately only with a precision that's limited to 1 digit.
```

exercises/concept/party-robot/party_robot_test.go
```diff
@@ -60,10 +65,10 @@ func TestAssignTable(t *testing.T) {
 			description:   "Greet the guest and give them directions to their seat",
 			name:          "Christiane",
 			direction:     "on the left",
-			tableNumber:   27,
+			tableNumber:   tableNumber,
 			distance:      23.7834298,
 			neighbourName: "Frank",
-			want:          "Welcome to my party, Christiane!\nYou have been assigned to table 1B. Your table is on the left, exactly 23.8 meters from here.\nYou will be sitting next to Frank.",
+			want:          fmt.Sprintf("Welcome to my party, Christiane!\nYou have been assigned to table %03d. Your table is on the left, exactly 23.8 meters from here.\nYou will be sitting next to Frank.", tableNumber),
```
junedev, line 71: It feels a little strange that we are using the same way here to test the result than to build it. Additionally you fixed the issue that hard coding the table number does not work anymore but hard coding all other values still works.

I would recommend instead of using the random number, we add another test case to the table.  

I saw the issue that there is only one test case also exist for all other tests.
Would you be able to extend this PR to add a second test case to all tests AND make sure all test cases are different from the example shown in the instructions.md file? That would be very helpful.

junedev (commented):
Thanks for taking care of this. I left some comments.

junedev:
@sharno Don't worry, I already did the remaining changes.

PR #1790, opened 2021-10-04
Add exercises' extra files to editor

andrerfcsantos:
While I was looking at `simple-cipher` using the editor, I was a bit frustrated I couldn't see an extra file that I was supposed to see, so I added it to the editor. I also went ahead and checked other exercises on the track that are in a similar situation and would benefit from having other files showing up in the editor.


jmrunkle (approved):
Thank you so much for doing this!

PR #1791, opened 2021-10-04
allergies: make tests not care about the order of elements

andrerfcsantos:
There are some solutions for this exercise that produce the right list of allergies Tom is allergic to, but do not produce them in the same order as the tests expect:

```
allergies_test.go:11: FAIL: Allergies(255): expected []string{"eggs", "peanuts", "shellfish", "strawberries", "tomatoes", "chocolate", "pollen", "cats"}, actual []string{"shellfish", "strawberries", "tomatoes", "chocolate", "pollen", "cats", "eggs", "peanuts"}
--- FAIL: TestAllergies (0.00s)
```

The `actual` list and the `expected` list contain the same elements, just ordered differently.

This PR makes the tests see if the lists have the same elements, regardless of their order. I think it's reasonable to allow for the elements to be ordered differently than the expected. The instructions mention returning a list of "all the allergens Tom is allergic to" without specifying any order.

### DIFF
--- a/exercises/practice/allergies/allergies_test.go
+++ b/exercises/practice/allergies/allergies_test.go
@@ -7,7 +7,7 @@ import (
 
 func TestAllergies(t *testing.T) {
 	for _, test := range listTests {
-		if actual := Allergies(test.score); !reflect.DeepEqual(actual, test.expected) && (len(actual) > 0 || len(test.expected) > 0) {
+		if actual := Allergies(test.score); !sameSliceElements(actual, test.expected) {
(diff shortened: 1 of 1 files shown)


jmrunkle (commented):
Any particular reason not to simply use reflect.DeepEqual on the maps to compare for equality?

andrerfcsantos:
> Any particular reason not to simply use reflect.DeepEqual on the maps to compare for equality?

I've changed it now.

It's mainly because as a general principle I try to stay away of the reflect package, too magical with unnecessary type assertions at runtime. But people here at Exercism seem to like it, so I won't go against that.

jmrunkle (approved):

PR #1798, opened 2021-10-05
Improve language

brandoncc:


jmrunkle (changes requested):
Fix the exemplar as well? https://github.com/exercism/go/blob/main/exercises/concept/party-robot/.meta/exemplar.go#L10

jmrunkle (approved):

PR #1809, opened 2021-10-06
Update instructions.md

sbromberger:
Double indirection is not correct for the function signature as described.

### DIFF
--- a/exercises/concept/election-day/.docs/instructions.md
+++ b/exercises/concept/election-day/.docs/instructions.md
@@ -51,11 +51,8 @@ Create a function `IncrementVoteCount` that will take a counter (`*int`) as an a
 var votes int
 votes = 3
 
-var voteCounter *int
-voteCounter = &votes
-
-IncrementVoteCount(voteCounter, 2)
-*voteCounter == 5 // true
+IncrementVoteCount(&votes, 2)
+*votes == 5 // true
 ```
 
 ## 4. Create the election results


junedev (commented):
On Slack you correctly said there is an incorrect indirection in the instructions for task 2 but you PR only shows a change for task 3. 🤔

sbromberger:
> On Slack you correctly said there is an incorrect indirection in the instructions for task 2 but you PR only shows a change for task 3. 🤔

Sorry about that. I edited within the github browser editor and wasn't expecting two errors. :)

junedev:
Also pinging the author @andrerfcsantos for the review here.

exercises/concept/election-day/.docs/instructions.md
```diff
@@ -51,11 +50,8 @@ Create a function `IncrementVoteCount` that will take a counter (`*int`) as an a
 var votes int
 votes = 3
 
-var voteCounter *int
-voteCounter = &votes
-
-IncrementVoteCount(voteCounter, 2)
-*voteCounter == 5 // true
+IncrementVoteCount(&votes, 2)
+votes == 5 // true
```
andrerfcsantos, line 54:
```suggestion
var voteCounter *int
voteCounter = &votes

IncrementVoteCount(voteCounter, 2)

votes == 5          // true
*voteCounter == 5   // true
```

I think being more explicit about the types is better. I did like the fact you showed that the variable was changed with `votes == 5`, so I think we should keep that.

exercises/concept/election-day/.docs/instructions.md
```diff
@@ -31,9 +31,8 @@ Create a function `VoteCount` that will take a counter (`*int`) as an argument a
 
 ```go
 votes := 3
```
andrerfcsantos, line 33:
```suggestion
var votes int
votes = 3
```

I think being explicit about the types is better.

exercises/concept/election-day/.docs/instructions.md
```diff
@@ -31,9 +31,8 @@ Create a function `VoteCount` that will take a counter (`*int`) as an argument a
...
-voteCounter := &votes
 
-VoteCount(&voteCounter)
+VoteCount(&votes)
```
andrerfcsantos, line 35:
```suggestion
var voteCounter *int
voteCounter = &votes

VoteCount(voteCounter)
```

andrerfcsantos (approved):
@sbromberger Thanks for this. I left some suggestions. The main thing is that to be consistent with the rest of the explanations, I think we should show the types of the variables explicitly, but I don't feel too strongly about this, feel free to disagree,

Issue #1806, opened 2021-10-05
Fix instruction "styling" for Annalyns Infiltration


junedev:
Since #1804 was merged, the different actions Annalyn can do now have different styling. One is italic, the others are normal font. I think that was not intended and it would be good to make all actions italic.

![image](https://user-images.githubusercontent.com/12543047/136072939-bd6ba3b1-f6f4-40e0-a290-443c25739979.png)

Source is here: https://github.com/exercism/go/blob/main/exercises/concept/annalyns-infiltration/.docs/instructions.md?plain=1

PR #1812, opened 2021-10-06
Update instructions.md

ishan2790:
https://github.com/exercism/go/issues/1806
Changing font of actions to italic
https://github.com/exercism/go/commit/928d39274c348bc0ba0eb8b5e782711a95a6c777


junedev (approved):
Thanks for the fix!

PR #1824, opened 2021-10-08
Fix the track distance in test case

vissree:
With a speed of 5, car can be driven 20 times before battery drains. So
the track distance to test CanFinish just before battery drained should
be 100.

### DIFF
--- a/exercises/concept/need-for-speed/need_for_speed_test.go
+++ b/exercises/concept/need-for-speed/need_for_speed_test.go
@@ -142,7 +142,7 @@ func TestCanFinish(t *testing.T) {
 			expected: true,
 		},
 		{
-			name: "Car can finish the track distance just before battery drained.",
+			name: "Car can finish the track distance just as the battery is drained.",
 			car: Car{
 				speed:        5,
 				batteryDrain: 5,


Lewiscowles1986:
This actually passes without change with learner implementations. Are you sure you are not falling into altering tests in order to make an implementation pass which should not?

Can you create a state output to show what you are trying to solve?

vissree:
Based on the description, my assumption is that with every invocation of Drive() the car covers Car.speed distance and battery drains by Car.batteryDrain.

My implementation of CanFinish was as below:

```go
	tries := track.distance / car.speed
	return car.battery-(tries*car.batteryDrain) > 0
}
```

My logic was to find out how many invocations of Drive() does it take to cover the given track distance, and then see if the car has enough battery to sustain.

But the problem here was that I was using '>' instead of '>=' which gives the incorrect result when the car has just enough battery to cover the track distance.

However, the test case does not catch this.

The test case "Car can finish the track distance just before battery drained." has the track distance set to 20, whereby, with a speed of 5 the car can cover a distance of 100 with full battery.
```go
{
    name: "Car can finish the track distance just before battery drained.",
    car: Car{
        speed:        5,
        batteryDrain: 5,
        battery:      100,
        distance:     20,
    },
    track: Track{
        distance: 20,
    },
    expected: true,
},
```

Again, this is all based on my assumption that with each Drive(), the car covers Car.speed distance.

Note 1:
In the meta implementation of the solution, Car.distance is used instead of Car.speed.

```go
	maxDistance := car.battery / car.batteryDrain * car.distance
	return track.distance <= maxDistance
}
```

Even if my assumption about what is expected in Drive() is wrong, the test case still doesn't cover the expected scenario where track.distance == maxDistance.
The maxDistance  will then be 400 and that should the track.distance in the test case.

Note 2:
If the meta implementation is the correct one, then the description here is possibly incorrect.
"Cars start with full (100%) batteries. Each time you drive the car using the remote control, it covers the car's speed in meters and decreases the remaining battery percentage by its battery drain."

Please let me know if this clarifies things. I am sorry for not including all the details in the pull request when I opened it.

vissree:
Understood, thank you!

The main concern I have (the reason I opened the PR at the first place) is that my solution was wrong, but then the test case didn't catch it. i.e. when the `track.distance == car.maxdistance` (maxdistance in the literal sense here) my implementation gives the wrong result, but there wasn't a test case to catch that.
And among the existing test cases, I thought `Car can finish the track distance just before battery drained` is the one that should have covered the scenario.

Or put it in a different way,
```go
	tries := track.distance / car.speed
	return car.battery-(tries*car.batteryDrain) > 0
}
```

and 

```go
	tries := track.distance / car.speed
	return car.battery-(tries*car.batteryDrain) >= 0
}
```

produces the same test results with the current test cases.

Issue #1605, opened 2021-09-15
concepts: review links


andrerfcsantos:
On a concept page, there's the "Learn More" section that contains links related to the concept. I found some problems in these links and we should do a pass to review them:

* The link naming convention is inconsistent across concepts. Some concepts capitalize each word and put spaces between them, others use snake case, others use kebab case. I think we use a naming convention that is consistent.
* Some link names are not descriptive enough of their content. If a link points to the section of maps in the Tour of Go, the link should be called `Tour of Go: Maps` instead of just `maps` or something similar.
* Some concepts don't have any links.

andrerfcsantos:
I can take care of this one - just waiting for all the concept exercises to settle down a bit before looking at this.

junedev:
I noticed this as well, thanks for creating the issue. Regarding the target naming convention I would recommend just what you suggested in the second point: `Source: Title`

amarlearning:
Hi, I was going through the slice documentation of Go Language [https://exercism.org/tracks/go/concepts/slices](https://exercism.org/tracks/go/concepts/slices). The `Learn more` text at the end of the document doesn't have any link.

![Screenshot from 2021-10-05 21-28-25](https://user-images.githubusercontent.com/9383897/136077945-e9bb4690-6c9b-4952-a584-2ece52cda3c9.png)

PR #1836, opened 2021-10-10
Update reference to Slices in Go

orsenthil:
### Update reference to Slices in Go

Part of #1605


andrerfcsantos:
A link to https://github.com/golang/go/wiki/SliceTricks would also be helpful.

PR #1841, opened 2021-10-11
interest-is-interesting: docstring update

sbromberger:
Update docstrings to reflect actual functions.


junedev:
FYI I added the same fix for the exemplar file as well. If you post more of those PRs, please look in both places. Usually they have the same errors.

PR #1843, opened 2021-10-11
Fix typo in code example

hepranputra:
That small mistake confused at first. I proposed the output is 14, not 7.

### DIFF
--- a/concepts/conditionals-if/introduction.md
+++ b/concepts/conditionals-if/introduction.md
@@ -39,7 +39,7 @@ if v := 2 * num; v > 10 {
 } else {
     fmt.Println(num)
 }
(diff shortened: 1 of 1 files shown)


hepranmarunduri:
> good catch!

Hehehe...

> @hepranmarunduri Could you please check if this issue also needs to be fixed in the introduction.md of the concept and the introduction.md of the cars-assemble exercise?

I could.
I may be confused where to find those files, but I will find it and I will fix it.

junedev:
@hepranmarunduri Could you please include the other two fixes here in this PR so they will be merged together? (You can do that by editing the files in the `patch-1` branch.)

PR #1845, opened 2021-10-13
Use more recognizable key value to demonstrate range iteration over map.

orsenthil:
Using 0, 1, and 2 confuses the first timer with the index values.


orsenthil:
@junedev - Sounds good. I will make the changes in other places. Thank you.

PR #1854, opened 2021-10-14
Update interest function output from instructions

SFEstrada:


junedev:
Thanks, but this was already fixed in this PR: https://github.com/exercism/go/pull/1841

PR #1866, opened 2021-10-17
Changed "negative" interest rate into positive.

### LINKED ISSUE #1840: Interest-is-interesting: negative interest rate on negative balances?
> In this exercise you'll be working with savings accounts. Each year, the balance of your savings account is updated based on its interest rate. The interest rate your bank gives you depends on the amount of money in your account (its balance):
>  * -3.213% for a negative balance.

I think this is ambiguous. Does this mean that if I have a negative balance of, say, ($100,000), then after one year I earn $3,213 on it, bringing my balance to ($96,787)?

If so, please name the bank that will let me take out a billion dollar loan on these terms :)

(I realize that the intent is probably to increase the size of the debt, but the way it's presented isn't clear.)

Dersivative:
Hi,
This pretty much fixes #1840 - negative rate on negative balance implies earning on it, whilst tests are constructed so they expect a loss.
It's confusing, forces the learner to add unnecessary "if" and as someone who deals with bank/broker statements professionally I can say you never see rates represented this way for negative balances.

I'd appreciate suggestions regarding the tests naming conventions - "[...] negative interest rate" did not fit well, since there's no negative rate to begin with.

Many thanks!

### DIFF
--- a/exercises/concept/interest-is-interesting/.docs/hints.md
+++ b/exercises/concept/interest-is-interesting/.docs/hints.md
@@ -8,7 +8,8 @@
 ## 2. Calculate the interest
 
 - When calculating interest, it might be helpful to notice that `InterestRate` returns a percentage.
-  One could use arithmetic here, or one of the methods in the [`math` package](https://pkg.go.dev/math).
+- You can use the function defined in the previous task (`InterestRate`) to calculate the interest.
+- You might need to convert the result `InterestRate` to a `float64`.
 
 ## 3. Calculate the annual balance update
 


andrerfcsantos:
I suggested some changes.

Also pinging @ErikSchierboom as the original author of the exercise in the C# track. @ErikSchierboom what do you think of these changes? Are they useful for the C# track too?

> Thanks, I totally missed that one. As for the #1870, I presume that at some point there will be a batch filename update for all exercises. Or should I change the filenames now?

If that PR is approved in its current state, it will change the filenames, the name of the package and the configuration files that also refer to the files. It's probably best to wait until #1870 gets merged and then apply your changes on top. This will result in conflicts with the main branch that must be solved, but we can deal with those later and we can help you solve those if you don't feel comfortable with solving git conflicts.
If you want to make the changes the other PR is making to this exercise now to avoid conflicts later, I'd just suggest making them in a separate commit just so they can be easily reversed if needed. But it's probably best to just wait.

Dersivative:
Caught another file I missed to update (hints.md) and I can't help but notice that merging this PR would mean no more opportunity to explore `math` package, as opposed to current exemplary solution with `math.abs()`.

junedev:
@Dersivative Regarding the math package, we can take care of that later. Just create a new issue saying to add a task to this exercise so the student practices using the math package.

ErikSchierboom:
@andrerfcsantos I like it! I've opened issues at the C# and F# repositories.

PR #1871, opened 2021-10-17
Make CI not ignore some exercises tests if file is not found

### LINKED ISSUE #1870: CI is silently skipping tests for exercises without a correctly named solution file
I noticed the CI tests were passing even when there were errors in some PR's.  I found the root of the problem to be CI silently ignoring some tests if it didn't find the solution file it expected.

If an exercise is in a folder `something-something`, CI expects a `something_something.go` file for the solution. Not all exercises in this repo follow this convention, and for those the tests are effectively being ignored.

As a concrete example of this, the exercise `interesting-is-interesting` has a `savings_account.go` file where the students must put the solution. Since the file is not named `interesting_is_interesting.go`, its tests are being ignored. This can be seen in [this execution example](https://pipelines.actions.githubusercontent.com/5AfNBF8j66vMo76tKsrtvdYzwQWP2AXsE4BmuxoVRYzUsORA8o/_apis/pipelines/1/runs/1740/signedlogcontent/3?urlExpires=2021-10-17T10%3A49%3A46.7723612Z&urlSigningMethod=HMACV1&urlSignature=NgRAR2NSJceqNLrqi5wsucXQwD06qvQIlBsJsg8rL%2B4%3D) where a search for "interesting" returns nothing.

As I see it, this issue requires 2 tasks:

* Make CI return an error if the solution file doesn't have the expected name
* Fix the exercises for which the folder name doesn't match the expected solution file name. The solution file name should be the same as the name of the folder, with all the `-` replaced by `_` and adding a `.go` extension at the end.

andrerfcsantos:
Fixes #1870


andrerfcsantos:
This PR:

* Makes CI throw an error when the files for the solution don't have the expected names, instead of silently skipping them (and consequently skipping the tests).

When fixing this, the exercises which were being skipped previously had small errors, which were also fixed in this PR, most notably:

* `lasagna_master_test.go` was not properly formatted with `gofmt -s`, which made gocritic complain.
* In the `exemplar.go` for `logs-logs-logs` there was a variable `new`, which was shadowing the bultin `new` function, which also made gocritic complain.

PR #1879, opened 2021-10-20
lasagna: argument clarity

kirkkaf13:


andrerfcsantos:
@kirkkaf13 It was a good first PR!

It's normal to miss something sometimes, but we are here to help you fix it!

If you are thinking in contributing more, I recommend the [building docs](https://exercism.org/docs/building) so you are more comfortable on how exercism works.

PR #1880, opened 2021-10-20
Fix typo

panchoh:
According to the exercise's notes, the ranks are named "A" trough "H".
The tests are coded with that convention in mind.
As the comment is now, the type for the Chessboard would have to be `map[int]Rank`, which clashes with both the tests and the notes.


andrerfcsantos:
Thanks for this! Left a minor suggestion.

The [exemplar solution](https://github.com/exercism/go/blob/main/exercises/concept/chessboard/.meta/exemplar.go) also has this error. Can you change there too?

PR #1916, opened 2021-10-29
Update scrabble_score.go

ysharlaimov:
Add function stub with expected name and signature

### DIFF
--- a/exercises/practice/scrabble-score/scrabble_score.go
+++ b/exercises/practice/scrabble-score/scrabble_score.go
@@ -1,5 +1,5 @@
 package scrabble
 
 func Score(word string) int {
-  panic("implement me")
+  panic("Please implement the Score function")
 }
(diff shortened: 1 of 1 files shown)


exercises/practice/scrabble-score/scrabble_score.go
```diff
@@ -1 +1,5 @@
 package scrabble
+
+func Score(word string) int {
+  panic("implement me")
```
andrerfcsantos, line 4:
```suggestion
  panic("Please implement the Score function")
```

A suggestion to make this comment similar to the other ones we have on the stubs.

andrerfcsantos (changes requested):
Thanks for this! Just left a minor siggestion.

andrerfcsantos (approved):
Thanks for this @ysharlaimov.

I went ahead and applied the suggested change and merged this in order to avoid possible conflicts for someone working on adding these stubs.

Issue #1893, opened 2021-10-23
Add stub functions for practice exercises


junedev:
### Issue
Currently most practice exercises have empty stub files. That means the student always has to start the task with guessing the function signature by looking at the test file. That is frustrating for students, does not teach a valuable skill and is not the point of the exercise. 

Examples of complains: https://github.com/exercism/go/issues/834, https://github.com/exercism/go/issues/842, https://github.com/exercism/go/issues/1337, ...

### Task
For now, we want to add stub functions to the stub files (`<exercise>.go`) at least for all practice exercises that have a small API (1-3 functions/methods/structs). Exercises with larger APIs can also be tackled later as they are more involved sometimes. E.g. they might contain the API information in the instructions.append.md file or in the test file so we need to make some more changes to avoid duplication.

#### Example of the change that should be made
Before:
```go
package binary
```

After:
```go
package binary

func ParseBinary(binary string, expected int, ok bool) {
	panic("please implement the ParseBinary function")
}
```

#### Notes
- Structs/Type definitions should only be part of the stub if their form is forced by the tests. Otherwise there should be a comment for the student to implement the type.
- The correct stub functions/stubs can be determined the easiest way by looking at the the .meta/example.go file. There you can copy the relevant parts.
- If you are not familiar with the anatomy of a practice exercise, you can find the general documentation here: https://exercism.org/docs/building/tracks/practice-exercises

### Exercises already done

- Allergies https://github.com/exercism/go/pull/1896
- Binary search tree https://github.com/exercism/go/pull/1894

junedev:
@PakkuDon Could you let us know whether you plan to work on the rest of the exercises as well or whether we should open this tasks up for other contributors?

PR #1921, opened 2021-11-01
Add stub functions to practice exercises Q-S

PakkuDon:
Related to #1893 

We would like to add stub functions to the practice exercises so that students know what API they are expected to implement. We would also like to remove any API descriptions from test files as these may go out of sync with the code over time

This PR adds stubs for practice exercises starting with letters Q through to S and updates the associated test files accordingly

The following practice exercises have been omitted because they already have stubs defined:
- [raindrops](https://github.com/exercism/go/blob/main/exercises/practice/raindrops/raindrops.go)
- [rna-transcription](https://github.com/exercism/go/blob/main/exercises/practice/rna-transcription/rna_transcription.go)
- [scrabble-score](https://github.com/exercism/go/blob/main/exercises/practice/scrabble-score/scrabble_score.go)
- [space-age](https://github.com/exercism/go/blob/main/exercises/practice/space-age/space_age.go)

### DIFF
--- a/exercises/practice/queen-attack/queen_attack.go
+++ b/exercises/practice/queen-attack/queen_attack.go
@@ -1,5 +1,5 @@
 package queenattack
 
-func CanQueenAttack(whiteQueen, blackQueen string) (bool, error) {
+func CanQueenAttack(whitePosition, blackPosition string) (bool, error) {
 	panic("Please implement the CanQueenAttack function")
 }
--- a/exercises/practice/queen-attack/queen_attack.go
+++ b/exercises/practice/queen-attack/queen_attack.go
@@ -1,1 +1,5 @@
 package queenattack
+
+func CanQueenAttack(whiteQueen, blackQueen string) (bool, error) {
+	panic("Please implement the CanQueenAttack function")
+}
--- a/exercises/practice/react/react.go
+++ b/exercises/practice/react/react.go
@@ -1,6 +1,7 @@
 package react
 
 // Define reactor, cell and canceler types here.
+// These types will implement the Reactor, Cell and Canceler interfaces, respectively.
 
 func (c *canceler) Cancel() {
 	panic("Please implement the Cancel function")
--- a/exercises/practice/react/react.go
+++ b/exercises/practice/react/react.go
@@ -1,1 +1,35 @@
 package react
+
+// Define reactor, cell and canceler types here.
+
+func (c *canceler) Cancel() {
+	panic("Please implement the Cancel function")
+}
+
+func (c *cell) Value() int {
+	panic("Please implement the Value function")
+}
+
+func (c *cell) SetValue(value int) {
+	panic("Please implement the SetValue function")
+}
+
+func (c *cell) AddCallback(callback func(int)) Canceler {
+	panic("Please implement the AddCallback function")
+}
+
+func New() Reactor {
+	panic("Please implement the New function")
+}
+
+func (r *reactor) CreateInput(initial int) InputCell {
+	panic("Please implement the CreateInput function")
+}
+
+func (r *reactor) CreateCompute1(dep Cell, compute func(int) int) ComputeCell {
+	panic("Please implement the CreateCompute1 function")
+}
+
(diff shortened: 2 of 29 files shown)


exercises/practice/queen-attack/queen_attack.go
```diff
@@ -1 +1,5 @@
 package queenattack
+
+func CanQueenAttack(whiteQueen, blackQueen string) (bool, error) {
```
PakkuDon, line 3: Wasn't sure what these parameters were supposed to represent. The example solution has the following:
https://github.com/exercism/go/blob/67cb0d27a99524dec7b8467f2b825a061e7a652c/exercises/practice/queen-attack/.meta/example.go#L5

Is there a more accurate name we could use?

exercises/practice/robot-simulator/robot_simulator.go
```diff
@@ -1 +1,42 @@
 package robot
+
+// See defs.go for other definitions
+
```
PakkuDon, line 4: robot-simulator has three different test files: (eg: [robot_simulator_step2_test.go](https://github.com/exercism/go/blob/main/exercises/practice/robot-simulator/robot_simulator_step2_test.go))

Should we add comments or something (eg: `// Step 1`) to signpost which parts of the stub file are linked to which test file?

exercises/practice/queen-attack/queen_attack.go
andrerfcsantos: Maybe `whitePosition` and `blackPosition`, since the strings are the positions of the queens?

exercises/practice/react/react.go
```diff
@@ -1 +1,35 @@
 package react
+
+// Define reactor, cell and canceler types here.
+
```
andrerfcsantos, line 4:
```suggestion
// Define reactor, cell and canceler types here.
// These types will implement the Reactor, Cell and Canceler interfaces, respectively.

```
I think this bit of clarification is important.

exercises/practice/robot-simulator/robot_simulator.go
andrerfcsantos: I think it is a good idea.

It's also probably worth to move the instructions that are in the test files to an `instructions.append.md` file, for better visibility.

exercises/practice/saddle-points/saddle_points.go
```diff
@@ -1 +1,11 @@
 package matrix
+
+// Define the Matrix and Pair type here.
```
andrerfcsantos, line 3:
```suggestion
// Define the Matrix and Pair types here.
```

exercises/practice/series/series_test.go
```diff
@@ -1,32 +1,3 @@
-// Define two functions:  (Two? Yes, sometimes we ask more out of Go.)
```
andrerfcsantos, line 1: We should probably move some parts of this to an `instructions.append.md` file.

andrerfcsantos (changes requested):
Thanks once again for doing this! Left some suggestions, feel free to discuss them if you don't agree.

exercises/practice/queen-attack/queen_attack.go
PakkuDon: I like those names better 👍 
Writing this as a suggestion so I can add it to the batch
```suggestion
func CanQueenAttack(whitePosition, blackPosition string) (bool, error) {
```

exercises/practice/series/series_test.go
PakkuDon: I forgot to mention that this exercise already has an [instructions append](https://github.com/exercism/go/blob/main/exercises/practice/series/.docs/instructions.append.md) with the same content as this comment. Sorry for the trouble

exercises/practice/robot-simulator/robot_simulator.go
PakkuDon: Good call! I'll move them across

andrerfcsantos (approved):

PR #1922, opened 2021-11-01
Added quotes to strings

### LINKED ISSUE #1920: atbash-cipher: quote strings in test output
Right now the tests for the `atbash-cipher` exercise are printing the input and output strings as is:

```go
// snippet of atbash_cipher_test.go
if actual != test.expected {
	t.Errorf("Atbash(%s): expected %s, actual %s", test.s, test.expected, actual)
}
```

This can make comparisons of the `test.expected` and `actual` strings a bit difficult, specially if the strings have trailing or leading spaces. This is relevant in this exercise since it requires building a string with spaces in it, and it's very easy for an almost-correct solution to accidentally add a trailing space. For such solutions, debugging the problem from the test outputs can be very hard, here's an example:

```console
$ go test
--- FAIL: TestAtbash (0.00s)
    atbash_cipher_test.go:9: Atbash(OMG): expected lnt, actual lnt 
```

## Tasks

- [ ] Update [atbash_cipher_test.go](https://github.com/exercism/go/blob/main/exercises/practice/atbash-cipher/atbash_cipher_test.go) so the output of the tests is more clear. A suggestion would be to quote the `%s` strings to make leading and trailing spaces more evident.

arjunraghurama:
Wrapped `%s` with single quotes `'%s'`.
Fixes #1920

### DIFF
--- a/exercises/practice/atbash-cipher/atbash_cipher_test.go
+++ b/exercises/practice/atbash-cipher/atbash_cipher_test.go
@@ -6,7 +6,7 @@ func TestAtbash(t *testing.T) {
 	for _, test := range tests {
 		actual := Atbash(test.s)
 		if actual != test.expected {
-			t.Errorf("Atbash(%s): expected %s, actual %s", test.s, test.expected, actual)
+			t.Errorf("Atbash('%s'): expected '%s', actual '%s'", test.s, test.expected, actual)
 		}
 	}
 }


andrerfcsantos (approved):

andrerfcsantos:
Thanks for this 👍

PR #1923, opened 2021-11-01
simple-linked-list: Add size test cases for Push and Pop methods

endlesslycurious:
It was possible to have an implementation for `simple-linked-list` that passed the existing tests but hung on the benchmarks; as Size was not verified with Push and Pop.  This PR adds individual Size tests for Push and Pop.

This worked locally for me via the exercise files from the Exercism CLI but I couldn't get `test-without-stubs` to run to verify it in the repro as it gave the following error which seemed unrelated:
`Missing examplar in exercises/concept/*/.meta called exemplar.go or example.go`

This was with go version 1.17.2 darwin/amd64 on MacOS 12.0.1.


exercises/practice/simple-linked-list/simple_linked_list_test.go
```diff
@@ -51,6 +51,30 @@ func TestNonEmptyListHasCorrectSize(t *testing.T) {
...
 
+func TestListHasCorrectSizeAfterPop(t *testing.T) {
+	list := New([]int{1, 2, 3})
+
+	_, _ = list.Pop()
+	_, _ = list.Pop()
+	_, _ = list.Pop()
+
+	if size := list.Size(); size != 0 {
+		t.Fatalf("Size of list from [1, 2, 3] after pop(), pop(), pop(): %d, expected: %d", size, 0)
```
andrerfcsantos, line 62:
```suggestion
		t.Fatalf("Size of list from [1, 2, 3] after 3 calls to pop(): got %d, expected: %d", size, 0)
```

exercises/practice/simple-linked-list/.meta/config.json
```diff
@@ -25,4 +26,4 @@
   },
   "source": "Inspired by 'Data Structures and Algorithms with Object-Oriented Design Patterns in Ruby', singly linked-lists.",
   "source_url": "https://web.archive.org/web/20160731005714/http://brpreiss.com/books/opus8/html/page96.html"
-}
+}
```
andrerfcsantos, line 29:
```suggestion
}

```

JSON files must end in a new line.

exercises/practice/simple-linked-list/simple_linked_list_test.go
```diff
@@ -51,6 +51,30 @@ func TestNonEmptyListHasCorrectSize(t *testing.T) {
...
+
+func TestListHasCorrectSizeAfterPush(t *testing.T) {
+	list := New([]int{})
+
+	list.Push(1)
+	list.Push(2)
+	list.Push(3)
+
+	if size := list.Size(); size != 3 {
+		t.Fatalf("Size of list from [] after push(1) push(2), push(3): %d, expected: %d", size, 3)
```
andrerfcsantos, line 74:
```suggestion
		t.Fatalf("Size of list from [] after 3 calls to push(): got %d, expected: %d", size, 3)
```

andrerfcsantos (changes requested):
Thanks for bringing this issue to our attention. Left some minor suggestions.

> It was possible to have an implementation for simple-linked-list that passed the existing tests but hung on the benchmarks; as Size was not verified with Push and Pop. This PR adds individual Size tests for Push and Pop.

For the sake of completeness, can you show an example of a solution that passes the tests but hungs up the benchmarks?

exercises/practice/simple-linked-list/.meta/config.json
endlesslycurious: Ah, that seems to have been the VSCode auto-formatter

andrerfcsantos:
Before merging this, can you post the solution that passed the previous tests but not this version of the tests?

endlesslycurious:
@andrerfcsantos It took me a while to type this up and double check.

What seems to have happened is I wrote the `New` factory function without using the `Push` method, my `New` method was correct but `Push` and `Pop` didn't modify size. As the [tests](https://github.com/exercism/go/blob/main/exercises/practice/simple-linked-list/simple_linked_list_test.go) use a combination of `New` to seed test data then `Array` to export the contents to compare against the expected results it was possible for my bad `Push` and `Pop` methods to pass the tests as the exported Array will be correct but it will hang in the [BenchmarkListPop](https://github.com/exercism/go/blob/6f0d7c87fee11797ee8afd30780374ca64346d5b/exercises/practice/simple-linked-list/simple_linked_list_test.go#L195) benchmark.

This is the code that would hang, the fixes to the size are commented out:
```
// func Pop removes and returns last value from list
func (l *List) Pop() (int, error) {
	var res int

	if l.size == 0 {
		return 0, errors.New("list is empty")
	}

	var prev *Element
	for element := l.head; element != nil; element = element.next {
		if element.next == nil {
			res = element.value

			if prev != nil {
				prev.next = nil
			}

			// l.size--
			break
		} else {
			prev = element
		}
	}

	return res, nil
}

// func push adds new element to end of list
func (l *List) Push(val int) {

	if l.head == nil {
		l.head = &Element{nil, val}
		// l.size++
	} else {
		for element := l.head; element != nil; element = element.next {
			if element.next == nil {
				pushed := Element{nil, val}
				element.next = &pushed
				// l.size++
				break
			}
		}
	}
}
```

The hang seems to happen when Push doesn't increment size, if I recall correctly there is an issue with bench's auto-scaling when applied to very fast calls, see [stack overflow](https://stackoverflow.com/a/37624250) and [Go Issue 27217](https://github.com/golang/go/issues/27217). I suspect the `l.size == 0` error case in `Pop` is short circuiting the method (both the popped value and error are not checked in the benchmark) and that's tormenting the auto-scaling during `BenchmarkListPop`

```
func BenchmarkListPop(b *testing.B) {
	if testing.Short() {
		b.Skip("skipping benchmark in short mode.")
	}
	for i := 0; i < b.N; i++ {
		b.StopTimer()
		list := New([]int{})
		for k := 0; k < 1000; k++ {
			list.Push(k)
		}
		b.StartTimer()
		for k := 0; k < 1000; k++ {
			_, _ = list.Pop()
		}
	}
}
```

andrerfcsantos (approved):
I see, thanks for the explanation.

endlesslycurious:
@andrerfcsantos I added my handle to the contributors list in `config.json` but I don't see it reflected on the [exercise's page](https://exercism.org/tracks/go/exercises/simple-linked-list) but I did see the exercise files update though. Does the contributors list get regenerated periodically?

andrerfcsantos:
@endlesslycurious Looks like your Github account is not linked to your exercism account. there's currently an issue about this: https://github.com/exercism/exercism/issues/5608

PR #1933, opened 2021-11-03
comments have sorted according to exercise order

eren-kara:


andrerfcsantos (approved):

andrerfcsantos:
Thanks for this!

junedev (commented):
It would be great if you could apply the same change for the .meta/exemplar.go file.

junedev (approved):

junedev:
Looked at good now, thanks for the fix!

PR #1935, opened 2021-11-03
Fix embedded mutex in `bank-account` example

andrerfcsantos:
This is causing CI jobs to fail because of [this check](https://go-critic.com/overview#exposedSyncMutex-ref).


junedev (approved):
Thanks!

PR #1937, opened 2021-11-03
orders of functions changed

eren-kara:


junedev:
@eren-kara Please include this change in your existing PR #1933. You can do that in the GitHub UI by selecting the patch-1 branch in your fork, editing the exemplar file their and then selecting "commit to existing branch" instead of creating a new branch.

PR #1939, opened 2021-11-04
error-handling: ensure `Close` is called only once

### LINKED ISSUE #1967: error-handling: ensure `Close` is called only once
Related to #1939 .

The exercise requires the usage of `defer` (the code can panic). This in turn raises a questions about some inner workings of `defer`. 

Answering to the comment in PR: https://github.com/exercism/go/pull/1939#issuecomment-961216677
> Is there any motivation behind this? Did you see any particular solution calling Close() more than once?

In one of the solutions I have mentored there was something like this:
```
...
defer r.Close()
defer func() {
   if recover() != nil {
      ...
      r.Close()
   }
}
```

> I don't believe this adds much or solves any problem with the exercise.

1. Knowing how multiple `defer`s are executed in case one of them handles panic with `recover` is something worth learning IMO.
2. If `r.Close` closes some channel we want to be sure it is called exactly once. So I think it is a reasonable requirement.

fyrchik:
Fixes #1967

I think it is a nice thing to check.

### DIFF
--- a/exercises/practice/error-handling/.docs/instructions.md
+++ b/exercises/practice/error-handling/.docs/instructions.md
@@ -37,4 +37,4 @@ There will be a few places in your Use function where errors may occur:
   function_. Either way `Use` should return the error.
 
 - _Also note_: if the Resource was opened successfully make sure to call its
-  Close function no matter what (even if errors have occurred).
+  Close function exactly once no matter what (even if errors have occurred).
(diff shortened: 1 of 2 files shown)


andrerfcsantos:
Is there any motivation behind this? Did you see any particular solution calling `Close()` more than once?

I appreciate you taking the time to submit a PR, but for things like this is probably better if you open an issue first if you find something we can improve. That way we can discuss the change first.

I don't believe this adds much or solves any problem with the exercise.

junedev:
I am closing this as it seems abandoned. Happy to re-open if the open questions are addressed.

andrerfcsantos (approved):
Thanks for this!

Issue #1925, opened 2021-11-02
Phone Number: Set up instructions append


junedev:
The Go version of the [phone-number](https://github.com/exercism/go/blob/main/exercises/practice/phone-number/.docs/instructions.md) practice exercise differs slightly from the original in the problem spec repo: https://github.com/exercism/problem-specifications/blob/main/exercises/phone-number/description.md
In Go, the student needs to write two additional functions.

We want to get the instructions.md file in sync with the problem spec repo again. Therefore the following tasks need to be done:
- [ ] Copy the content of the problem spec description file (link above) into the instructions.md file of the exercise.
- [ ] Add an instructions.append.md file that contains instructions for the two additional functions `AreaCode` and `Format`. The content of this file will be rendered right below the instructions later on.

You can find general information about the structure of practice exercises here: https://exercism.org/docs/building/tracks/practice-exercises
 
Sidenote: This issue originates from the discussion here https://github.com/exercism/go/issues/1915.

PR #1943, opened 2021-11-06
phone-number: move go specific instruction to instructions.append.md (#1925)

kahgoh:
This is for #1925

This is to sync the instructions.md to back up to the problem specification.


junedev (approved):

PR #1944, opened 2021-11-08
Fix typo in concepts/pointers

greenren:


andrerfcsantos (approved):
Awesome! Thanks for this and for taking extra care to change the typo in all places.

PR #1947, opened 2021-11-08
Annalyns infiltration: test updates

TheArchDev:


andrerfcsantos:
Thanks for your interest in contributing. What motivated this test update?

TheArchDev:
> Thanks for your interest in contributing. What motivated this test update?

Thanks for the comment! Apologies for missing those details.

For the test class of `TestCanSpy`, there are 8 possible input parameter combinations and only 6 were covered in the test classes, and both of the missing combinations were focused on the state of `archerIsAwake`, and so I added them.

For the test class of `TestCanFreePrisoner`, there are 16 possible input parameter combinations and only 14 were covered in the test classes. Of the 16, only 5 result in `CanFreePrisoner` returning True, and one of the missing test cases was one of those 5 True scenarios.

andrerfcsantos (approved):
Thanks for this.

PR #1948, opened 2021-11-08
Interest-is-Interesting: Fix float comparison

### LINKED ISSUE #1842: Interest is interesting:  Fix float comparison logic
The `Interest()` function unfortunately requires that the rates be in percentage, not decimal (i.e., "three percent" is `3.0`, not `0.03`), and in `float32` (as opposed to `float64`). This has the additional unfortunate consequence of introducing the student unfamiliar with IEEE 754 to the fact that 
` n * float64(r) / 100` may not equal `n * float64(r / 100)`. (See https://play.golang.org/p/148rIMlhIHQ for the demonstration.)

It is almost never a good idea to perform exact equality comparisons on floating point numbers due to float imprecision. In this case, the former calculation is what the test suite expects; the latter fails, even though the difference is within some small epsilon in every case.

I would suggest modifying the test framework to check to see whether `math.Abs(expected - actual) < epsilon` instead of `expected == actual`.

(Edited to add: this checking against epsilon is precisely what we're doing in the Space Age exercise.)

Satho-M:
Fixes #1842 
Changed the comparison between floats to work with bigger values, as suggested by @junedev 

Also found the following options that ended up being more strict:
```go
(math.Abs(x - y) <= epsilon * math.Max(math.Abs(x), math.Abs(y)))
```

And the following one that accepted two different thresholds:
```go
((math.Abs(x-y) <= epsilonTol) || (math.Abs(x-y) <= epsilonRel*math.Max(math.Abs(x), math.Abs(y))))
```

All of the options above also worked with the tests on the Interest-is-Interesting exercise and fixed the issue, but june's suggestion accepted a wider range of values.
The solutions above were found here: http://realtimecollisiondetection.net/blog/?p=89

### DIFF
--- a/exercises/concept/interest-is-interesting/interest_is_interesting_test.go
+++ b/exercises/concept/interest-is-interesting/interest_is_interesting_test.go
@@ -8,8 +8,9 @@ import (
 const floatEqualityThreshold = 1e-5
 
 func floatingPointEquals(got, want float64) bool {
-	return math.Abs(got-want) <= floatEqualityThreshold ||
-		math.Abs(got-want)/(math.Abs(got)+math.Abs(want)) <= floatEqualityThreshold
+	absoluteDifferenceBelowTreshold := math.Abs(got-want) <= floatEqualityThreshold
+	relativeDifferenceBelowTreshold := math.Abs(got-want)/(math.Abs(got)+math.Abs(want)) <= floatEqualityThreshold
+	return absoluteDifferenceBelowTreshold || relativeDifferenceBelowTreshold
 }
 
 func TestInterestRate(t *testing.T) {
--- a/exercises/concept/interest-is-interesting/interest_is_interesting_test.go
+++ b/exercises/concept/interest-is-interesting/interest_is_interesting_test.go
@@ -8,7 +8,8 @@ import (
 const floatEqualityThreshold = 1e-5
 
 func floatingPointEquals(got, want float64) bool {
-	return math.Abs(got-want) <= floatEqualityThreshold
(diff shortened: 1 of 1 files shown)


andrerfcsantos (approved):
Thanks for fixing this!

Looks good to me. WDYT @junedev?

exercises/concept/interest-is-interesting/interest_is_interesting_test.go
```diff
@@ -8,7 +8,8 @@ import (
 const floatEqualityThreshold = 1e-5
 
 func floatingPointEquals(got, want float64) bool {
-	return math.Abs(got-want) <= floatEqualityThreshold
+	return math.Abs(got-want) <= floatEqualityThreshold ||
+		math.Abs(got-want)/(math.Abs(got)+math.Abs(want)) <= floatEqualityThreshold
```
junedev, line 12: I would recommend to improve readability by creating two variables, "absoluteDifferenceBelowTreshold" and "relativeDifferenceBelowTreshold" or similar and then using those in the actual return statement with the `||` operator.

junedev (approved):
I would also be ok with one of the other formulas you presented but I think the current solution you have is fine for the purpose. I just made a suggestion regarding readability. (Mentoring feedback if you will. :wink:)

exercises/concept/interest-is-interesting/interest_is_interesting_test.go
Satho-M: Sent the version with two variables :) Thank you, mentoring feedback is exactly what I'm looking for.

PR #1949, opened 2021-11-09
party-robot: clarify difference between Println and Printf

TheArchDev:
This additional comment in the Introduction section of the party-robot exercise aims to capture an important difference in usage of fmt.Println and fmt.Printf.

### DIFF
--- a/exercises/concept/party-robot/.docs/introduction.md
+++ b/exercises/concept/party-robot/.docs/introduction.md
@@ -42,4 +42,4 @@ fmt.Sprintf("%.2f", number)
 // Returns: 4.32
 ```
 
-`fmt` contains other functions for working with strings, such as `Println` which simply prints the arguments it receives to the console and `Printf` which formats the input in the same way as `Sprintf` before printing it.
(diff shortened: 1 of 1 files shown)


exercises/concept/party-robot/.docs/introduction.md
```diff
@@ -42,4 +42,4 @@ fmt.Sprintf("%.2f", number)
 // Returns: 4.32
 ```
 
-`fmt` contains other functions for working with strings, such as `Println` which simply prints the arguments it receives to the console and `Printf` which formats the input in the same way as `Sprintf` before printing it.
+`fmt` contains other functions for working with strings, such as `Println` which simply prints the arguments it receives to the console and `Printf` which formats the input in the same way as `Sprintf` before printing it. Note that `Println` has built-in whitespace and newline logic when printing output to the console, whereas `Printf` does not.
```
junedev, line 45: This is difficult to understand for someone who doesn't know what that logic is already. I would recommend concretely list the main differences (maybe with a bullet list even).

Also please put every new sentence on a new line (it looks the same when rendered but results in better diffs e.g. here for the reviews.)

junedev (commented):
I agree it would be nice to explain the difference. I left an idea for improving the sentence for that.

Make sure to apply the change to all the files mentioned in the automated PR comment.

junedev:
I am closing this as it seems abandoned. Happy to re-open in case it is picked up again.

PR #1951, opened 2021-11-09
Elons toys & Need for Speed

KalebHawkins:
There was no test case to see if the car could be driven if the battery was below the batteryDrain percentage. Added this test case to both the Elon's-Toys and Need for Speed exercises.


junedev (dismissed):

junedev:
Good catch!
You still need to address the CI failure. Probably you need to adjust the exemplar solution to pass your new test.

KalebHawkins:
The examplar.go files were correct. I just misplaced my tests in need for speed. So many brackets... but that is what the tests for the tests are for I guess :D.

junedev (approved):

PR #1952, opened 2021-11-09
party-robot: update tests

TheArchDev:
This change does the following:
* extend an existing testcase to include non-standard ASCII character values for other string parameters as well. Note that this was already being checked for the `name` parameter, but was not checked for the `direction` and `seatmate` parameters.
* add a new testcase which checks additional value scenarios for `tableNumber` (an int parameter) and `distance` (a float64 parameter)

### DIFF
--- a/exercises/concept/party-robot/party_robot_test.go
+++ b/exercises/concept/party-robot/party_robot_test.go
@@ -81,11 +81,11 @@ func TestAssignTable(t *testing.T) {
 		{
 			description: "Greet Xuân Jing and give them directions to their seat",
 			name:        "Xuân Jing",
-			direction:   "on the right",
+			direction:   "by the façade",
 			tableNumber: 4,
 			distance:    23.470103,
-			seatmate:    "Eike",
-			want:        "Welcome to my party, Xuân Jing!\nYou have been assigned to table 004. Your table is on the right, exactly 23.5 meters from here.\nYou will be sitting next to Eike.",
+			seatmate:    "Renée",
+			want:        "Welcome to my party, Xuân Jing!\nYou have been assigned to table 004. Your table is by the façade, exactly 23.5 meters from here.\nYou will be sitting next to Renée.",
 		},
 	}
 	for _, tt := range tests {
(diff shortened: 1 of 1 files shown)


junedev:
I am sorry, I have trouble imagining why these additional cases are necessary. Could you show me a student solution that would pass the old set of tests but would fail your new version?

TheArchDev:
> I am sorry, I have trouble imagining why these additional cases are necessary. Could you show me a student solution that would pass the old set of tests but would fail your new version?

Hi @junedev, thank you for the review! I believe something like the code below would fail on the new/3rd test case within the `TestAssignTable` tests, on both the displayed table number and distance.
I think this new test class could help make sure that if someone _didn't_ use the helpful built-in formatting logic of the `fmt` package, that they didn't miss any edge cases for formatting `int`s and `float64`s if trying to handle the width and precision logic themselves.

```
package partyrobot

import "fmt"
import "math"

// Welcome greets a person by name.
func Welcome(name string) string {
	return fmt.Sprintf("Welcome to my party, %v!", name)
}

// HappyBirthday wishes happy birthday to the birthday person and exclaims their age.
func HappyBirthday(name string, age int) string {
	return fmt.Sprintf("Happy birthday %v! You are now %v years old!", name, age)
}

// AssignTable assigns a table to each guest.
func AssignTable(name string, table int, neighbor, direction string, distance float64) string {
	message := fmt.Sprintf("Welcome to my party, %v!\n", name)
	var table_output string
	if table < 10 {
		table_output = "00" + fmt.Sprint(table)
	} else {
		table_output = "0" + fmt.Sprint(table)
	}
	distance_output := math.Round(distance*10)/10

	message += fmt.Sprintf("You have been assigned to table %v. Your table is %v, exactly %v meters from here.\n", table_output, direction, distance_output)
	message += fmt.Sprintf("You will be sitting next to %v.", neighbor)
	return message
}

```

In terms of failing on the string fields `direction` or `seatmate`, any solution which would've caused a failure on `name: "Xuân Jing"` should trigger a failure for those now as well.

junedev (approved):

junedev:
Thanks for the explanation regarding the table number. I understand now.

> In terms of failing on the string fields direction or seatmate, any solution which would've caused a failure on name: "Xuân Jing" should trigger a failure for those now as well.

If I understand this correctly, it means the old version of the tests was sufficient to catch the type of error we are aiming for here. I merged your change nevertheless but it would be great if you could refrain from adding such test cases in the future. Maintainers need to invest time to review them but they don't add value for the student.

PR #1953, opened 2021-11-09
Basics concepts - add named and naked return

gilvansfilho:
Add explanation of named and naked return.

### DIFF
--- a/concepts/basics/about.md
+++ b/concepts/basics/about.md
@@ -79,3 +79,27 @@ func hi (name string) string {
     return "hi " + name
 }
 ```
+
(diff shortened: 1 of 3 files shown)


junedev:
I am sorry but this content is not a good fit for `basics`. It is meant to teach only the bare minimum to solve other concept exercises. The content you added is already covered in the concept `functions`.

Please consider opening an issue first next time so you don't have unnecessary work.

gilvansfilho:
Sorry @junedev 
My bad.

I did not see it on functions but really are there.

Issue #1881, opened 2021-10-10
Upgrade to Go 1.16+


rootulp:
My solution to [tournament](https://exercism.org/tracks/go/exercises/tournament) passes locally but fails on exercism.com due to:

```
# tournament [tournament.test]
./tournament.go:88:16: undefined: io.ReadAll
FAIL	tournament [build failed]
'/usr/local/go/bin/go test --short --json .' returned exit code 2: exit status 2
```
 
It looks like `io.ReadAll` was added in Go 1.16 per [this issue](https://github.com/golang/go/issues/44568#issue-815252790). However, the test-runner appears to run with Go 1.15 per https://github.com/exercism/go-test-runner/blob/b6393ab75da7b6bd6d0526e55608da8009f6dc34/go.mod#L3

Should we consider upgrading the test runner to Go 1.16+?

andrerfcsantos:
@junedev What problems do you anticipate from this upgrade?

The way I see it, it's just an upgrade in the Dockerfile of the test runner and with a script we can upgrade all the `go.mod` files to target a specific version. Do you have something else in mind?

As a general strategy, I think we should keep the test runner image always running the latest major release of Go.

Since the `go.mod` file only specifies a minimum version, the version specified there doesn't need to exactly match the one on the test runner. For that minimum version, I suggest it to be at least one version behind from the one on the test runner. The _rationale_ for this is that Go keeps their official support for 2 major releases, the current and the previous one. But we probably should be a more lenient than that and make the minimum version in the `go.mod` file lag more than one version behind. This is because people on linux might be using the go version that comes with their package manager, and sometimes those versions are a bit behind. As a concrete example, right now the latest major version of Go is 1.17, but looking at the [ubuntu packages](https://packages.ubuntu.com/search?keywords=golang) repository, their most recent LTS is focal (20.04) and it's is running go 1.14, which is 3 major versions behind. Of course there are some exercises that can be exceptions. For instance, if one exercise uses a concept introduced in a specific version, its `go.mod` file should be set to that version, regardless of everything else.

[...]

PR #1954, opened 2021-11-10
Add gomod-sync

andrerfcsantos:
Part of #1881 

This PR introduces a new tool to the repo: `gomod-sync`. This is a little program to:

* List `go.mod` files and the Go versions they are declaring via the `gomod-sync list` command.
* Check if all `go.mod` files from the exercises are set to a specific version, with the `gomod-sync check` command. The desired version can be set with the `--goversion` flag.
* Update all `go.mod` files to a specific version with the `gomod-sync update` command. The desired version for the update can be set with the `--goversion` flag.

Right now these 3 commands are working. The `update` command was ran against the repo to update all exercises to 1.16, and the result is https://github.com/exercism/go/pull/1955

## Tasks

- [x] `list`, `check` and `update` commands
- [x] Setup the use of the `check` command in CI
- [x] Documentation for the tool
- [x] Add a way to allow specific go versions for specific exercises
- [x] Tests


andrerfcsantos:
Ready for review 🚀

This one is on the bigger side, so feel free to take your time, there is no rush.

This is what the output of the check looks like when it fails: https://github.com/exercism/go/runs/4662874598?check_suite_focus=true

For the output in case everything is fine, see the `go.mod check` below ⬇️

.github/workflows/gomod-sync.yml
```diff
@@ -0,0 +1,27 @@
+name: go.mod check
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - master
```
junedev, line 8: Do we need this line?

.github/workflows/gomod-sync.yml
```diff
@@ -0,0 +1,27 @@
...
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
+
+      - name: Check go.mod files
+        shell: bash
+        run: |
+          cd gomod-sync && go build
+          ./gomod-sync check --goversion 1.16 --exercises ../exercises
```
junedev, line 27: Couldn't we use `go run` here?

gomod-sync/README.md
```diff
@@ -0,0 +1,101 @@
+# gomod-sync
+
+Utility tool to check and update the Go version specified in the `go.mod` files of all exercises.
+It works by specifying the desired Go version for all the `go.mod` files to be in: the `check` command
```
junedev, line 4: I don't understand the `:` after "to be in".

junedev:
Thanks a lot for taking care of this. I only read the README so far (which is great) and it sounds like it does all the great things we need it to do. However, I am contemplating whether it is possible to reduce the complexity of the tool a bit. You built something quite "general purpose" which is nice but since it also means more code, it means there is more code for future maintainers/contributors to read, understand and maintain. Also the various versions in which the command can be invoked make it a bit harder for users.

Here some concrete points:
* The tool could assume the default Go version is always provided via the config file, then you would not need the additional goversion flag. Currently, it is confusing what happens if the file says one thing and the flag says something else. So one way of doing things would reduce potential problems and complexity. The readme then would only say: "want to configure? change the config file".
* In our context, the folder that contains the exercises and the folder that contains the config will rarely ever change. It would be ok if those were hardcoded or part of the config file so they don't have to be passed when invoking the command. I think your tool is fast enough that we don't need the ability to one check individual exercise folders.
* Do we need the `list` command?

Point 1 and 2 together would then make the invocation as simple as `gomod-sync check` (same for the other commands) which would be a very user friendly API.

If you really want the tool to be used for other Go projects, then it would make more sense to keep in a separate repository and only `go get/install` it here.

gomod-sync/.gitignore
```diff
@@ -0,0 +1,15 @@
...
+*.dylib
+
+# Test binary, built with `go test -c`
+*.test
+
+# Output of the go coverage tool, specifically when used with LiteIDE
+*.out
+
+# Dependency directories (remove the comment below to include it)
+# vendor/
```
junedev, line 15: Do we really need a separate gitignore file here? Couldn't we update the existing root level file instead?

gomod-sync/README.md
```diff
@@ -0,0 +1,101 @@
...
+that must not have the default version.
+
+## Usage
+
+```
+  gomod-sync command [flags]
+
+Available Commands:
+  check       Checks if all go.mod files are in the target version
+  completion  generate the autocompletion script for the specified shell
```
junedev, line 19: I don't understand what this does/means, can you help me? :thinking: :see_no_evil:

gomod-sync/README.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+With such configuration, all `go.mod` files will be expected to have the `1.16` version of Go,
+except the exercise `strain`, which must have version `1.18` in its `go.mod`.
+
+## Examples
+
+* Check if all `go.mod` files of exercises in the `exercises` folder have the `1.16` Go version:
+
+  * `gomod-sync check --goversion 1.16 --exercises ./exercises`
```
junedev, line 93: If someone would just copy and paste this it would not work (because the executable does not exist). It would be good to give complete instructions that work for a new contributor that just cloned this repository. (Doesn't have to be here in "Examples", could also be some "install" part or similar.)

gomod-sync/.gitignore
junedev, line 15: I think it would be good to include vendor in the gitignore just in case someone runs in vendored mode.

.github/workflows/gomod-sync.yml
andrerfcsantos: Probably not. I copied that part from [exercises-tests.yml](https://github.com/exercism/go/blob/main/.github/workflows/exercise-tests.yml) where we also have that line.

.github/workflows/gomod-sync.yml
andrerfcsantos: Yes, we can. I'll change it.

gomod-sync/README.md
andrerfcsantos: This seems like a copy-paste gone wrong. Thanks for pointing out. Will fix.

gomod-sync/.gitignore
andrerfcsantos: I thought about it and I don't have a strong opinion. In the end I decided to put it here because I didn't want to pollute the global gitignore, and helps this folder to me more self-contained - we can move the folder around without the need to update the gitignore.

But I don't have a strong opinion either way.

andrerfcsantos: I thought about it and I don't have a strong opinion. In the end I decided to put it here because I didn't want to pollute the global gitignore, and helps this folder to me more self-contained - we can move the folder around without the need to update the gitignore.

But I don't have a strong opinion either way.

gomod-sync/README.md
andrerfcsantos: Which part? The `completion` command?

andrerfcsantos: Which part? The `completion` command?

gomod-sync/README.md
andrerfcsantos: Sounds good!

andrerfcsantos: Sounds good!

gomod-sync/.gitignore
andrerfcsantos: Sounds good!

gomod-sync/README.md
junedev: Yes, the completion command.

.github/workflows/gomod-sync.yml
junedev: Probably from a time when there was a transition between main and master ...

gomod-sync/README.md
andrerfcsantos: Most shells like bash or zsh allow users to provide auto completion files that tell the shell how to autocomplete custom commands. The idea here is that if you write `gomod-sync ` and then press tab, your shell will autocomplete with some random file you have in the current directory, because the shell doesn't know better. What you want is the shell to cycle the list of commands that are specific to `gomod-sync` when you write `gomod-sync ` and then press tab. But for the shell to be able to do this, it has to know the syntax for `gomod-sync`, the list of commands,  flags, etc. These completion files allow a user to teach their shell about `gomod-sync` so it provides auto completions that make sense for the command.

What the `completion` command does is generate one of these completion files for `gomod-sync`. A user can then plug this file into their shell to provide auto completion for `gomod-sync`.

That command is implemented automatically by cobra. I think we can disable it if needed.

andrerfcsantos:
> Thanks a lot for taking care of this. I only read the README so far (which is great) and it sounds like it does all the great things we need it to do. However, I am contemplating whether it is possible to reduce the complexity of the tool a bit. You built something quite "general purpose" which is nice but since it also means more code, it means there is more code for future maintainers/contributors to read, understand and maintain. Also the various versions in which the command can be invoked make it a bit harder for users.

Thank you for reviewing! Specially such a big PR.

The tool ended being a bit bigger than I anticipated originally, but I think it's still at a manageable size. The alternative would be to have a bash script, but my experience is that bash scripts also become easily unmanageable when they grow a little. And with a language like Go we can at least write tests easily for it. So I believe it's a good trade-off. Plus, we know how to write Go, right? :P 

> Point 1 and 2 together would then make the invocation as simple as `gomod-sync check` (same for the other commands) which would be a very user friendly API.

You can already do this. If no flags are specified, it'll grab the target version from the config file. Not specifying the path to the exercises is also OK, as the default value for the flag is already hardcoded to be `..\exercises`. 

The reason I made the script use all the flags was just so it is more explicit about what the program is doing. The idea is that just by seeing the flags, you get a sense on what the program is trying to do, without the need to see the documentation or actually digging into the code to see how it works. But we can just remove the use of the flags without changing the program. I see merits in both approaches.

> * The tool could assume the default Go version is always provided via the config file, then you would not need the additional goversion flag. Currently, it is confusing what happens if the file says one thing and the flag says something else. So one way of doing things would reduce potential problems and complexity. The readme then would only say: "want to configure? change the config file".

If the version is not specified via the flag, it'll use the version in the configuration file.

If you specify the version flag, it will override the configuration file. I'm not sure if this is confusing, as it is the default behavior of every program I know that uses configuration files. I can make the documentation be more explicit about it though.

> * In our context, the folder that contains the exercises and the folder that contains the config will rarely ever change. It would be ok if those were hardcoded (for the config) or part of the config file (for the exercises) so they don't have to be passed when invoking the command. I think your tool is fast enough that we don't need the ability to one check individual exercise folders.

Both the path to the config file flag and the path to the exercises folder flag have [default values](https://github.com/andrerfcsantos/exercism-go/blob/3672d22c6a356b0110f21fdd5c4ddf690fc0f122/gomod-sync/cmd/root.go#L18) hardcoded, which means that you can omit both of those flags and the default value will be used.

The version flag has no default value, but the program always reads the config file to look for the target version, which means that the version flag can also be omitted.

> * Do we need the `list` command?

I used it in the beginning to get an idea of the global state of the repository, and I think it can be useful for that purpose. I don't think maintaining this command is expensive, as it makes use of a function that the `check` and `update` commands are also using and has only a for loop beyond that.

> If you really want the tool to be used for other Go projects, then it would make more sense to keep in a separate repository and only `go get/install` it here. However, that wouldn't solve the problem of keeping this as easy/maintainable as possible for Exercism folks.

Well, I don't want this tool to be used for other Go projects, so I think it's fine to leave it as is :) I don't think the tool is that general purpose in its current form, as it relies on having the folder structure we do and on the notion of exercises.

gomod-sync/.gitignore
andrerfcsantos: Comment removed.

.github/workflows/gomod-sync.yml
andrerfcsantos: Done.

.github/workflows/gomod-sync.yml
andrerfcsantos: Just removed and seems to be working fine.

gomod-sync/README.md
andrerfcsantos: Turns out that was supposed to be a period `.`.

gomod-sync/README.md
andrerfcsantos: I added an install section. I also added a simpler example without any flags just to showcase the defaults values in action.

gomod-sync/.gitignore
junedev: If you don't mind it would be great if you could remove this file and update the root gitignore instead. Imo, this is not polluting it but instead this file here mostly contains the standard entries that should be in the root level file anyway.

Also please include "gomod-sync" in the list of files to ignore so the binary that gets build does not show up in the files to commit.

gomod-sync/README.md
junedev: Thanks for the info!

junedev:
> Thank you for reviewing! Specially such a big PR.

Don't thank me yet. :slightly_smiling_face: As I said, I have not reviewed the code, mainly only the README. (This is why I didn't know that you already had good defaults in place.)

I totally agree a bash script wouldn't be a good option. I was just asking myself whether we really need that much code for this task or whether there is significant overhead here because you wrote a proper command line application instead of a "go script" that does 2 things with some hard coded assumptions. The answer might very well be "no".

Let me try to group the other things we discussed above into some categories given the information you provided above. I am also ok if this gets merged without addressing any of the things mentioned below.

**I am strongly in favor of ...**
- Adding a section `## Quick Start` above `## Installing` section in the gomod-sync README that says something like this:
  ```
  To update all go.mod files according to the config file (gomod-sync/config.json) run `cd gomod-sync && go run main.go update`.
  
  To check all exercise go.mod files specify the correct Go version, run `cd gomod-sync && go run main.go check`.
  ```
- Add a section `## Managing the Go version` or similar to the root level readme (maybe below where the generator is explained) that contains the following
  - a very short explanation that we have the this tool, that it is built in Go and what it does
  - a link to the tool readme
  - the two lines mentioned above

**I am mid-level in favor of ...**
- Removing the list command entirely. Imo the less that needs to be explained in the readme and needs to be maintained in the code the better, even if its only some lines that are removed. In the future, the functionality of "getting an overview" is provided by looking at the config file.
- Removing the completion command from the readme. If there is really no extra code involved to support it, we could leave it as undocumented feature in case anyone ever tries it. Since the tool will be used by human like once a year and you need to type like 10 letters, I don't think anyone will miss the auto-completion thingy.

The two things above would make the command list `check`, `update`, `help` which would make it very clear what the tool is for.

**I am slightly in favor of but its totally up to you ...**
- Removing the flags

andrerfcsantos:
* Removed the `.gitignore` on the `gomod-sync` folder and moved the relevant settings to the global `.gitignore`
* Removed the `list` command
* Removed the `completion` command from the README
* Updated both README with the suggested changes

gomod-sync/.gitignore
andrerfcsantos: Done.

gomod-sync/gomod/version.go
```diff
@@ -0,0 +1,76 @@
...
+	// Truncate the file
+	if err = file.Truncate(0); err != nil {
+		return fmt.Errorf("error truncating %q: %w", path, err)
+	}
+	if _, err := file.Seek(0, 0); err != nil {
+		return fmt.Errorf("error seeking %q: %w", path, err)
+	}
+
+	// Write the file
+	_, err = file.WriteString(content)
```
junedev, line 71: What about https://pkg.go.dev/os#WriteFile instead of doing the file writing steps manually?

junedev (approved):
I tried to do a review. I don't know anything about cobra though so I can't comment on that. The part that I could understand looked good to me, just some minor comments.

gomod-sync/cmd/config/load_test.go
```diff
@@ -0,0 +1,82 @@
...
+		ExpectError bool
+	}{
+		{
+			Name:        "Loading non-existent config",
+			Path:        filepath.Join("..", "..", "testdata", "non_existent.json"),
+			Expected:    config.VersionConfig{},
+			ExpectError: true,
+		},
+		{
+			Name: "Loading non-existent config",
```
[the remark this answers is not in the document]
andrerfcsantos: It was indeed a copy/paste error. Well spotted, thanks! Fixed now.

gomod-sync/cmd/config/version.go
```diff
@@ -0,0 +1,60 @@
...
+			return false
+		}
+	}
+
+	return true
+}
+
+// ExerciseVersion represents the version one particular exercise should
+// have in its go.mod file
+type ExerciseVersion struct {
```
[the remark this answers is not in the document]
andrerfcsantos: Done.

gomod-sync/cmd/config/version.go
```diff
@@ -0,0 +1,60 @@
...
+
+// Equal tells if this VersionConfig is equal to another VersionConfig
+func (vc *VersionConfig) Equal(other VersionConfig) bool {
+	return vc.Default == other.Default && equalExceptions(vc.Exceptions, other.Exceptions)
+}
+
+// equalExceptions compares two lists of exercise versions and tells if they are equal.
+// Two exercise list versions are considered equal if they contain the same elements,
+// regardless of their order in both lists.
+func equalExceptions(a, b []ExerciseVersion) bool {
```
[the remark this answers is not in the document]
andrerfcsantos: We don't strictly need it. At the time I thought it was a good idea to make this equal function not care about the order of the elements, but I understand it might complicate things unnecessarily. I removed that logic - the elements now must be in the same order in the JSON and in the config struct.

gomod-sync/cmd/config/version_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+func TestExpectedVersion(t *testing.T) {
+	tests := []struct {
+		Name     string
+		Config   config.VersionConfig
+		Exercise string
+		Expected string
+	}{
+		{
+			Name: "Expected version just with a default should return the default",
```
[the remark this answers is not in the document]
andrerfcsantos: I agree this was confusing. I changed this first case and reworded the others to make things more clear.

gomod-sync/cmd/config/version.go
```diff
@@ -0,0 +1,60 @@
...
+	for _, exception := range vc.Exceptions {
+		if exception.Exercise == exercise {
+			return exception.Version
+		}
+	}
+	return vc.Default
+}
+
+// Equal tells if this VersionConfig is equal to another VersionConfig
+func (vc *VersionConfig) Equal(other VersionConfig) bool {
```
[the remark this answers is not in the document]
andrerfcsantos: I had my doubts when writing this too.

Since we were indeed using these functions only in tests, I moved them to the relevant test files.

gomod-sync/config.sample.json
```diff
@@ -0,0 +1,9 @@
+{
```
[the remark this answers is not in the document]
andrerfcsantos: I made this file inspired by other pieces of software I used before that provided a file like this. But this might not be necessary in our case since we have the README, and I understand how it can be confusing.

The file is now removed.

gomod-sync/gomod/find.go
```diff
@@ -0,0 +1,30 @@
...
+		if info.Name() == "go.mod" {
+			files = append(files, path)
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("walking dir %q: %w", dirPath, err)
+	}
+
+	return files, err
```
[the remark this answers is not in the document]
andrerfcsantos: :facepalm: Thanks :)

gomod-sync/gomod/version.go
```diff
@@ -0,0 +1,76 @@
...
+	"os"
+	"regexp"
+)
+
+var versionRegex *regexp.Regexp
+
+func init() {
+	// versionRegex is a regex that matches the Go version
+	// information in a go.mod file
+	versionRegex = regexp.MustCompile(`go (\d+\.\d+)`)
```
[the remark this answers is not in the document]
andrerfcsantos: You are right. I modified it to check for the start and end of the line, optionally allowing for empty space at the beginning and  end of the line.

gomod-sync/gomod/version.go
andrerfcsantos: Done.

gomod-sync/gomod/version.go
```diff
@@ -0,0 +1,76 @@
...
+func GoVersion(gomodFilePath string) (string, error) {
+	// Open gomodFilePath for reading
+	file, err := os.Open(gomodFilePath)
+	if err != nil {
+		return "", fmt.Errorf("error opening %q: %w", gomodFilePath, err)
+	}
+	defer file.Close()
+
+	// Read go.mod file
+	rawContent, err := io.ReadAll(file)
```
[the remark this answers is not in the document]
andrerfcsantos: Done. There was one other place where this was present  - the `Update` function of this file. I fixed there too.

andrerfcsantos:
@junedev Thanks! Applied all the changes. Left comments in the conversations, feel free to un-resolve any of them or suggest further changes if there's something that can be further improved.

junedev:
@andrerfcsantos Thanks for addressing all the comments. Feel free to merge whenever you like. (I don't think there will be more reviews if we wait longer tbh.)

PR #1955, opened 2021-11-10
Update exercises' go.mod to 1.16

andrerfcsantos:
PR made by running:

`gomod-sync.exe update --exercises "..\exercises" --goversion 1.16`

See https://github.com/exercism/go/pull/1954

This must be merged only after https://github.com/exercism/go-test-runner/pull/57

### DIFF
--- a/bin/test-without-stubs
+++ b/bin/test-without-stubs
@@ -77,6 +77,13 @@ cleanup () {
 
 trap cleanup EXIT INT TERM
 
+# Since Go 1.16 all go commands run in module-aware mode,
+# which would require a go.mod file to be present at the root
+# of the project.
+# Setting GO111MODULE to 'auto' makes the go commands run in
+# module-aware mode only in places where a go.mod is present.
+export GO111MODULE=auto
+
 # This is the last command in the file,
 # so the exit status of this script is the exit status of go test.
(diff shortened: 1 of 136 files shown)


junedev (approved):

junedev:
In case someone else is looking for this info: The go version in the go.mod file only affects vendoring (which we don't use) and the use of new language features. You can have go 1.13 installed locally and run a program with a go.mod file that specifies go 1.16 without any problems as long as you don't use any language features that are not available in go 1.13. 
https://golang.org/ref/mod#go-mod-file-go

bin/test-without-stubs
```diff
@@ -83,4 +83,5 @@ trap cleanup EXIT INT TERM
 # Otherwise CI may falsely report a test as passed when it has failed.
 # Run lint which is configured by .golangci.yml at the root of the repo
 # and if it's successful, run tests
+export GO111MODULE=auto
```
junedev, line 86: I would recommend moving this command above line 80 and adding a comment for it to explain why it is needed currently.

junedev:
LGTM, feel free to merge

junedev:
@andrerfcsantos Feel free to change the size label to "medium" here. I feel with all the debugging it was more work than "small".

PR #1958, opened 2021-11-14
Add stub functions to practice exercises T-Z

### LINKED ISSUE #1893: Add stub functions for practice exercises

PakkuDon:
Closes #1893 

We would like to add stub functions to the practice exercises so that students know what API they are expected to implement. We would also like to remove any API descriptions from test files as these may go out of sync with the code over time

This PR adds stubs for practice exercises starting with letters T through to Z and updates the associated test files accordingly

The following practice exercises have been omitted because they already have stubs defined:
- [tree-building](https://github.com/exercism/go/blob/main/exercises/practice/tree-building/tree_building.go)
- [triangle](https://github.com/exercism/go/blob/main/exercises/practice/triangle/triangle.go)
- [trinary](https://github.com/exercism/go/blob/main/exercises/practice/trinary/trinary.go) - this exercise is also deprecated https://github.com/exercism/go/blob/d8d2082055301d6edc2c6040f894111d17e3c1c2/config.json#L1769-L1778
- [two-fer](https://github.com/exercism/go/blob/main/exercises/practice/two-fer/two_fer.go)

### DIFF
--- a/exercises/practice/word-count/word_count.go
+++ b/exercises/practice/word-count/word_count.go
@@ -1,1 +1,7 @@
 package wordcount
+
+type Frequency map[string]int
+
+func WordCount(phrase string) Frequency {
+	panic("Please implement the WordCount function")
+}
--- a/exercises/practice/zebra-puzzle/zebra_puzzle.go
+++ b/exercises/practice/zebra-puzzle/zebra_puzzle.go
@@ -1,1 +1,7 @@
 package zebra
+
+// Define the Solution type here
+
+func SolvePuzzle() Solution {
+	panic("Please implement the SolvePuzzle function")
+}
--- a/exercises/practice/zebra-puzzle/zebra_puzzle.go
+++ b/exercises/practice/zebra-puzzle/zebra_puzzle.go
@@ -1,6 +1,9 @@
 package zebra
 
-// Define the Solution type here
+type Solution struct {
+	DrinksWater string
+	OwnsZebra   string
+}
 
 func SolvePuzzle() Solution {
 	panic("Please implement the SolvePuzzle function")
--- a/exercises/practice/tournament/tournament.go
+++ b/exercises/practice/tournament/tournament.go
@@ -1,1 +1,7 @@
 package tournament
+
+import "io"
+
+func Tally(reader io.Reader, writer io.Writer) error {
+	panic("Please implement the Tally function")
+}
--- a/exercises/practice/transpose/transpose.go
+++ b/exercises/practice/transpose/transpose.go
@@ -1,1 +1,5 @@
 package transpose
+
+func Transpose(input []string) []string {
+	panic("Please implement the Transpose function")
+}
--- a/exercises/practice/twelve-days/twelve_days.go
+++ b/exercises/practice/twelve-days/twelve_days.go
@@ -1,1 +1,9 @@
 package twelve
+
+func Verse(i int) string {
+	panic("Please implement the Verse function")
+}
+
+func Song() string {
+	panic("Please implement the Song function")
+}
--- a/exercises/practice/two-bucket/two_bucket.go
+++ b/exercises/practice/two-bucket/two_bucket.go
@@ -1,1 +1,5 @@
 package twobucket
+
+func Solve(sizeBucketOne, sizeBucketTwo, goalAmount int, startBucket string) (string, int, int, error) {
+	panic("Please implement the Solve function")
+}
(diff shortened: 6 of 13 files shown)


exercises/practice/word-count/word_count.go
```diff
@@ -1 +1,7 @@
 package wordcount
+
+type Frequency map[string]int
+
+func WordCount(phrase string) Frequency {
```
PakkuDon, line 5: I tried to replace the `Frequency` type with `map[string]int` but the tests failed locally.

```sh
go test
# wordcount [wordcount.test]
./cases_test.go:10:14: undefined: Frequency
./cases_test.go:15:3: undefined: Frequency
./cases_test.go:20:3: undefined: Frequency
./cases_test.go:25:3: undefined: Frequency
./cases_test.go:30:3: undefined: Frequency
./cases_test.go:35:3: undefined: Frequency
./cases_test.go:40:3: undefined: Frequency
./cases_test.go:45:3: undefined: Frequency
./cases_test.go:50:3: undefined: Frequency
./cases_test.go:55:3: undefined: Frequency
./cases_test.go:55:3: too many errors
FAIL	wordcount [build failed]
```

Commands to reproduce:

```sh
$ cd exercises/practice/word-count
$ cat .meta/example.go \
| sed "s/type Frequency.*//g" \
| sed "s/Frequency/map\[string\]int/g" \
> word_count.go
```

What do we think? Should we leave this as is? Should we remove the references to `Frequency` in `cases_test.go`?

andrerfcsantos: I think we should leave it as-is. In the API comment on the test file, the type was given to the student, so it's ok to give away the type here too.

What are you trying to test by replacing the `Frequency` type with `map[string]int`?  For your example to work we would have to do that replacement also in `cases_test.go`, which in turn would require to adjust the `gen.go` test generator for the exercise, but I don't see why that replacement would be useful in the first place.

andrerfcsantos (approved):
Last stretch 🎉
Thanks again for doing this!

Looks good to me, but I'll allow @junedev to review it too.

exercises/practice/zebra-puzzle/zebra_puzzle.go
```diff
@@ -1 +1,7 @@
 package zebra
+
+// Define the Solution type here
+
+func SolvePuzzle() Solution {
+	panic("Please implement the SolvePuzzle function")
+}
```
junedev, line 7: For this stub, the `Solution` definition should be part of the stub (defining that type is not part of the exercise). Additionally, since we have the proper stub now, that part of the instructions.append.md file that shows the content of the stub can be removed.

exercises/practice/word-count/word_count.go
junedev: I agree to keep `type Frequency map[string]int` in there as is.

junedev:
FYI I updated the PR description so that the issue will be closed after this is merged. Solving that issue is really a big improvement for the Go track imo. Thanks again @PakkuDon!

exercises/practice/word-count/word_count.go
PakkuDon: Gotcha. Thanks for clarifying!

exercises/practice/zebra-puzzle/zebra_puzzle.go
PakkuDon: Thanks! I've updated the stub and instructions. Let me know how it looks.

PR #1961, opened 2021-11-16
Fix typo

rootulp:
This test invokes `MapInt`. It looks like a copy + paste mistake because previous tests invoked `MapString`


junedev (approved):
Good catch! Thanks for the fix.

PR #1962, opened 2021-11-16
Explain --bench and --benchmem results by example

stmllr:
For the absolute beginners which are not familiar with go test benchmarking, some basic explanations might help to quickly understand test results.


junedev (approved):
That's a great idea. Thanks for the PR!

PR #1966, opened 2021-11-17
vehicle-purchase: correct spelling of `License`. - #1965

### LINKED ISSUE #1965: Vehicle-Purchase exercise typo
In the Vehicle Purchase exercise there are repeat occurences of the word `Licence`. 

I believe in some cases this may be a grammatical error and the correct spelling should be `License`. 

A License is a permit granting authority to own or do something. Where as Licence is the act of granting a license. 
So for example in the stub file the `NeedsLicence(kind string) bool`. Should probably be `NeedsLicense(kind string) bool`.

This isn't a big deal and maybe I am wrong but I thought it was a little confusing when running tests through the exercise.

KalebHawkins:
Replace the occurence of the word Licence with License.
closes #1965

### DIFF
--- a/exercises/concept/vehicle-purchase/.docs/hints.md
+++ b/exercises/concept/vehicle-purchase/.docs/hints.md
@@ -1,6 +1,6 @@
 # Hints
 
-## 1. Determine if you will need a driver's licence
+## 1. Determine if you will need a driver's license
 
 - Use the [equals comparison operator][comparison-operators] to check whether your input equals a certain string.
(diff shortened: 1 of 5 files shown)


junedev (approved):
Thanks for fixing this. The reason this was wrong was a mix up of American and British English.

KalebHawkins:
No worries! I know it is international language so I opened the issue before hand not entirely sure if it was controversial or not.

PR #1976, opened 2021-12-02
the-farm: teach about interfaces

jmrunkle:
Addresses #1638

### DIFF
--- a/exercises/concept/the-farm/.docs/introduction.md
+++ b/exercises/concept/the-farm/.docs/introduction.md
@@ -2,7 +2,35 @@
 
 ## Interfaces
 
-TODO
+An **interface type** is effectively a set of method signatures.
+Any type that defines those methods "implements" the interface implicitly.
+There is no `implements` keyword in Go.
+Here is what an interface definition might look like:
+
+```go
+type InterfaceName interface {
+    MethodOne() MethodOneReturnType
+    MethodTwo(paramOne ParamOneType, paramTwo ParamTwoType) MethodTwoReturnType 
+    ...
+}
+```
+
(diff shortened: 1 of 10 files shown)


exercises/concept/the-farm/.docs/introduction.md
```diff
@@ -1,20 +1,59 @@
...
+// Not this:
+func BadFoo() (int, error) {
   return 10, errors.New("Error")
 }
 ```
 
-When no errors have to be returned, return `nil`:
+> NOTE: the standard library has an `errors` package can be used to create and check errors.
+
+When no errors have to be returned, return `nil` for the error:
```
kahgoh, line 56:
```suggestion
Return `nil` for the error when there are no errors:
```

exercises/concept/the-farm/.docs/introduction.md
```diff
@@ -33,31 +72,41 @@ if err != nil {
...
 } else {
   // do something with file
 }
 ```
 
 ## Scope of error
 
-In Go, it is also a best practice to avoid declaring variables in a scope if
-these are not required later.
+In Go, it is also a best practice to avoid declaring variables in a scope if these are not required later.
```
kahgoh, line 109:
```suggestion
In Go, it is also a best practice to avoid declaring variables in a scope if they are not required later.
```

kahgoh (changes requested):
Just adding some suggestions to fix or improve some wording.

junedev (approved):

junedev:
For some reason I can't push my fixes to the branch here. I will merge this as is and create an additional PR for the updates.

PR #1977, opened 2021-12-02
Non-zero values returned wit non-zero errors

### LINKED ISSUE #1972: Non-zero values returned wit non-zero errors in `protein_translation_test.go`
This test here

https://github.com/exercism/go/blame/main/exercises/practice/protein-translation/protein_translation_test.go#L145

seems to be incorrect, at least this is not idiomatic in Go, return values with non-nil errors.
There is one important exception, sentinel errors (like `io.EOF` for example), but here it is not the case, as far as I can judge.

Maybe we should not encourage learning writing non-idiomatic (though not directly erroneous) code,
and eliminate or correct this test?

ppmasa8:
Fixed #1972.

### DIFF
--- a/exercises/practice/protein-translation/.meta/example.go
+++ b/exercises/practice/protein-translation/.meta/example.go
@@ -70,7 +70,7 @@ func FromRNA(s string) ([]string, error) {
 		case ErrStop:
 			return proteins, nil
 		case ErrInvalidBase:
-			return proteins, err
+			return nil, err
 		default:
 			proteins = append(proteins, p)
 		}
--- a/exercises/practice/protein-translation/protein_translation_test.go
+++ b/exercises/practice/protein-translation/protein_translation_test.go
@@ -159,11 +159,29 @@ func TestProtein(t *testing.T) {
 			t.Fatalf("FAIL: RNA translation test: %s\nExpected: %s\nGot error: %q",
 				test.input, test.expected, err)
 		}
-		if !reflect.DeepEqual(actual, test.expected) {
+		if !slicesEqual(actual, test.expected) {
 			t.Fatalf("FAIL: RNA Translation test: %s\nExpected: %q\nActual %q", test.input, test.expected, actual)
 		}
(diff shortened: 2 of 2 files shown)


andrerfcsantos (changes requested):
Thanks for this.

However, this don't fully solve our problems. With the current code, when the function return the error, we are checking the error and the non-error value. This can be problematic with the current check, because we are using `reflect.DeepEquals` to compare both slices, and if a student returns an empty slice when an error occurs, the `nil` slice of the tests won't be `DeepEquals` to an empty slice - but for the context of this problem I think it's fine to treat them as the same. I suggest one of two approaches to solve this:

* When are expecting and error and the function returns an error, we only check if it is the error we are expecting, like we are doing now, but unlike we are doing now, we don't check for the non-error value and we let the user student return whatever they want. I think this is fine, as if a student is not returning `nil` or the empty slice in an error case, this can be a point of a mentor discussion.
* If we still want to check the non-error value when an error is returned, we must treat the nil slice and the empty slice as equal.  We did something similar to this in the [cards-trick tests](https://github.com/exercism/go/blob/main/exercises/concept/card-tricks/card_tricks_test.go#L256), where we replaced the use of `reflect.DeepEquals` with that `slicesEqual` function that treats the nil and empty slices as the same.

andrerfcsantos:
The tests are not passing, this is because the `example.go` file must also be updated to return a nil or empty slice in case of an error. Alternatively, we can also change the test to not care about the slice returned when an error is returned.

andrerfcsantos (changes requested):
Sorry it took me a while to get back to you. The tests are not passing, this is because the example.go file must also be updated to return a nil or empty slice in case of an error. Alternatively, we can also change the test to not care about the slice returned when an error is returned.

ppmasa8:
Sorry, I deleted the file by mistake, I'll restore it as soon as I can.

andrerfcsantos (approved):

PR #1979, opened 2021-12-09
Wrapped CalculateResellPrice under Test.Run

RajivTS:
The tests for CalculateResellPrice vary from the other tests in that it just prints whether all the tests were Pass or Fail. Its missing the following:
1. List of individual tests that were executed for the CalculateResellPrice method
2. The status at individual test level (Pass or Fail)
3. The time taken to execute the individual tests (e.g. 0.01s)


andrerfcsantos (approved):
Thanks for this!

PR #1981, opened 2021-12-10
Pythagorean triplets exercise could use more detailed explanation

### LINKED ISSUE #1963: Pythagorean triplets exercise could use more detailed explanation
The current explanation for this exercise is extremely short and leaves the user (me) guessing what is exactly expected in the three functions that are presented. I've personally done this exercise in at least 2 other languages where the tests were sufficient to figure out what was expected. For Go, it's unclear what Sum/Range should do. I also found that Sum is sensitive to the order in which the elements are returned.

ppmasa8:
Fixed #1963.


junedev (approved):

junedev:
@ppmasa8 Thanks! I fixed the formatting in the append file a little and added a short function comment back in as we usually provide those.

PR #1982, opened 2021-12-10
Instructions for missed case in PrefilledSlice 

RajivTS:
The instructions and comments for PrefilledSlice method do not mention the intended behavior when the length of the slice is non-positive. However, the test suite for Prefilled slice contains a case where the length is -3. Updating the instructions to include an example and updating the comments to match the instructions.


junedev (approved):
Thanks for bringing this up! I am ok with the current version but I would prefer not to spoil too much and just add the new line 59 to the instructions instead of also having an example and an extra comment.
If someone does not get it 100% right and the test case then points that out, that's fine

RajivTS:
Updated based on review comments.

PR #1992, opened 2021-12-13
Add test to the_farm.go

tbrisker:
Added a test for negative fodder with ScaleError, which should result in a "Negative Fodder" error.


junedev (approved):
Good catch! Thanks!

PR #1994, opened 2021-12-13
Included TimeZone in AnniversaryDate Instructions

RajivTS:
Based on the test suite, the AnniversaryDate function needs to return the current year's anniversary date in UTC timezone. However, the current instructions have no mention of any such requirement. Updating the instruction to fix this.


junedev (approved):

junedev:
Thanks for the fix!

PR #2001, opened 2021-12-14
Extended Instructions for AddItem Function

RajivTS:
Based on the test suite, the expectation from the AddItem function is to add the input item to the bill and if its already present, to increase its quantity by unit amount. The current instructions mention only the former part of the requirement and leave out the latter. Included an additional line to fix this.

### DIFF
--- a/exercises/concept/gross-store/.docs/instructions.md
+++ b/exercises/concept/gross-store/.docs/instructions.md
@@ -41,7 +41,8 @@ fmt.Println(bill)
 To implement this, you'll need to:
 
 - Return `false` if the given `unit` is not in the `units` map.
-- Otherwise add the item to the customer `bill`, indexed by the item name, then return `true`. If the item is already present in the bill, increase its quantity by `unit` amount
+- Otherwise add the item to the customer `bill`, indexed by the item name, then return `true`.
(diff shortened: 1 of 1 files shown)


exercises/concept/gross-store/.docs/instructions.md
```diff
@@ -41,7 +41,7 @@ fmt.Println(bill)
 To implement this, you'll need to:
 
 - Return `false` if the given `unit` is not in the `units` map.
-- Otherwise add the item to the customer `bill`, indexed by the item name, then return `true`.
+- Otherwise add the item to the customer `bill`, indexed by the item name, then return `true`. If the item is already present in the bill, increase its quantity by `unit` amount
```
junedev, line 44:
```suggestion
- Otherwise add the item to the customer `bill`, indexed by the item name, then return `true`.
- If the item is already present in the bill, increase its quantity by the amount that belongs to the provided `unit`.
```

junedev (approved):
Good catch! What do you think about adding it as a separate point? See suggestion.

exercises/concept/gross-store/.docs/instructions.md
RajivTS: Yes that looks better.

PR #2002, opened 2021-12-15
sorting-room: Align hints.md content to the sorting_room.go

jasonwucj:
In the hints.md, the hint 3 describes that we can check if this is `FancyBox`.
That is inconsistent to the type in the sorting_room.go source.
I think `FancyNumber` is supposed to be the correct word in the hint.


jasonwucj:
Sorry, I used wrong branch.
Please ignore this PR and I will create correct one instead.

PR #2003, opened 2021-12-15
sorting-room: Align hints.md content to sorting_room.go source

jasonwucj:
In the hints.md, it describes that we can check if this is `FancyBox`.
That is inconsistent to the type in sorting_room.go source.
I think `FancyNumber` is supposed to be the correct word in the hint.


junedev (approved):
Yes, good catch! We changed the exercise at one point but forgot to change that hint.

jasonwucj:
Thanks for the review and merge. :)

PR #2009, opened 2021-12-18
Instructions fix for The Farm

RajivTS:
This PR contains the following minor changes:

- Renamed `weightFodder` to `FodderAmount` when referencing the method that provides the fodder amount
- Renamed instances of `function` with `method` when referencing the `FodderAmount` method
- Included the keyword `struct` when referring to the `weightFodder` instance
- Updated the names of `WeightFodder` instances to maintain consistency between comments and function arguments (e.g. replaced `twentyFodderNoError says there are -5.0 fodder` with `negativeFiveFodder says there are -5.0 fodder`)

This PR contains the following major change:

The sequence of instructions do not reflect the sequence of cases that need to be handled by the `DivideFood` function. For example, if the amount of fodder is negative then the error returned by `FodderAmount` method (e.g. `ErrScaleMalfunction` or `nil`) is inconsequential since in this case `DivideFood` should always return `Negative fodder` error [Inferred from test cases]. From observation, it appears that negative fodder, no cows and negative cows take precedence over `ErrScaleMalfunction` and other errors which in turn take precedence over the happy-path case. The instructions have been rearranged (and modified a bit) to reflect this sequence.

### DIFF
--- a/exercises/concept/the-farm/.docs/instructions.md
+++ b/exercises/concept/the-farm/.docs/instructions.md
@@ -12,7 +12,7 @@ While this is good for your pocket, you want to catch the division by zero retur
 Also, your silly nephew (who has just learned about negative numbers) sometimes will say that there are a negative number of cows.
 You love your nephew so you want to return a helpful error when he does that.
 
-## 1. Get the amount of fodder from the `weightFodder` method
+## 1. Get the amount of fodder from the `FodderAmount` method
 
 You will be passed a `WeightFodder` struct which has a method called `FodderAmount` that returns the amount of fodder available and possibly an error.
 
--- a/exercises/concept/the-farm/.docs/instructions.md
+++ b/exercises/concept/the-farm/.docs/instructions.md
@@ -14,7 +14,7 @@ You love your nephew so you want to return a helpful error when he does that.
 
 ## 1. Get the amount of fodder from the `FodderAmount` method
 
-You will be passed a `WeightFodder` struct which has a method called `FodderAmount` that returns the amount of fodder available and possibly an error.
+You will be passed a `WeightFodder` interface instance which has a method called `FodderAmount` that returns the amount of fodder available and possibly an error.
 
 ## 2. Return an error for negative fodder
 
--- a/exercises/concept/the-farm/.docs/instructions.md
+++ b/exercises/concept/the-farm/.docs/instructions.md
@@ -12,33 +12,16 @@ While this is good for your pocket, you want to catch the division by zero retur
 Also, your silly nephew (who has just learned about negative numbers) sometimes will say that there are a negative number of cows.
 You love your nephew so you want to return a helpful error when he does that.
 
-## 1. Get the amount of fodder from the `weightFodder` function
+## 1. Get the amount of fodder from the `weightFodder` method
 
-You will be passed a `WeightFodder` which has a function called `FodderAmount` which returns the amount of fodder available and possibly an error.
-
-```go
-// twentyFodderNoError says there are 20.0 fodder
-fodder, err := DivideFood(twentyFodderNoError, 10)
-// fodder == 2.0
-// err == nil
-```
-
-If `ErrScaleMalfunction` is returned by `FodderAmount`, double the fodder amount returned by `FodderAmount`.
-For any other error, return `0` and the error.
-
-```go
-// twentyFodderNoError says there are 20.0 fodder and a ErrScaleMalfunction
-fodder, err := DivideFood(twentyFodderWithErrScaleMalfunction, 10)
-// fodder == 4.0
-// err == nil
-```
+You will be passed a `WeightFodder` struct which has a method called `FodderAmount` that returns the amount of fodder available and possibly an error.
 
 ## 2. Return an error for negative fodder
 
 If the scale is broken and returning negative amounts of fodder, return an error saying "Negative fodder":
 
 ```go
-// twentyFodderNoError says there are -5.0 fodder
+// negativeFiveFodder says there are -5.0 fodder
 fodder, err := DivideFood(negativeFiveFodder, 10)
 // fodder == 0.0
 // err.Error() == "Negative fodder"
@@ -68,4 +51,25 @@ You can see the format of the error message in the example below.
 fodder, err := DivideFood(twentyFodderNoError, -5)
 // fodder == 0.0
 // err.Error() == "silly nephew, there cannot be -5 cows"
+```
+
+## 5. Handle ErrScaleMalfunction and other errors
+
+If `ErrScaleMalfunction` error is returned by `FodderAmount`, double the fodder amount returned by `FodderAmount` before dividing it between the cows.
+For any other error, return `0` and the error. 
+
+```go
+// twentyFodderWithErrScaleMalfunction says there are 20.0 fodder and a ErrScaleMalfunction
+fodder, err := DivideFood(twentyFodderWithErrScaleMalfunction, 10)
+// fodder == 4.0
+// err == nil
+```
+
+In case of `nil` error, return the fodder amount equally divided between the input number of cows.
+
+```go
+// twentyFodderNoError says there are 20.0 fodder
+fodder, err := DivideFood(twentyFodderNoError, 10)
+// fodder == 2.0
+// err == nil
 ```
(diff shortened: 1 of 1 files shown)


exercises/concept/the-farm/.docs/instructions.md
```diff
@@ -12,33 +12,16 @@ While this is good for your pocket, you want to catch the division by zero retur
...
-If `ErrScaleMalfunction` is returned by `FodderAmount`, double the fodder amount returned by `FodderAmount`.
-For any other error, return `0` and the error.
-
-```go
-// twentyFodderNoError says there are 20.0 fodder and a ErrScaleMalfunction
-fodder, err := DivideFood(twentyFodderWithErrScaleMalfunction, 10)
-// fodder == 4.0
-// err == nil
-```
+You will be passed a `WeightFodder` struct which has a method called `FodderAmount` that returns the amount of fodder available and possibly an error.
```
jasonwucj, line 17: Hi, @RajivTS , after checking the source types.go, I think `WeightFodder` is interface rather than struct. It would be better to describe it as interface type here. What do you think? :)

RajivTS: Thanks for pointing this out! Yes I agree. Calling it a struct wouldn't be entirely accurate since the underlying type is an interface even though the instance passed as argument to the function `DivideFood` would be a struct. Updated it to "interface instance" for now until someone provides a better suggestion.

exercises/concept/the-farm/.docs/instructions.md
```diff
@@ -12,33 +12,16 @@ While this is good for your pocket, you want to catch the division by zero retur
...
-If `ErrScaleMalfunction` is returned by `FodderAmount`, double the fodder amount returned by `FodderAmount`.
-For any other error, return `0` and the error.
-
-```go
-// twentyFodderNoError says there are 20.0 fodder and a ErrScaleMalfunction
-fodder, err := DivideFood(twentyFodderWithErrScaleMalfunction, 10)
-// fodder == 4.0
-// err == nil
-```
+You will be passed a `WeightFodder` interface instance which has a method called `FodderAmount` that returns the amount of fodder available and possibly an error.
```
junedev, line 17: In your new version, task 1 does not contain any actual task anymore. There is only this one sentence. That is confusing.

exercises/concept/the-farm/.docs/instructions.md
```diff
@@ -69,3 +52,24 @@ fodder, err := DivideFood(twentyFodderNoError, -5)
 // fodder == 0.0
 // err.Error() == "silly nephew, there cannot be -5 cows"
 ```
+
+## 5. Handle ErrScaleMalfunction and other errors
```
junedev, line 56: Doing this part last might reflect your implementation but it is not good practice. To decide whether to return the negative fodder error in task 2, you need to look at `fodder`. That is only "allowed" when you first ensured that `err` is `nil`. If the error is not nil, you should never blindly use the result of a function call, it can contain whatever random value.

`ErrScaleMalfunction` is an exception as the instructions that that is ok to work with the result but we should probably change that at some point so it does not set a bad example anymore.

junedev (changes requested):

exercises/concept/the-farm/.docs/instructions.md
junedev: "interface instance" does not sound much like Go. Maybe something like this instead:

You will be passed a value that fulfills the `WeightFodder` interface. `WeightFodder` includes a method called `FodderAmount` that ...

junedev:
@RajivTS Thanks for raising all these points. I agree with most of the minor fixes. I left a suggestion for improving the sentence in task 1 above.

I don't think changing the order of the tasks is a good idea though, see comment above. If you revert that part, I am happy to merge get the other fixes merged.

junedev:
FYI I will create a separate issue for adding a test case where something like `-1, errors.New("not a scale error")` is returned to encourage handling the error before evaluating the result.

exercises/concept/the-farm/.docs/instructions.md
RajivTS: Agreed. Proposed verbiage sounded more C# than Go :)

exercises/concept/the-farm/.docs/instructions.md
RajivTS: > `ErrScaleMalfunction` is an exception as the instructions that that is ok to work with the result but we should probably change that at some point so it does not set a bad example anymore.

Absolutely. In fact, the case of `ErrScaleMalfunction` is the one that led me to the incorrectly conclude on the prior handling of values before error. As per instructions, `ErrScaleMalfunction` error requires the implementation to double the fodder amount and return a nil error. However, if the fodder amount is negative then this rule is bypassed by the `Negative fodder` case (with or without an error) implying that inspection of the value should be given precendence.

exercises/concept/the-farm/.docs/instructions.md
RajivTS: Agreed. With the new set of changes, this is no longer needed and hence removed.

RajivTS:
> FYI I will create a separate issue for adding a test case where something like `-1, errors.New("not a scale error")` is returned to encourage handling the error before evaluating the result.

Sounds great 👍🏼  One thing to consider though, if the fodder amount is negative and the error returned is non-`ErrScaleMalfunction` error then based on which part of the instructions we focus on, this can be handled in either of the following two ways.

1. Line 26 of the instructions specifies that in case of any non-`ErrScaleMalfunction` error, return the error and `0`. Thus the final return value would be (`0`, err)
2. Line 36 of the instructions specifies that in case of negative fodder amount, we should return `0` and `Negative fodder` error.

Based on the practice of checking the error before the value, option 1 seems to the valid choice. Is there a case here for adding extra instructions to clarify the expected response? Or is the `wanted` output of the test case sufficient instruction for the user?

RajivTS:
@junedev I have reverted the reordering part and only kept the minor changes as suggested.

exercises/concept/the-farm/.docs/instructions.md
```diff
@@ -12,9 +12,9 @@ While this is good for your pocket, you want to catch the division by zero retur
 Also, your silly nephew (who has just learned about negative numbers) sometimes will say that there are a negative number of cows.
 You love your nephew so you want to return a helpful error when he does that.
 
-## 1. Get the amount of fodder from the `weightFodder` function
+## 1. Get the amount of fodder from the `FodderAmount` method
 
-You will be passed a `WeightFodder` which has a function called `FodderAmount` which returns the amount of fodder available and possibly an error.
+You will be passed a value that fulfills the `WeightFodder` interface. `WeightFodder` includes a method called `FodderAmount` that returns the amount of fodder available and possibly an error.
```
junedev, line 17:
```suggestion
You will be passed a value that fulfills the `WeightFodder` interface.
`WeightFodder` includes a method called `FodderAmount` that returns the amount of fodder available and possibly an error.
```

exercises/concept/the-farm/.docs/instructions.md
```diff
@@ -23,11 +23,10 @@ fodder, err := DivideFood(twentyFodderNoError, 10)
 // err == nil
 ```
 
-If `ErrScaleMalfunction` is returned by `FodderAmount`, double the fodder amount returned by `FodderAmount`.
-For any other error, return `0` and the error.
+If `ErrScaleMalfunction` is returned by `FodderAmount` and the fodder amount is positive, double the fodder amount returned by `FodderAmount` before dividing it equally between the cows. For any other error, return `0` and the error.
```
junedev, line 26:
```suggestion
If `ErrScaleMalfunction` is returned by `FodderAmount` and the fodder amount is positive, double the fodder amount returned by `FodderAmount` before dividing it equally between the cows.
For any other error, return `0` and the error.
```

junedev (approved):

junedev:
@RajivTS Thanks a lot for being so cooperative here! 

Regarding your question above: Since it is a general guideline in Go to not use the result if there was an error, I think the test would be enough guidance here. Rather than adding more instructions we should make sure the introduction (concept) includes that info about not using the result.

FYI I added 2 missing line breaks. See our markdown guidelines for reference: https://exercism.org/docs/building/markdown/markdown#h-layout

PR #2012, opened 2021-12-21
list-ops: fix typo in stub function

jasonwucj:
I noticed that there is a typo in the stub function when I was doing list-ops exercise.
Fix it with this small PR.


junedev (approved):
Thanks for the fix!

jasonwucj:
Thanks for the approval and merge. :)

PR #2015, opened 2021-12-22
Removing int from `DescribeAnything`

mdrakiburrahman:
`DescribeAnything` should not route `int` to  `DescribeNumber` because `DescribeNumber` can only handle `float64` - here's the function signature:

`func DescribeNumber(f float64) string {...`

Also, none of the above mini-challenges (1 to 4) build up to handling `int`, but rather `float64`s, so it makes sense to remove the `int` verbiage.

### DIFF
--- a/exercises/concept/sorting-room/.docs/instructions.md
+++ b/exercises/concept/sorting-room/.docs/instructions.md
@@ -65,7 +65,7 @@ This is the main function Jen needs which takes any input (the empty interface m
 `DescribeAnything` should delegate to the other functions based on the type of the value passed in.
 More specifically:
 
-- `int` and `float64` should both delegate to `DescribeNumber`
+- `float64` should delegate to `DescribeNumber`
 - `NumberBox` should delegate to `DescribeNumberBox`
 - `FancyNumberBox` should delegate to `DescribeFancyNumberBox`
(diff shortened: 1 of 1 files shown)


jmrunkle:
This concept exercise is teaching about type conversion and type assertions. The point here is to force them to cast from an int to a float64 (as the exemplar does [here](https://github.com/exercism/go/blob/aab8c439abfa989fdf779fe42ac6d2eef6341136/exercises/concept/sorting-room/.meta/exemplar.go#L58)). The [hints](https://github.com/exercism/go/blob/aab8c439abfa989fdf779fe42ac6d2eef6341136/exercises/concept/sorting-room/.docs/hints.md) verify this if they get confused.

What we probably *should* do is add a test for the int case which apparently was missed (or possibly removed) from the [DescribeAnything tests](https://github.com/exercism/go/blob/aab8c439abfa989fdf779fe42ac6d2eef6341136/exercises/concept/sorting-room/sorting_room_test.go#L145).

junedev:
@mdrakiburrahman Thanks for bringing this up but like @jmrunkle I also think including `int` is good practice here.

I created an issue for the missing test case: https://github.com/exercism/go/issues/2016. I will close this PR in turn.

PR #2017, opened 2021-12-22
Integer Test case added

### LINKED ISSUE #2016: Sorting Room: Add test case
The `DescribeAnything` function is supposed to be able to handle `int`s but this is currently not covered in the test cases: https://github.com/exercism/go/blob/aab8c439abfa989fdf779fe42ac6d2eef6341136/exercises/concept/sorting-room/sorting_room_test.go#L145-L176

A test case for this should be added. Its probably good to explicitly mention the type in value for `input` to avoid it only works due to numeric constant auto-magic.

VanSHOE:
Fixes #2016 

Please have a look and let me know if this is what was required

### DIFF
--- a/exercises/concept/sorting-room/sorting_room_test.go
+++ b/exercises/concept/sorting-room/sorting_room_test.go
@@ -155,7 +155,7 @@ func TestDescribeAnything(t *testing.T) {
 		},
 		{
 			description: "Describe 42",
-			input:       "42",
+			input:       42,
 			want:        "This is the number 42.0",
 		},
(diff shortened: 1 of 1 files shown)


VanSHOE:
Might want to flatten the commits as well, as I made a small mistake by making the input a string before

junedev (approved):
Yes, this work, thanks! I double checked the code coverage to be sure we have 100% now.

junedev:
We always squash commits, so no need to worry about it.

PR #2018, opened 2021-12-22
tree-building: Fix exercise description

stmllr:
The exercise is about implementing a tree building algorithm from scratch, not refactoring an existing one.

### DIFF
--- a/exercises/practice/tree-building/.meta/config.json
+++ b/exercises/practice/tree-building/.meta/config.json
@@ -1,5 +1,5 @@
 {
-  "blurb": "Refactor a tree building algorithm.",
+  "blurb": "Implement a tree building algorithm.",
   "authors": [
     "pminten"
   ],


jmrunkle:
So... the blurb here actually comes from the [problem-specifications repo](https://github.com/exercism/problem-specifications/tree/main/exercises/tree-building). I understand that the current implementation over here actually does not do what is described there, but I am not sure if we should "fix" our exercise to actually do what was described in the problem-specifications repo. @junedev for more thinking about this. I am not certain we want to veer too far from what is in the problem-specifications repo.

stmllr:
Thanks for your reply. 

The phrase also appears in the problem-specifications exercise description and was removed in the go exercise instructions:

```diff
--- problem-specifications/blob/main/exercises/tree-building/description.md      2021-12-23 07:38:11.941182328 +0100
+++ go/blob/main/exercises/practice/tree-building/.docs/instructions.md     2021-12-23 07:38:33.605817991 +0100
@@ -1,18 +1,15 @@
-# Description
-
-Refactor a tree building algorithm.
+# Instructions
 
 Some web-forums have a tree layout, so posts are presented as a tree. However
 the posts are typically stored in a database as an unsorted set of records. Thus
 when presenting the posts to the user the tree structure has to be
 reconstructed.
 
-Your job will be to refactor a working but slow and ugly piece of code that
-implements the tree building logic for highly abstracted records. The records
-only contain an ID number and a parent ID number. The ID number is always
-between 0 (inclusive) and the length of the record list (exclusive). All records
-have a parent ID lower than their own ID, except for the root record, which has
-a parent ID that's equal to its own ID.
+Your job will be to implement the tree building logic for these records. The
+records only contain an ID number and a parent ID number. The ID number is
+always between 0 (inclusive) and the length of the record list (exclusive). All
+records have a parent ID lower than their own ID, except for the root record,
+which has a parent ID that's equal to its own ID.
```
https://github.com/exercism/problem-specifications/blob/main/exercises/tree-building/description.md
https://github.com/exercism/go/blob/main/exercises/practice/tree-building/.docs/instructions.md

junedev:
@jmrunkle For tree building, some other tracks also have this diverged version and I think it is fine to change the blurb as well here, so feel free to merge the PR.

As for fixing our exercise to conform to problem spec: Every now and then there is a new discussion whether these refactoring exercises make sense in their current form because if you would encounter that messy function in a real world scenario, you could also just rewrite it completely. For refactoring to make sense, it would need some constrains (e.g. @andrerfcsantos suggested a scenario where you have multiple functions and you need to fix the "down-stream" ones in a way that the upper ones still work). But setting this up is a lot of work (I haven't seen it on Exercism yet). So as long as we don't have a refactoring scenario that makes sense, I rather keep the normal exercise even if it diverges from problem spec.

jmrunkle (approved):

PR #2019, opened 2021-12-23
isbn-verifier: Enhance invalid characters test

lcarva:
Signed-off-by: Luiz Carvalho <lucarval@redhat.com>

### DIFF
--- a/exercises/practice/isbn-verifier/cases_test.go
+++ b/exercises/practice/isbn-verifier/cases_test.go
@@ -25,5 +25,6 @@ var testCases = []struct {
 	{"", false, "empty isbn"},
 	{"134456729", false, "input is 9 characters"},
 	{"3132P34035", false, "invalid characters are not ignored"},
+	{"3598P215088", false, "extraneous invalid characters are not ignored"},
 	{"98245726788", false, "input is too long but contains a valid isbn"},
 }


andrerfcsantos:
Hello,

Thanks for the interest in contributing to the Go track!

Usually, the tests cases for practice exercises are centralized in the [problem-specifications](https://github.com/exercism/problem-specifications) repo, so changes to the test cases should be suggested there and not on any specific language track. The idea is to have the test cases defined in one place so all tracks can then benefit from it.

I suggest opening a PR in the problem specifications repo with this change. In the PR, it is also a good idea to mention _why_ you think this test case is important and what it is trying to test that it's not tested already by the other test cases. Taking a quick look at the test you added, I believe what you are trying to test is if a solution correctly checks the length of the input when it has an invalid character which makes the input an invalid ISBN, when it would otherwise be a valid one. I believe it might be a good idea to add this test indeed, but please open a PR in the problem spec repo. After being accepted there, we can import it here to the Go track.

lcarva:
@andrerfcsantos, thank you for the prompt feedback. As suggested, I created the following: https://github.com/exercism/problem-specifications/pull/1898.

What's the next step for this PR? I don't know if it's still needed.

andrerfcsantos:
Great, thanks!

I'll close this PR for now.

PR #2023, opened 2021-12-25
the-farm: fix go-staticcheck warning

Elfayer:
Task2 asks for:

```go
// negativeFiveFodder says there are -5.0 fodder
fodder, err := DivideFood(negativeFiveFodder, 10)
// fodder == 0.0
// err.Error() == "Negative fodder"
```

When I type `return 0, errors.New("Negative fodder")`, VSCode warns out: `error strings should not be capitalized (ST1005) go-staticcheck`
Same for Task3.


andrerfcsantos (approved):
Thanks for this!

PR #2024, opened 2021-12-25
sorting-room: fix test description typo

Elfayer:
Sending value `12`, while description says `11`


andrerfcsantos (approved):
Nice catch, thanks for this!

PR #2026, opened 2021-12-26
Update sorting_room_test.go

### LINKED ISSUE #2022: sorting-room: TestExtractFancyNumber doesn't fully test for incorrect type
In the `sorting-room` exercise, the `ExtractFancyNumber` function must check the type of the argument given:

> // ExtractFancyNumber should return the integer value for a FancyNumber and 0 if any other FancyNumberBox is supplied.

However, the tests still pass with this function that doesn't check any type. Just checking if `Atoi` returns an error is enough:

```go
func ExtractFancyNumber(fnb FancyNumberBox) int {
	v, err := strconv.Atoi(fnb.Value())
	if err != nil {
		return 0
	}
	return v
}
```

This happens because the only test we have that tests a non-FancyNumber value is also a test that provides an invalid string number, which makes `Atoi` fail:

 ```go
{
	description: "Extract a different fancy number returns 0",
	input:       differentFancyNumber{"two"},
	want:        0,
},
```

To fix this issue, a possible solution is to add a test that creates a `differentFancyNumber` with a string that is a valid number different than zero. This way `Atoi` will happily return a value different than zero, when the test will expect zero to be returned because `differentFancyNumber` is not a `FancyNumber`. This will force solutions to actually check the type of the argument.

LH-ctrl:
Fixes #2022

I added a test case in TestDescribeFancyNumberBox that passes a valid Number in ASCII-Format to the differentFancyNumber struct, so solutions have to check for valid arguments

### DIFF
--- a/exercises/concept/sorting-room/sorting_room_test.go
+++ b/exercises/concept/sorting-room/sorting_room_test.go
@@ -135,7 +135,7 @@ func TestDescribeFancyNumberBox(t *testing.T) {
 		{
 			description: "Describe a valid different fancy number",
 			input:       differentFancyNumber{"4"},
-			want:        "This is a fancy box containing the number 4.0",
+			want:        "This is a fancy box containing the number 0.0",
 		},
 	}
 	for _, test := range tests {
--- a/exercises/concept/sorting-room/sorting_room_test.go
+++ b/exercises/concept/sorting-room/sorting_room_test.go
@@ -132,6 +132,11 @@ func TestDescribeFancyNumberBox(t *testing.T) {
 			input:       differentFancyNumber{"three"},
 			want:        "This is a fancy box containing the number 0.0",
 		},
+		{
+			description: "Describe a valid different fancy number",
+			input:       differentFancyNumber{"4"},
+			want:        "This is a fancy box containing the number 4.0",
+		},
 	}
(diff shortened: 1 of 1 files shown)


andrerfcsantos (changes requested):

CI is failing because the test is not 100% correct, please see my suggestion below for a fix.

andrerfcsantos (changes requested):
**EDIT**: Seems that CI failures were temporary indeed, disregard advice below and this request for changes.

Original comment:

> Ok, so the tests seem to be passing now, but CI seems to be complaining. However, the errors seem to be unrelated to the changes. I re-ran all the jobs to discard the possibility of it being a temporary failure but the errors persist.
>After taking a closer look, I noticed this branch is 30+ commits behind our main. Even though I don't immediately see anything in those 30+ commits that might explain this failure, I think updating this branch to be up-to-date with our main might help, as the jobs in other PR's seem to be working fine.
> Can you please update the main branch of your fork to be up-to-date with the main branch of this repository? My suggestion to do this would be to rebase your main branch into this repository main branch and force-push the changes to your main. That will update this PR and will force the CI to re-run, hopefully making all tests green.

andrerfcsantos (approved):
It does seem that the CI failures were temporary after all.

PR #2029, opened 2021-12-26
Do not define to Board type here.

bobtfish:
This type is already defined in the file board.go

### DIFF
--- a/exercises/practice/minesweeper/minesweeper.go
+++ b/exercises/practice/minesweeper/minesweeper.go
@@ -1,7 +1,5 @@
 package minesweeper
 
-// Define the Board type here.
-
 func (b Board) Count() error {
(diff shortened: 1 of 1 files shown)


junedev:
Note: Before this is merged we should check whether the type should be provided in that file or whether it should be part of the solution the student writes. (Depends on how the tests are set up etc.)

jmrunkle (approved):
The board is defined in a separate file board.go and listed as an ["editor"](https://github.com/exercism/go/blob/c0bd1966118cab3913c55284e9b0521e270e3525/exercises/practice/minesweeper/.meta/config.json#L28-L30) file which means it cannot be edited in the online editor, but will be shown. This seems to be working as designed, we just need to remove the comment. Thanks!

PR #2031, opened 2021-12-27
Update design.md

LH-ctrl:
Added notes about the use of for-loops at the bottom of the document, to encourage a different approach to the exercise


junedev (approved):
Thanks for the note. That will be valuable when we work on the analyzer later. Would be great if you could make the following adjustments:
* Use "Analyzer" instead if "Notes" as header
* "teached" should be "taught"
* Include the name of the specific function that can be solved using the for loop.

Issue #2011, opened 2021-12-19
The Farm: Improve introduction


junedev:
Follow up from #2009 

The tests for "The Farm" currently don't cover what should happen when a negative fodder amount and a non-scale error is returned. Go's general guideline is to not use the result if there was an error so passing on the error takes precedence over handling the negative amount.

## Tasks
- [x] Add a test case for a negative fodder amount and a non-scale error. The expected outcome of `DevideFood` is 0 and the error that was returned from FodderAmount.
- [ ] Make sure the errors concept and the exercise introduction mention the guideline for usually not using the result in case there was error.

junedev:
A quote from @andrerfcsantos on Slack regarding the topic of the open task here:
> In Go, when you return a non-nil error in a function that also returns other things, the caller of the function is supposed to ignore all the other things except the error. This is unless the documentation of the function says otherwise, but I never really saw a function return an error but make you pay attention to the other values.
This means that n theory, when you return an error, you can return anything in the other values - those other values aren't supposed to be looked at anyway because you are returning an non-nil error. But returning the zero value makes the code easier to read, as you are signaling there's nothing special going on there.

Lewiscowles1986:
To be quite honest, I think this might be a problem with the errors introduction, that it does not support or show learners the "Handle specific errors, but defer others" explicitly.

```
    if err != nil && err != ErrScaleMalfunction {
    	return 0.0, err
    }
```

I Was most frustrated as I did not see language I recognised as saying "Hey, we want to handle this error, but early exit all others."

I'd in-fact read the exercise as "We will only expect this error".

How to communicate this...

> ## Defining behavior for specific error types
> While we can and should, early return most errors that occur. We may in some special cases, wish to define behavior upon a specific type of error, that we know we can handle. When that happens, we still return all other errors early (code example from above). But we might choose to return a value without error for a specific error.

junedev:
@Lewiscowles1986 You raise a good point there. However, I think what you are bringing up is unrelated to the open task in this issue. It would be great, if you could open a new issue instead and copy your comment in there. Then we can discuss a bit more in that new issue.

junedev:
[...]

PR #2032, opened 2021-12-27
added test case for negative amount with non-scale error #2011

est5:
Part of #2011

### DIFF
--- a/exercises/concept/the-farm/the_farm_test.go
+++ b/exercises/concept/the-farm/the_farm_test.go
@@ -73,6 +73,14 @@ func TestDivideFood(t *testing.T) {
 			wantErr:                 nonScaleError,
 		},
 		{
+			description:             "Negative fodder with generic error from the scale is returned",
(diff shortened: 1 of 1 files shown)


andrerfcsantos (approved):
Thanks for this!

enarha:
Hi, as I understand it, this test contradicts to the exercise instructions. The instructions say "If the scale is broken and returning negative amounts of fodder, return an error saying "negative fodder". This test checks for negative value, but expects "non-scale error"

junedev:
@enarha It might seem a bit strange but it is all correct. The test in this PR tests that the solution checks the error first and only evaluates the actual result of the function of the error is either `nil` or is the special `ErrScaleMalfunction`. See https://github.com/exercism/go/issues/2011#issuecomment-1001638982 for details.

junedev:
@enarha There was a task to update the concept introduction accordingly but I think there was some issue with merging that change. I will re-open the respective issue and we will make sure the documentation on this will be improved.

PR #2033, opened 2021-12-27
Add field names to structs when failing tests for Structs tutorial

2121159i:
Hopefully, this will make it easier for learners to understand which values refer to which fields, especially if there are multiple fields of the same type.

Originally did this only for the `Drive` function, but it might be useful in all the other places as well. Lmk if that is fine with you.

Before:
```Car{5 95 2}```

After
```Car{"speed":5, "battery":95,  "batteryDrain":2}```

### DIFF
--- a/exercises/concept/need-for-speed/need_for_speed_test.go
+++ b/exercises/concept/need-for-speed/need_for_speed_test.go
@@ -28,7 +28,7 @@ func TestNewCar(t *testing.T) {
 			got := NewCar(tt.car.speed, tt.car.batteryDrain)
 
 			if got != tt.expected {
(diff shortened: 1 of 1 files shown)


junedev (approved):
I think this is a good improvement, thanks!

2121159i:
@junedev Sweet, thanks for taking a look 🙇 . I see that a build step is failing, though when tested locally seems to pass.
Also I do not have write access to the repo, so feel free to merge whenever is suitable for you.

andrerfcsantos:
@2121159i These failures are unrelated with your change. This seems to be a temporary problem with Github Actions: https://www.githubstatus.com/incidents/gh0vvxtlj5v7

We are monitoring it and we are performing re-runs on the jobs periodically. When one of the runs passes, we'll merge this!

andrerfcsantos (approved):

PR #2040, opened 2021-12-28
Correct typo in Chessboard exercise

jessealter:
The exemplar says that ranks are accessed by values from "A" to "H",
but maps are accessed by keys.


andrerfcsantos (approved):
Thanks for this.

PR #2044, opened 2021-12-30
Escape asterisks

rogeriobastos:
Escaping asterisks in instruction text cause it is being interpreted as special caracter.


junedev (approved):
That's a very helpful fix, thanks!

PR #2045, opened 2021-12-30
Fix broken link in nucleotide-count's description

telemachus:
The link at the end of the description goes to a GitHub page that is out
of date—the link is a 404. In addition, I think it makes more sense to link
to the leap exercise on Exercism's own site rather than to a raw markdown
document on GitHub.

I've updated the link accordingly.


exercises/practice/nucleotide-count/.docs/instructions.append.md
```diff
@@ -16,7 +16,7 @@ Take a look at the test cases to get a hint about what could be the possible inp
 
 ## note about the tests
```
junedev, line 17:
```suggestion
## Note about the tests
```

junedev (approved):
Thanks! I added a fix for the header.

PR #2048, opened 2021-12-30
Simplified Linked List Reverse

lpedras:
Simplified Linked List Reverse
The new implementation only iterates over the list once and it's easier to understand (IMO)
Feel free to suggest improvements or reject the change.

### DIFF
--- a/exercises/practice/linked-list/.meta/example.go
+++ b/exercises/practice/linked-list/.meta/example.go
@@ -100,31 +100,12 @@ func (ll *List) Reverse() {
 		return
 	}
 
-	// construct singly-linked list from the back
-	dummy := NewNode(-1)
-	cur := dummy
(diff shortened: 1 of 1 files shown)


andrerfcsantos (approved):
Thanks! Since the solution you propose is a common solution for this problem, it's a good idea to be represented in the exemplar.

junedev:
@lpedras Please be aware that the example files for practice exercises solely serve the purpose of running the tests in the CI process. There is no requirement for them to be good solutions, they are not presented anywhere on the website.
You can read the official documentation on this here: https://exercism.org/docs/building/tracks/practice-exercises#h-file-example-implementation

This is different for learning exercises, there the example solution should be a good solution and it is presented to the mentor.

I am ok with this being merged, I just want you to be aware of the bigger context.

PR #2051, opened 2022-01-01
Add benchmarks for `run-length-encoding` exercise

### LINKED ISSUE #2050: Run Length Encoding: Add Benchmarks
I noticed the `run-length-encoding` is missing benchmarks.

Benchmarks for this exercise have been requested by users before, and it's one of the exercises where it actually makes sense to have them.

The exercise requires some string building, and benchmarks can help to tell the difference between solutions using simple string concatenation + conversion of runes to string and solutions that just use [strings.Builder](https://pkg.go.dev/strings#Builder).

Tasks:

- [ ] Add benchmarks to [run_length_encoding_test.go](https://github.com/exercism/go/blob/main/exercises/practice/run-length-encoding/run_length_encoding_test.go)

jessealter:
Resolves #2050

### DIFF
--- a/exercises/practice/run-length-encoding/run_length_encoding_test.go
+++ b/exercises/practice/run-length-encoding/run_length_encoding_test.go
@@ -31,6 +31,9 @@ func TestRunLengthEncodeDecode(t *testing.T) {
 }
 
 func BenchmarkRunLengthDecode(b *testing.B) {
+	if testing.Short() {
+		b.Skip("skipping benchmark in short mode.")
+	}
 	for i := 0; i < b.N; i++ {
 		for _, test := range decodeTests {
 			RunLengthDecode(test.input)
@@ -39,6 +42,9 @@ func BenchmarkRunLengthDecode(b *testing.B) {
 }
 
 func BenchmarkRunLengthEncode(b *testing.B) {
+	if testing.Short() {
+		b.Skip("skipping benchmark in short mode.")
+	}
 	for i := 0; i < b.N; i++ {
 		for _, test := range decodeTests {
 			RunLengthEncode(test.input)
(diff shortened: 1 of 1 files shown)


exercises/practice/run-length-encoding/run_length_encoding_test.go
```diff
@@ -29,3 +29,19 @@ func TestRunLengthEncodeDecode(t *testing.T) {
 		t.Logf("PASS %s", test.description)
 	}
 }
+
+func BenchmarkRunLengthDecode(b *testing.B) {
+	for i := 0; i < b.N; i++ {
```
andrerfcsantos, line 34:
```suggestion
func BenchmarkRunLengthDecode(b *testing.B) {
	if testing.Short() {
		b.Skip("skipping benchmark in short mode.")
	}
	for i := 0; i < b.N; i++ {
```

exercises/practice/run-length-encoding/run_length_encoding_test.go
```diff
@@ -29,3 +29,19 @@ func TestRunLengthEncodeDecode(t *testing.T) {
...
+		for _, test := range decodeTests {
+			RunLengthDecode(test.input)
+		}
+	}
+}
+
+func BenchmarkRunLengthEncode(b *testing.B) {
+	for i := 0; i < b.N; i++ {
```
andrerfcsantos, line 42:
```suggestion
func BenchmarkRunLengthEncode(b *testing.B) {
	if testing.Short() {
		b.Skip("skipping benchmark in short mode.")
	}
	for i := 0; i < b.N; i++ {
```

andrerfcsantos (changes requested):
Awesome, thanks! Just checked and the tests run in a reasonable time (<2s) and do allow to see the differences in solutions.

There's just a little detail that I forgot to mention in the issue: for the test runner, we are running the tests in short mode, which allows to skip benchmarks while testing.

https://github.com/exercism/go-test-runner/blob/5496a4d0f33880ab0b332d20af9aca811bc506fa/testrunner/execute.go#L259

However, this requires each benchmark to check for this flag and skip the benchmark if it is present. See below the suggestions adding the code to do this.

jessealter:
Thank you for the feedback and info!

andrerfcsantos (approved):

PR #2052, opened 2022-01-01
Update introduction.md

piyush1104:
Should be `PascalCase` instead of `CamelCase`.


andrerfcsantos (approved):
Good catch, thanks for this!

junedev:
@piyush1104 As the automated comment explained, this mistake also exists in two other places. As this PR was already merged, maybe you could create another one that applies the fix to the remaining files.

Issue #2047, opened 2021-12-30
elons-toys: minor tweaks needed


fmmatheus:
### Description

Minor tweaks found on `elons-toys` concept exercise:

1. Typo on `README.md` file, line 56. "On" instead of "own": 

> "All the methods we have seen so far have a value receiver which means they `on` a copy of the value passed to the method, meaning that any modification done to the receiver inside the method is not visible to the caller."

2. Wrong return example on items 2 and 3. Instructions and tests require `string` type return:

```go
car.DisplayBattery()
// Output: "Battery at 100%"

//and

car.DisplayDistance()
// Output: "Driven 0 meters"
``` 

3. Tests are sorted differently from the exercise order:

```
|       Exercise       |      |           Tests          |
| 1. Drive()           |      | 1. TestDrive()           |
| 2. DisplayDistance() |  --> | 2. TestCanFinish()       |
| 3. DisplayBattery()  |      | 3. TestDisplayDistance() |
| 4. CanFinish()       |      | 4. TestDisplayBattery()  |
```

### Suggestions

1. Simple typo fix
2. Either remove the returning `string` type and require student to use `fmt.Print()` to display the message or change the examples to:
```go
msg := car.DisplayBattery()
fmt.Print(msg)
// Output: "Battery at 100%"
```
3. Use exercise order on tests

andrerfcsantos:
These are good points.

1. I'll probably change the text to something simpler like "they will receive a copy...". I don't think "own" is clear language, as it is open to interpretation and in other languages it has a different meaning, i.e, in Rust variables can be owned by functions, but that is not related with copy (at least directly).
2. I would just change the example
3. Sound good.

PR #2053, opened 2022-01-01
elons-toys: minor tweaks

est5:
FIxes: #2047
1. Text changed to "...will receive a copy..."
2. fmt.Println is used to display a message
3. The order of the tests has been changed


andrerfcsantos (approved):
Thanks for working on this!

PR #2054, opened 2022-01-02
Adjust generators to v3 folder structure

### LINKED ISSUE #2043: Make generators run under new directory structure
run-generators script is not aware of the new directory structure; seems it expects to find all exercises under exercises/* , while they have been splitted in concept, practice, shared folders. 
When ran manually, each generator gives error about missing import
```
Generating acronym
gen.go:7:2: cannot find package "." in:
	/home/andrea/projects/exercism-go/exercises/gen

```

andrerfcsantos:
Fixes #2043

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,7 +222,7 @@ $ tree -L 1 .
 ```
 
 To regenerate the test cases, navigate into the **go** directory and run
-`go run exercises/<exercise>/.meta/gen.go`. You should see that the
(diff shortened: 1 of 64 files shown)


junedev (approved):
Thanks for fixing this! Merge whenever you like.

We might want to revisit the "module off" flag once the support for workspaces with multiple modules landed in Go. Maybe then there is a better solution than the relative import path and turning off modules.

PR #2057, opened 2022-01-03
Lasagna Master: Improve task 3 (AddSecretIngredient)

### LINKED ISSUE #2036: Lasagna Master: Improve task 3 (AddSecretIngredient)
Lasagna Master was ported from JavaScript. Task 3 (`AddSecretIngredient`) was meant to make the student practice that in some cases the argument can/will be modified and the modification is visible outside of the function. In contrast, in task 4 the student practices NOT to modify the argument.

Because of the 1to1 port from JavaScript, the practice goal for task 3 got lost because currently `append` is used which creates a new slice so the arguments are not modified. The practice goal can be restored by changing the task so that the first list always ends with an element `"?"` and the secret ingredient is supposed to be inserted in that place. The function should not have any return value.
The solution could look like this:
```go
func AddSecretIngredient(friendsList, myList []string) {
	myList[len(myList)-1] = friendsList[len(friendsList)-1]
}
```

To implement this change, make sure to adjust all relevant exercise files.
See https://exercism.org/docs/building/tracks/concept-exercises for the general structure.

sougat818:
Fix #2036

1. Update to README.md to install golint
2. Update AddSecretIngredient tests to update myList instead of returning 
3. Update exemplar, instructions and hints

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,7 @@ You should have a [recent version of Go](http://golang.org/doc/install) installe
 You will need a github account and you will need to fork exercism/go to your account.
 See [GitHub Help](https://help.github.com/articles/fork-a-repo/) if you are unfamiliar with the process.
 Clone your fork with the command: `git clone https://github.com/<you>/go`.
-Test your clone by cding to the go directory and typing `bin/fetch-golangci-lint` and then
-`bin/test-without-stubs`. You should see tests pass for all exercises.
+Test your clone by cding to the go directory and typing `bin/test-without-stubs`. You should see tests pass for all exercises.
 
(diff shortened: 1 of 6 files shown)


andrerfcsantos (changes requested):
Thanks for working on this!
Left some suggestions.

andrerfcsantos (approved):

andrerfcsantos:
One last thing: for your contribution to this exercise, feel free to add your github handle to the contributors of the exercise: https://github.com/exercism/go/blob/main/exercises/concept/lasagna-master/.meta/config.json#L4

PR #2058, opened 2022-01-03
Replace CamelCase with PascalCase

piyush1104:
Follow-up of: #2052


andrerfcsantos (approved):
Thanks for fixing this typo it in these places too - my bad for missing them originally.

PR #2062, opened 2022-01-05
Lasagna Master: Fix references to "array"

csutter:
The instructions for _Lasagna Master_ say "array(s)" in several places
when referring to slices, which might be confusing given that they are
different, incompatible data structures in Go.

### DIFF
--- a/exercises/concept/lasagna-master/.docs/hints.md
+++ b/exercises/concept/lasagna-master/.docs/hints.md
@@ -2,7 +2,7 @@
 
 ## 1. Estimate the preparation time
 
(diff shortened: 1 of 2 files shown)


andrerfcsantos (changes requested):
Thanks for this!

References to arrays instead of slices also appear in [hints.md](https://github.com/exercism/go/blob/main/exercises/concept/lasagna-master/.docs/hints.md). Can you please also change the references to arrays there to references to slices?

andrerfcsantos (approved):

PR #2063, opened 2022-01-08
Reference correct function name

hardbap:
The function is named `incrementPeterAge` (not `addPeterAge`).


andrerfcsantos (approved):
Thanks for this, well spotted! This function was previously called `addPeterAge` and we missed to update this referenced to it.

PR #2064, opened 2022-01-08
prime-factors: allow empty nil slices

andrerfcsantos:
For this exercise, a solution that simply creates a slice with its default value fails the tests:

```go
func Factors(n int64) []int64 {
	var factors []int64
	for i := int64(2); i <= n; i++ {
		for n%i == 0 {
			factors = append(factors, i)
			n /= i
		}
	}
	return factors
}
```

```console
$ go test
--- FAIL: TestPrimeFactors (0.00s)
    prime_factors_test.go:15: FAIL no factors
        Factors(1) = []int64(nil);
        expected []int64{}
FAIL
exit status 1
FAIL    prime   0.001s
```

In the previous solution, changing the declaration of `factors` to `factors :=  make([]int64, 0)` will make the tests pass.

This PR makes the slice comparison not care if an empty slice is created explicitly created with make, or has its `nil` default value.

### DIFF
--- a/exercises/practice/prime-factors/prime_factors_test.go
+++ b/exercises/practice/prime-factors/prime_factors_test.go
@@ -1,7 +1,6 @@
 package prime
 
 import (
-	"reflect"
 	"sort"
 	"testing"
 )
@@ -11,7 +10,7 @@ func TestPrimeFactors(t *testing.T) {
 		actual := Factors(test.input)
 		sort.Slice(actual, ascending(actual))
 		sort.Slice(test.expected, ascending(test.expected))
-		if !reflect.DeepEqual(actual, test.expected) {
+		if !slicesEqual(actual, test.expected) {
 			t.Fatalf("FAIL %s\nFactors(%d) = %#v;\nexpected %#v",
 				test.description, test.input,
 				actual, test.expected)
@@ -29,6 +28,24 @@ func BenchmarkPrimeFactors(b *testing.B) {
 			Factors(test.input)
 		}
 	}
+}
+
+func slicesEqual(a, b []int64) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	if len(a) == 0 {
+		return true
+	}
+
+	for i := 0; i < len(a); i++ {
+		if a[i] != b[i] {
+			return false
+		}
+	}
+
+	return true
 }
 
 func ascending(list []int64) func(int, int) bool {


junedev (commented):
@andrerfcsantos I think it is valuable to force the student to really return the empty slice here. This has nothing to do with using make or not, it's the difference between `var a []int` and `a := []int{}`.

In the real world, e.g. when you later produce JSON output, it makes a difference whether you used the nil or empty version (one will become `null`, the other one `[]`) so it is good for a student to practice how to set up an actual empty slice. 

Additionally, I think the output of the function is more consistent with the empty slice instead of nil. At work, we also return an empty slice for example when searching for a collection in the database and there is no result with the given WHERE constraints.

Imo the solution you showed should really fail the test. If we want to help the student understanding the error message, we could add an append file explicitly explaining that an empty slice should be returned in case there are no factors. Additionally, we could double check that the slices concept explains the difference between the two options.

andrerfcsantos:
I understand the difference, but I don't think the difference should matter for this exercise. Your argument for wanting a non-nil empty slice is valid, but there's also the idea that in Go you should take advantage of the default value of the types. As an example, the [Uber Go Style Guide](https://github.com/uber-go/guide/blob/master/style.md#nil-is-a-valid-slice) is explicitly in favor of returning a nil slice instead of a non-nil slice of length 0. I'm not saying this advice should always apply, which is exactly why I think the difference should not matter for this exercise and we should not enforce a particular style.

The "right answer" will always depend on the context, and for this exercise there's not much context outside the function the student has to write. Forcing the student to return a non-nil empty slice would make more sense on an exercise where that actually matters, like the example you gave where the student would have to produce json.

junedev:
Since we are in disagreement here, I asked on Slack what others think. Maybe wait for a couple of days. If there was no additional objecting feedback then, you can merge.

junedev (approved):
As explained, I personally disagree here but I don't feel I have the right to block this change as @andrerfcsantos arguments are equally valid.

NobbZ:
Personally I'm in favour of *never* returning `nil` from a function, unless it returns a `*T`/interface or also returns an `error`/`bool` *and* `nil` would be the zerovalue of the primary return

tehsphinx:
I'm pretty sure the "rule" is to return `nil` over an empty slice. There is no downside to using `nil`, there are some downsides to using an empty slice:
- it takes more memory as it has a `len` and `cap` set, maybe even some capacity reserved
- it does not work with `omitempty` (json tags)
- ...

IIRC, one of the regrets of the Go team is to not have done the same with the `map` (to make it work even if `nil`). Then we'd also be able to pass around nil maps, the same as we do with `nil` slices.

marcofeltmann:
The main difference is: 
Ther *is* a list of prime numbers which has zero items **vs.** there is *no* list of prime numbers.

It only depends on business logic which solution is right and the test case should not assume any underlying business logic.

Therefor the test should check correctness of the result independently of its type: It should length-check the result for 0 entries.

—-
My opinion indeed is bound to years of knowledge in dynamically typed languages, so it feels natural to me handling empty slices and nil values the same way.

NobbZ:
But the type says `[]T`, so we should get a `[]T`, and not something that behaves like it most of the time…

We use a statically typed language, and not a duck-typed ruby/python here.

marcofeltmann:
Guess the "`nil` is evil" fraction is confusing things:
A `nil` slice is not a `nil` pointer. A `nil` slice is a `nil` slice.

Implementation detail `nil` slice:

```
[]int { pointsTo: nil, cap: 0, len 0}
```
Implementation detail empty slice:

```
[]int { pointsTo: &[0]int{}, cap: 0, len 0}
```

So it's kinda the same thing with some single improvement: Memory that has not been allocated, has not to be maintained by the garbage collector.

In addition Rob Pike says in his "Go Proverbs" talk: [Make the zero value useful.](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=6m25s)
Since Rob is 1/3 of the people who initially designed the language, we (a.k.a. "the test") should "make the zero value useful".

tehsphinx:
If the question is `Should we return nil over empty slice in production code`, the answer is: `Yes, it is convention to return nil rather than an empty slice`. That is the consensus in the Go community.

If the question is `Should we allow both cases to pass the tests`, then I think the answer should also be `Yes`. The tests should not force any student to return an empty slice over a nil (absolutely not!) and we might also not want to force the student to return `nil` over an empty slice, as this is more of a convention than a strict rule.

junedev:
@andrerfcsantos Judging by the discussion above, the solution you proposed seems the best way forward. Feel free to merge.

andrerfcsantos:
@junedev I'd like to keep this open for a few more days to let the chance for other people to participate in the discussion. I didn't imagine this as being such a polarizing issue, but the discussion so far has been productive in my opinion and I want to give the chance for more people to chime in. Even though the right decision does seem to allow both options, hearing more opinions can be beneficial to give us a better understanding of the issue and on how people feel about it.

I just want to add one more thing to the discussion: in allowing the students to write both options, there's also a great opportunity for a mentor to comment on that and to show an alternative way, explaining the differences. I see this as an advantage over a student just being forced into a particular solution just to make the tests pass without really understanding why.

bitfield:
As usual with issues like this, I think there's the answer that we would use in "real programs", and the answer that we want Exercism students to produce. For some people, those are often very different things.

It's a pretty well-established convention in Go to return zero _in the error case_. For functions that return `(T, error)`, that is, when the error is non-nil, the function should return the zero value of T. For slice types, that's `nil`. There's no argument about that, I'm sure. And in any case it doesn't really matter, because when we're _testing_ a function like that, we shouldn't even assert anything about the slice value if the error is non-nil. The error is _signalling_, if you like, that the other value is invalid and shouldn't be used.

There's a distinct, but related question of what should happen when the operation succeeds, but there are simply no elements in the answer. For example, listing the prime factors of 3. That's a valid question to _ask_, it's just that the answer contains no values.

If the function is supposed to return a slice, given valid input, then it must return a slice. `nil` is not a slice, any more than `nil` is a pointer, so `nil` wouldn't be appropriate here. An empty slice is what we're looking for.

Now, the issue of the difference between `nil` and an empty slice is a tricky one for Go beginners. They can get quite confused by it, especially since they _print_ the same. So I think it's reasonable to say this might be an opportunity to help make the difference clear to Exercism learners.

On the other hand, I can see the argument that we shouldn't knock people's confidence by failing their solution for what is, in practice, a pretty minor mistake, and one that not even all Go programmers would agree _is_ a mistake. (I'm sure there are some people out there happily returning nil slices in the non-error case, and their programs work just fine.)

So, while I think it's worth making sure _we're_ all clear about whether `nil` is a slice (it's not), I'm happy to leave the decision about what to do in this exercise to the hard-working volunteers who are actually implementing it.

marcofeltmann:
> `nil` is not a slice, any more than `nil` is a pointer, so `nil` wouldn't be appropriate here. An empty slice is what we're looking for.

Wrong. Totally and completely wrong.
`nil` is a totally perfect valid slice.

I repeat myself:
> Implementation detail nil slice:
> 
> ```
> []int { pointsTo: nil, cap: 0, len 0}
> ```
> 
> Implementation detail empty slice:
> 
> ```
> []int { pointsTo: &[0]int{}, cap: 0, len 0}
> ```

If you don't like to take my word for granted, than this is perfectly finde.
But what about [Fracesc Campoy's word](https://youtu.be/ynoY2xz-F8s?t=1217)?
Or ask the [corresponding official example](https://youtu.be/ynoY2xz-F8s?t=1217).
Or head over to the [official Golang blog entry](https://youtu.be/ynoY2xz-F8s?t=1217).
Or at least ask the [specification](https://go.dev/ref/spec#Slice_types)

But maybe you simply try it for yourself in this [playground](https://go.dev/play/p/Fa1mgDsdpEQ) I hacked together…

Whatever you do: please stop making assumptions sound like facts if they are not!

iHiD:
@marcofeltmann Turn the strength of your language down a notch here please. Bitfield has laid out his opinions here as he's entirely entitled to (and in fact, we definitely want his opinions!). He's clearly laid out two different ways of reasoning about this stuff (real world vs beginners, which is useful) and concluded that he doesn't have a strong enough opinion to warrant an opinion as he's not the one implementing this. His strength of wording is exactly the same as your strength of wording (He says "nil is a slice (it's not)", you say "nil is a totally perfect valid slice") - there's clearly a difference of opinion but @bitfield is not presenting himself any more forcefully than anyone else here. Please calmly explain why you disagree with the opinion stated, but please don't attack his post or anyone else's post because they disagree with you.

---

(Sidenote, the top three links all point to the same youtube video, and the four says that an "The value of an uninitialized slice is nil" - it doesn't say that `nil` is a valid slice and more than `nil` is a valid pointer - it uses exactly the same wording as it does for pointers - "The value of an uninitialized pointer is nil". However, I've never written a line of Go in my life and don't know what a "Slice" is, so I have no value or opinion to add on the subject).

mcastorina:
I think for a learning environment, allowing both solutions (`nil` and empty slice) is the way to go as it correctly implements the function definition. Returning a "list of primes with zero items" vs "no list of primes" is effectively the same to a student, and a mentor can provide guidance on Golang best practices as they see fit.

As a student I think it's too pedantic to differentiate between the two, at least for this exercise.

tehsphinx:
I was wondering if I should go deeper into the discussion if nil is a slice or not. It seems to me that both parties here agree that we shouldn't enforce either way on the student in the tests. I propose to move on with this MR, as that seems to be something we can agree on. 😄 

---

As for the discussion:

> nil is not a slice, any more than nil is a pointer

This is absolutely correct.

In Go however we cannot have `nil` without a type. `var1 := nil` does not compile. In the current case we have the type `[]int64`. Returning `nil` will make that a `nil slice of int64`.  Any way the Go language provides (that I of) on how to check if this is a slice or not always says: Yes, it is.

For example:
```
fmt.Println(reflect.TypeOf(a).Kind())
// Output: slice
```

Note: Yes, we can create a variable of type `interface{}` that seems to be just `nil`. But that is not the concrete case we have here. Then the function declaration would have to be `func Factors(n int64) interface{}`,  not `func Factors(n int64) []int64`. Won't go into more details on this as I don't think it makes sense to argue about the `interface{}` case.

andrerfcsantos:
I'm going to merge this as I feel we reached a satisfactory conclusion and we heard both sides of the argument. Thanks to everyone who took the time to contribute to share their opinion and provide valuable feedback. If anyone feels the need to add something relevant that we might have missed here, feel free to open a new issue referencing this PR.

PR #2066, opened 2022-01-11
Remove index 2 value from example output

bilal-sharifeh:
Remove index 2 value from example output as the ending index is exclusive in about.md

Fix the typo in the example output.
https://go.dev/play/p/6YZqXuuvvXa


andrerfcsantos (approved):
Thanks for this!

For future readers, the context for this PR is https://github.com/exercism/go/commit/1fe6434197d38f985d9755cd1c5b37f07936cdb4

PR #2069, opened 2022-01-16
Changed test inputs and example code for for parallel letter frequency

### LINKED ISSUE #1941: parallel-letter-frequency: have bigger test cases to show the benefits of concurrency
On `parallel-letter-frequency`, one thing that often students find weird and talk about is how the sequential solution is often more performant than the concurrent one.

This happens because we have small test cases. Not only we have few strings for which to concurrently compute the letter frequencies, but the strings themselves are small. This makes the overhead of creating goroutines/channels/waitgroups not worth performance-wise.

With this issue I intend to open the discussion if we should do this. I understand it's not essential for the exercise but I think it would be a nice touch for students paying attention to the benchmarks to see that their efforts to make the solution concurrent paid off, not only as a learning experience, but also in terms of performance.

kudligi:
Fixes #1941


kudligi:
Im unsure how to add labels to the PR for x:size  and the issue this related to.

exercises/practice/parallel-letter-frequency/parallel_letter_frequency_test.go
```diff
@@ -12,32 +12,38 @@ import (
...
-O'er the land of the free and the home of the brave?`
+	mark_twain = `You don't know about me without you have read a book by the name of The Adventures of Tom Sawyer;
+but that ain't no matter. That book was made by Mr. Mark Twain, and he told the truth, mainly.
+There was things which he stretched, but mainly he told the truth. That is nothing. I never seen anybody but lied one time or another,
+without it was Aunt Polly, or the widow, or maybe Mary. Aunt Polly - Tom's Aunt Polly, she is - and Mary,
+and the Widow Douglas is all told about in that book, which is mostly a true book, with some stretchers, as I said before.`
+
+	gary_provost = `This sentence has five words. Here are five more words. Five-word sentences are fine.
+But several together become monotonous. Listen to what is happening. The writing is getting boring.
+The sound of it drones. It’s like a stuck record. The ear demands some variety. Now listen. I vary the sentence length,
```
andrerfcsantos, line 42:
```suggestion
The sound of it drones. It's like a stuck record. The ear demands some variety. Now listen. I vary the sentence length,
```

Since you are using `'` for prepositions in the other texts, let's use it here too instead of `’`. `'` is more common in source code. They look very similar, but both vim and vscode were giving me warnings about this.

exercises/practice/parallel-letter-frequency/parallel_letter_frequency_test.go
```diff
@@ -12,32 +12,38 @@ import (
...
-Tochter aus Elysium,
-Wir betreten feuertrunken,
-Himmlische, dein Heiligtum!
-Deine Zauber binden wieder
-Was die Mode streng geteilt;
-Alle Menschen werden Brüder,
-Wo dein sanfter Flügel weilt.`
+	ralph_ellison = `I am an invisible man. No, I am not a spook like those who haunted Edgar Allan Poe;
+nor am I one of your Hollywood-movie ectoplasms. I am a man of substance, of flesh and bone,
+fiber and liquids—and I might even be said to possess a mind. I am invisible, understand,
```
andrerfcsantos, line 17:
```suggestion
fiber and liquids - and I might even be said to possess a mind. I am invisible, understand,
```

Replaced `—` by the more standard and regular dash `-`. You are using the regular dash in the other texts, so let's use it here too.

exercises/practice/parallel-letter-frequency/parallel_letter_frequency_test.go
```diff
@@ -12,32 +12,38 @@ import (
...
+simply because people refuse to see me. Like the bodiless heads you see sometimes in circus sideshows,
+it is as though I have been surrounded by mirrors of hard, distorting glass.
+When they approach me they see only my surroundings, themselves, or figments of their imagination—indeed,
```
andrerfcsantos, line 20:
```suggestion
When they approach me they see only my surroundings, themselves, or figments of their imagination - indeed,
```

Same as the previous comment.

andrerfcsantos (approved):
Thanks for working on this and to contribute to Exercism!

This looks good, and it seems to do what we want. I tested this code in my two computers and both report better results for the concurrent version:

```console
$ go test -bench .
goos: linux
goarch: amd64
pkg: letter
cpu: 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
BenchmarkSequentialFrequency-8             23349             49937 ns/op
BenchmarkConcurrentFrequency-8             26884             43835 ns/op
```

```console
$ go test -bench .
goos: windows
goarch: amd64
pkg: letter
cpu: Intel(R) Core(TM) i5-7400 @ 3.00GHz
BenchmarkSequentialFrequency-8             15970             74560 ns/op
BenchmarkConcurrentFrequency-8             19448             61857 ns/op
```

Just left some minor change suggestions related to less standard characters in the texts.

> Im unsure how to add labels to the PR for x:size and the issue this related to.

That instruction in the previous comment by the bot is aimed at mantainers/reviewers - you don't need to worry about it :)

exercises/practice/parallel-letter-frequency/parallel_letter_frequency_test.go
```diff
@@ -12,32 +12,38 @@ import (
...
 var (
-	euro = `Freude schöner Götterfunken
-Tochter aus Elysium,
-Wir betreten feuertrunken,
-Himmlische, dein Heiligtum!
-Deine Zauber binden wieder
-Was die Mode streng geteilt;
-Alle Menschen werden Brüder,
-Wo dein sanfter Flügel weilt.`
+	ralph_ellison = `I am an invisible man. No, I am not a spook like those who haunted Edgar Allan Poe;
```
junedev, line 15: Variable names should be camelCase.

andrerfcsantos (approved):

junedev:
@andrerfcsantos Can you wait with merging this for a bit please. I still see concurrent as slower on my machine with these changes.

kudligi:
@junedev @andrerfcsantos If we reuse the same 4 strings multiple times as inputs to increase the size of input, the performance gain is more significant.

This way the size to download also wont increase. 

I tried with 2 copies of every string

exercism-pr-benchmark

junedev:
@kudligi You are correct this would help but also it does not make a lot of sense to count something twice from the context of the exercise. Sometimes these things confuse students when they look at the tests. So instead I would suggest we use real longer texts. I tried with ca. 4 times the text length we have currently and that makes the concurrent solution about 30% faster and the test file is 12kB which is fine. Would you be willing to change the code to use the longer texts? Otherwise we can also merge this as is and I or someone else can make another PR later.

kudligi:
@junedev yeah that makes sense. Ill make the change.

andrerfcsantos:
@junedev I also tested in my server and could reproduce the concurrent version being slower there with the previous version of the code. It seems fixed with the most recent changes, now I see the concurrent version being faster everywhere, do you see these results on your side too?

andrerfcsantos (commented):
The weird dashes `–` are back. If you can, please replace them with regular dashes`-` with a search+replace. You should find 15 instances of `–`.

exercises/practice/parallel-letter-frequency/parallel_letter_frequency_test.go
```diff
@@ -12,32 +12,17 @@ import (
...
...ly be included under any classification? You see, you gentlemen have, to the best of my knowledge, taken your whole register of human advantages from the averages of statistical figures and politico-economical formulas. Your advantages are prosperity, wealth, freedom, peace–and so on, and so on. So that the man who should, for instance, go openly and knowingly in opposition to all that list would to your thinking, and indeed mine, too, of course, be an obscurantist or an absolute madman: would not he? But, you know, this is what is surprising: why does it so happen that all these statisticians, sages and lovers of humanity, when they reckon up human advantages invariably leave out one? They don't even take it into their reckoning in the form in which it should be taken, and the whole reckoning depends upon that. It would be no greater matter, they would simply have to take it, this advantage, and add it to the list. But the trouble is, that this strange advantage does not fall under any classification and is not in place in any list. I have a friend for instance ... Ech! gentlemen, but of course he is your friend, too; and indeed there is no one, no one to whom he is not a friend!`
```
andrerfcsantos, line 23: Remove a newline after the ` just so it's consistent with the others. In addition, I'm also wondering if we should split the text into several lines or alternatively have each text take just 1 line. What do you think @junedev ?

```suggestion
	dostoevsky3 = `But these are all golden dreams. Oh, tell me, who was it first announced, who was it first proclaimed, that man only does nasty things because he does not know his own interests; and that if he were enlightened, if his eyes were opened to his real normal interests, man would at once cease to do nasty things, would at once become good and noble because, being enlightened and understanding his real advantage, he would see his own advantage in the good and nothing else, and we all know that not one man can, consciously, act against his own interests, consequently, so to say, through necessity, he would begin doing good? Oh, the babe! Oh, the pure, innocent child! Why, in the first place, when in all these thousands of years has there been a time when man has acted only from his own interest? What is to be done with the millions of facts that bear witness that men, consciously, that is fully understanding their real interests, have left them in the background and have rushed headlong on another path, to meet peril and danger, compelled to this course by nobody and by nothing, but, as it were, simply disliking the beaten track, and have obstinately, wilfully, struck out another difficult, absurd way, seeking it almost in the darkness. So, I suppose, this obstinacy and perversity were pleasanter to them than any advantage.... Advantage! What is advantage? And will you take it upon yourself to define with perfect accuracy in what the advantage of man consists? And what if it so happens that a man's advantage, sometimes, not only may, but even must, consist in his desiring in certain cases what is harmful to himself and not advantageous. And if so, if there can be such a case, the whole principle falls into dust. What do you think–are there such cases? You laugh; laugh away, gentlemen, but only answer me: have man's advantages been reckoned up with perfect certainty? Are there not some which not only have not been included but cannot possibly be included under any classification? You see, you gentlemen have, to the best of my knowledge, taken your whole register of human advantages from the averages of statistical figures and politico-economical formulas. Your advantages are prosperity, wealth, freedom, peace–and so on, and so on. So that the man who should, for instance, go openly and knowingly in opposition to all that list would to your thinking, and indeed mine, too, of course, be an obscurantist or an absolute madman: would not he? But, you know, this is what is surprising: why does it so happen that all these statisticians, sages and lovers of humanity, when they reckon up human advantages invariably leave out one? They don't even take it into their reckoning in the form in which it should be taken, and the whole reckoning depends upon that. It would be no greater matter, they would simply have to take it, this advantage, and add it to the list. But the trouble is, that this strange advantage does not fall under any classification and is not in place in any list. I have a friend for instance ... Ech! gentlemen, but of course he is your friend, too; and indeed there is no one, no one to whom he is not a friend!`
```

exercises/practice/parallel-letter-frequency/parallel_letter_frequency_test.go
```diff
@@ -12,32 +12,15 @@ import (
...
...ctively afraid of attaining his object and completing the edifice he is constructing? Who knows, perhaps he only loves that edifice from a distance, and is by no means in love with it at close quarters; perhaps he only loves building it and does not want to live in it, but will leave it, when completed, for the use of les animaux domestiques - such as the ants, the sheep, and so on. Now the ants have quite a different taste. They have a marvellous edifice of that pattern which endures for ever - the ant-heap. With the ant-heap the respectable race of ants began and with the ant-heap they will probably end, which does the greatest credit to their perseverance and good sense. But man is a frivolous and incongruous creature, and perhaps, like a chess player, loves the process of the game, not the end of it. And who knows (there is no saying with certainty), perhaps the only goal on earth to which mankind is striving lies in this incessant process of attaining, in other words, in life itself, and not in the thing to be attained, which must always be expressed as a formula, as positive as twice two makes four, and such positiveness is not life, gentlemen, but is the beginning of death.`
```
junedev, line 19: @andrerfcsantos What do you think about the formatting here? Should line breaks be used? Not sure how the editor will display this on the website.

(I will run the benchmarks again later.)

andrerfcsantos: I was wondering the same thing. I think we should either put all the texts in one line to allow an easier access to the test code, or split it into newlines with 80ish characters per line to make the strings more readable. I'm fine with either option more tempted to go with the second one, as the texts we have are not **that** big.

kudligi: reduced the max character count per line to around 100 charcters

junedev:
@andrerfcsantos Looks really good with the latest version on my machine:
```
BenchmarkSequentialFrequency
BenchmarkSequentialFrequency-12             2168            467806 ns/op
BenchmarkConcurrentFrequency
BenchmarkConcurrentFrequency-12             3704            279518 ns/op
```

junedev (approved):

junedev:
@kudligi Thanks for bearing with us here as we sort this out and for the quick iterations. Your PR is merged now. :tada:

kudligi:
Thanks @junedev  @andrerfcsantos ! :)

PR #2073, opened 2022-01-20
Fix the type conversion example

orsenthil:
The original example was using a type `int` and will fail with compiler error : `type int is not an expression`


andrerfcsantos (approved):
Thanks for this, good catch!

PR #2076, opened 2022-01-23
remove valid roman numeral marked as invalid

joaofnds:
I know the instruction states that _There is no need to be able to convert numbers larger than about 3000_, but I shouldn't be forced to return an error for a valid number. IMO if there's no need to convert numbers higher than 3000, we should just not test them.

### DIFF
--- a/exercises/practice/roman-numerals/.docs/instructions.md
+++ b/exercises/practice/roman-numerals/.docs/instructions.md
@@ -20,8 +20,7 @@ into stone tablets).
  7  => VII
 ```
 
-There is no need to be able to convert numbers larger than about 3000.
-(The Romans themselves didn't tend to go any higher)
+The largest number that can be represented in this notation is 3,999.
 
 Wikipedia says: Modern Roman numerals ... are written by expressing each
 digit separately starting with the left most digit and skipping any
--- a/exercises/practice/roman-numerals/cases_test.go
+++ b/exercises/practice/roman-numerals/cases_test.go
@@ -30,4 +30,5 @@ var romanNumeralTests = []romanNumeralTest{
 	{911, "CMXI", false},
 	{1024, "MXXIV", false},
 	{3000, "MMM", false},
+	{3999, "MMMCMXCIX", false},
 }
--- a/exercises/practice/roman-numerals/roman_numerals_test.go
+++ b/exercises/practice/roman-numerals/roman_numerals_test.go
@@ -6,7 +6,6 @@ func TestRomanNumerals(t *testing.T) {
 	tc := []romanNumeralTest{
 		{0, "", true},
 		{-1, "", true},
-		{3001, "", true},
 	}
 	tc = append(tc, romanNumeralTests...)
 
--- a/exercises/practice/roman-numerals/roman_numerals_test.go
+++ b/exercises/practice/roman-numerals/roman_numerals_test.go
@@ -6,6 +6,7 @@ func TestRomanNumerals(t *testing.T) {
 	tc := []romanNumeralTest{
 		{0, "", true},
 		{-1, "", true},
+		{4000, "", true},
 	}
 	tc = append(tc, romanNumeralTests...)
 


andrerfcsantos (changes requested):
Technically, the biggest number that can be written with roman numerals is 3999, so we shouldn't give an error for numbers that are lower than this. However, I think we should still test for some upper limit, because converting a number bigger than 3999 should indeed be an error as any produced result will be necessarily wrong otherwise. 

I suggest changing the test to something like `{4000, "", true}`. This way we are testing for the correct upper limit.

junedev:
@joaofnds Please read the first comment in the automated feedback above. This applies to the exercise you are trying to change. The test case are even generated automatically here that means next time someone runs the generator,  manual changes (as the ones proposed here) might be overwritten.

We want to keep our exercises in sync with the problem-spec version to benefit from the improvements made their. In problem spec, the problem you mentioned was addressed just like you suggested it should be, it does not test any "out of range" numbers, there is no error handling required.

So the actual task here is to sync the exercise with the problem spec version. We want to check and do this for all exercises soon but we are waiting for a new release of our tooling for that (configlet).

I will leave it up to @andrerfcsantos to merge these changes nevertheless because the current version is wrong but you need to be aware your code will vanish when we run the next sync.

exercises/practice/roman-numerals/.docs/instructions.md
```diff
@@ -20,8 +20,7 @@ into stone tablets).
  7  => VII
 ```
 
-There is no need to be able to convert numbers larger than about 3000.
-(The Romans themselves didn't tend to go any higher)
+The largest number that can be represented in this notation is 3,999.
```
andrerfcsantos, line 23: Like @junedev  mentioned previously, changes to this file must be discussed first in the problem specifications repo, so we won't make them here.

exercises/practice/roman-numerals/cases_test.go
```diff
@@ -30,4 +30,5 @@ var romanNumeralTests = []romanNumeralTest{
 	{911, "CMXI", false},
 	{1024, "MXXIV", false},
 	{3000, "MMM", false},
+	{3999, "MMMCMXCIX", false},
```
andrerfcsantos, line 33: The `cases_test.go` files are automatically generated with our test generator, which in turn takes as reference the test cases specified in problem specs. For that reason, like in the previous change, tests to this file must also be discussed and added first in the problem spec repo.

Personally, I'm against adding this test right now. Since the instructions mention to not go higher than `3000`, most people I've seen solving this exercise logically do check if the value is less than `3000`. Adding this test will necessarily break all those solutions and will force people to revisit the exercise if they want to keep it up-to-date, which I think it's unnecessary for such a small change. Our exemplar solution also does this check, that's why CI is failing right now.

I would consider adding this test if the description of the problem would change too, but like previously mentioned, both the change to the description and the new test must be discussed in the problem spec repo.

exercises/practice/roman-numerals/roman_numerals_test.go
```diff
@@ -6,7 +6,7 @@ func TestRomanNumerals(t *testing.T) {
 	tc := []romanNumeralTest{
 		{0, "", true},
 		{-1, "", true},
-		{3001, "", true},
+		{4000, "", true},
```
andrerfcsantos, line 9: I agree with adding this test, as it won't break any existing good solutions and does properly check if the solution is doing some kind of checks to the number.

Numbers equal or bigger than `4000` will be an error, whether someone is just checking for values lower `3000` influenced by the description or for values lower than `4000` if they read a bit more on the topic.

andrerfcsantos (changes requested):
Suggested some changes. In summary, we should just keep the new test in `roman_numerals_test.go`

joaofnds:
I wasn't aware of all this generated code (: The right solution, as I see it, is to discuss the upper bound in the problem-spec repo, so I'll just close this PR for now as this is not the place for this discussion. Thanks for all the info, you guys are awesome!

PR #2079, opened 2022-01-25
Modifying a test case to verify that car cannot finish the track distance

neelflash07:
There are multiple community solutions where the _**"CanFinish"**_ method has a check to keep driving the car until the battery is less than 1. But that logic is incorrect because the car can be driven only when:
`car.battery > car.batteryDrain`.

This case does not get tested because all the test cases are such that either battery is a multiple of batteryDrained or 
`battery - (batteryDrained*some_multiple)=1`.  Hence the modification to the test case.

Here, for initial condition of,
`car: Car{
				speed:        2,
				batteryDrain: 6,
				battery:      100,
				distance:     5,
			},
			track: Track{
				distance: 80,
			}`
After car is driven completely,
`car: Car{
				speed:        2,
				batteryDrain: 6,
				battery:      4,
				distance:     37,
			},`
For solutions with logic "drive car until battery is less than 1" will fail as battery != 1 but battery < batteryDrain.

### DIFF
--- a/exercises/concept/need-for-speed/need_for_speed_test.go
+++ b/exercises/concept/need-for-speed/need_for_speed_test.go
@@ -173,7 +173,7 @@ func TestCanFinish(t *testing.T) {
 			name: "Car cannot finish the track distance.",
 			car: Car{
 				speed:        2,
-				batteryDrain: 10,
+				batteryDrain: 6,
 				battery:      100,
 				distance:     5,
 			},


junedev:
@neelflash07 Welcome to contributing to the Go track!

I am not 100% sure I understood you correctly. Could you provide a link to one of those community solutions you mentioned that are wrong but still pass the tests currently? That would be very helpful.

neelflash07:
@junedev With reference to your previous comment, PFA the link to the community solutions:
[Solution1](https://exercism.org/tracks/go/exercises/need-for-speed/solutions/andybroger)
[Solution2](https://exercism.org/tracks/go/exercises/need-for-speed/solutions/lnds)

In each of these solutions, the car is driven as long as `car.battery > 1`. That is a wrong condition. In reality, the car should only be driven as long as `car.battery > car.batteryDrain`. The update to the test case made will check for this condition.

junedev (approved):

junedev:
Thanks for the links and good catch on noticing the issue with the test case here! Your change is merged now. It will be available on the website shortly. If someone started the exercise already, they will see this orange banner/icon that there is an update available.

Issue #1997, opened 2021-12-13
Implement new Concept Exercise: Error Wrapping


junedev:
## Getting Started

If you have not yet contributed to concept exercises before, this task requires some upfront reading to acquire the necessary background knowledge.

Here you can read about what Concept Exercises are and how they are structured:

- [What are Concept Exercises?](https://exercism.org/docs/building/product/concept-exercises)
- [Concept Exercises](https://exercism.org/docs/building/tracks/concept-exercises)
- [Concepts](https://exercism.org/docs/building/tracks/concepts)

Also be aware of these general guidelines.

- [Exercism Markdown Specification](https://exercism.org/docs/building/markdown/markdown)
- [Exercism Formatting and Style Guide](https://exercism.org/docs/building/markdown/style-guide)

## Goal

The goal here is to create a new concept exercise that teaches error wrapping. That includes writing a concept and creating a new exercise.

## Concepts

The following concept needs to be created. You can use the introduction.md file of the concept also as introduction.md file of the exercise. No need to create different content at this point. Additionally, if you want to save some time it is ok to not have an extensive about.md at this point. It can also be mainly the instruction.md content, maybe with some additions you would like to make.

- `error-wrapping`

## Learning Objectives

In the concepts the student should learn about the following topics and then practice them in the concept exercise.

- understanding the error chain
- how/when to use `errors.Is` and `errors.As`
- how to wrap an error
- the `Unwrap` method for custom errors

The about.md file could additionally mention ...

- how to get a stack strace

We could potentially add some other advanced error topics if there are ideas.

## Out of Scope

- error basics ([Concept errors](https://github.com/exercism/go/blob/main/concepts/errors/introduction.md) for what was already covered)

## Prerequisites

- `errors`
- `interfaces`
- `type-assertions`

Other prerequisites should be added as needed to solve the specific exercise.

## Story Idea

There is probably no exercise that can be ported directly because Go's error handling is so special but here are some error related exercises from other tracks. The story could maybe be re-used even if we need to change the actual tasks.

- [JavaScript "Factory Sensors" Exercise](https://github.com/exercism/javascript/blob/main/exercises/concept/factory-sensors/.docs/instructions.md)
- [C# "Calculator Conundrum" Exercise](https://github.com/exercism/csharp/blob/main/exercises/concept/calculator-conundrum/.docs/instructions.md)
[...]

PR #2080, opened 2022-01-27
Error encapsulation

Lewiscowles1986:
Part 1 of addressing #2011 

The exercise that follows this document asks to encapsulate one error, but not any others.
Both the knowledge of how to encapsulate, as well as that it is a requirement seemed absent to me.

### DIFF
--- a/concepts/errors/about.md
+++ b/concepts/errors/about.md
@@ -163,7 +163,7 @@ if _, err := os.Getwd(); err != nil {
 
 ## Encapsulating error types and values
 
-While it is advised above, to early return most errors that occur. There are occasions where defining behavior upon a specific type, or exported value of error is more appropriate. When that happens, only specific error types would not be immediately returned. This Encapsulation of errors is not unique to golang, but can be thought of as analogous to, catching an Exception in other languages. The purpose might be to differentiate between things in-scope of a library or framework; and things that have not (perhaps yet) been considered; or are deliberately excluded from a domain or context.
+While it is advised above, to early return most errors that occur. There are occasions where defining behavior upon a specific type, or exported value of error is more appropriate. When that happens, only the specified error types or values would not be immediately returned. This Encapsulation of errors is not unique to golang, but can be thought of as analogous to, catching an Exception in other languages. The purpose might be to differentiate between things in-scope of a library or framework; and things that have not (perhaps yet) been considered; or are deliberately excluded from a domain or context.
 
 ```go
     if err != nil && err != ErrSpecificCaseWeHandleOrEncapsulateWithOurOwnError {
--- a/concepts/errors/about.md
+++ b/concepts/errors/about.md
@@ -163,7 +163,7 @@ if _, err := os.Getwd(); err != nil {
 
 ## Encapsulating error types and values
 
-While it is advised above, to early return most errors that occur. There are occasions where defining behavior upon a specific type, or exported value of error is more appropriate. When that happens, only specific error types would not be immediately returned. This Encapsulation of errors is not unique to golang, but can be thought of as catching an Exception in other languages. The purpose might be to differentiate between things in-scope of a library or framework; and things that have not (perhaps yet) been considered.
+While it is advised above, to early return most errors that occur. There are occasions where defining behavior upon a specific type, or exported value of error is more appropriate. When that happens, only specific error types would not be immediately returned. This Encapsulation of errors is not unique to golang, but can be thought of as analogous to, catching an Exception in other languages. The purpose might be to differentiate between things in-scope of a library or framework; and things that have not (perhaps yet) been considered; or are deliberately excluded from a domain or context.
 
 ```go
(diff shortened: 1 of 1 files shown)


Lewiscowles1986:
If acceptable to exercism; I can squash the commits. I was using exercism on the web and decided to stay in the web, where I'm not aware of squash or commit amend functionality.

andrerfcsantos:
Hi,

Thanks for the PR.

In my opinion, this PR doesn't add much, and I'm having some issues trying to see what a student might get from this. My first issue is with the term "encapsulation". For me, encapsulation means hiding implementation details by only exposing behavior, which ideally is detached from a concrete implementation and I don't see how it applies here.

I find the description to be a bit vague by not providing a concrete example on how this can be useful. The description tries to sell the idea that we must check for a particular error and do something else with it other than returning it. My question when reading this is: ok, so what should we do instead? By not providing a concrete example of that "something else" we should do with the error, In the example from the "Checking for Errors" section of this file, we already provide an example on how to do something else when the error is of a particular error and I'm failing to see how this is different.

I think you are trying to introduce the student to the concept that we can check for certain error types and do something different with them based on its type, and I think that's why you are comparing them to exceptions. However, this topic is probably a better fit for the [Error Wrapping](https://github.com/exercism/go/issues/1997) concept where we can properly introduce the student to error type checking.

> If acceptable to exercism; I can squash the commits. I was using exercism on the web and decided to stay in the web, where I'm not aware of squash or commit amend functionality.

In your branch, it's personal choice whether you create several commits or squash/amend them to make sure it all fits into just one. When merging, we squash all the commits, so all PR's end up being just 1 commit regardless.

Lewiscowles1986:
Hi.

> trying to see what a student might get from this

The wish/intent is for a student to understand a thing they will be asked to do in the task which covers errors as a concept, which I believe was not explained enough for new learners. In that task, a specific type of error is encapsulated and hidden within the task; which does not mention to forward all errors, but one; but readily requires it as far as I can see.

> I find the description to be a bit vague by not providing a concrete example on how this can be useful.

My apologies for failing to capture the use of this. My attempt was to give enough information, without doing the work for the learner. As I have before, and I am sure I will again, I clearly failed to convey this:

> The purpose might be to differentiate between things in-scope of a library or framework; and things that have not (perhaps yet) been considered; or are deliberately excluded from a domain or context.

To my mind, the only reasons not to forward an error to a user, are to:

* Substitute it for your own internal error. Perhaps raising a socket closed to be a HTTP exception in a library _encapsulating_ the complexities of low-level sockets. (I did not want to give such a concrete example of this, as it's not why the exception is caught in the follow-up exercise.
* Define behavior, which is appropriate to a specific domain, or context.

I did not stack the error, by passing it to an error, because the task does not demonstrate this.

Can you think of other reasons this is a useful pattern? I Too am always trying to learn new things 😉. 

In any case. I Can accept this isn't what is needed, but I might need some help to improve it. Alternatively if nobody has time, I am happy to close this, as I know it's not quite enough, but don't think I can think of a clearer, more concise explanation; than I've provided.

Best

junedev:
Just as @andrerfcsantos, I also think "Encapsulation" and the coding pattern shown in this PR are two separate concerns. For example, in the following code, there is no encapsulation happening.
```
var ErrNotFound = errors.New("not found")

// ...

if err != nil && err != ErrNotFound {
  // e.g. write a 500 status code for an http response
}
```
Encapsulation should be discussed in the error wrapping concept.

Putting that aside, the other question remains whether or not we want to present the pattern of "returning for all but one specific error" (`if err != nil && err != ErrSomethingSpecific`) in the introduction. This pattern avoids nested if statements in the error handling in the case of this specific exercise (see [examplar solution](https://github.com/exercism/go/blob/main/exercises/concept/the-farm/.meta/exemplar.go#L21) vs [common community solution](https://exercism.org/tracks/go/exercises/the-farm/solutions/bndkn24)). My feeling is that this pattern is not very common as usually you can just handle the specific errors first and handle the "rest of the errors" last. It is only in our very specific case in this exercise that this does not work as we don't want to return in the case of the ErrScaleMalfunction but continue with the program execution. (Which leads us to yet another questions: Is it a good idea that the exercise requires the student to use a result even though there was an error?)

So I would vote for closing this PR and having a discussion about changing the exercise so that the student does not have to work with a result of a function even though there was an error. (I will leave a comment about this in #2011.)

andrerfcsantos:
> The wish/intent is for a student to understand a thing they will be asked to do in the task which covers errors as a concept

@Lewiscowles1986 I see. Then you are right in making a PR like this one, as the introduction must indeed have everything the student needs to solve the exercise.

However, like @junedev suggested in #2011, I also agree that instead of adding this to the description, we should just change the exercise so it doesn't need this.

For this reason I'll close this PR and forward further discussion to #2011.

Many thanks @Lewiscowles1986 for bringing this to our attention!

PR #2082, opened 2022-01-29
the-farm: fixes typo in one of the test descriptions

cwithmichael:
One of the tests returns 1, but in the description says it should return 5. This patch set just updates the description to say 1 instead.
```
{
	description:             "Scale returns 5 with ErrScaleMalfunction for 10 cows",
	weightFodder:            testWeightFodder{fodder: 5, err: ErrScaleMalfunction},
	weightFodderDescription: "5 fodder, ErrScaleMalfunction",
	cows:                    10,
	wantAmount:              1,
	wantErr:                 nil,
},
```


andrerfcsantos (approved):
Nice catch, thanks!

Issue #2085, opened 2022-02-01
Wrong mention of current domain being golang.org


davidandradeduarte:
In https://exercism.org/tracks/go/concepts/basics there's a mention to `golang.org` being the current domain, which isn't true at this point, since `go.dev` is the new domain.

Should be rephrased to:
... _Golang because of its **previous** domain name, golang.org, but the proper name is Go._ ...

junedev:
Good point. Would you like to provide a PR?

Here are the files where this needs to be changed:
- https://github.com/exercism/go/blob/main/concepts/basics/introduction.md
- https://github.com/exercism/go/blob/main/concepts/basics/about.md

PR #2086, opened 2022-02-01
fix: rephrasing introduction.md and about.md since golang.org was replaced by go.dev

davidandradeduarte:
Rephrasing introduction and about, since golang.org was replaced by go.dev.

Fixes https://github.com/exercism/go/issues/2085


junedev (approved):

junedev:
Thanks for the fix!

PR #2087, opened 2022-02-03
Readme row and column to be zero-indexed

mdrakiburrahman:
`saddle_points_test.go` assumes row and column is zero indexed. Readme starts at 1.

This is confusing - so changing readme to align to test suite.

### DIFF
--- a/exercises/practice/saddle-points/.docs/instructions.md
+++ b/exercises/practice/saddle-points/.docs/instructions.md
@@ -5,14 +5,14 @@ Detect saddle points in a matrix.
 So say you have a matrix like so:
 
 ```text
-    1  2  3
+    0  1  2
   |---------
-1 | 9  8  7
-2 | 5  3  2     <--- saddle point at column 1, row 2, with value 5
-3 | 6  6  7
+0 | 9  8  7
+1 | 5  3  2     <--- saddle point at column 0, row 1, with value 5
+2 | 6  6  7
 ```
 
-It has a saddle point at column 1, row 2.
+It has a saddle point at column 0, row 1.
 
 It's called a "saddle point" because it is greater than or equal to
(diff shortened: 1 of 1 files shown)


andrerfcsantos:
Hi @mdrakiburrahman,

The instructions from practice exercises come from the [problem-specifications](https://github.com/exercism/problem-specifications) repo, so any changes to the instructions of practice exercises should be suggested there. All tracks get their problem descriptions from there and the only thing each track can do is to append some additional instructions to the end of the exercise. Here in the Go track, we used this to add additional instructions to clarify that Go uses zero-based indexes:

> For all operations on the matrix, assume that rows and columns are zero-based. This means that first row will be row 0, the second row will be row 1, and so on.

The rationale for having the `1 2 3` notation on the description is because when talking about matrices we start at 1, and not 0. Start counting at 0 is very much a programmer thing. Historically, we did have `0 1 2` there, but that was changed [here](https://github.com/exercism/problem-specifications/pull/1387) to follow the canonical matrix notation. In short, these numbers aren't supposed to represent indexes, only the positions in the matrix, following matrix conventions. Even if they were indexes I'm not sure changing this to `0 1 2` would be the best approach, as some languages like Lua, do start their indexes at 1.

However, I understand this can still be confusing and maybe there's something we can do to clarify that those numbers don't represent indexes. Those changes should be suggested in the [problem-specifications](https://github.com/exercism/problem-specifications) repository and not here. For that reason, I'll close this PR. Thanks @mdrakiburrahman for pointing this out, and by all means feel free to open an issue in the problem specification repository for a follow-up discussion on what should be done here to improve the instructions.

mdrakiburrahman:
@andrerfcsantos - appreciate your thoughtful feedback, your thought process here makes sense.

Issue #1990, opened 2021-12-13
Card Tricks: Adapt to current concept tree progression


junedev:
## Getting Started

If you have not yet contributed to concept exercises before, this task requires some upfront reading to acquire the necessary background knowledge.

Here you can read about what Concept Exercises are and how they are structured:

- [What are Concept Exercises?](https://exercism.org/docs/building/product/concept-exercises)
- [Concept Exercises](https://exercism.org/docs/building/tracks/concept-exercises)
- [Concepts](https://exercism.org/docs/building/tracks/concepts)

## Motivation

Currently the concept exercise "Card Tricks" has some issues.
1. It requires the student to know about `for-loops` which are taught later in the [syllabus](https://exercism.org/tracks/go/concepts) and depend on `slices`.
2. It requires the student to know about variadic functions (`append` in this case) which were not introduced yet.
3. `multiple-return-values` is now part of the `functions` concept and should not be taught as part of Card Tricks anymore.

These issues should be addressed.

## Tasks

- [ ] Remove the task about `PrefilledSlice` from all the relevant exercise files (it requires for loops).
- [ ] Optional (can also be done later as a separate issue/PR): As replacement for `PrefilledSlice` it would be good to have another task that requires the student to create a slice. Additionally, the student should get a task that requires to define a variadic function.
- [ ] Change `GetItem` to have only one return value. In case the index is out of range, `-1` should be returned. Make sure to apply the change consistently (check hints, instructions etc.).
- [ ] Add the content of `concepts/variadic-functions/introductions.md` to the introductions file of the exercise. A good structure would be
  - `# Introduction`
  - `## Slices` (the content in slices should mention that append is a variadic function and refer to the details below)
  - `## Variadic Functions`
- [ ] Update the root level config.json file:
  - Remove `multiple-return-values` from the concepts taught by `card-tricks`.
  - Add `variadic-functions` to the concepts taught by `card-tricks`.
  - There are 2 other exercises that have `multiple-return-values` as prerequisite. Change the concept name to `functions` for those.
  
## Out of Scope

- Writing the `variadic-functions` concept is a separate issue.
- Entirely removing the `multiple-return-values` concept will be separate issue.

---

If you want to work on this issue, just leave a comment below and go ahead. No need to wait for a reply.

junedev:
I noticed another thing regarding this exercise. It would be good if `SetItem` would have a test case that checks that the student did not copy the slice.

shakhzodkudratov:
[...]

Issue #1373, opened 2021-01-28
[V3] Implement new Concept Exercise: panic-and-recover


tehsphinx:
This issue describes how to implement the `panic-and-recover` concept exercise for the Go track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

Please also watch the following video:

- [The Anatomy of a Concept Exercise](https://www.youtube.com/watch?v=gkbBqd7hPrA).

## Goal

The goal of this exercise is to teach the student how to work with `panic` and `recover` in Go.

## Learning objectives

- `panic` with error or other value
- `recover` a panic
- check the value type of the recover value
- `don't panic`: if, then only in `main` package to crash the program (e.g. if not all required services are reachable)
- other things that can panic: nil pointer dereference, index out of bounds, etc.
- A panic in the code is always a bug --> fix it! (don't use it as `try .. catch` alternative!)
- global panic handling in e.g. web-servers: their reason is to prevent potential bugs from crashing the entire server

## Out of scope

Nothing yet.

## Concepts

- `panic`
- `recover`

## Prerequisites

- `defer`
- `type-casting`

## Resources to refer to

- https://blog.golang.org/defer-panic-and-recover

## Implementing

Some function(s) that panic and the panic must be recovered in another function and turned into a proper error.

As an idea on how to teach "when not to use a panic/not to handle a panic but instead fix the bug":
Some functions could be already in the stub that panic on running the tests and the students task is to fix the problem. (they should fix the bug, not use `recover`)

## Help

If you have any questions while implementing the exercise, please post the questions as comments in this issue.

myworkcircle:
Hey guys,
Im a goloang developer, and also a learner from you website. I want to contribute to this issue

junedev:
[...]

PR #2091, opened 2022-02-06
add variadic function concept exercise

### LINKED ISSUE #1989: Create new Concept: Variadic Functions
## Getting Started

If you have not yet contributed to concept exercises before, this task requires some upfront reading to acquire the necessary background knowledge.

Here you can read about what Concept Exercises are and how they are structured:

- [What are Concept Exercises?](https://exercism.org/docs/building/product/concept-exercises)
- [Concept Exercises](https://exercism.org/docs/building/tracks/concept-exercises)
- [Concepts](https://exercism.org/docs/building/tracks/concepts)

## Motivation

The current exercise "Card Tricks" includes `append` (which is a variadic function) and the `x...` syntax when passing arguments to a variadic function. None of this is properly introduced currently. We want to add another concept to fix this. "Card Tricks" will then later be changed to teach `slices` and this new concept at the same time.

## Task

The task here is to create a new concept `variadic-functions` that explains what it means to prefix the last function parameter with `...` and what it means if you pass an argument with `...` as suffix.

## Out of Scope

Changing the "Card Tricks" exercise is out of scope for this issue, there will be a separate issue for this.

## Resources
Before writing this concept, please take a close look at the Go spec so the content you write will be correct:
- https://go.dev/ref/spec#Function_types
- https://go.dev/ref/spec#Passing_arguments_to_..._parameters

This article might be good for the links section:
- https://golangbot.com/variadic-functions/

myworkcircle:
+ Add variadic function concept exercise

Resolves #1989


concepts/variadic-functions/about.md
```diff
@@ -0,0 +1,57 @@
...
+        find(89, 89, 90, 95)
+        find(45, 56, 67, 45, 90, 109)
+        find(78, 38, 56, 98)
+        find(87)
+    }
+```
+The way variadic functions work is by converting the variable number of arguments to a slice of the type of the variadic parameter. 
+For instance, in line no. 22 of the program above, the variable number of arguments to the find function are 89, 90, 95. The find function expects a variadic int argument. 
+Hence these three arguments will be converted by the compiler to a `slice of type int []int{89, 90, 95}` and then it will be passed to the find function.
+
```
junedev, line 52: I think this file is missing an explanation for the second part of the task in https://github.com/exercism/go/issues/1989

> and what it means if you pass an argument with ... as suffix.

https://go.dev/ref/spec#Passing_arguments_to_..._parameters <- Second part of this section.

concepts/variadic-functions/about.md
```diff
@@ -0,0 +1,57 @@
+# About
+Functions in general accept only a fixed number of arguments. A variadic function is a function that accepts a variable number of arguments. If the last parameter of a function definition is prefixed by ellipsis ..., then the function can accept any number of arguments for that parameter.
+
+```go
+func hello(a int, b ...int) {
```
junedev, line 5: Maybe it makes sense to use `find` here as well. `hello` with these int params seems a bit strange. Sometimes these example that don't make sense make it harder for students to understand the content.

concepts/variadic-functions/about.md
```diff
@@ -0,0 +1,57 @@
+# About
+Functions in general accept only a fixed number of arguments. A variadic function is a function that accepts a variable number of arguments. If the last parameter of a function definition is prefixed by ellipsis ..., then the function can accept any number of arguments for that parameter.
```
junedev, line 2: We apply "one sentence per line" and code elements in the text should be written in backticks. See https://exercism.org/docs/building/markdown/markdown for reference.

I applied this here but this feedback also applies to some other places I think.

```suggestion
Functions in general accept only a fixed number of arguments.
A variadic function is a function that accepts a variable number of arguments.
If the last parameter of a function definition is prefixed by ellipsis `...`, then the function can accept any number of arguments for that parameter.
```

concepts/variadic-functions/about.md
```diff
@@ -0,0 +1,57 @@
...
+In the above function, parameter b is variadic and we can pass 0 or more arguments to b.
+
+```go
+    func hello(5, 6){}
+    func hello(5, 6, 7){}
+    func hello(5){}
+```
+
+Note:
+    Variadic paramter should be the last parameter of the function.
```
junedev, line 18: For this, you could use one of these "caution" blocks here: https://exercism.org/docs/building/markdown/markdown#h-special-blocks-sometimes-called-admonitions

concepts/variadic-functions/about.md
```diff
@@ -0,0 +1,57 @@
+# About
+Functions in general accept only a fixed number of arguments. A variadic function is a function that accepts a variable number of arguments. If the last parameter of a function definition is prefixed by ellipsis ..., then the function can accept any number of arguments for that parameter.
+
+```go
+func hello(a int, b ...int) {
+}
```
junedev, line 6: You could add `// ...` inside the function to indicate the function body.

concepts/variadic-functions/about.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+In the above function, parameter b is variadic and we can pass 0 or more arguments to b.
+
+```go
+    func hello(5, 6){}
+    func hello(5, 6, 7){}
+    func hello(5){}
```
junedev, line 14: There should be no `func` and `{}` for these as you want to show the function call.

concepts/variadic-functions/about.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+Note:
+    Variadic paramter should be the last parameter of the function.
+
+```go
+    func hello(...a int, b int) {}
+```
+In the above function, we can pass multiple arguments but all those arguments will assigned to a.
+`Hence, variadic parameters should be passed at the end`. The above function will fail to compile with error syntax error: cannot use ... with non-final parameter b
```
junedev, line 24: It is great that you included what happens when using this wrong here. However, I would recommend to start this section with the sentence that says it does not compile. The current first sentence sounds like it would work.

concepts/variadic-functions/about.md
```diff
@@ -0,0 +1,57 @@
...
+        fmt.Println(num, "not found in ", nums)
+    }
+    fmt.Printf("\n")
+    }
+    func main() {
+        find(89, 89, 90, 95)
+        find(45, 56, 67, 45, 90, 109)
+        find(78, 38, 56, 98)
+        find(87)
+    }
```
junedev, line 47: The indentation seems off in this example. The two `func` lines should not have any indentation.

Also an empty line between the two functions would help readability.

concepts/variadic-functions/about.md
```diff
@@ -0,0 +1,57 @@
...
+```
+The way variadic functions work is by converting the variable number of arguments to a slice of the type of the variadic parameter. 
+For instance, in line no. 22 of the program above, the variable number of arguments to the find function are 89, 90, 95. The find function expects a variadic int argument.
```
junedev, line 50: There are no line numbers on the website for these examples. Maybe you can just copy the line into inline code (backticks) here.

junedev (commented):
Well done with writing up this concept! I left some comments for minor improvement and I think one part is still missing (also mentioned in a comment).

junedev:
The new concept also needs to be added to the root level config.json file. See https://exercism.org/docs/building/tracks/config-json#h-concepts.

concepts/variadic-functions/about.md
myworkcircle: got it will work on it

concepts/variadic-functions/about.md
myworkcircle: will make changes to it

concepts/variadic-functions/.meta/config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "blurb": "Structs group pieces of related data together.",
+  "authors": [],
+  "contributors": [
+    "myworkcircle"
```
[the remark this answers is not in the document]
myworkcircle: resolved

concepts/variadic-functions/.meta/config.json
```diff
@@ -0,0 +1,7 @@
+{
+  "blurb": "Structs group pieces of related data together.",
```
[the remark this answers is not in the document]
myworkcircle: resolved

concepts/variadic-functions/about.md
```diff
@@ -0,0 +1,57 @@
...
+```
+The way variadic functions work is by converting the variable number of arguments to a slice of the type of the variadic parameter. 
+For instance, in line no. 22 of the program above, the variable number of arguments to the find function are 89, 90, 95. The find function expects a variadic int argument. 
+Hence these three arguments will be converted by the compiler to a `slice of type int []int{89, 90, 95}` and then it will be passed to the find function.
+
+To dive deeper into this type, you can check these resources:
+
+[Go by example: Variadic Functions](https://gobyexample.com/variadic-functions)
+
+[Medium: Variadic functions](https://medium.com/rungo/variadic-function-in-go-5d9b23f4c01a)
```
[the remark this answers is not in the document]
myworkcircle: resolved

concepts/variadic-functions/about.md
myworkcircle: resolved

concepts/variadic-functions/about.md
myworkcircle: resolved

concepts/variadic-functions/about.md
myworkcircle: resolved

concepts/variadic-functions/about.md
myworkcircle: resolved

concepts/variadic-functions/about.md
myworkcircle: resolved

concepts/variadic-functions/about.md
myworkcircle: resolved

concepts/variadic-functions/about.md
myworkcircle: resolved

myworkcircle: resolved

myworkcircle:
@junedev have made all the changes requested by you

concepts/variadic-functions/about.md
myworkcircle: resolved

junedev: It seems there was a misunderstanding here. I meant this syntax from the link above is missing in the concept. I added a section for this to your PR.

```
s := []string{"James", "Jasmine"}
Greeting("goodbye:", s...)
```

junedev:
@myworkcircle Thanks for the update. I did some more word smithing and added a missing part in https://github.com/exercism/go/pull/2091/commits/2e1c8400ff85543205cbf1213742d82254068e25.
I also removed some unneeded indentation in the code blocks and in the exercism block and I simplified the big example code a bit to make it more readable. Hope this is ok with you. I wanted to get this over the line quickly so the follow up task https://github.com/exercism/go/issues/1990 is not blocked anymore.

junedev (approved):

myworkcircle:
https://github.com/exercism/go/issues/1990 @junedev is it my next task?

junedev:
You can work on whatever you like. :slightly_smiling_face: You can pick up that follow-up task #1990 but if you prefer, you can also work on https://github.com/exercism/go/issues/1373 as you originally planned.

Issue #2089, opened 2022-02-05
Remove old Go version tests from CI


andrerfcsantos:
As suggested by @junedev in https://github.com/exercism/go/issues/1881#issuecomment-959705810, we should stop running the tests with older versions of go in our CI.

Right now in [exercise-tests.yml](https://github.com/exercism/go/blob/main/.github/workflows/exercise-tests.yml) we have a matrix of tests that runs the tests with some older versions of Go.

Due to the [Go 1 compatibility promise](https://go.dev/doc/go1compat), this is largely redundant, as new versions of Go "promise" to compile old programs. Not testing with old versions of Go also solves a problem that could happen when using a newer feature of Go in our exercises, as we would have to make sure those specific cases weren't run with older versions of Go. By simply not testing with older versions, we remove the need to handle that special case.

jarbhav:
Hey, if I understand this correctly, only go versions 1.17.+ should be there in the [exercise-tests.yml](https://github.com/exercism/go/blob/main/.github/workflows/exercise-tests.yml) file.

andrerfcsantos:
@jarbhav Yes, exactly.

jarbhav:
Made the changes. @andrerfcsantos it would be great if you review them!

PR #2093, opened 2022-02-10
modify exercise-tests.yml

jarbhav:
Removed version 1.16.+ from [exercise-tests.yml](https://github.com/exercism/go/blob/main/.github/workflows/exercise-tests.yml) as suggested by @andrerfcsantos in issue #2089


andrerfcsantos (approved):
Thanks for this!

Issue #2095, opened 2022-02-10
Need for Speed: Improve tests for the Drive function


andrerfcsantos:
I recently came across a solution that writes `Drive` as:

```go
func Drive(car Car) Car {
    drive := Car {
        speed : car.speed,
        batteryDrain : car.batteryDrain,
        battery : car.battery,
        distance : car.speed,
    }
    if car.battery < car.batteryDrain {
        drive.distance = 0
    } else {
        drive.battery = car.battery - car.batteryDrain
    }
    return drive
}
```

This solution passes the tests, but has 2 problems:

* When the car is out of battery, it sets the `distance` to `0`, when none of the variables should be touched in that situation. This passes the tests because our test cases for battery drained also happen to have `distance` set to `0`. The fix here is to create or change an existing test where a car with a battery drained has `distance != 0`.
* It sets `distance` to be `speed`, instead of incrementing distance by speed. This passes the tests because the only test case he have where the car doesn't have the battery drained, also starts with a `distance` of `0`. The fix here is to create a test case where the car has enough battery, but `distance != 0`.

LeamHall:
There's a conflict between the instructions and the tests. The instructions say: "Assume the car is just starting the race:", which implies a fully charged battery. However, some of the tests assume the car starts with less than a full battery.

PR #2097, opened 2022-02-10
need-for-speed: Add test cases for car with positive initial distance

mhorod:
Fixes issue #2095


andrerfcsantos (approved):
Thanks for taking care of this! Looks great!

PR #2099, opened 2022-02-12
docs(gross-store): Typo in map literal

rawalplawit:
There proper format is key: value not key, value


andrerfcsantos (approved):
Thanks for this!

PR #2101, opened 2022-02-14
hamming: Update link leading to leap exercise

mhorod:
From the [Hamming](https://github.com/exercism/go/blob/main/exercises/practice/hamming/.docs/instructions.append.md) exercise:
> You may be wondering about the `cases_test.go` file. We explain it in the [leap exercise](https://github.com/exercism/go/blob/main/exercises/practice/leap/.docs/instructions.md).

The link leads to the `instructions.md` file which has no explanation about the `cases_test.go` and I found it very confusing.
I suggest changing it so it leads to the actual exercise on the website.


andrerfcsantos (approved):
Thanks for this!

Issue #1565, opened 2021-09-14
card-tricks: should have for-loops as a prerequisite


ErikSchierboom:
The [exemplar solution](https://github.com/exercism/go/blob/main/exercises/concept/card-tricks/.meta/exemplar.go) of the `card-tricks` exercise uses [for loops](https://github.com/exercism/go/blob/main/exercises/concept/card-tricks/.meta/exemplar.go#L29).

This concepts should be added as prerequisite. (This relies on [Bird Watcher being implemented](https://github.com/exercism/go/issues/1617))

jmrunkle:
Stuck on this one until bird-watcher is implemented. I guess I can take that one since I need it to finish this one.

jmrunkle:
Actually nevermind, that one is assigned to sachinmk27. Just waiting for that to be implemented.

junedev:
There is a problem with this task. It is very hard to make a sensible for loop exercise without slices. That is why the exercise that is currently in the making has slices as prerequisite. That means card tricks that teaches slices cannot have for loops as prerequisite.

I would recommend solving this by removing the for loop task from card tricks as this is much easier than removing slices from the for loop exercise. Maybe someone has a good idea for a replacement task.

junedev:
There are more things around card tricks that need to be fixed. I am closing this issue and create a new one that fits the current state of things.

PR #2102, opened 2022-02-15
interest-is-interesting: added for-loop prerequisite 

Zakini:
Similar to #1565, the exemplar for interest-is-interesting uses a for loop, so should have for-loops as a prerequisite.


andrerfcsantos (approved):
Thanks for this!

Issue #2107, opened 2022-02-26
test case expects 0.0 == parseInt("4")


venkateshpitta:
`        if ExtractFancyNumber(fnb) == 4 {
                // the test case says so --shrug--
                return fmt.Sprintf("This is a fancy box containing the number 0.0")
        }
`
is required to satisfy
`                {
                        description: "Describe a valid different fancy number",
                        input:       differentFancyNumber{"4"},
                        want:        "This is a fancy box containing the number 0.0",
                },
`

andrerfcsantos:
Closing this for the same reason as https://github.com/exercism/go/pull/2108

Let's discuss this in https://github.com/exercism/go/issues/2110

PR #2108, opened 2022-02-27
Changing 0.0=="4" to 4.0=="4"

venkateshpitta:
If this patch looks good, please accept the PR
Patch is for [issue #2107](https://github.com/exercism/go/issues/2107)

### DIFF
--- a/exercises/concept/sorting-room/sorting_room_test.go
+++ b/exercises/concept/sorting-room/sorting_room_test.go
@@ -135,7 +135,7 @@ func TestDescribeFancyNumberBox(t *testing.T) {
 		{
 			description: "Describe a valid different fancy number",
 			input:       differentFancyNumber{"4"},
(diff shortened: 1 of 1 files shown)


andrerfcsantos:
Thanks for the interest in contributing!

The tests here are correct as is. Currently the instructions say:

> If the FancyNumberBox is a FancyNumber, Jen wants strings saying "This is a fancy box containing the number 4.0". Any other type of FancyNumberBox should say "This is a fancy box containing the number 0.0".

`anotherFancyNumber` is not a `FancyNumber`, that's why it must return 0. The test case has a string with `"4"` to catch solutions that ignore this fact and convert the number to a string regardless.

However, this is a common issue that comes up a lot in this exercise and we should address it. I created a new issue for it, and would love to hear your feedback there: https://github.com/exercism/go/issues/2110

Issue #2110, opened 2022-02-27
sorting-room: make edge case of "Describe a fancy number box" clearer


andrerfcsantos:
I've seen some reports of students this being confusing by this and it is something we can improve.

Right now we have the following test for task 4 "Describe a fancy number box" of the exercise `sorting-room`:

https://github.com/exercism/go/blob/8437f72dd89fc9d743927e88fcd0c5cef75b08dc/exercises/concept/sorting-room/sorting_room_test.go#L135-L139

The point of this test is to check if the student is correctly checking for the `FancyNumber` type. Since `differentFancyNumber` is not a `FancyNumber`, solutions should not attempt to parse the string as a number. This test tries to catch that by having something that is not a `FancyNumber` with a string with a valid number.

Event though the instructions clearly say that something that is not a `FancyNumber`, clearly the message is not getting through, as we have a lot of people just assuming that the test case is wrong.

Some ways I can see this being improved (but suggestions are also welcome):

* Change the description of the test `"Describe a valid different fancy number"`. I understand this description can be misleading as it hints that the number is "valid". However, the "valid" means that the string is a valid number, not that `differentFancyNumber` is a valid `FancyNumber`. Making the wording more clear would help I think to clear some of the confusion.
* Maybe change the wording in the task description itself to make this extra clear.

norbs57:
Currently the example  in the instructions is as follows: 

```
fmt.Println(DescribeFancyNumberBox(FancyNumber{"10"}))
// Output: This is a fancy box containing the number 10.0
fmt.Println(DescribeFancyNumberBox(AnotherFancyNumber{"one"}))
// Output: This is a fancy box containing the number 0.0
```

Maybe there would be less confusion if the second example was changed so that the string parsed as an int? 

```
fmt.Println(DescribeFancyNumberBox(AnotherFancyNumber{"4"}))
// Output: This is a fancy box containing the number 0.0
```

I think there is a similar issue in Section 3 "...extracting the number from a fancy number box" , but maybe students did not stumble about this because there was no corresponding test case.

junedev:
@andrerfcsantos @norbs57 Is this issue resolved by the PR that was merged or is there more we want to do here?

andrerfcsantos:
@junedev Yes, this solved. Not entirely sure why he PR being merged didn't close this, but thanks for pointing it out.

PR #2111, opened 2022-02-27
Sorting room: instructions and test cases.

### LINKED ISSUE #2106: Sorting Room - suggestions re task description and additional test case
In the sorting-room exercise task description, "Section 2. Describe a number box",  it says:
 > Jen wants numbers to...

I wonder if this should be:  
 > Jen wants number boxes to...

In the same section, the sample code  refers to 
> numberBoxContaining{12.345}

This looks odd as number boxes supply int- values, not floats. This could be fixed by using 12 instead of 12.345, for example. 

In sorting_room_test.go, function TestExtractFancyNumber currently has only one non-FancyNumber  test with a string "two" that fails conversion to int. Because of this, the test can be passed without the code actually checking the type of the fnb argument. Adding a test case similar to below should fix this: 
```
{ description: "Extract 4 packed as a different fancy number returns 0",
  input:       differentFancyNumber{"4"},
  want:        0,
 } 
```

norbs57:
Resolves #2106

Sorting room instructions: 

- fix in first sentence in Section 2 #2106
- changes to function arguments in example in Sections 2 #2106
- changes to function arguments in examples in Sections 3 and 4. #2110 

sorting_room_test.go, func TestExtractFancyNumber: 

- changed "differentFancyNumber" test case #2106
- added a new test case for "FancyNumber with invalid string"

### DIFF
--- a/exercises/concept/sorting-room/.docs/instructions.md
+++ b/exercises/concept/sorting-room/.docs/instructions.md
@@ -25,11 +25,11 @@ fmt.Println(DescribeNumber(-12.345))
 
 ## 2. Describe a number box
 
-Jen wants numbers to return strings like `"This is a box containing the number 2.0"` (again, including one digit after the decimal):
+Jen wants number boxes to return strings like `"This is a box containing the number 2.0"` (again, including one digit after the decimal):
 
 ```go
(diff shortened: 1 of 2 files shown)


andrerfcsantos (approved):
Thanks for this!

PR #2119, opened 2022-03-04
Clarify instructions

### LINKED ISSUE #2109: The Farm: Instructions can be more clear
Hi!
I hope all are doing great!

In the instructions, I've found that it is not explicitly clear what should happen in the case of `ErrScaleMalfunction` and `fodder < 0`. It can mislead one to believe it is counted as "any other error" or all cases of `fodder < 0` is a `negative fodder` error.
To make this clear in instructions, I think it is a good idea to add a new subsection in instructions on specifying different cases for `ErrScaleMalfunction`.

I would like to know what are your thoughts on this.
Thank you!

MikaeelMF:
Resolves #2109


andrerfcsantos (approved):
Thanks a lot for working on this!

PR #2121, opened 2022-03-05
Pov tree logic ops

### LINKED ISSUE #1918: POV: Change exercise from graph to tree logic
[Original Issue Titel: **POV exercise passes all tests with a noop mandatory function**]

The `AddNode` function is mandatory however a noop implementation satisfies all the tests... Apparently the information passed in is never needed/tested.

here is my solution (I cannot find a way to link to it)

``` go
package pov

type arcList map[string]string

type Graph struct {
	arcs arcList
}

func New() *Graph {
	return &Graph{arcs: make(arcList)}
}

func (g *Graph)AddNode(id string) {
}

func (g *Graph)AddArc(a, b string) {
	g.arcs[b] = a
}

func (g *Graph)ArcList() (list []string) {
 	for k, v := range g.arcs {
		 list = append(list, v + " -> " + k)
	}
	return list
}

func (g *Graph)ChangeRoot(oldRoot, newRoot string) (newGraph *Graph) {
	onPath := make(map[string]bool)
	newGraph = &Graph{arcs: make(arcList)}

	// find path between oldRoot and newRoot
	for p := newRoot; ; p = g.arcs[p] {
		onPath[p] = true
		if p == oldRoot {
			break
		}
	}

	// reverse arrow for edges on path, copy rest
	for child, parent := range g.arcs {
		if onPath[child] && onPath[parent] {
			newGraph.arcs[parent] = child
		} else {
			newGraph.arcs[child] = parent
		}
	}

	return newGraph
}
```

norbs57:
Pov exercise using tree logic ops. Resolves #1918
Largely based on the C# POV exercise.


exercises/practice/pov/TODO.md
```diff
@@ -0,0 +1,30 @@
+## Tree logic POV exercise
```
junedev, line 1: This file needs to be removed from the current folder because otherwise it might be included in the download "package" the student gets.

@norbs57 If you think all or some of those thoughts/ideas are important to keep, please create an issue (e.g. "POV: Ideas for further improvements") and put them in there. Then people that have opinions on this can find it. We will label the issue as "discussion".
In general I trust your judgement on these. If you think the exercise works as you have it now, lets keep it as is and see how it goes. If you would prefer to make some more changes than go ahead.

exercises/practice/pov/pov.go
```diff
@@ -1,23 +1,52 @@
 package pov
 
-// Define the Graph type here.
+// Define the Tree type here
 
-func New() *Graph {
-	panic("Please implement the New function")
+type Tree struct{}
```
junedev, line 5: Line 3 and 5 are a bit contradictory. In other exercises we only have the equivalent of line 3 and nothing like line 5 so the student needs to decide that they want a struct here. Alternatively, can a comment like "Add the needed fields here" as body of the struct instead of line 3.

exercises/practice/pov/pov.go
```diff
@@ -1,23 +1,52 @@
...
+
+// New creates and returns a new Tree with the given root value and children
+// You can assume that there will be no duplicate values in test trees.
```
junedev, line 8: Line 8 should go into the instructions (append), not into the function comment.

exercises/practice/pov/pov.go
```diff
@@ -1,23 +1,52 @@
 package pov
 
-// Define the Graph type here.
+// Define the Tree type here
 
-func New() *Graph {
-	panic("Please implement the New function")
+type Tree struct{}
+
+// New creates and returns a new Tree with the given root value and children
```
junedev, line 7: All comments that are full sentences should end with a period. https://github.com/golang/go/wiki/Comments#comments

exercises/practice/pov/pov_test.go
```diff
@@ -1,260 +1,156 @@
 package pov
 
 import (
-	"reflect"
 	"sort"
 	"testing"
 )
 
-type arc struct{ fr, to string }
+// auxiliary functions
```
junedev, line 8: I think this comment is not needed because its clear from the file structure and what is exported and not exported, that these are the helpers. 

Besides that, I would recommend to put the helpers either below the place where they are used or below all of the `Test` functions in case they are used in multiple ones. This top down sorting of functions is more consistent with the code we have in the other exercises I think.

exercises/practice/pov/test_cases.go
```diff
@@ -0,0 +1,156 @@
+package pov
```
junedev, line 1: The (sub-)test names (and test cases) from the canonical test data https://github.com/exercism/problem-specifications/blob/main/exercises/pov/canonical-data.json should be respected. It's ok to add more tests but if it is very hard to match the code here to the canonical data we have a hard time updating the code when the canonical test data was improved.

junedev (commented):
I finally got around to review this. Great work overall!

I am not sure everyone agrees but I personally really like the new instructions append file.

exercises/practice/pov/.meta/config.json
```diff
@@ -1,7 +1,7 @@
 {
-  "blurb": "Reparent a graph on a selected node",
+  "blurb": "Reparent a tree on a selected node",
```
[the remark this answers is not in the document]
norbs57: Thanks - I was not aware of that. I have reverted to the old version.

exercises/practice/pov/.docs/instructions.append.md
```diff
@@ -3,29 +3,42 @@
...
+The basic tree construction and deconstruction must be working before you start
+on the interesting part of the exercise, so it is tested separately in the first
+three tests.
+
+API methods `FromPov` and  `PathTo` are the interesting part of the exercise.
+ 
+Method `FromPov` takes a string argument `from` which specifies a node in the
+tree via its value. It should return a tree with the value `from` in the root.
+
+- You can modify the original tree and return it or create a new tree and return
```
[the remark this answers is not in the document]
norbs57: Thanks for the pointer to the "note" block. I was trying to create a "side note"  so semantically, the "note block" sounds right. However, looking at the preview, it feels a bit heavy in such a short text - it gives too much emphasis to something that is more of an aside.  The same goes for a "blockquote" although it looks a bit better. So I think I will just move those three sentences to the end of the previous paragraph.

exercises/practice/pov/.docs/instructions.append.md
```diff
@@ -3,29 +3,42 @@
 ## Implementation Notes
 
 POV / reparent / change root of a tree
```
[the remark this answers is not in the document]
norbs57: I don't think so. I just kept it from the previous version. 
I have now also removed the next line "POV/...".
Please let me know if you would like to keep that.

exercises/practice/pov/pov.go
norbs57: Thanks. I like the second option.

exercises/practice/pov/pov.go
norbs57: Done :)

exercises/practice/pov/pov_test.go
norbs57: I have put all the helpers below all of the `Test` functions - I think that may help students who are looking at this file.

exercises/practice/pov/TODO.md
norbs57: Thanks.  This is my first more major PR and I kept it in to highlight some of the design decisions where I thought there were several viable options.

exercises/practice/pov/pov.go
norbs57: Done, I think.

exercises/practice/pov/test_cases.go
norbs57: I assume you are referring to the "description" field and maybe also the "property" field of each of the "cases", is that correct?  I could  add this information as additional fields to the structs `fromPovTestCase` and `pathToTestCases`. Would that be sufficient?

junedev: Pretty much. Since the people working on problem spec usually put quite some thought into making these names meaningful for the students, I would recommend you use the new description field in t.Run of `TestFromPov` and `TestPathTo` instead of your self-constructed names.

exercises/practice/pov/.docs/instructions.append.md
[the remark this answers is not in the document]
junedev: It looks like you also removed "## Implementation Notes". This line is needed because otherwise the append content is rendered directly below the instructions from problem spec. Depending on what the last sentence/section is there, that can lead to very confusing texts. (We had this issue with other exercises before.)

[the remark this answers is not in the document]
norbs57: Sorry, I did not mean to remove that line.

junedev (approved):

junedev:
When I looked at the append file locally, my editor corrected some issues with excess whitespace. Additionally, the code examples do not have to be indented. I pushed a commit to fix the issue. Maybe check whether your editor provides some plugin that can fix such formatting issues automatically on save.

junedev:
I have a feeling the test runner output on the website for this will not look very nice because it can only handle exactly this structure well https://github.com/exercism/go-test-runner/#subtest-format-specification. But this is something we can address in another PR if we feel like it should be improved.

PR #2123, opened 2022-03-07
Binary-search-tree #2113

### LINKED ISSUE #2113: Binary Search Tree: map functions - mismatch with specs
File binary_search_tree.go defines two (higher-order) functions MapString and MapInt.  There seems to be a mismatch between the type of these functions and their specs in that source file:

- The type (and name) of the Map functions suggests that the argument f will be applied to each of the elements of the tree that is to be transformed into a slice.
- The specs do not mention the function argument f at all. Instead, these specs are more in line with Map-functions that do not take a function argument.   

There are at least four different ways in which we could address this issue: 

(A) Keep the function types the same, update the specs and add some test cases for different f-arguments. 

(B) Keep the specs, change both Map-functions to no-argument functions, and update the tests accordingly. 

(C) Keep function MapInt as it is and replace MapString with a no-argument String function. The latter could be implemented with or without invocation of MapInt. This would entail adding test cases for MapInt and updating MapString tests to String tests. 

(D) Remove the Map-functions from this exercise - arguably they do not go all that well with the topic of binary **search** trees. Instead, we could reuse some of the material for a new, separate exercise on map-functions for different types of trees, graphs, slices, etc.  In this context, it is worthwhile remembering that Go is about to acquire generics. Once this has happened, functions like MapString and MapInt can be defined as special instances of more generic Map functions. 

If we keep the Map-functions, then it would be good to mention them in the instructions and maybe also have something about them in README.md.

norbs57:
Resolves #2113
Replaced two map functions with one SortedData() function
Updated tests and example accordingly
Renamed type SearchTreeData to BinarySearchTree and changed file names accordingly
Changed testing to a table-driven approach and added a SortedData() benchmark
Removed some generic comments from binary_search_tree.go

### DIFF
--- a/exercises/practice/binary-search-tree/binary_search_tree.go
+++ b/exercises/practice/binary-search-tree/binary_search_tree.go
@@ -1,54 +1,26 @@
-// This is a "stub" file.  It's a little start on your solution.
-// It's not a complete solution though; you have to write some code.
-
-// Package binarysearchtree should have a package comment that summarizes what it's about.
-// https://golang.org/doc/effective_go.html#commentary
 package binarysearchtree
 
-type SearchTreeData struct {
-	left  *SearchTreeData
+type BinarySearchTree struct {
+	left  *BinarySearchTree
 	data  int
-	right *SearchTreeData
+	right *BinarySearchTree
 }
 
-// NewBst creates and returns a new SearchTreeData.
-func NewBst(i int) SearchTreeData {
-	// Write some code here to pass the test suite.
-	// Then remove all the stock comments.
-	// They're here to help you get started but they only clutter a finished solution.
-	// If you leave them in, reviewers may protest!
-	return SearchTreeData{}
+// NewBst creates and returns a new BinarySearchTree.
+func NewBst(i int) *BinarySearchTree {
+	panic("Please implement the NewBst function")
 }
 
-// Insert inserts an int into the SearchTreeData.
-// Inserts happen based on the rules of a BinarySearchTree
-func (std *SearchTreeData) Insert(i int) {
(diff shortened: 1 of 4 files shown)


exercises/practice/binary-search-tree/binary_search_tree.go
```diff
@@ -1,54 +1,26 @@
...
-// SearchTreeData that has the numbers [1,3,7,5] added will return the
+// A BinarySearchTree that has the numbers [1,3,7,5] added will return the
 // []int [1,3,5,7].
-func (std *SearchTreeData) MapInt(func(int) int) (result []int) {
-	// Write some code here to pass the test suite.
-	// Then remove all the stock comments.
-	// They're here to help you get started but they only clutter a finished solution.
-	// If you leave them in, reviewers may protest!
-	return []int{}
+func (bst *BinarySearchTree) SortedData() (result []int) {
```
andrerfcsantos, line 24:
```suggestion
func (bst *BinarySearchTree) SortedData() []int {
```

Since we try to discourage named returns, let's remove them here. Please update the exemplar solution too.

andrerfcsantos (changes requested):
Look great! Just left a minor comment.

exercises/practice/binary-search-tree/binary_search_tree.go
norbs57: Done :)

andrerfcsantos: This change seems to not have gone through. Maybe you forgot to push this change?

andrerfcsantos (approved):
🚀

PR #2124, opened 2022-03-11
Use a map of test cases rather than a slice

sudomateo:
Using `map[string]struct` allows the name of the test case to be clearly seen and also gives the guarantee that no two test cases will have the same name.

### DIFF
--- a/exercises/concept/census/census_test.go
+++ b/exercises/concept/census/census_test.go
@@ -8,16 +8,13 @@ import (
 
 // TestNewResident tests the NewResident function.
 func TestNewResident(t *testing.T) {
-	tests := []struct {
-		name     string
+	testCases := map[string]struct {
 		resident *Resident
 	}{
-		{
-			name:     "no data collected",
+		"no data collected": {
 			resident: &Resident{},
 		},
-		{
-			name: "all data collected",
+		"all data collected": {
 			resident: &Resident{
 				Name: "Matthew Sanabria",
 				Age:  29,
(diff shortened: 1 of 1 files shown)


andrerfcsantos:
The problem with having a map with the tests is that the iteration order is not guaranteed on a map, and for Exercism it's desirable to run the tests by the same order.  There are several reasons for this:

1. Students are able to see the subtests in the the online editor. They might want to fix the errors one by one in order. However if in each run the "Test 1" they see is different, it might be hard for them to see if the old error is still there or if they fixed it and that's the reason they are having a new error.
2. I think point 1 also applies for people using the command line. Potentially seeing the errors in a different order each time might make it harder for them to identify if they fixed anything.
3. Usually in Exercism basic tests are ran first and tests for edge cases are last. This allows the student to fix the tests one by one in increasingly difficulty, leaving the weird edge cases for last. This doesn't work if the test order can be randomized and a test for a weird edge case might appear first.

I also think that having duplicate names is not a big issue. Of course it's something that shouldn't happen, but I don't see it happening often. It's something that can be detected in review and even if it happens, it can be easily fixed and doesn't have any major consequences. In the online editor students can disambiguate the test by its number. We also have good prints about the arguments and output of "expected vs got" that we show when it fails - this helps both students using the online editor and students using the CLI to disambiguate between them.

sudomateo:
I was unaware Exercism wanted deterministic test execution. Thanks!

PR #2136, opened 2022-03-20
Overhaul of BlackJack concept exercise

### LINKED ISSUE #1731: black-jack: LargeHand lacks card information
https://exercism.org/tracks/go/exercises/blackjack

The large hand description states "two aces must always split", however the function signature provided only has 

```go
func LargeHand(isBlackjack bool, dealerScore int) string
```

I think the API should be changed to the following so the function can check the cards. Either that or match `SmallHand` and provide the total score since `22` implies two aces.

```go
func LargeHand(card1, card2 string, dealerScore int) string
// or
func LargeHand(handScore, dealerScore int) string
```

Unless I'm missing something here..

P.S. I'm happy to submit a fix if this really is a mistake

norbs57:
Fixes #1731

* Replaced functions `LargeHand`, `SmallHand` and `IsBlackJack` with a single function `FirstTurn` that is to be implemented by the students
* Removed `IsBlackJack` tests
* Adapted hints and instructions accordingly
* Simplified sample solution

### DIFF
--- a/exercises/concept/blackjack/.docs/instructions.md
+++ b/exercises/concept/blackjack/.docs/instructions.md
@@ -18,29 +18,22 @@ You will receive two cards and will be able to see the face up card of the deale
 
 Depending on your two cards and the card of the dealer, there is a strategy for the first turn of the game, in which you have the following options:
 
-    - Stand (S)
-    - Hit (H)
-    - Split (P)
-    - Automatically win (W)
+- Stand (S)
+- Hit (H)
+- Split (P)
+- Automatically win (W)
 
 Although not optimal yet, you will follow the strategy your friend Alex has been developing, which is as follows:
 
-Category: Large Hand
-
 - If you have a pair of aces you must always split them.
 - If you have a Blackjack (two cards that sum up to a value of 21), and the dealer does not have an ace, a figure or a ten then you automatically win. If the dealer does have any of those cards then you'll have to stand and wait for the reveal of the other card.
-
-Category: Small Hand
-
-- If your cards sum up to 17 or higher you should always stand.
-- If your cards sum up to 11 or lower you should always hit.
+- If your cards sum up to a value within the range [17, 20] you should always stand.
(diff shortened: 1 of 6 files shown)


exercises/concept/blackjack/.docs/instructions.md
```diff
@@ -18,66 +18,45 @@ You will receive two cards and will be able to see the face up card of the deale
...
+Implement a function to calculate the numerical value of a card:
 
 ```go
 value := ParseCard("ace")
 fmt.Println(value)
 // Output: 11
 ```
 
-## 2. Determine if two cards make up a Blackjack.
+## 2. Implement the strategy.
```
junedev, line 44: the header here needs to match the header in the hints, otherwise the connection between task and hint might not work on the website

exercises/concept/blackjack/.docs/instructions.md
```diff
@@ -18,66 +18,45 @@ You will receive two cards and will be able to see the face up card of the deale
...
-choice := SmallHand(handScore, dealerScore)
-fmt.Println(choice)
-// Output: "H"
+FirstTurn("ace", "ace", "jack") == "P"
+FirstTurn("queen", "ace", "eight") == "W"
+FirstTurn("ace", "king", "ace") == "S"
+FirstTurn("eight", "ten", "jack") == "S"
+FirstTurn("seven", "seven", "five") == "S"
+FirstTurn("five", "queen", "ace") == "H"
+FirstTurn("six", "five", "ten") == "H"
```
junedev, line 61: I would recommend to give max 3 examples max because the real source of lots of examples are the test cases the student will work through.

junedev (approved):
I really like the new version of the exercise you created here. Only 2 minor comments.

exercises/concept/blackjack/.docs/instructions.md
norbs57: Done.

Issue #2137, opened 2022-03-22
Blackjack: Add test case for "other card" case


junedev:
I checked the code coverage for the blackjack exercise and noticed that the ["default" case](https://github.com/exercism/go/blob/39b054e0c79d38aa6f42196a1c6f236cc34f52dc/exercises/concept/blackjack/.meta/exemplar.go#L26-L27) for `ParseCard` is not covered by a test case currently. That should be fixed.

The test file can be found here: https://github.com/exercism/go/blob/main/exercises/concept/blackjack/blackjack_test.go

General information about the structure of concept exercises can be found here: https://exercism.org/docs/building/tracks/concept-exercises
If you are not familiar with contributing on GitHub, see https://exercism.org/docs/building/github.

PR #2138, opened 2022-03-23
Add missing test case for Blackjack (#2136)

TristanAppDev:
Added default test case to TestParseCard. 
name: "parse invalid", card: "joker", want: 0

Related to Issue2137: https://github.com/exercism/go/issues/2137


junedev (commented):
Welcome to contributing to the Go track and thanks for working on this!

The exercise description talks about "other cards" so maybe "parse other" would be a better name for the test. That would also fit your card name very well.

junedev (approved):

PR #2146, opened 2022-03-31
Update for go-1.18

parasense:
* TL;DR: `any` is an alias for `interface{}`.
* Update test cases to use `any` instead of `interface{}`.
* Users may still use `interface{}` in their solutions.
* This sould maintain backwards compatibility with old solutions.
* Small docs ammendment to clearly specify the expected error handling.

Signed-off-by: Jon Disnard <jdisnard@redhat.com>

### DIFF
--- a/exercises/practice/linked-list/.docs/instructions.append.md
+++ b/exercises/practice/linked-list/.docs/instructions.append.md
@@ -24,3 +24,7 @@ Your `List` should have the following methods:
 * `PushFront(v interface{}) `: insert value at the front of the list.
 * `PopFront() (interface{}, error)`: remove value from the front of the list.
 * `Reverse()`: reverse the linked list.
+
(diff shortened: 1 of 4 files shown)


junedev:
@parasense I am sorry but I cannot accept this PR. If we merge these changes, students that don't have Go 1.18 installed on their machines would not be able to solve e.g. "Linked List" locally anymore. "any" would not compile for them. So while we want to ensure that solutions written in Go 1.18 work on the platform, we want to stick with Go 1.16 as minimal Go version required to solve all exercises for now. Once Go 1.18 is the lowest still maintained version, we will make it the minimal requirement and that would then be the right time for your PR (so about a year from now).

Feel free to submit a separate PR for the suggested improvements to the instructions though. That can of course be considered right now.

Issue #2142, opened 2022-03-28
Tree building exercise test results unclear


JayKayy:
The test results show a failure on all of the test cases but the actual and expected results appear to be identical and I'm not sure what to do.

https://github.com/exercism/go/blob/b531203995d0e9a805bdcbb0a2f8b54d0b37a68f/exercises/practice/tree-building/tree_building_test.go#L14

```▶ go test
--- FAIL: TestMakeTreeSuccess (0.00s)
    --- FAIL: TestMakeTreeSuccess/one_node (0.00s)
        tree_building_test.go:258: Build for test case "one node" returned 0:[] but was expected to return 0:[].
    --- FAIL: TestMakeTreeSuccess/three_nodes_in_order (0.00s)
        tree_building_test.go:258: Build for test case "three nodes in order" returned 0:[1:[] 2:[]] but was expected to return 0:[1:[] 2:[]].
    --- FAIL: TestMakeTreeSuccess/three_nodes_in_reverse_order (0.00s)
        tree_building_test.go:258: Build for test case "three nodes in reverse order" returned 0:[1:[] 2:[]] but was expected to return 0:[1:[] 2:[]].
    --- FAIL: TestMakeTreeSuccess/three_levels_of_nesting (0.00s)
        tree_building_test.go:258: Build for test case "three levels of nesting" returned 0:[1:[2:[3:[]]]] but was expected to return 0:[1:[2:[3:[]]]].
    --- FAIL: TestMakeTreeSuccess/more_than_two_children (0.00s)
        tree_building_test.go:258: Build for test case "more than two children" returned 0:[1:[] 2:[] 3:[]] but was expected to return 0:[1:[] 2:[] 3:[]].
    --- FAIL: TestMakeTreeSuccess/binary_tree (0.00s)
        tree_building_test.go:258: Build for test case "binary tree" returned 0:[1:[4:[] 5:[]] 2:[3:[] 6:[]]] but was expected to return 0:[1:[4:[] 5:[]] 2:[3:[] 6:[]]].
    --- FAIL: TestMakeTreeSuccess/unbalanced_tree (0.00s)
        tree_building_test.go:258: Build for test case "unbalanced tree" returned 0:[1:[4:[]] 2:[3:[] 5:[] 6:[]]] but was expected to return 0:[1:[4:[]] 2:[3:[] 5:[] 6:[]]].
FAIL
exit status 1
FAIL	tree	0.005s
```
[...]

PR #2151, opened 2022-04-03
Tree-building exercise: added custom nodeEqual method, added type def

norbs57:
Fixes issue #2142 "Tree building exercise test results unclear"
* Added declaration of types `Node` and `Record` in `tree_building.go`
* Defined a `nodeEqual` method and used this for node-equality comparisons in tests


andrerfcsantos (approved):
Thanks!

PR #2153, opened 2022-04-04
POV tests - changed to more usual pattern

### LINKED ISSUE #2134: POV: Make test structure more test runner friendly
Currently, the structure of the tests differs from our usual patterns and is not very good for the test runner.

We should investigate, whether we can bring it more into this format: https://github.com/exercism/go-test-runner#subtest-format-specification (This might mean that this causes some duplication which is ok.)

Even if we can't reach that ideal structure, we should make sure that the cases_test.go file only contains the test case slices, not also parts of the test code. Currently, test code being in that helper file leads to issues where the error in the test output is not properly associated with the actual test case that caused it.

norbs57:
Fixes #2134
Created individual test structs for each test with explicitly constructed *Tree fields


andrerfcsantos (changes requested):
You can move the tests in `cases_test.go` to `pov_test.go`. The idea here is to make these tests be a slice declared inside the functions they are needed. For instance, the `TestFromPov` function uses the slice `pathToTestCases`. The idea would be to have this slice declared inside `TestFromPov` and this slice can be called just `tests`. A bit of context on why this is desirable: the test runner tries to parse the code in the test file of the exercise and looks for a slice in the function test to extract each test case. This is used in the output of the tests to show the students using the online editor exactly which test cases passed/failed.

The `cases_test.go` is usually created by the test generator, but we don't have a test generator for this exercise (no `.meta/gen.go` file), so it's fine to remove it and move its tests to the functions they are used - this way the test runner can provide better feedback to the student.

When removing the `cases_test.go` file, also don't forget to remove it for the `config.json`:

https://github.com/exercism/go/blob/920eb320d4dc6f569f55d65c4b4a15f5acc3401d/exercises/practice/pov/.meta/config.json#L30-L32

Other than that, looks good!

andrerfcsantos (approved):

PR #2154, opened 2022-04-04
Vehicle purchase exercise - removed UTF-16 ref, renamed ordering

### LINKED ISSUE #2147: Vehicle purchase exercise, introduction: UTF-16 link and string order
The [introduction to this exercise](https://github.com/exercism/go/blob/main/exercises/concept/vehicle-purchase/.docs/introduction.md) contains the following paragraph: 
```text 
The comparison operators above can also be used to compare strings.
In that case a dictionary (lexicographical) order is applied.
You can find a list of the exact order of all the characters [here][utf-16-list].
```

One of our students raised two issues about this: 

1.  The link is to a UTF-16 resource. This is confusing as Go strings are UTF-8 encoded. 

2. The last two sentences suggest that the ordering of Go strings is lexical based on **characters**. On the other hand, the Go spec says: ["string values are ... ordered, lexically **byte-wise**." ](https://go.dev/ref/spec#Comparison_operators)
 
I am unsure how to best resolve this. According to the Wikipedia UTF-8 page, sorting UTF-8 strings by byte or code-points gives the same result. But I don't think we want to explain that in this document. 

Perhaps the easiest solution would be to remove the last sentence in that paragraph. As an introduction to the sample string comparisons below, we could add a phrase such as `For example:`.  

Also, should we change the second sentence so that it reads: 
```text
In that case a lexicographical (dictionary) order is applied.
```
and replace the term "dictionary order" with "lexicographic order" throughout in `hints.md` and `introduction.md`?

norbs57:
Fixes #2147
Removed spurious link to UTF-16
Renamed "dictionary order" to "lexicographical order" in several places

### DIFF
--- a/exercises/concept/vehicle-purchase/.docs/introduction.md
+++ b/exercises/concept/vehicle-purchase/.docs/introduction.md
@@ -23,8 +23,8 @@ a > 5  // false
 ```
 
 The comparison operators above can also be used to compare strings.
-In that case a dictionary (lexicographical) order is applied.
-You can find a list of the exact order of all the characters [here][utf-16-list].
+In that case a lexicographical (dictionary) order is applied.
+For example:
 
 ```Go
 	"apple" < "banana"  // true
@@ -75,6 +75,5 @@ if v := 2 * num; v > 10 {
 // Output: 14
 ```
 
-> Note: any variables created in the initialization statement go out of scope after the end of the if statement.
+> Note: any variables created in the initialization statement go out of scope after the end of the if statement. 
 
-[utf-16-list]: https://www.fileformat.info/info/charset/UTF-16/list.htm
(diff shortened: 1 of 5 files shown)


exercises/concept/vehicle-purchase/.docs/introduction.md
```diff
@@ -23,8 +23,8 @@ a > 5  // false
 ```
 
 The comparison operators above can also be used to compare strings.
-In that case a dictionary (lexicographical) order is applied.
-You can find a list of the exact order of all the characters [here][utf-16-list].
```
andrerfcsantos, line 27: Remove the link also from the bottom of this file where it is defined.

andrerfcsantos (changes requested):

andrerfcsantos (approved):

PR #2157, opened 2022-04-06
Typo: e.g.

jraregris:
Add a "." to "e.g." to make it consistent with the other uses.

### DIFF
--- a/concepts/arithmetic-operators/about.md
+++ b/concepts/arithmetic-operators/about.md
@@ -12,7 +12,7 @@ Go supports many standard arithmetic operators:
 | `/`      | `13 / 3 == 4`  |
 | `%`      | `13 % 3 == 1`  |
 
-For integer division, the remainder is dropped (eg. `5 / 2 == 2`).
+For integer division, the remainder is dropped (e.g. `5 / 2 == 2`).
 
 ## Arithmetic operations on different types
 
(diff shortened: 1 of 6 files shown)


jraregris:
As suggested by the bot, i searched the repo for more of the same typo. I found a very similar one and made a commit for that.

I also found the same thing in the code of conduct:
https://github.com/exercism/go/blob/main/CODE_OF_CONDUCT.md?plain=1#L46

I'm a bit vary to touch the CoC, and it seems to be synced from a different repo?

andrerfcsantos (changes requested):
I found these other occurrences that are not included in this PR:

```
➜  exercism-go git:(patch-1) find . -type f -exec  grep 'eg\.' {} /dev/null --color=auto \;
./CODE_OF_CONDUCT.md:- We refrain from using language that can be considered offensive or oppressive (systemically or otherwise), eg. sexist, racist, homophobic, transphobic, ableist, classist, etc. - this includes (but is not limited to) various slurs.
./exercises/concept/party-robot/.docs/instructions.md:If the number is less than 3 digits it gets extra leading zeroes to become 3 digits (eg. 3 becomes 003).
./exercises/concept/sorting-room/.docs/hints.md:- either use type assertions (eg. `i.(someType)`) or a type switch (`switch v := i.(type)`)
./exercises/concept/election-day/.docs/hints.md:- You can get the value stored in a pointer by using the `*` operator on a pointer, eg. `var i int = *aPointerToInt`. This is called dereferencing the pointer.
./concepts/arithmetic-operators/introduction.md:For integer division, the remainder is dropped (eg. `5 / 2 == 2`).
```

Can you update these too?

You are right in that you shouldn't update the code of conduct. That file must be changed in https://github.com/exercism/org-wide-files

andrerfcsantos (approved):
Thanks for this!

PR #2160, opened 2022-04-07
Update introduction.md

ahmedsameha1:
var n1, n2 int
var n1, n2 int = 10, 20

### DIFF
--- a/concepts/first-class-functions/about.md
+++ b/concepts/first-class-functions/about.md
@@ -98,6 +98,6 @@ for i := 0; i < N; i++ {
 }
 ```
 
-A call to `fib` initializes the variables `n1` and `n2` and returns an anonymous function that, in turn, changes the values of these variables each time the function is called. Nth calls of the anonymous function return the Nth number of the Fibonacci sequence starting from 0. The anonymous inner function has access to the local variables (`n1` and `n2`) of the enclosing function `fib`. This is a great way to have function values keep state between calls. We say that the anonymous function is a closure of the variables `n1` and `n2`. [Closures][closure] are widely used in programming and you might see other languages supporting them.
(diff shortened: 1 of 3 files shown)


andrerfcsantos (changes requested):
Thanks for this!

There are 2 other files that also must be changed:


```console
$ find . -type f -exec grep --color=auto 'A call to `fib` initializes' {} /dev/null \;
./exercises/concept/expenses/.docs/introduction.md:A call to `fib` initializes the variables `n1` and `n2` and returns an anonymous function that, in turn, changes the values of these variables each time the function is called. Nth calls of the anonymous function return the Nth number of the Fibonacci sequence starting from 0. The anonymous inner function has access to the local variables (`n1` and `n2`) of the enclosing function `fib`. This is a great way to have function values keep state between calls. We say that the anonymous function is a closure of the variables `n1` and `n2`. [Closures][closure] are widely used in programming and you might see other languages supporting them.
./concepts/first-class-functions/introduction.md:A call to `fib` initializes the variables `n1` and `n2` and returns an anonymous function that, in turn, changes the values of these variables each time the function is called. Nth calls of the anonymous function return the Nth number of the Fibonacci sequence starting from 0. The anonymous inner function has access to the local variables (`n1` and `n2`) of the enclosing function `fib`. This is a great way to have function values keep state between calls. We say that the anonymous function is a closure of the variables `n1` and `n2`. [Closures][closure] are widely used in programming and you might see other languages supporting them.
./concepts/first-class-functions/about.md:A call to `fib` initializes the variables `n1` and `n2` and returns an anonymous function that, in turn, changes the values of these variables each time the function is called. Nth calls of the anonymous function return the Nth number of the Fibonacci sequence starting from 0. The anonymous inner function has access to the local variables (`n1` and `n2`) of the enclosing function `fib`. This is a great way to have function values keep state between calls. We say that the anonymous function is a closure of the variables `n1` and `n2`. [Closures][closure] are widely used in programming and you might see other languages supporting them.
```

ahmedsameha1:
I have done the requested changes.

andrerfcsantos:
@ahmedsameha1 Can you apply the changes to this PR? You can do that by pushing to the `patch-1` branch of your fork - no need to change each file in its own individual PR.

ahmedsameha1:
I'm totally new about this.
I thought it simple to mention that there is a possible mistake.

andrerfcsantos (approved):

PR #2161, opened 2022-04-07
Removed excess percentage signs

jonasbn:
The output from the tests confused me as there was double percentage signs (`%%`) in the description. For formatted print  statements, like `Printf` etc., `%` is used do indicate a placeholder and hence `%%` is used to print a percentage sign. This does not seem necessary in this case and due to the use of the double percentage signs both are emitted as output.


andrerfcsantos (approved):
Thanks for this!

PR #2163, opened 2022-04-07
Update introduction.md

ahmedsameha1:
var n1, n2 int
var n1, n2 int = 10, 20


andrerfcsantos:
As suggested, apply this change to the original PR.

PR #2166, opened 2022-04-09
Pointers concept - Change comments about declaration and initialization

ahmedsameha1:
var n1, n2 int
var n1, n2 int = 10, 20


andrerfcsantos (approved):
Thanks for this!

PR #2167, opened 2022-04-09
Methods concept: make it clear that there are two types of receivers

ahmedsameha1:


andrerfcsantos (approved):
This is helpful information to add!

PR #2168, opened 2022-04-09
Methods concept: two minor fixes about grammar and punctuation

ahmedsameha1:
Sorry, this is my fault.


andrerfcsantos:
My bad for not noticing it too initially :)

This is conflicting with the main branch of the repository because the main branch already has these 2 lines. My suggestion to fix this would be to first update your branch that is linked with this PR `patch-5` to be up-to-date with the main branch and then perform the changes.

ahmedsameha1:
What about now?

andrerfcsantos:
The files you were supposed to change look good, but it also looks like you've included some extra files - I assume these were auxiliary files you forgot to exclude from your commit. Remove those extra files and we'll merge!

ahmedsameha1:
Sorry for committing those extra files.

andrerfcsantos (approved):

PR #2186, opened 2022-04-14
Add instructions to create errors

scubbo:
At this point in the syllabus, error-creation has not yet been
introduced!

### DIFF
--- a/config.json
+++ b/config.json
@@ -398,35 +398,36 @@
           "for-loops",
           "functions",
           "structs",
-          "type-definitions"
+          "type-definitions",
+          "errors"
         ],
         "status": "beta"
       },
       {
         "slug": "animal-magic",
         "name": "Animal Magic",
-        "uuid": "71a1be49-daf1-4043-af6e-33cd44481bb5", 
+        "uuid": "71a1be49-daf1-4043-af6e-33cd44481bb5",
         "concepts": [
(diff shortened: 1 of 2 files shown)


scubbo:
image

See image screenshotted from [here](https://exercism.org/tracks/go/concepts). Alternatively, the `errors` exercise could be made a prerequisite of the `First class functions` exercise, but I don't know how to go about that.

junedev:
@scubbo I think your second idea (adding errors to the prequesites) is the better course of action here. It does not hurt that this concept/exercise moved further down in the tree.

Would be great if you could adjust the PR accordingly.

scubbo:
I believe I've updated the track such that the "First-class functions" exercise is dependent upon "Errors". However, I should stress that this is my first time contributing to an Exercism track - although I have read through [this page](https://github.com/exercism/go), I don't know how to confirm that this change is correct (in particular, I don't know how I'd detect a circular dependency).

junedev (approved):
Looks all correct. We have a tool in the CI pipeline that would throw an error if there would be an issue with building up the concept tree. Since your PR passes all the checks, you can rest assured its all good. 🙂

scubbo:
Sweet - it's great to be working in a project with robust CI/CD practices! Thanks for your help :)

PR #2187, opened 2022-04-15
docs(grains): unify enumerations

magicrc:
Little modification to keep enumerations consistent, across whole file.


andrerfcsantos:
@magicrc Changes to the introductions/tests of practice exercises must be done in the problem-specs repository: [Grains folder in the problem specs repository](https://github.com/exercism/problem-specifications/tree/main/exercises/grains).

After the changes are merged there, we can import them to the Go track. Check the links on the previous comment from the bot for more information.

PR #2190, opened 2022-04-17
Car assembly: Escape $ in instructions.md

dirkkul:
In markdown with math mode enabled the dollar sign enables this mode and results in italic text with a different font that is hard to read. Escaping with \ fixes this.

With math mode disabled this should not make a difference.

### DIFF
--- a/exercises/concept/cars-assemble/.docs/instructions.md
+++ b/exercises/concept/cars-assemble/.docs/instructions.md
@@ -31,8 +31,8 @@ rate := CalculateWorkingCarsPerMinute(1105, 90)
 
 ## 3. Calculate the cost of production 
 
-Each car normally costs $10,000 to produce individually, regardless of whether it is successful or not.
-But with a bit of planning, 10 cars can be produced together for $95,000.
+Each car normally costs \$10,000 to produce individually, regardless of whether it is successful or not.
+But with a bit of planning, 10 cars can be produced together for \$95,000.
 
 For example, 37 cars can be produced in the following way:
 37 = 3 x groups of ten + 7 individual cars


andrerfcsantos:
Thanks for bringing this to our attention.

Just for a little bit more context, I've never heard of markdown math mode, can you tell us an example where It is used?

If we are going making this change, we should also make it in all places the `$` is used. I've found that `$` is also used in the [book store](https://github.com/exercism/go/tree/main/exercises/practice/book-store/.docs) exercise, where it is unescaped too.

dirkkul:
Hi,

I installed [this VS code extension](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one) and it had the math mode enabled by default for side-by-side previews. I am not too much into markdown so I don't know how common it is. 

If that's enough for you to merge the change, I'll see if my grep/sed-foo is sufficient to find all occurances.

andrerfcsantos:
@dirkkul I did a global search in the repository and the [book-store](https://github.com/exercism/go/tree/main/exercises/practice/book-store/.docs) also has some instances of `$` that need escaping. Would you like to update this PR with the needed changes escaping the `$` on that exercise too?

junedev:
@andrerfcsantos Book store originates from the problem spec repo and I don't think we have a good case getting this change through there.

My opinion on the whole issue here would be to not apply any change. The markdown extension is not needed to work on Exercism exercises so if it is causing an issue it can be disabled. VSCode allows to disable an extension just for a certain workspace (click on the extensions icon, select the extension, then click on the down arrow next to "disable" and select "disable (workspace)"). Without that extension, dollar signs are rendered just fine in the markdown preview pane as far as I can see.

dirkkul:
alright, I don't care too much :)

PR #2193, opened 2022-04-20
Issue #2181: Improve functions introduction

### LINKED ISSUE #2181: Lasagna Master: Improve functions introduction
https://exercism.org/tracks/go/exercises/lasagna-master

Introduction for exercise says:

> Functions in Go pass their arguments by value. This means that the arguments are copied, and any changes to those arguments within the function will not be seen outside the function.

Task 3 says:

> Write a function AddSecretIngredient that accepts two slices of ingredients of type []string as parameters. The first parameter is the list your friend sent you, the second is the ingredient list of your own recipe. The last element in your ingredient list is always "?". The function should replace it with the last item from your friends list. Note: AddSecretIngredient does not return anything - you should modify the list of your ingredients directly. The list with your friend's ingredients should not be modified.

Resulting code for task 3:
```
func AddSecretIngredient(friendsList []string, myList []string) {
	myList[len(myList)-1] = friendsList[len(friendsList)-1]
}
```

I don't understand, how newbie will understand why they pass the slice by value to the function but myList will be changed outside. 

I suppose it will be better to clarify why it happens.

nahuakang:
## Motivation

Fixes #2181

This PR addresses [#2181](https://github.com/exercism/go/issues/2181) and adds more educational information on `function` to help students complete the [Lasagna Master challege](https://exercism.org/tracks/go/exercises/lasagna-master).

## Notes
- I added a clarification on `argument` vs `parameter` since previous authors have used both terms without explanation.
- I added modified pointer-related content to function concept's `about.md` but not `introduction.md`.

### DIFF
--- a/concepts/functions/about.md
+++ b/concepts/functions/about.md
@@ -32,12 +32,41 @@ func PrintGreetingName(greeting, name string) {
 }
 ```
 
-In the example above both `greeting` and `name` are parameters of type `string`.
-When we pass primitive types such as `string` as arguments into a function, 
-we essentially make a copy of the original value or underlying data so that only this copy is accessed or modified by our function.
+## Parameters vs. Arguments
 
-In Go, not only can we pass primitive values into a function, but we can also pass a `pointer`, as in a reference to some data, to the function.
+Let's quickly cover two terms that are often confused together: `parameters` and `arguments`. 
+Function parameters are the names defined in the function's signature, such as `greeting` and `name` in the function `PrintGreetingName` above.
+Function arguments are the concrete values passed to the function parameters when we invoke the function.
+For instance, in the example below, `"Hello"` and `"Katrina"` are the arguments passed to the `greeting` and `name` parameters:
+
+```go
+PrintGreetingName("Hello", "Katrina")
+```
+
+## Pass by Value & Pass by Pointer
+
+It is also important to clarify the concept of passing by value as well as passing by pointer (sometimes referred to as _by reference_).
+In the example below, `name`, the `string` parameter to the function `PrintHelloName`, is passed into the function by value.
+This means that Go essentially makes a copy of the original value so that only this copy is accessed or modified by our function.
+
+```go
+import "fmt"
+
+func PrintHelloName(name string) {
+  fmt.Println("Hello " + name)
+}
+```
+
+Strictly speaking, all arguments are passed by value in Go, i.e. a copy is made of the value or data provided to the function.
+But if you don't want to make a copy of the data that you choose to pass to a function and
+want to actually access and change the data in the function,
+then you should use pointers as arguments, a.k.a. pass by reference.
+
+In Go, not only can we pass primitive values into a function, but we can also pass a `pointer` of these values
+(a reference to the data) to the function.
 This means that the function receives only a copy to the pointer instead of the underlying data that this pointer refers to.
+Note that we say _a copy to the pointer_ since everything is passed by value to functions in Go,
+but a pointer copy, in contrast to a value copy, gives us an "address" that points to the data we want to access and modify.
 
 If the concept of `pointer` is confusing, no worries.
 We have more details below as well as a dedicated section to help you understand pointers.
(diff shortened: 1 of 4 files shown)


andrerfcsantos (changes requested):
Thanks for working on this!

I like what you did here 👍 

Left some suggestions.

concepts/functions/about.md
```diff
@@ -19,6 +22,25 @@ func PrintHelloName(name string) {
...
+
+```go
+import "fmt"
+
+func PrintGreetingName(greeting, name string) {
+  fmt.Println(greeting + " " + name)
+}
+```
+
+When we pass primitive types such as `string` as arguments into a function,
```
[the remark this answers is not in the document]
nahuakang: Thanks! I agree with you and made some changes now :) Let me know what you think?

concepts/functions/about.md
```diff
@@ -19,6 +22,55 @@ func PrintHelloName(name string) {
...
+Function arguments are the concrete values passed to the function parameters when we invoke the function.
+For instance, in the example below, `"Hello"` and `"Katrina"` are the arguments passed to the `greeting` and `name` parameters:
+
+```go
+PrintGreetingName("Hello", "Katrina")
+```
+
+## Pass by Value & Pass by Pointer
+
+It is also important to clarify the concept of passing by value as well as passing by pointer (sometimes referred to as _by reference_).
```
andrerfcsantos, line 48:
```suggestion
## Passing by value vs passing by reference

It is also important to clarify the concept of passing by value as well as passing by reference.
```

The term is "passing by reference". We can clarify in the text that reference here means passing a pointer (you are already doing that though).

concepts/functions/about.md
```diff
@@ -19,6 +22,55 @@ func PrintHelloName(name string) {
...
+
+func PrintHelloName(name string) {
+  fmt.Println("Hello " + name)
+}
+```
+
+Strictly speaking, all arguments are passed by value in Go, i.e. a copy is made of the value or data provided to the function.
+But if you don't want to make a copy of the data that you choose to pass to a function and
+want to actually access and change the data in the function,
+then you should use pointers as arguments, a.k.a. pass by reference.
```
andrerfcsantos, line 63: I feel like this explanation deserves an example. We give an example later in the text in the "Pointer" section which has some overlap with this one.

My suggestion here would be to join this explanation with the explanation in the "Pointers" section and just have one section talking about this. And maybe this explanation should be at the end of the document.

If you still think it makes sense to have to have 2 separate sections, I think they should at least be together and not separated like they are now.

concepts/functions/about.md
```diff
@@ -19,6 +22,55 @@ func PrintHelloName(name string) {
...
+
+In Go, not only can we pass primitive values into a function, but we can also pass a `pointer` of these values
+(a reference to the data) to the function.
+This means that the function receives only a copy to the pointer instead of the underlying data that this pointer refers to.
+Note that we say _a copy to the pointer_ since everything is passed by value to functions in Go,
+but a pointer copy, in contrast to a value copy, gives us an "address" that points to the data we want to access and modify.
```
andrerfcsantos, line 69: While correct, I don't think going into this level of detail of trying to explain why a pointer is in reality just a value too is helpful here. Might just confuse the students even more. I like the simple explanation of the "Pointers" section, and we probably should just leave as that.

andrerfcsantos (changes requested):
Some more comments.

concepts/functions/about.md
nahuakang: I think this is related to the other comment you made: There's only pass by value in Go so I chose to say "pass by pointer", which is technically correct but makes the concept a bit more complex. Happy to change it to simplify it a bit :)

nahuakang:
@andrerfcsantos Thanks for the feedback. Let me know if it's okay that I leave out the two updated sections in function concept's `introduction.md`? :)

andrerfcsantos (approved):
Awesome!

I think leaving those two sections out of the concepts' introduction is a very sensible decision.

PR #2198, opened 2022-04-21
Fix error in TestRemainingOvenTime, fix typos

deasa:
There was an error where `tt.expected` was being sent in as the `actualTimeInOven`, where it should have been `tt.time`

I also fixed a couple of typos in the test names.


andrerfcsantos (approved):
Thanks for this!

PR #2201, opened 2022-04-24
Interest is interesting does not specify zero

lucassperez:
This may sound pedantic, but the "interest is interesting" exercise does not specify what is the desired behaviour of the program when the **balance** is zero, because it talks about when this number is **negative** or **positive**, but formally speaking, zero is neither.

I genuinely had this doubt when solving the exercise, and since there are automated tests that put balance as specifically zero, I thought the instructions could be more explicit.

There are two ways to reword it that I thought of:

First:

- 0.5% for a **non negative** balance less than `1000` dollars.

Second:

- 0.5% for a balance greater or equal than `0` dollars and less than `1000` dollars.

How it is today, including

- 3.213% for a negative balance (balance gets more negative).
- 0.5% for a positive balance less than `1000` dollars.


exercises/concept/interest-is-interesting/.docs/instructions.md
```diff
@@ -5,7 +5,7 @@ Each year, the balance of your savings account is updated based on its interest
 The interest rate your bank gives you depends on the amount of money in your account (its balance):
 
 - 3.213% for a negative balance (balance gets more negative).
-- 0.5% for a positive balance less than `1000` dollars.
+- 0.5% for a non negative balance less than `1000` dollars.
```
andrerfcsantos, line 8:
```suggestion
- 0.5% for a positive balance greater or equal to 0 and less than `1000` dollars.
```

I actually like your second suggestion better so I applied it here.

Saying "non negative" for me doesn't fully solve the issue because while technically correct, it's still not explicit about 0.

andrerfcsantos (changes requested):
Thanks for raising this issue, you have a good point.

Left a minor suggestion.

exercises/concept/interest-is-interesting/.docs/instructions.md
lucassperez: Good idea! The more explicit, the better. :+1:

exercises/concept/interest-is-interesting/.docs/instructions.md
```diff
@@ -5,7 +5,7 @@ Each year, the balance of your savings account is updated based on its interest
 The interest rate your bank gives you depends on the amount of money in your account (its balance):
 
 - 3.213% for a negative balance (balance gets more negative).
-- 0.5% for a positive balance less than `1000` dollars.
+- 0.5% for a positive balance greater or equal to 0 and less than `1000` dollars.
```
junedev, line 8: @andrerfcsantos You suggested the sentence above but wouldn't this be more accurate?
```suggestion
- 0.5% for a balance greater or equal to 0 and less than `1000` dollars.
```

lucassperez: Well, I guess this is more accurante mathematically speaking and is also explicit about the number 0.

But now I realized that the other lines say:

```
- 1.621% for a positive balance greater or equal than `1000` dollars and less than `5000` dollars.
- 2.475% for a positive balance greater or equal than `5000` dollars.
```

Is the correct way **greater or equal to** or **greater or equal than**? English is not my first language, but I think it is the first one. Anyways, I think we should write all the sentences in the same way.

And maybe the correct way of saying this in english is **greater than or equal to**.

What do you guys think?

junedev: @ee7 Thanks for the help! @lucassperez Would be great if you could apply the fixes as suggested by @ee7.

(@lucassperez As background info, I asked in our Slack channel for a native speaker to help us out and @ee7 stepped up, Andre and me are both not native speakers as well.)

lucassperez: Applied the suggestions. Thanks for all the suggestions, everyone!

kotp (approved):
Definitely better.

PR #2207, opened 2022-04-28
#2155: Clarify when an exercise is tested with the `-race` flag

### LINKED ISSUE #2155: Clarify when an exercise is tested with the `-race` flag.
For some exercises, we are running the tests with the `-race` flag. The tests that run with this flag have a custom property in its `config.json`. Here's `bank-account` for example:

https://github.com/exercism/go/blob/920eb320d4dc6f569f55d65c4b4a15f5acc3401d/exercises/practice/bank-account/.meta/config.json#L34-L37

However, this can create some confusion among the students, as they can run the tests normally and have them working, but have them fail on the online editor. To make this a bit clearer I'd suggest the following:

* Add a clarification on [docs/TESTS.md](https://github.com/exercism/go/blob/main/docs/TESTS.md#tests) that for some exercises the tests might be run with this flag active.
* Add a sentence to the `instructions.append.md` of the exercise saying the particular exercise is run with this flag and provide the command to run the tests with the flag.

Right now the only exercise we have using this flag is `bank-account`.

nahuakang:
## Motivation

Fixes #2155.

## Note
Workflow job for Windows failed due to timeout. Cannot re-run jobs so maybe a maintainer could help out? 🙏

### DIFF
--- a/exercises/practice/bank-account/.docs/instructions.append.md
+++ b/exercises/practice/bank-account/.docs/instructions.append.md
@@ -14,6 +14,14 @@ For example: multiple goroutines may be depositing and withdrawing money
 simultaneously, two withdrawals occurring concurrently should not be able
 to bring the balance into the negative.
 
+To test that your code handles concurrency properly and does not introduce
+data races, run tests with `-race` flag on.
+
+```bash
+$ cd exercism/project/directory/go/bank-account
+$ go test -race
+```
+
 If you are new to concurrent operations in Go it will be worth looking
 at the sync package, specifically Mutexes:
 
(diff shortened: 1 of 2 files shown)


exercises/practice/bank-account/.docs/instructions.append.md
```diff
@@ -14,6 +14,14 @@ For example: multiple goroutines may be depositing and withdrawing money
 simultaneously, two withdrawals occurring concurrently should not be able
 to bring the balance into the negative.
 
+To test that your code handles concurrency properly and does not introduce
```
junedev, line 17: Since this content is visible for people working on the website as well as working locally, I would recommend to change it slightly so website users are not confused, e.g.
```suggestion
When working locally, you can test that your code handles concurrency properly and does not introduce
```

junedev (approved):
Thanks! This will avoid some more issues on the matter. 👏‍

Issue #2202, opened 2022-04-25
Format the output of code examples consistently across exercises/concepts


andrerfcsantos:
After some discussion in #2156, it was decided that to show the output of code examples, we must follow a standard format across exercises and concepts. The agreed format is to put the output of a piece of code as a comment after the code. The comment should start with an arrow `=>`. It looks like this:

```go
NeedsLicense("car")
// => true
```

Examples that have some specific reason to include a print statement should have the following format:
```go
result := NeedsLicense("car")
fmt.Println(result)
// Output: true
```

## Scope

The idea is to use this format across all exercises. Some exercises are not following this format right now, and we'll use this issue as a general tracking issue for PRs addressing this issue.

## Contributing to this issue

If you'd like to contribute to this issue, feel free to open PRs. PRs addressing this issue for a single exercise or just a few exercises or concepts are welcome - you don't need to fix this across all instances before opening a PR.

## Related PRs

- https://github.com/exercism/go/pull/2195
- https://github.com/exercism/go/pull/2197
- https://github.com/exercism/go/pull/2200

junedev:
Judging from https://github.com/exercism/go/pull/2472, it looks to me like most files are fixed by now. To avoid more PRs like https://github.com/exercism/go/pull/2472 that go in the wrong direction, I will close this issue for now. If we find more concrete places that need fixing, we can re-open it.

PR #2209, opened 2022-04-29
Fixed "Output" - Expenses exercise 

GabrielNSD:
Replace "Output:" for "=>" in example codes.

Related issue: #2202


andrerfcsantos (approved):
Thanks for this!

PR #2215, opened 2022-05-11
Tree Building: Tests accept tree in any order

dirkkul:
Currently the tests assume an order of a node's children, however the description does not specify anything about the order.

Example:
Build for test case "three nodes in order" returned 0:[2:[] 1:[]] but was expected to return 0:[1:[] 2:[]].

I added an additional step in the test-code where the expected nodes and the actual nodes are sorted by their ID before comparing them and now the above example is accepted. I think that the order shouldn't matter in a tree, if anyone thinks otherwise I'm happy to extend the description instead.

### DIFF
--- a/exercises/practice/tree-building/tree_building_test.go
+++ b/exercises/practice/tree-building/tree_building_test.go
@@ -3,6 +3,7 @@ package tree
 import (
 	"fmt"
 	"math/rand"
+	"sort"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


dirkkul:
Now linting should be fine, made the edit at first via web-UI :)

andrerfcsantos:
@dirkkul Thanks for raising an issue about this! I agree that the order on a tree shouldn't matter. Not having this, requires having to sort the records before building the tree, which I agree might be busy work for the exercise. However, I took a peek at how other tracks were implementing this exercise and in their exemplar solutions, they also sort the records before building the tree. So this looks like it's something all tracks "suffer" from.

Instead of fixing this just in the Go track, I believe we should change the description in all tracks to make it clearer that some kind of sorting is needed. You can suggest a change to the description of the exercise by opening an issue in the [problem-specifications](https://github.com/exercism/problem-specifications) about this asking if the description should be changed or if requiring some kind of sorting must be something that each track must decide individually. If the changes are agreed upon there, feel free to open a PR there implementing the changes, and that way all tracks will be able to benefit from it.

I'll close this for now, but if a conclusion is reached in problem-specifications is reached about this, feel free to ask for this PR to be reopened or create a new one bringing the changes from problem-specifications here to the Go track.

PR #2217, opened 2022-05-16
Tree Building: add test case

btl01:
failureTestCases: one root node and has parent

### DIFF
--- a/exercises/practice/tree-building/tree_building_test.go
+++ b/exercises/practice/tree-building/tree_building_test.go
@@ -170,6 +170,12 @@ var failureTestCases = []struct {
 	input []Record
 }{
 	{
+		name: "one root node and has parent",
+		input: []Record{
+			{ID: 0, Parent: 1},
+		},
+	},
+	{
 		name: "root node has parent",
 		input: []Record{
 			{ID: 0, Parent: 1},


andrerfcsantos:
Hi!

Thanks for this contribution.

Just curious, did you see any solution for which this test is needed, i.e a solution that passes without this but fails when this test is present? At first glance, there seems to be some overlap with what the test `root node has parent` below is trying to test.

btl01:
My contribution is meaningful to this
```go
func Build(records []Record) (*Node, error) {
	if len(records) == 0 {
		return nil, nil
	}

	rootNode := &Node{
		ID:       0,
		Children: []*Node{},
	}
	// Here is the problem sir
	if len(records) == 1 && records[0].ID == 0 {
		return rootNode, nil
	}
// More code here
}
```

full solution
```go
package tree

import (
	"errors"
	"sort"
)

type Record struct {
	ID     int
	Parent int
}

type Node struct {
	ID       int
	Children []*Node
}

func (parent *Node) findNode(ID int) *Node {
	if parent != nil {
		if parent.ID == ID {
			return parent
		}
		for _, n := range parent.Children {
			node := n.findNode(ID)
			if node == nil {
				continue
			}
			return node
		}
	}
	return nil
}

func Build(records []Record) (*Node, error) {
	if len(records) == 0 {
		return nil, nil
	}
	rootNode := &Node{
		ID:       0,
		Children: []*Node{},
	}

	// Here is the problem sir
	if len(records) == 1 && records[0].ID == 0 {
		return rootNode, nil
	}
	sort.Slice(records, func(i, j int) bool {
		return records[i].ID < records[j].ID
	})

	if records[0].ID > 0 {
		return nil, errors.New("no root node")
	}
	if records[0].Parent != 0 {
		return nil, errors.New("root has parent")
	}
	if len(records) == 1 {
		return rootNode, nil
	}
	for i := 1; i < len(records); i++ {
		if records[i].ID != 0 && records[i].ID <= records[i].Parent {
			return nil, errors.New("higher id parent of lower id")
		}
		if records[i].ID == records[i-1].ID {
			return nil, errors.New("duplicate node")
		}
		if records[i].ID-1 != records[i-1].ID {
			return nil, errors.New("non-continuous")
		}
		parent := rootNode.findNode(records[i].Parent)
		parent.Children = append(parent.Children, &Node{
			ID:       records[i].ID,
			Children: []*Node{},
		})
	}
	return rootNode, nil
}

```

andrerfcsantos (approved):
Thanks for this!

PR #2218, opened 2022-05-18
Update instructions.md

arthis:
Missing information on the instructions. It drives people nut :), at least me. The part "If the scale is broken" might not be obvious to understand that it concerns the ErrScaleMalfunction, especially for non native english speaker.


andrerfcsantos (approved):
Thanks for this!

This is a common source of confusion for students and I think this can help indeed.

PR #2220, opened 2022-05-21
Fix typo in slices and maps paragraph

patoui:
Description
----
There's a small typo in the explanation for the "Slices and maps are already pointers" paragraph `that not` becomes `than not` :)


andrerfcsantos (approved):
Thanks a lot for this!

Nice cleanup.

PR #2223, opened 2022-05-26
Fixed "Output"- All remaining files

GabrielNSD:
Replace "Output:" for "=>" in example codes.

Related issue: https://github.com/exercism/go/issues/2202

### DIFF
--- a/concepts/randomness/about.md
+++ b/concepts/randomness/about.md
@@ -64,7 +64,7 @@ func main() {
 }
 ```
 
-Every time this program runs, it will produce the same output:
+Every time this program runs, it will produce the same =>
 
 ```text
 498081 727887 131847
--- a/concepts/booleans/about.md
+++ b/concepts/booleans/about.md
@@ -5,16 +5,16 @@ Booleans in Go are represented by the `bool` type, which values can be either `t
 Go supports three [boolean operators][logical operators]: `!` (NOT), `&&` (AND), and `||` (OR).
 
 ```go
-true || false // Output: true
-true && false // Output: false
-!true // Output: false
+true || false // => true
+true && false // => false
+!true // => false
 ```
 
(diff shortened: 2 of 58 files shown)


concepts/randomness/about.md
```diff
@@ -64,7 +64,7 @@ func main() {
 }
 ```
 
-Every time this program runs, it will produce the same output:
+Every time this program runs, it will produce the same =>
```
andrerfcsantos, line 67: The substitution here is not needed

andrerfcsantos (changes requested):
Thanks a lot for taking care of this issue! This is a big one and will be great to keep consistency on the track.

I assume to do this you made a simple substitution from `Output:` to `=>`, which is a good starting point, but has some issues:

- The case I pointed out in the suggestion, where `output:` is part of a paragraph and must not be replaced with `=>`
- It was agreed upon that if a print statement is present, we should keep using `Output:` (https://github.com/exercism/go/issues/2202#:~:text=Examples%20that%20have%20some%20specific%20reason%20to%20include%20a%20print%20statement%20should%20have%20the%20following%20format%3A). Here's an example:

```go
result := NeedsLicense("car")
fmt.Println(result)
// Output: true
```

Arguably, the variable here is not needed and it is used just for the print statement, in which case, we can consider removing the print statement entirely and use the `=>` syntax:

```go
NeedsLicense("car")
// => true
```

So, my suggested changes would be:

- Fix the instance where `output:` is part of a paragraph.
- For instances where the output follows print statements, either keep using `Output:` or check if we can remove the print statement/variables and simplify the examples like shown above. For now, I think we can keep using `Output:` in those situations as it is easier. The simplification of the examples can be a separate issue, as it will require a lot more work and more careful checking of each exercise because the variables and the print statements might be important in some exercises. However, if you want to do that simplification in this PR, go for it and I'll make sure to size it accordingly.

andrerfcsantos:
There has been some interest by other contributors in fixing this issue in individual exercises, which is making this PR conflict with the main branch. The conflicts will become even more as new contributors fix this issue in more exercises.

For that reason, and since reproducing this PR can be done easily with a find and replace, I'm going to close this PR.

If you are still interested in solving this issue across all exercises, please create a new PR integrating the feedback in https://github.com/exercism/go/pull/2223#pullrequestreview-985873826.

PR #2225, opened 2022-05-27
Improve exercise nth-prime

### LINKED ISSUE #2224: Improve exercise nth-prime
Hi all!

While I was solving nth-prime exercise I figured out there might be two problems regarding this exercise:
1) The description is really vague regarding what the main function should return. It is not clear why there is a boolean field,
2) Looking at test cases, one can figure out that the solution should return if the input is less than 1 or not (since there is no 0th prime number!) and it does not contribute to the main point of exercise at all!

The first problem arises because of the second one and I believe two solutions can help the exercise become better here:
1) Either set the input limit from n>=1 so there would be no need to return a boolean (which makes the exercise much cleaner and does not hurt the main goal of itx)
2) Require to panic or return an error in case of an input less than 1 and make it clear in the description that the input must be bigger than 1 and in other cases it should return an error

I was wondering what is your thoughts on this.
P.S. I did not mention this problem in [problem-specifications](https://github.com/exercism/problem-specifications/tree/main/exercises/nth-prime) because although the description is vague there as well but looking at [canonical-data.json](https://github.com/exercism/problem-specifications/blob/main/exercises/nth-prime/canonical-data.json) you can see there need to be an error in case of n=0

MikaeelMF:
Hi!
Fixes #2224
According to issue #2224, I have changed this exercise to require an error instead of a boolean in case of input out of range.

### DIFF
--- a/exercises/practice/nth-prime/cases_test.go
+++ b/exercises/practice/nth-prime/cases_test.go
@@ -1,43 +1,44 @@
 package prime
 
 // Source: exercism/problem-specifications
-// Commit: 4a3ba76 nth-prime: Apply new "input" policy
-// Problem Specifications Version: 2.1.0
+// Commit: 42dd0ce Remove version (#1678)
+
+import "errors"
 
 var tests = []struct {
 	description string
 	n           int
 	p           int
-	ok          bool
+	err         error
 }{
 	{
 		"first prime",
 		1,
 		2,
-		true,
+		nil,
 	},
 	{
 		"second prime",
 		2,
 		3,
-		true,
+		nil,
 	},
 	{
 		"sixth prime",
 		6,
 		13,
-		true,
+		nil,
 	},
 	{
 		"big prime",
 		10001,
 		104743,
-		true,
+		nil,
 	},
 	{
 		"there is no zeroth prime",
 		0,
 		0,
-		false,
+		errors.New("Input must be greater than 1"),
 	},
 }
--- a/exercises/practice/nth-prime/cases_test.go
+++ b/exercises/practice/nth-prime/cases_test.go
@@ -39,6 +39,6 @@ var tests = []struct {
 		"there is no zeroth prime",
 		0,
 		0,
-		errors.New("Input must be greater than 1"),
+		errors.New("input must be greater than 1"),
 	},
 }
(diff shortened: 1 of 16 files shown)


MikaeelMF:
I don't understand why the linter things `err` is a boolean.

andrerfcsantos:
@MikaeelMF For changes like this, you must also change the [exemple.go](https://github.com/exercism/go/blob/main/exercises/practice/nth-prime/.meta/example.go) of the solution. It's this solution that is being used for the tests. If you are curious, [this is the script](https://github.com/exercism/go/blob/main/bin/test-without-stubs) that is being ran to perform the tests. Note that it looks for an `.meta/example.go` or an `.meta/exemplar.go` and copies it to a temporary dir and then runs the tests on that dir. This is most likely the reason the tests are failing.

Keeping your branches up-to-date with this repo is a good idea. However, you don't necessarily need a new PR if your version is outdated. You can rebase the branch of this pr (your `main` branch) with the main branch of the remote you have configured for the `exercism/go` remote. You most likely won't have conflicts, and if you do, I don't think they'll be big/hard to solve. Then force pushing your main after that rebase will update the PR.

And don't worry about the commit history. Even if you do several commits for the PR, when merging, we always squash them into a single one automatically, so the history always looks clean. So make as many commits as you want. Your branch, your rules :)

exercises/practice/nth-prime/cases_test.go
```diff
@@ -1,43 +1,44 @@
...
 		104743,
-		true,
+		nil,
 	},
 	{
 		"there is no zeroth prime",
 		0,
 		0,
-		false,
+		errors.New("Input must be greater than 1"),
```
andrerfcsantos, line 42:
```suggestion
		errors.New("input must be greater than 1"),
```

In Go, error messages are usually not capitalized.

andrerfcsantos (changes requested):
Some more suggestions, after that we can merge :)

MikaeelMF:
@andrerfcsantos 
All done! Thank you for the suggestions!

andrerfcsantos (approved):

andrerfcsantos:
@MikaeelMF Looks great! Made just a little change to add you to the list of contributors to the exercise - the changes you made were are significant enough to justify it :)

MikaeelMF:
> @MikaeelMF Looks great! Made just a little change to add you to the list of contributors to the exercise - the changes you made were are significant enough to justify it :)

Wow, Awesome! Thanks!

PR #2229, opened 2022-06-02
Improve Linked List

### LINKED ISSUE #2219: Linked List: Inconsistency between method names in global and track specific instructions
Hi,

I'm currently doing the linked list exercise and noticed that some methods are named differently in the global and track specific instructions. In specific, in the readme part that comes from the global problem specification (instructions.md), the List should have the the methods:
* `push` (*insert value at back*);
* `pop` (*remove value at back*);
* `shift` (*remove value at front*).
* `unshift` (*insert value at front*);

However in the stub file and the track-specific part of the readme (instructions_append.md) they are called:
- PushFront
- PushBack
- PopFront
- PopBack

Moreover, In the description the Node field should have a field "Val", while the tests check for a member "Value" (this was recently changed [here](https://github.com/exercism/go/issues/2090)). 

My proposal would be to:
1) change the Val to Value in instructions_append.md
2) name the List methods consistently with the problem-specification (eg push, pop etc), even though I like the track-specific names more

I'm happy to do an MR in case this is ok.

EldoranDev:
Hi,

this should fix #2219.

Adapted the task according to the comment https://github.com/exercism/go/issues/2219#issuecomment-1139120968 .

### DIFF
--- a/exercises/practice/linked-list/linked_list.go
+++ b/exercises/practice/linked-list/linked_list.go
@@ -16,7 +16,7 @@ func (n *Node) Prev() *Node {
 }
 
 func (l *List) Unshift(v interface{}) {
-	panic("Please implement the PushFront function")
+	panic("Please implement the Unshift function")
 }
 
 func (l *List) Push(v interface{}) {
--- a/exercises/practice/linked-list/linked_list.go
+++ b/exercises/practice/linked-list/linked_list.go
@@ -20,7 +20,7 @@ func (l *List) Unshift(v interface{}) {
 }
 
 func (l *List) Push(v interface{}) {
-	panic("Please implement the PushBack function")
+	panic("Please implement the Push function")
 }
 
 func (l *List) Unshift() (interface{}, error) {
--- a/exercises/practice/linked-list/linked_list.go
+++ b/exercises/practice/linked-list/linked_list.go
@@ -28,7 +28,7 @@ func (l *List) Shift() (interface{}, error) {
 }
 
 func (l *List) Pop() (interface{}, error) {
-	panic("Please implement the PopBack function")
+	panic("Please implement the Pop function")
 }
 
 func (l *List) Reverse() {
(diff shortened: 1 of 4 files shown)


exercises/practice/linked-list/linked_list.go
```diff
@@ -15,19 +15,19 @@ func (n *Node) Prev() *Node {
 	panic("Please implement the Prev function")
 }
 
-func (l *List) PushFront(v interface{}) {
+func (l *List) Unshift(v interface{}) {
 	panic("Please implement the PushFront function")
```
andrerfcsantos, line 19:
```suggestion
	panic("Please implement the Unshift function")
```

exercises/practice/linked-list/linked_list.go
```diff
@@ -15,19 +15,19 @@ func (n *Node) Prev() *Node {
...
 }
 
-func (l *List) PushBack(v interface{}) {
+func (l *List) Push(v interface{}) {
 	panic("Please implement the PushBack function")
```
andrerfcsantos, line 23:
```suggestion
	panic("Please implement the Push function")
```

exercises/practice/linked-list/linked_list.go
```diff
@@ -15,19 +15,19 @@ func (n *Node) Prev() *Node {
...
 }
 
-func (l *List) PopFront() (interface{}, error) {
+func (l *List) Unshift() (interface{}, error) {
 	panic("Please implement the PopFront function")
 }
```
andrerfcsantos, line 28:
```suggestion
func (l *List) Shift() (interface{}, error) {
	panic("Please implement the Shift function")
}
```

The Unshift function is repeated here. I think you meant Shift.

exercises/practice/linked-list/linked_list.go
```diff
@@ -15,19 +15,19 @@ func (n *Node) Prev() *Node {
...
 
-func (l *List) PopBack() (interface{}, error) {
+func (l *List) Pop() (interface{}, error) {
 	panic("Please implement the PopBack function")
```
andrerfcsantos, line 31:
```suggestion
	panic("Please implement the Pop function")
```

andrerfcsantos (changes requested):
A few changes to the stub so it conforms with the new names.
After that we can merge 🚀

andrerfcsantos (approved):

PR #2233, opened 2022-06-07
✏️ Fix typo

vdavez:
`ane` -> `are`

### DIFF
--- a/exercises/concept/meteorology/.docs/introduction.md
+++ b/exercises/concept/meteorology/.docs/introduction.md
@@ -31,7 +31,7 @@ type Distance struct {
 } 
 ```
 
-In the example above, `Kilometer` and `Mile` ane constants of type `DistanceUnit`.
+In the example above, `Kilometer` and `Mile` are constants of type `DistanceUnit`.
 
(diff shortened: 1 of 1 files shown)


imattf:
Possible additional typo in...

```
mileUnit := Mile
fmt.Sprint(mileUnit)
// => 1
// The result is '1' because that is the underlying value of the 'Mile' contant (see contant declarations above) 
```

...thinking last comment should be "constant" ?

andrerfcsantos:
@imattf Well spotted. Feel free to open a PR fixing that typo! If you could fix the typo in theses places would be great too:

```console
$ find . -type f -exec grep -n 'contant' {} /dev/null \;
./concepts/stringers/about.md:43:// The result is '1' because that is the underlying value of the 'Mile' contant (see contant declarations above)
./concepts/stringers/introduction.md:43:// The result is '1' because that is the underlying value of the 'Mile' contant (see contant declarations above)
./exercises/concept/meteorology/.docs/introduction.md:43:// The result is '1' because that is the underlying value of the 'Mile' contant (see contant declarations above)
```

PR #2235, opened 2022-06-07
Need for Speed: Minor improvements (#1755)

### LINKED ISSUE #1755: Need for Speed: Apply minor improvements
A few things I've encountered when solving the Need for Speed exercise:

- I can't solve the tasks in order due to `undefined: Track` errors. I have to do part of Task 1, then part of Task 2, then I can resume with Task 1. Since there's no extra knowledge required for defining the track struct, this could be moved into Task 1. Also related to https://github.com/exercism/go-test-runner/issues/48
- What is the point of `NewTrack`? Why can't I just call `Track{}` directly?
- Task 4: My naïve ansatz would be to loop and `Drive` until the battery runs out and break out of the loop with `true` if the track has been finished. I don't know loops yet, though. I suspect this might lead to some people searching for loops and using those instead. Perhaps this can be prevented through the instructions.
- Task 4: Instructions call the parameter `raceTrack`, the code calls it `track`.

eklatzer:
Fixes #1755

Hey, I have done the following:

- adjusted the `instructions.md` to clarify the struct-initialization and the usage of constructors
- renamed variables for consistency

### DIFF
--- a/concepts/structs/about.md
+++ b/concepts/structs/about.md
@@ -13,6 +13,8 @@ type Shape struct{
 }
 ```
 
+Field names in structs follow the Go convention - fields whose name starts with a lower case letter are only visible to code in the same package, whereas those whose name starts with an upper case letter are visible in other packages.
+
 Once you have defined the `struct`, you need to create a new instance defining the fields using their field name
 in any order:
 
@@ -36,47 +38,16 @@ fmt.Printf("name: %s size: %d\n", s.name, s.size)
 Fields that don't have an initial value assigned, will have their zero value. For example:
 
 ```go
-person:= Person{}
-fmt.Printf("name: %s age: %d\n", person.name, person.age)
-// Output: name: age: 0
-```
-
-Another way of creating a new instance of a struct is by using the `new` built-in: 
-```go
-s2 := new(Shape)
-fmt.Printf("name: %s size: %d\n", s2.name, s2.size)
-// Output: name:  size: 0
-```
-
-Sometimes it can be nice to have a _constructor_, which is a function to create an instance of a struct.
-The following constructor can be used to create a new instance of `Shape` and automatically set a default-value for the size of the shape:
-
-```go
-func NewShape(name string) Shape {
-	return Shape{
-		name: name,
-		size: 100, //default-value for size is 100
-	}
-}
-```
-
-Using a constructor can have the following advantages:
-* validation of the given values
-* handling of default-values
-
-Another advantage will be shown with an example. When importing the `package shapes` containing the
-struct `Shape` the unexported fields `name` and `size` cannot be accessed. Field names in structs follow the Go convention - fields whose name starts
-with a lower case letter are only visible to code in the same package, whereas those whose name starts with an upper case letter are visible in other packages.
-Using a constructor the initial value can nevertheless be set:
-```go
-rect := shapes.NewShape("Rectangle")
+s := Shape{name: "Circle"}
+fmt.Printf("name: %s size: %d\n", s.name, s.size)
+// Output: name: Circle size: 0
 ```
 
 You can create an instance of a `struct` without using the field names, as long as you define the fields _in order_:
 
 ```go
-person := Person{
-	"Mary",
+s := Shape{
+	"Oval",
 	20,
 }
 ```
@@ -91,7 +62,7 @@ type Shape struct {
 	size        int
 }
 
-person := Shape{
+s := Shape{
     "Circle",
     20,
 }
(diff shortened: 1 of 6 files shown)


eklatzer:
Hey, I adjusted the format of the outputs because I have seen #2202

eklatzer:
The other changes should fix the TODO's from #1755

andrerfcsantos (changes requested):
I'm just going to suggest a change in the order we talk about these concepts. I think we should talk about string literals first and then the `New()` or `New***()` functions to create structs. I would leave the explanation of the `new` builtin for last as it is not as relevant for the exercise .

Also note that the `introduction.md` for the exercise is often a copy of the `introduction.md` and `about.md` of the concept they teach, in this case [structs](https://github.com/exercism/go/tree/main/concepts/structs), so the changes must also be reflected there. You can leave this for last, just mentioning so it's not forgotten.

eklatzer:
I hope I have covered everything you wanted me to change

eklatzer:
@andrerfcsantos I am a bit confused, have you already checked my changes, is something missing or what is the current state?

andrerfcsantos:
I did take a look at this and I have some changes in mind, but I want to think about them a little bit better before actually suggesting them. This is why this is taking longer to review, didn't have the chance yet to fully gather and formalize my thoughts - but it is not forgotten! :)

concepts/structs/about.md
```diff
@@ -39,6 +41,37 @@ fmt.Printf("name: %s age: %d\n", person.name, person.age)
...
 
+Another way of creating a new instance of a struct is by using the `new` built-in: 
+```go
+s2 := new(Shape)
+fmt.Printf("name: %s size: %d\n", s2.name, s2.size)
+// Output: name:  size: 0
+```
+
+Sometimes it can be nice to have a _constructor_, which is a function to create an instance of a struct.
+The following constructor can be used to create a new instance of `Shape` and automatically set a default-value for the size of the shape:
```
andrerfcsantos, line 52:
```suggestion
Sometimes it can be nice to have functions that help us create struct instances. By convention, these functions usually are called `New` or have their names starting with `New`, but since they are just regular functions, you can give them any name you want.  They might remind you of constructors in other languages, but in Go they are just regular functions.

In the following example, one of these `New` functions is used to create a new instance of `Shape` and automatically set a default value for the `size` of the shape:
```

Avoided the word "constructor" here because they are not really constructors in the sense they are "attached" to a particular type. I think it's important to empathize that these are just regular functions.

concepts/structs/about.md
```diff
@@ -39,6 +41,37 @@ fmt.Printf("name: %s age: %d\n", person.name, person.age)
...
+	return Shape{
+		name: name,
+		size: 100, //default-value for size is 100
+	}
+}
+```
+
+Using a constructor can have the following advantages:
+* validation of the given values
+* handling of default-values
```
andrerfcsantos, line 65:
```suggestion
Using `New` functions can have the following advantages:

* validation of the given values
* initialization of the struct fields with default values other than the zero value for the type
* since `New` functions are often declared in the same package of the structs they initialize, they can initialize even private fields of the struct
```

concepts/structs/about.md
```diff
@@ -39,6 +41,37 @@ fmt.Printf("name: %s age: %d\n", person.name, person.age)
...
+
+Another advantage will be shown with an example. When importing the `package shapes` containing the
+struct `Shape` the unexported fields `name` and `size` cannot be accessed. Field names in structs follow the Go convention - fields whose name starts
+with a lower case letter are only visible to code in the same package, whereas those whose name starts with an upper case letter are visible in other packages.
+Using a constructor the initial value can nevertheless be set:
+```go
+rect := shapes.NewShape("Rectangle")
+```
```
andrerfcsantos, line 73: I don't think we need to go into this much detail and the example doesn't really show the initialization of private fields.

Just mentioning briefly is enough. I tried to summarize this bit in the previous suggestion.

concepts/structs/about.md
```diff
@@ -39,6 +41,37 @@ fmt.Printf("name: %s age: %d\n", person.name, person.age)
...
+
 You can create an instance of a `struct` without using the field names, as long as you define the fields _in order_:
```
andrerfcsantos, line 75: This bit makes more sense to be together with the part where we talk about how to initialize structs using struct literals. After showing how to initialize the struct with the fields, we can say that including the name of the fields is optional.

concepts/structs/about.md
```diff
@@ -52,23 +85,21 @@ However, this syntax is considered _brittle code_ (breaks when a field is added)
...
+    "Circle",
     20,
 }
 // The compiler will fail with a similar error:
 // Output: cannot use 20 (type untyped int) as type string in field value
-// Output: too few values in Person{...}
+// Output: too few values in Shape{...}
 ```
 
-Field names in structs follow the Go convention - fields whose name starts with a lower case letter are only visible to code in the same package, whereas those whose name starts with an upper case letter are visible in other packages.
```
andrerfcsantos, line 70: I think this was removed by accident?

andrerfcsantos (changes requested):
Left my review in `about.md` just to make things simpler, but please also apply the changes in the other files.

concepts/structs/about.md
```diff
@@ -1,34 +1,36 @@
 # About
+
 In Go, a `struct` is a sequence of named elements called _fields_, each field has a name and type. The name of a field must be unique within the struct.
 Structs can be compared with _classes_ in the Object-Oriented Programming paradigm.
 
 You create a new struct by using the `type` keyword and the **_built-in type_** `struct`, and explicitly define the name and type of the fields.
-For example, to define a `Person` struct:
+For example, to define a `Shape` struct:
```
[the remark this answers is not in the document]
eklatzer: I changed the struct to `Shape`, because later in the constructor, I use a default-value:
```
func NewShape(name string) Shape {
	return Shape{
		name: name,
		size: 100, //default-value for size is 100
	}
}
```

It was hard to find a default value for `Person` that is not somehow political or racist or something like this.

concepts/structs/about.md
```diff
@@ -1,34 +1,36 @@
...
-fmt.Printf("name: %s age: %d\n", person.name, person.age)
-// Output: name: John age: 23
+// Update the name and the size
+s.name = "Circle"
+s.size = 35
+fmt.Printf("name: %s size: %d\n", s.name, s.size)
+// Output: name: Circle size: 35
 ```
 
 Fields that don't have an initial value assigned, will have their zero value. For example:
```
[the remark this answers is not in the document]
eklatzer: OK like this?

Fields that don't have an initial value assigned, will have their zero value. For example:

```go
s := Shape{name: "Circle"}
fmt.Printf("name: %s size: %d\n", s.name, s.size)
// Output: name: Circle size: 0
```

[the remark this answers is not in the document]
andrerfcsantos: Yes :)

eklatzer:
@andrerfcsantos Thanks for the feedback, I hope I have addressed all of your messages.

andrerfcsantos (approved):

Issue #2232, opened 2022-06-05
logs-logs-logs: Additional tests to ensure the student is checking the first application character


andrerfcsantos:
This solution has a bug, but passes our current tests:

```go
type Applications struct {
	Key   rune
	Value string
}

//mapping based on struct to have an as defined ordered array
var mapping = []Applications{
	{ Key:   '❗', Value: "recommendation" },
	{Key:   '🔍', Value: "search"},
	{Key:   '☀',Value: "weather"},
}

// Application identifies the application emitting the given log.
func Application(log string) string {
	for _, val := range mapping {
		if strings.ContainsRune(log, val.Key) {
			return val.Value
		}
	}
	return "default"
}
```

The bug with this solution is that it goes to a list of application characters and returns the application of the _first application character on that list_ that also appears in the string. However, the specifications say that it must be the first application character that appear _on the log line_ that must determine the application of the log:

> If a log line contains more than one character in the above table, return the application corresponding to the first character found in the log line starting from left to right.

We only have 1 test that contains multiple application characters in it:

https://github.com/exercism/go/blob/b048fb5b594390bc70580656eeda63710cc190b5/exercises/concept/logs-logs-logs/logs_logs_logs_test.go#L33-L37

The solution above passes because the first application character on the log line (`❗`) also happens to appear before the 🔍 in the `mapping` slice. But if we were to have a test like:

```go
		{
			name: "multiple characters search",
			log:  "🔍 search about recommendation ❗",
			want: "search",
		},
```

Then the solution above would fail.

My suggestion would be to add additional tests to catch this. The test above is one that can be added, but I think having another one with multiple application characters where `☀` is the first application character that appears would also be good.

andrerfcsantos:
@Jasstkn Go ahead! Let us know if you have any questions.

andrerfcsantos:
We should, thanks for remembering me.

This can be done automatically when the linked PR is merged, but we didn't use any of the [magic words](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) so it wasn't.

PR #2236, opened 2022-06-07
logs-logs-logs: add more tests to better cover multiple characters case

Jasstkn:

Add more tests to cover case that is described here https://github.com/exercism/go/issues/2232

With logic from issue:
```golang
type Applications struct {
	Key   rune
	Value string
}

//mapping based on struct to have an as defined ordered array
var mapping = []Applications{
	{ Key:   '❗', Value: "recommendation" },
	{Key:   '🔍', Value: "search"},
	{Key:   '☀',Value: "weather"},
}

// Application identifies the application emitting the given log.
func Application(log string) string {
	for _, val := range mapping {
		if strings.ContainsRune(log, val.Key) {
			return val.Value
		}
	}
	return "default"
}
```

With new tests it catches case where the first element in the variable `mapping` isn't the same as in the input log for `Application` function.
```golang
--- FAIL: TestApplication (0.00s)
    --- FAIL: TestApplication/multiple_characters_recommendation(2/3) (0.00s)
        logs_logs_logs_test.go:53: Application("🔍 search recommended product ❗") = "recommendation", want "search"
    --- FAIL: TestApplication/multiple_characters_recommendation(3/3) (0.00s)
        logs_logs_logs_test.go:53: Application("☀ weather is sunny ❗") = "recommendation", want "weather"
```

Also it fixes formatting that is described here: #2202

### DIFF
--- a/exercises/concept/logs-logs-logs/.docs/introduction.md
+++ b/exercises/concept/logs-logs-logs/.docs/introduction.md
@@ -38,7 +38,7 @@ Since `rune` is just an alias for `int32`, printing a rune's type will yield `in
 ```go
 myRune := '¿'
 fmt.Printf("myRune type: %T\n", myRune)
-// Output: myRune type: int32
+// => myRune type: int32
 ```
 
 Similarly, printing a rune's value will yield its integer (decimal) value:
@@ -46,7 +46,7 @@ Similarly, printing a rune's value will yield its integer (decimal) value:
 ```go
 myRune := '¿'
 fmt.Printf("myRune value: %v\n", myRune)
-// Output: myRune value: 191
+// => myRune value: 191
 ```
 
 To print the Unicode character represented by the rune, use the `%c` formatting verb:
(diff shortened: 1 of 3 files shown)


Jasstkn:
hey @andrerfcsantos! thanks for providing me opportunity to contribute. I decided to add all of the three emoji in the tests so no matter what is the order in the mapping variable - we catch wrong behaviour. 
Looking forward to your feedback!

exercises/concept/logs-logs-logs/.docs/introduction.md
```diff
@@ -38,31 +38,31 @@ Since `rune` is just an alias for `int32`, printing a rune's type will yield `in
 ```go
 myRune := '¿'
 fmt.Printf("myRune type: %T\n", myRune)
-// Output: myRune type: int32
+// => myRune type: int32
```
andrerfcsantos, line 41:
```suggestion
// Output: myRune type: int32
```

As discussed here https://github.com/exercism/go/issues/2202 when an instruction is a print statement, we actually want to keep the `Output:` wording.

Can you change this back in the other print statements?

andrerfcsantos (changes requested):
Thanks for this!

Just left a minor comment.

exercises/concept/logs-logs-logs/.docs/introduction.md
Jasstkn: Sorry, my bad. Reverted changes in https://github.com/exercism/go/pull/2236/commits/c773cd8c1bac7a02048db3e92e7abd3620ee32b6.

andrerfcsantos (approved):

Issue #2237, opened 2022-06-07
Need for speed: car already in on the track?


vanclmil:
It is not clear for me whether the car may be already on the track or not (from this [information](https://github.com/exercism/go/blob/37fda27323803a9c15a88ef1e756b13f51c51d7d/exercises/concept/need-for-speed/.docs/instructions.md?plain=1#L59)).

If so, there is missing test case to check it, see example:
```go
speed := 5
batteryDrain := 10
car := NewCar(speed, batteryDrain)
car.distance = 50
distance := 100
raceTrack := NewTrack(distance)
CanFinish(car, raceTrack)
// Output: true
```
In that case, I think that most of the community solutions would return false, so fail.

andrerfcsantos:
Thanks for opening an issue!

In our test suite, we do have tests that have `distance > 0` and `battery < 100%`, see [need_for_speed_test.go](https://github.com/exercism/go/blob/37fda27323803a9c15a88ef1e756b13f51c51d7d/exercises/concept/need-for-speed/need_for_speed_test.go#L169). Do you think we are missing a test there?

eklatzer:
Hey, the following test-case nearly covers this case:
```
		{
			name: "Car can finish the track distance just as the battery is drained.",
			car: Car{
				speed:        5,
				batteryDrain: 5,
				battery:      100,
				distance:     20,
			},
			track: Track{
				distance: 100,
			},
			expected: true,
		},
```
But the possible distance is 100: `car.battery / car.batteryDrain * car.speed`
(100/5) * 5 = 20 * 5 = 100
 
For the example of this issue:
(100/10) * 5=10 * 5=50

The point: as the car already drove a distance of 50, the total distance is still 100, because:
- already driven: 50
- still possible: 50
- -->sum: 100
To cover this case, `func CanFinish` would need the following refactoring:
```
func CanFinish(car Car, track Track) bool {
	maxDistance := car.battery / car.batteryDrain * car.speed + car.distance
	return track.distance <= maxDistance
}
```
It depends whether you count the distance on the car for the current track or not

vanclmil:
@eklatzer thank you very much for your sumarization, that's exactly what I wanted to point out

andrerfcsantos:
The idea of the exercise is to not consider `distance` in the computations. This means that if the car has `distance > 0`, this distance was driven before the track. The `Assume that you start the engine of the car for the race` is meant to make it more obvious that you are just starting the race, but might not be enough?

[...]

PR #2241, opened 2022-06-09
clarification of the instructions for Need For Speed #2237

eklatzer:
Hey, I changed the text for the instructions of Need For Speed as discussed in #2237


andrerfcsantos (approved):
Thanks!

PR #2244, opened 2022-06-10
fix function arguments and parameters missmatch

vanclmil:
I am afraid that there is a mismatch of "arguments" and "parameters" in the Functions introduction. Regarding the description above (Parameters vs. Arguments) the sentence discusses "parameters" instead of "arguments".


andrerfcsantos (approved):

andrerfcsantos:
Thanks for this and for taking extra care to fix it in all 3 places!

PR #2248, opened 2022-06-10
Minor mistake in exercise 3. example

jonasbn:
It seemed there was a minor mistake where the wrong variable was used, compared to the outlined result


andrerfcsantos (approved):

andrerfcsantos:
Thanks for this!

PR #2249, opened 2022-06-11
elons-toys: improve description for first stage

Jasstkn:
It seems to be not clear what does the note mean:
> Note: If a car's battery is below **its battery drain percentage**, you can't drive the car anymore.

Changed to the more clear definition. 

### DIFF
--- a/exercises/concept/elons-toys/.docs/instructions.md
+++ b/exercises/concept/elons-toys/.docs/instructions.md
@@ -27,7 +27,7 @@ car.Drive()
 // car is now Car{speed: 5, batteryDrain: 2, battery: 98, distance: 5}
 ```
 
-Note: If as a result of execution `car.Drive()` car's battery is below 0, you can't drive the car anymore.
+Note: You should not try to drive the car if doing so will cause the car's battery to be below 0.
 
 ## 2. Display the distance driven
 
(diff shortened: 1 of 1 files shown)


exercises/concept/elons-toys/.docs/instructions.md
```diff
@@ -27,7 +27,7 @@ car.Drive()
 // car is now Car{speed: 5, batteryDrain: 2, battery: 98, distance: 5}
 ```
 
-Note: If a car's battery is below its battery drain percentage, you can't drive the car anymore.
+Note: If as a result of execution `car.Drive()` car's battery is below 0, you can't drive the car anymore.
```
andrerfcsantos, line 30:
```suggestion
Note: You should not try to drive the car if doing so will cause the car's battery to be below 0.
```

Let me know if this suggestion clears up your confusion.

The goal of that sentence is to tell the student to not drive the car if the battery would be depleted as a result.

In that regard, I think your original suggestion makes it worse because the wording `If as a result of execution car.Drive() car's battery is below 0 ...` suggests that you should try to drive the car and only then check if the battery is less than 0, and if it is, only after that you can't drive anymore, which is not what we are trying to say.

andrerfcsantos (changes requested):

exercises/concept/elons-toys/.docs/instructions.md
Jasstkn: Yes, it's much better. Sorry, I was trying to clear up the same confusion but failed a bit :c

andrerfcsantos (approved):

PR #2255, opened 2022-06-17
Added the test-cases from canonical-data.json to exercise diamond

### LINKED ISSUE #2222: Improvement idea for Diamond tests
While solving the Diamond exercise, I had a difficult time iteratively solving the exercise because [diamond_test.go]( https://github.com/exercism/go/blob/main/exercises/practice/diamond/diamond_test.go) is all or nothing. In other words, the entire solution is implemented and all tests pass or the solution isn't implemented and all tests fail.

While solving, I found it helpful to iteratively solve against test cases like:

```go
func TestDiamond(t *testing.T) {
	tests := []test{
		{'A', "A\n"},
		{'B', " A \nB B\n A \n"},
		{'C', "  A  \n B B \nC   C\n B B \n  A  \n"},
		{'D', "   A   \n  B B  \n C   C \nD     D\n C   C \n  B B  \n   A   \n"},
	}
	for _, test := range tests {
		got, err := Gen(byte(test.input))

		if err != nil {
			t.Errorf("Gen(byte(%v)) err %v", test.input, err)
		}
		if got != test.want {
			t.Errorf("Gen(byte(%v) got\n%v want\n%v", test.input, got, test.want)
		}
	}
}
```

If maintainers are open to a test like the above being added to the exercise, I think it could help future solvers and I'm happy to create a PR for it.

eklatzer:
Hey, I have done the following steps:

- added a test-case generator for the test-cases defined in [canonical-data.json](https://github.com/exercism/problem-specifications/blob/main/exercises/diamond/canonical-data.json) in `gen.go`
- regenerated the `tests.toml` using configlet
- generated the `cases_test.go` using the `gen.go`
- added `func TestDiamond` using this test-cases

I additionally adjusted the existing tests, as they expected a `\n` at the end. Regarding to the `instructions.md` there should not be a new-line at the end, for example:
Diamond for letter 'E':
```text
····A····
···B·B···
··C···C··
·D·····D·
E·······E
·D·····D·
··C···C··
···B·B···
····A····
```

Reach out to me if this change should be reverted and a `\n` should be expected at the end.

Closes #2222

### DIFF
--- a/exercises/practice/diamond/diamond_test.go
+++ b/exercises/practice/diamond/diamond_test.go
@@ -1,6 +1,7 @@
 package diamond
 
 import (
+	"fmt"
 	"math/rand"
 	"reflect"
 	"strings"
@@ -221,12 +222,13 @@ func TestCharOutOfRangeShouldGiveError(t *testing.T) {
 
 func TestDiamond(t *testing.T) {
 	for _, testCase := range testCases {
-		got, _ := Gen([]byte(testCase.input)[0])
+		t.Run(testCase.description, func(t *testing.T) {
+			expected := fmt.Sprintf("%s\n", strings.Join(testCase.expected, "\n"))
+			got, _ := Gen([]byte(testCase.input)[0])
 
-		if got != strings.Join(testCase.expected, "\n") {
-			t.Fatalf("Gen(%s): %s\n\t Expected: %s\n\t Got: %s", testCase.input, testCase.description, strings.Join(testCase.expected, "\n"), got)
-		} else {
-			t.Logf("PASS: Gen(%s)", testCase.input)
-		}
+			if got != expected {
+				t.Fatalf("Gen(%q): %q\n\t Expected: %q\n\t Got: %q", testCase.input, testCase.description, expected, got)
+			}
+		})
 	}
 }
--- a/exercises/practice/diamond/diamond_test.go
+++ b/exercises/practice/diamond/diamond_test.go
@@ -28,10 +28,10 @@ func checkCorrect(requirement func(byte, []string) bool, keepSeparator bool, t *
 			separator = strings.SplitAfter
 		}
 		rows := separator(d, "\n")
-		if len(rows) < 2 {
+		if len(rows) < 1 {
 			return false
 		}
-		return requirement(byte(char), rows[:len(rows)-1])
+		return requirement(byte(char), rows)
 	}
 	if err := quick.Check(assertion, config); err != nil {
 		t.Error(err)
@@ -103,7 +103,7 @@ func TestDiamondIsHorizontallySymmetric(t *testing.T) {
 func TestDiamondIsVerticallySymmetric(t *testing.T) {
 	requirement := func(char byte, rows []string) bool {
 		for i, j := 0, len(rows)-1; i < j; i, j = i+1, j-1 {
-			if rows[i] != rows[j] {
+			if strings.TrimSpace(rows[i]) != strings.TrimSpace(rows[j]) {
 				return false
 			}
 		}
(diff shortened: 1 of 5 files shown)


exercises/practice/diamond/diamond_test.go
```diff
@@ -103,7 +103,7 @@ func TestDiamondIsHorizontallySymmetric(t *testing.T) {
 func TestDiamondIsVerticallySymmetric(t *testing.T) {
 	requirement := func(char byte, rows []string) bool {
 		for i, j := 0, len(rows)-1; i < j; i, j = i+1, j-1 {
-			if rows[i] != rows[j] {
+			if strings.TrimSpace(rows[i]) != strings.TrimSpace(rows[j]) {
```
andrerfcsantos, line 106: Why is this needed?

andrerfcsantos (changes requested):
Left some comments.

> Reach out to me if this change should be reverted and a \n should be expected at the end.

Let's revert it from now. The change might have some merits, but it is a big change, since it will make all current solutions fail the tests. Let's discuss it separately. Feel free to create a new issue about this after reverting the change.

exercises/practice/diamond/diamond_test.go
eklatzer: Not needed anymore as we reverted the `\n` at the end

exercises/practice/diamond/diamond_test.go
```diff
@@ -218,3 +219,16 @@ func TestCharOutOfRangeShouldGiveError(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+func TestDiamond(t *testing.T) {
+	for _, testCase := range testCases {
+		t.Run(testCase.description, func(t *testing.T) {
+			expected := fmt.Sprintf("%s\n", strings.Join(testCase.expected, "\n"))
+			got, _ := Gen([]byte(testCase.input)[0])
```
andrerfcsantos, line 227:
```suggestion
			got, err := Gen(testCase.input[0])
```

Indexing a string directly returns a byte already.

We must also test the presence of an error. Even though these test cases expect `Gen` to return `err == nil`, we must explicitly check that the student solution in fact does that.

This requires changing the generator to also include this in the test case struct, and in this test, add another if statement to check the error. We don't need to inspect the error message produced, checking just for the presence/lack of error is what the test expected is enough.

exercises/practice/diamond/diamond_test.go
```diff
@@ -218,3 +219,16 @@ func TestCharOutOfRangeShouldGiveError(t *testing.T) {
...
+
+			if got != expected {
+				t.Fatalf("Gen(%q): %q\n\t Expected: %q\n\t Got: %q", testCase.input, testCase.description, expected, got)
```
andrerfcsantos, line 230:
```suggestion
				t.Fatalf("Gen(%q)\nExpected:\n%s\n(len=%d)\nGot:\n%s\n(len=%d)",
					testCase.input,
					expected, len(expected),
					got, len(got))
```



- Removed the test description that is redundant, as Go will print the subtest names next to each test.
- Did some testing and came up with this string to make the output look nice:

image

`%s` ended up being better here just so the `\n` are properly interpreted. The problem with using `%s` is that makes whitespace differences hard to see, so I also included the length of the string in the output.

exercises/practice/diamond/diamond_test.go
```diff
@@ -218,3 +219,16 @@ func TestCharOutOfRangeShouldGiveError(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+func TestDiamond(t *testing.T) {
```
andrerfcsantos, line 223: Let's move this up in the file, so this is the first suite of tests that runs.

andrerfcsantos (changes requested):
Some more suggestions

andrerfcsantos (approved):
Great!

PR #2253, opened 2022-06-16
Update of the "generating test cases" section of the readme

eklatzer:
Hey, I have adjusted the documentation about generating test cases

Closes #2252

PR #2261, opened 2022-06-19
Updated anchors in readme.md

eklatzer:
Hey, as discussed in #2253 I have updated the anchors


andrerfcsantos (approved):
Thanks! I imagine the old link is still used somewhere, but we'll fix it as we see it.

PR #2265, opened 2022-06-22
Hello World - introduce users to example tests for debugging/testing of solutions

vpayno:
We can't have a `main` package in the same directory as the solution package so we can't build and debug an executable when we're having trouble with a solution. 

Copy pasting one or more Go files into the [Go Playground](https://go.dev/play/) isn't very practical either.

I found example tests are a good compromise for testing, experimenting and debugging solutions. Plus they help one learn both sides of the equation, writing the library and using the library.

### DIFF
--- a/exercises/practice/hello-world/.docs/instructions.md
+++ b/exercises/practice/hello-world/.docs/instructions.md
@@ -13,3 +13,35 @@ The objectives are simple:
 - Submit your solution and check it at the website.
 
 If everything goes well, you will be ready to fetch your first real exercise.
+
+You may find Go's [testable documentation examples](https://go.dev/blog/examples) useful for testing, experimenting an debugging your solution.
+
+Just treat the inside of the example function as a main function and run it with `go test -run Example…` instead of `go run main.go`.
+
+1. Before trying to solve the hello world exercise, create a `hello_world_examples_test.go` with these contents.
+
+```go
+package greeting
+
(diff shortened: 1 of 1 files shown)


junedev:
@vpayno I am sorry but I have to reject this PR for a couple of reasons.
* The "hello world" exercise is deliberately very minimal. It is only to familiarize yourself with the submit flow on Exercism, either via the CLI or via the website.
* When solving the exercise on the website, the existing test that is executed does pretty much the same thing that your example code does. It executes the function, tells you the result and tells you the expected result. Examples in Go are fundamentally not really different from tests, the only difference is that the assertion part is done by comparing the result to the string in the comment which makes them a bit easier to read in documentation.
* When working locally, most IDEs offer a "debug test" feature that allows you do step through any test (and the code under test) with a click of a button. Also in this case, adding an example is not necessary.

I hope you can understand the decision.

In general, I think teaching about examples in Go is a good idea. This is why it is already mentioned on our roadmap (https://exercism.org/docs/tracks/go/roadmap#h-potential-future-concepts). However, currently concepts and learning exercise need to be accompanied with tests and we currently don't have a good system to test whether the student wrote the examples correctly if we were to teach examples as a concept. So this is a bit further out on the roadmap. Maybe we will get the ability to add concepts without exercises, that would also give us the possibility to add examples to the syllabus.

PR #2268, opened 2022-06-23
parallel-letter-frequency: remove redandunt link, add more

### LINKED ISSUE #2257: Remove suggestion of `sync.Map` from Parallel Letter Frequency
In [instructions.append.md](https://github.com/exercism/go/blob/main/exercises/practice/parallel-letter-frequency/.docs/instructions.append.md) for `parallel-letter-frequency`, we have a link for the docs of `sync.Map`.

Some students actually read this and get the idea they should use a `sync.Map`. This is terrible advice for this exercise because:

1. Even the docs discourage the use of this map, and mention that when types for the map are known, a regular and properly typed map + mutex is better:

> The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.
Source: [sync.Map](https://pkg.go.dev/sync#Map)

2. Using a map and protecting it with locks (either by using sync.Map or a regular map + mutexes) is not the best way to solve the exercise. It's best to avoid lock contention altogether by letting each goroutine have its own map and communicate these maps through channels.

My suggestions here are:

- Remove the link to the `sync.Map`
- Add more links about waitgroups, channels and goroutines. Some students find this article very useful and directly applicable to the exercise: [Synchronizing Go Routines with Channels and WaitGroups
](https://dev.to/sophiedebenedetto/synchronizing-go-routines-with-channels-and-waitgroups-3ke2), maybe we can include it too.

Jasstkn:

fixes #2257

### DIFF
--- a/exercises/practice/parallel-letter-frequency/.docs/instructions.append.md
+++ b/exercises/practice/parallel-letter-frequency/.docs/instructions.append.md
@@ -30,7 +30,10 @@ you started. We recommend looking over these before starting this exercise:
 
 - [Concurrency in the Golang Book](https://www.golang-book.com/books/intro/10)
 - [A Tour of Go's concurrency section](https://tour.golang.org/concurrency/1)
-- [Go's sync.Map](https://medium.com/@deckarep/the-new-kid-in-town-gos-sync-map-de24a6bf7c2c)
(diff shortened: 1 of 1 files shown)


andrerfcsantos (approved):
Thanks!

PR #2270, opened 2022-06-25
Fixed "Output" format (#2202) -  constants and conditionals-if

yogeshCt3:
Replaced `Output:` to `=>` in :
**concepts/constants/about.md**

Relates to issue: https://github.com/exercism/go/issues/2202

### DIFF
--- a/concepts/conditionals-if/about.md
+++ b/concepts/conditionals-if/about.md
@@ -53,7 +53,7 @@ if v := 2 * num; v > 10 {
 } else {
     fmt.Println(num)
 }
-// => 14
+// Output: 14
 ```
 
 > Note: any variables created in the initialization statement go out of scope after the end of the if statement.
--- a/concepts/conditionals-if/about.md
+++ b/concepts/conditionals-if/about.md
@@ -53,7 +53,7 @@ if v := 2 * num; v > 10 {
 } else {
     fmt.Println(num)
 }
-// Output: 14
+// => 14
 ```
 
 > Note: any variables created in the initialization statement go out of scope after the end of the if statement.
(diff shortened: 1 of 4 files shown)


concepts/conditionals-if/introduction.md
```diff
@@ -39,7 +39,7 @@ if v := 2 * num; v > 10 {
 } else {
     fmt.Println(num)
 }
-// Output: 14
+// => 14
```
andrerfcsantos, line 42: Since this is an instance where print functions are used, we should keep the `Output:` wording.

concepts/constants/introduction.md
```diff
@@ -25,5 +25,5 @@ const (
     g
 )
 fmt.Print(a, b, c, d, e, f, g)
-// Output: 9 9 9 3 4 5 6
+// => 9 9 9 3 4 5 6
```
andrerfcsantos, line 28: Since this is an instance where print functions are used, we should keep the `Output:` wording.

concepts/conditionals-if/about.md
```diff
@@ -53,7 +53,7 @@ if v := 2 * num; v > 10 {
 } else {
     fmt.Println(num)
 }
-// Output: 14
+// => 14
```
andrerfcsantos, line 56: Since this is an instance where print functions are used, we should keep the `Output:` wording.

andrerfcsantos (changes requested):
Thanks for the interest in doing this!

Left some comments.

yogeshCt3:
updated as per suggestion :-)

andrerfcsantos (approved):
🚀

PR #2275, opened 2022-06-29
Update Roman Numerals with https link/better info

apolo49:
Nova Roma's use of http is a security vulnerability over Imperivm Romanvm's use of https.
Additionally Imperivm Romanvm's wiki page on Roman Numerals offers more information, and a converter than supplies information on why the result is what it is and breaks down the number.


andrerfcsantos:
Since this is a practice exercise, changes to the instructions should be made in the problem specifications repo, specifically to this file [description.md](https://github.com/exercism/problem-specifications/blob/main/exercises/roman-numerals/description.md).

Open a PR/issue there and after the changes are approved there we can bring them to the Go track.

Issue #2055, opened 2022-01-02
Generator ignoring `include=false` of `tests.toml`


andrerfcsantos:
`include=false`  is a directive that can be used in the `tests.toml` file of exercises to tell the tooling that a particular test is not implemented by the track. However, the test generator is ignoring this attribute when generating the tests, when it shouldn't:

> If a track uses a test generator to generate an exercise's test suite, it must use the contents of the tests.toml file to determine which tests to include in the generated test suite.
from [problem-specifications](https://github.com/exercism/problem-specifications#track-test-data)

I ran into this issue when going through the process of updating the tests for `scale-generator` using configlet. This particular exercise had test cases that were replaced using the [reimplements field](https://github.com/exercism/problem-specifications#changing-tests). I told configlet to chose the new version of all the tests, and what configlet did was to update the `tests.toml` with the new tests (by adding them to `tests.toml`), and mark the tests that were reimplemented with `include=false`. However, the generator then proceeded to generate all the tests, regardless of whether they were reimplemented by another test case or not. 

The proposed solution is to follow the recommendation in problem-specifications and make the generator read `tests.toml` and skip generating tests marked with `include=false`.

eklatzer:
Hey, I will try to fix this

eklatzer:
Hey, I came across the following problem. Most of exercise-specific-generators look something like this:
```
func main() {
	t, err := template.New("").Parse(tmpl)
	if err != nil {
		log.Fatal(err)
	}
	var j js
	if err := gen.Gen("luhn", &j, t); err != nil {
		log.Fatal(err)
	}
}

// The JSON structure we expect to be able to unmarshal into
type js struct {
	Cases []struct {
		Description string
		Input       struct {
			Value string
		}
		Expected bool
	}
}
```

All of them define the structure for the `canonical-data.json`. Later when comparing the test-cases to the data from the `tests.toml` every test-case would need the uuid to get the corresponding entry. What is the recommended way of handling this problem to follow guidelines of this project, as there is currently no info about the uuid of each test-case.

eklatzer:
Another question: how should external dependencies be handled? I have created a draft code to play around with the handling of the `.toml` as I have never used one in Go. Therefore I use the external dependency `github.com/pelletier/go-toml/v2`. What is the recommended way of handling this dependency?

```
	testsFilePath := filepath.Join("tests.toml")
	testsFileContent, err := ioutil.ReadFile(testsFilePath)
	if err != nil {
[...]

Issue #2170, opened 2022-04-10
Add new practice exercise: state-of-tic-tac-toe


andrerfcsantos:
## Getting Started

If you have not yet contributed to concept exercises before, this task requires some upfront reading to acquire the necessary background knowledge.

Here you can read about what Concept Exercises are and how they are structured:

- [What are Practice Exercises?](https://exercism.org/docs/building/product/practice-exercises)
- [Anatomy of a Practice Exercise](https://exercism.org/docs/building/tracks/practice-exercises)

Also be aware of these general guidelines.

- [Exercism Markdown Specification](https://exercism.org/docs/building/markdown/markdown)
- [Exercism Formatting and Style Guide](https://exercism.org/docs/building/markdown/style-guide)

## Problem Specifications

The description and test cases for all practice exercises are taken from the [problem specifications](https://github.com/exercism/problem-specifications) repository. 

- [Problem specifications PR](https://github.com/exercism/problem-specifications/pull/1991) - This is the PR in problem specifications that created this exercise. Use it to read more about the decisions taken about the exercise.
- [Problem specifications exercise directory](https://github.com/exercism/problem-specifications/pull/1991) - This folder contains the description and test cases you must have for the exercise.

For more information on how to use the information from problem specifications when building the exercise, check [Anatomy of a Practice Exercise](https://exercism.org/docs/building/tracks/practice-exercises).

## How to proceed

1. **First accept this issue by saying "I'd like to work on this"**. After posting that, feel free to start working on the issue right away. A maintainer will later assign put you as an assignee on the issue, but you don't have to wait for that to work on the issue. 
2. Use this issue to discuss any questions you have, what should be included in the content and shouldn't, and to collect more reference material.
3. Create a PR and set "exercism/go" as reviewers. Additionally, you can write in #maintaining-go on Slack that your PR is ready for review. Once you incorporated any critical feedback that the reviewer might give you and the PR is approved, it will be merged by a maintainer.

andrerfcsantos:
@AbhilashJN Assigned the issue to you, thanks a lot for the interest in collaborating.

> In the canonical data for this problem specification, there are two tests with the same input, but expect different outputs (lines 367 and 382 in canonical-data.json). Should one of the tests be removed?

[...]

PR #2276, opened 2022-06-30
Remove excluded test cases based on tests.toml when generating cases_test.go

AbhilashJN:
#2055 

Added a way to recursively parse the `canonical-data.json` for a problem specification, and remove test cases which are to be excluded as per the `tests.toml` file for that exercise.

### DIFF
--- a/gen/filter.go
+++ b/gen/filter.go
@@ -1,35 +1,28 @@
 package gen
 
-import "encoding/json"
+import (
+	"encoding/json"
+)
 
 func filterTestsJson(jsrc []byte, excludeList map[string]struct{}) ([]byte, error) {
-	var s map[string](json.RawMessage)
-	var filteredData []byte
-	err := json.Unmarshal(jsrc, &s)
-	if err != nil {
-		return filteredData, err
-	}
+	// put the json object in an array to match the recursive structure
+	jsrcArr := make([]byte, 1, len(jsrc)+2)
+	jsrcArr[0] = '['
+	jsrcArr = append(jsrcArr, jsrc...)
+	jsrcArr = append(jsrcArr, ']')
 
 	// recursively remove excluded cases from json source, starting from top level
-	filteredCases, err := recursiveFilterCases(s["cases"], excludeList)
+	filteredCases, err := recursiveFilterCases(json.RawMessage(jsrcArr), excludeList)
 	if err != nil {
-		return filteredData, err
+		return nil, err
 	}
 
-	// convert filteredCases map to json string
-	filteredCasesJson, err := json.Marshal(filteredCases)
+	// remove the json object back out of the array
+	filteredData, err := json.MarshalIndent(filteredCases[0], "", "\t")
 	if err != nil {
-		return filteredData, err
+		return nil, err
 	}
-
-	// typecast it as json.RawMessage to match types
-	s["cases"] = json.RawMessage(filteredCasesJson)
-
-	resp, err := json.Marshal(s)
-	if err != nil {
-		return filteredData, err
-	}
-	return resp, nil
+	return filteredData, nil
 }
 
 func recursiveFilterCases(cases json.RawMessage, excludeList map[string]struct{}) ([]map[string](json.RawMessage), error) {
--- a/gen/filter_test.go
+++ b/gen/filter_test.go
@@ -1,0 +1,204 @@
+package gen
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+)
+
+var (
+	validInputJson = `
+{
+	"exercise": "some-exercise-name",
+	"comments": [
+		"comment 123",
+		"comment 456"
+	],
+	"cases": [
+		{
+		"description": "test case 1",
+		"expected": "abcde",
+		"input": {
+			"some": "inp"
+		},
+		"uuid": "alskjdb-f781-4c52-b73b-d4b867f41540"
+		},
+		{
+		"description": "test case 2",
+		"expected": "rvedv",
+		"input": {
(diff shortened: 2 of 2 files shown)


eklatzer (commented):
Hey, as this is already doing some stuff mentioned in #2055 I will not start on my own implementation, but I will give you feedback about your implementation. 
* I would personally rename `parseTestsToml.go` to `toml.go` and `filterTests.go` to `filter.go`
* I do not think using a `map[string]bool{}` is needed, as we only need something like a set datastructure, which can be done by using `map[string]struct{}{}` (here a discussion: https://stackoverflow.com/questions/34018908/golang-why-dont-we-have-a-set-datastructure)
* instead of parsing to string I would use the following snippet:
```go
for scanner.Scan() {
		line := scanner.Text()
		line = strings.ReplaceAll(line, " ", "")
		if len(line) == 0 || strings.HasPrefix(line, "#") {
			continue
		}

		if strings.HasPrefix(line, "[") {
			currentCaseUUID = strings.Trim(line, "[]")
			continue
		}

		if isExcluded(line) {
			excludedList[currentCaseUUID] = struct{}{}
		}
	}

func isExcluded(line string) bool {
	parts := strings.Split(line, "=")
	key := parts[0]
	value := parts[1]
	if key == "include" {
		return value == "false"
	}
	return false
}
```

gen/filterTests.go
```diff
@@ -0,0 +1,87 @@
...
+	filteredCasesJson, err := json.Marshal(filteredCases)
+	if err != nil {
+		return filteredData, err
+	}
+
+	// typecast it as json.RawMessage to match types
+	x := json.RawMessage(filteredCasesJson)
+
+	// add filtered cases back into top level json
+	s["cases"] = &x
```
eklatzer, line 29: This part is not needed if `var s` is of type `map[string]json.RawMessage`

gen/gen.go
```diff
@@ -124,8 +124,22 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
+	log.Printf("[LOCAL] reading tests.toml file from exercise directory %s\n", tomlFile)
+	excludedTests, err := getExcludedTestCases(tomlFile)
+	if err != nil {
+		return fmt.Errorf("[LOCAL] unable to read tests.toml file : %v", err)
+	}
+
+	// remove the excluded test cases from the source json
+	filteredJSrc, err := filterTestsJson(jSrc, excludedTests)
+	if err != nil {
+		return fmt.Errorf("Unable to filter tests cases : %v", err)
```
eklatzer, line 138: Error messages should not be capitalized: https://stackoverflow.com/questions/68792696/why-error-messages-shouldnt-end-with-a-punctuation-mark-in-go#:~:text=Error%20strings%20should%20not%20be%20capitalized%20(unless%20beginning%20with%20proper,usually%20printed%20following%20other%20context.

eklatzer (commented):
Hey, thanks for the quick response. I think the `toml.go` is pretty clean now. Using `strings.TrimSpace` seems to be better than my suggestion.

gen/filter.go
```diff
@@ -0,0 +1,84 @@
...
+			}
+
+			// convert filteredCases map to json string
+			filteredCasesJson, err := json.Marshal(filteredCases)
+			if err != nil {
+				return validCases, err
+			}
+
+			// typecast it as json.RawMessage to match types
+			j["cases"] = json.RawMessage(filteredCasesJson)
```
eklatzer, line 65: Can we somehow improve this? It is basically a duplicate of lines 14-26.

junedev:
@AbhilashJN @eklatzer Great work on this you two! @eklatzer You cannot officially approve this but please let me know when you are ok with the result. Then I will give it a final check and merge.

Great that there was a test case added. I was just about to suggest that. I will also create an issue for the future about adding tests for other parts of the generator.

eklatzer:
Hey, thanks for the feedback. I don't know if we want to merge this one. I have used the code of this PR in #2279 as these two topics are strongly connected. I hope it is OK that I used the code there, if not...sorry

junedev:
@eklatzer The issue is that if I only merge your PR, @AbhilashJN will not get the credit (e.g. reputation on the website) they deserve for this so that would be unfair. This should be merged and then your PR can iterate on it as you please. I'm sorry for the inconvenience that causes you in resolving merge conflicts etc.

eklatzer:
OK, then I will wait for this PR to be merged. Could you review the code? I will also go through it later

junedev (approved):
I think there are some more minor improvements that could be done in the code here but no need to further block this from being merged. Also it does not do anything to the production code so if we find issues later when using the modified generator, we can always fix them later on. Also further improvements of the code can be made in future PRs.

I tested the full generation process with a simple example and one with sub groups and everything seems to work as expected.

junedev:
@eklatzer Sorry, I only saw your additional comments after I clicked merge. Feel free to simply fix those things as part of your own PR.

eklatzer:
Hey, is it OK if I create a new branch based on the current `main` and then add my changes with better commit messages than the ones I added when refactoring the test case generator? So #2279 could be closed without being merged

junedev:
@eklatzer Whatever you prefer but be aware that we squash commits so the individual commits get lost anyway.

junedev:
@AbhilashJN Thanks for getting this done. You should now be able to continue with #2170 after updating your branch with the latest changes from the upstream main branch.

andrerfcsantos:
Huge thanks @AbhilashJN and @eklatzer!

This was not something trivial and you handled it really well <3

PR #2277, opened 2022-07-01
changed t.Skip to t.Fatalf if input of test is invalid

### LINKED ISSUE #1911: Tests being skipped instead of giving an error
In the `kindergarten-garden` and `matrix` exercises, tests are being skipped (with `t.Skip()`) in situations where they should error, example:

```go
func TestNamesNotModified(t *testing.T) {
	cp := append([]string{}, test6names...)
	_, err := NewGarden(test6.diagram, cp)
	if err != nil {
		t.Skip("TestNamesNotModified requires valid garden")
	}
	if !reflect.DeepEqual(cp, test6names) {
		t.Fatalf("NewGarden modified children argment.  " +
			"Arguments should not be modified.")
	}
	sort.Strings(cp)
	if reflect.DeepEqual(cp, test6names) {
		t.Skip("TestNamesNotModified requires names out of order")
	}
}
```

We should investigate if these tests should really be skipped and possibly convert the calls to `Skip` to calls to `Fail`.

eklatzer:
Hey @junedev, here the PR, I hope the message indicates that the problem is not the student code. Basically this error only occurs in two cases:
* student changes the input of the test
* contributor changes the input and wants to make a PR-->should fail, as configlet should fail as the example-solution does not pass, as this case fails

Closes #1911

### DIFF
--- a/exercises/practice/kindergarten-garden/kindergarten_garden_test.go
+++ b/exercises/practice/kindergarten-garden/kindergarten_garden_test.go
@@ -153,13 +153,12 @@ RVGCCGCV`
 	g1, err1 := NewGarden(diagram, []string{"Alice", "Bob", "Charlie", "Dan"})
 	g2, err2 := NewGarden(diagram, []string{"Bob", "Charlie", "Dan", "Erin"})
 	if err1 != nil || err2 != nil {
-		t.Skip("Two garden test needs valid gardens")
+		t.Fatalf("error in test setup: Two garden test needs valid gardens")
 	}
 	tf := func(g *Garden, n int, child string, expPlants []string) {
 		switch plants, ok := g.Plants(child); {
 		case !ok:
-			t.Skipf("Garden %d lookup %s returned ok = false, want true.",
-				n, child)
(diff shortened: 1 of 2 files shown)


junedev (commented):
The code looks good. However, the description says this closed #1911 but that issue mentioned 2 exercises that need a fix and this PR only changes the code in 1 exercise.

Please either remove the "closes x" statement or add the fix for the other exercise to the PR, whatever you prefer.

eklatzer (commented):
Hey, I have seen that `kinddergarten` had more skips, so I have changed them. For `matrix` I am not completely sure, because these errors can have two causes:
* the needed functions of the user do not work properly
* the input is wrong

junedev (approved):
@eklatzer I think the code is fine as is. At the beginning of the test file, `New` is already tested extensively and the test execution stops if a test fails there. That means the error check for the New calls below is mainly a formality in regards to whether `New` actually works or not. So the most likely case that those fail is indeed the test setup as you have written in the current error message.

PR #2278, opened 2022-07-01
Update about.md in About Slices concept

Inkognitoo:
Add information about merging slices with append

### DIFF
--- a/concepts/slices/about.md
+++ b/concepts/slices/about.md
@@ -21,23 +21,38 @@ withData[1] = 5
 x := withData[1] // x is now 5
 ```
 
-You can create a new slice from an existing slice by getting a range of elements, once again using square-bracket notation, but specifying both a starting (inclusive) and ending (exclusive) index.
+You can create a new slice from an existing slice by getting a range of elements.
+Once again using square-bracket notation, but specifying both a starting (inclusive) and ending (exclusive) index.
 If you don't specify a starting index, it defaults to 0.
 If you don't specify an ending index, it defaults to the length of the slice.
 
 ```go
-newSlice := withData[2:4] // newSlice == []int{2,3}
-newSlice := withData[:2]  // newSlice == []int{0,1}
-newSlice := withData[2:]  // newSlice == []int{2,3,4,5}
-newSlice := withData[:]   // newSlice == []int{0,1,2,3,4,5}
+newSlice := withData[2:4]
+// => []int{2,3}
+newSlice := withData[:2] 
+// => []int{0,1}
+newSlice := withData[2:]
+// => []int{2,3,4,5}
+newSlice := withData[:]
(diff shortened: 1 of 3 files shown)


concepts/slices/about.md
```diff
@@ -32,6 +32,12 @@ newSlice := withData[2:]  // newSlice == []int{2,3,4,5}
 newSlice := withData[:]   // newSlice == []int{0,1,2,3,4,5}
 ```
 
+You can merge two or more slices with the `append` function. For this, you need to pass next slices as argument of variadic function.
+```go
+nextSlice := [int]{100,101,102}
+newSlice  := append(withData, nextSlice...) // newSlice == []int{0,1,2,3,4,5,100,101,102}}
```
andrerfcsantos, line 38:
```suggestion
append(withData, nextSlice...)
// => []int{0,1,2,3,4,5,100,101,102}}
```

Since we wrote this exercise, we have agreed on a new format to present these kind of results. You can read more about this [here](https://github.com/exercism/go/issues/2202). This suggestion makes this example follow the new format.

However, if we make this change here, I think it's best to make the other examples in this exercise be consistent with this new format. If you want, feel free to make this change of format in the other examples too in this concept/exercise. This change to the examples can be another PR if you prefer, in which case feel free to ignore this particular suggestion.

andrerfcsantos (changes requested):
Thanks for this!

I think this is a great change since the exercise directly requires this and you kept it brief.

You made this change in `about.md` which is fine, but this file is only shown when the student completes the exercise/concept, which is late to show this information. My suggestion is to also make this change in the concept's introduction [concepts/slices/introduction.md](https://github.com/exercism/go/blob/main/concepts/slices/introduction.md) and in the exercise's introduction [exercises/concept/card-tricks/.docs/introduction.md](https://github.com/exercism/go/blob/main/exercises/concept/card-tricks/.docs/introduction.md).

The exercise's introduction also includes a simpler example on how to use append with multiple arguments without using the `...` notation. I think we should also include that in this explanation:

```go
a := []int{1, 3}
b := append(a, 4, 2) // b == []int{1,3,4,2}
```

I'll let you choose how you want to incorporate this example in the explanation, but I'm thinking of maybe having a paragraph + code sample explaining this simple example first and then another paragraph + code sample explaining the `...` notation. And to be clear, this extra example should be included in the 3 files, not just in the exercise's introduction.

andrerfcsantos:
@Inkognitoo Let us know if you are still interested in working on this. It's totally fine if you are not, we can finish this for you. If you are still interested in working on this feel free to take your time, there's no rush.

andrerfcsantos (approved):
It has been a while. Since the changes of this PR were small and the suggestions easy to implement, I made the final changes to finish the PR.

Thanks!

Issue #1678, opened 2021-09-22
annalyn-filtration: heading in hints doesn't match task headings in instructions


kahgoh:
I noticed the hints for the _Check if the group can be spied upon_ and _Check if the prisoner can be signalled_ weren't showing on the exercism editor page. 

I think this is because the headings for these tasks aren't the same in the exercise's `hints.md` and `instructions.md`. According to the [tracks/concept-exercises](https://exercism.org/docs/building/tracks/concept-exercises) page, the task hints should be the same as the task heading in `instructions.md`:

> Task-specific hints should appear as a Markdown list underneath headings that match their task heading in the `instructions.md`

Issue #2056, opened 2022-01-02
Implemented tests not matching `tests.toml`


andrerfcsantos:
The [saddle-points](https://github.com/exercism/go/tree/main/exercises/practice/saddle-points) exercise implements [4 tests](https://github.com/exercism/go/blob/main/exercises/practice/saddle-points/saddle_points_test.go), however [tests.toml](https://github.com/exercism/go/blob/main/exercises/practice/saddle-points/.meta/tests.toml) specify a lot more. There are probably more exercises in this situation.

This might throw off configlet, which will read `tests.toml` to know which tests are implemented, when those tests might not be actually implemented.

eklatzer:
Hey, I think this will be done when the discussion in #2279 is finished so you could assign me

PR #2279, opened 2022-07-03
Update of testcase generator to latest JSON format

eklatzer:
Hey, as already discussed in Slack I have created a draft for the new version of the test case generator. I have also included the code covering #2055 (I have used code from #2276 and adjusted it a bit). Summary of what I have done:
* Collection of all test cases from the `canonical-data.json` and group by the value of the field `property`
* Removement of the test cases with `inclue=false` in the `tests.toml`
* Removement of the version info in the header of  `cases_test.go` as the info is not provided in the `canonical-data.json` anymore [(#1678)](https://github.com/exercism/problem-specifications/pull/1678)
* Generator expects now a struct for each of the groups defined by `property` (fields that have multiple data-types for one field have to be handled with an `interface{}` and have then to be handled by each specific generator)
* I have removed `property.json` as I think this is not needed anymore

To check if we have forgotten anything I have refactored the following generators:
* `bowling`: shows how multiple values for `property` can be handled to generate multiple cases with different structures and also how to handle an `interface{}` directly in the generator
* `diamond`: no special purpose, but I tried it as I made some changes to this one some days ago
* `forth`: shows how to handle an `interface{}` in the generator (I have also refactored the `forth_test.go` a bit as I don't think the info about the groups is really needed here and I also made use of `t.Run` and improved the error-messages)
 
**NOTE:** The use of multiple data types for one field mostly happens values and errors are used like here (`expected` is a struct and a number):
```
 {
      "uuid": "29220245-ac8d-463d-bc19-98a94cfada8a",
      "description": "an unstarted game cannot be scored",
      "property": "score",
      "input": {
        "previousRolls": []
      },
      "expected": {
        "error": "Score cannot be taken until the end of the game"
      }
    },
    {
      "uuid": "656ae006-25c2-438c-a549-f338e7ec7441",
      "description": "should be able to score a game with all zeros",
      "property": "score",
      "input": {
        "previousRolls": [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
      },
      "expected": 0
    },
```


eklatzer:
Hey @junedev and @andrerfcsantos, could you please have a look at the implementation and the changes on the test case generator and give me some feedback

eklatzer:
Hey regarding #2056: I have done the following things:
- updated `tests.toml`
- created a new test-case generator
- implemented the tests of the generator

Here is what I faced: when using the code from the `example.go` to run the tests, all tests of the `problem-specifications` failed and all with the same problem. Here an example:
Input:
```
4 5 4
3 5 5
1 5 4
```
Expected output in `problem-specifications`:
```
 "expected": [
        {
          "row": 1,
          "column": 2
        },
        {
          "row": 2,
          "column": 2
        },
        {
          "row": 3,
          "column": 2
        }
      ]

-->[]Pair{{1, 2}, {2, 2}, {3, 2}
```

As this test was also implemented in the existing test-cases here the expected value there:
```
[]Pair{{0, 1}, {1, 1}, {2, 1}
```

As you can see each value is incremented by 1, compared to the `problem-specifications` (index instead of position). This is also the case for all other test cases. Possible solutions:
- we adjust the example solution to match the expected `problem-specifications`
- we increment each value of the expected result by 1, by using the following template:
```
{{printf "%d+1" .Row}},
{{printf "%d+1" .Column}},
```

Please tell me what you expect here

andrerfcsantos (changes requested):
Thank you so much for this! ❤️

Left some suggestions.

Other than the inline suggestions, two general comments about the PR:

- Right now CI is failing with this error:

```
tmp/lint/practice/saddle_points/cases_test.go:14: File is not `gofmt`-ed with `-s` (gofmt)
		[][]int{[]int{9, 8, 7}, []int{5, 3, 2}, []int{6, 6, 7}},
```

To fix this, in the root of the repository run `gofmt -s -w ./..`.

- The `Gen` function is getting quite big and it's doing a lot. It might be worth separating it into several smaller functions that are more testable. However, I'm fine with this being a separate PR - that shouldn't block this one. Let me know what you think, and if you agree in making this a separate PR, I'll create the issue for it.

exercises/practice/bowling/cases_test.go
```diff
@@ -95,6 +94,13 @@ var scoreTestCases = []struct {
 		30,
 		"",
 	},
+	{
+		"last two strikes followed by only last bonus with non strike points",
+		[]int{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 10, 10, 0, 1},
+		true,
+		31,
+		"",
```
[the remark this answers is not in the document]
eklatzer: Added it for all the generators I have already updated to the new generator

gen/gen.go
```diff
@@ -98,167 +72,122 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
+		}
+		header, err = getRemoteCommit(exercise)
 		if err != nil {
 			return err
 		}
 	}
 
+	if !isJSON(jTestData) {
+		return fmt.Errorf("[ERROR] canonical-data.json seems not to be valid json")
+	}
```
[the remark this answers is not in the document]
eklatzer: Thanks for the suggestion, didn't know about that. I am now using `json.Valid`

eklatzer:
You are right, we should split `Gen` a bit up and add some tests. Please open a new issue for that

exercises/practice/bowling/.meta/gen.go
```diff
@@ -1,81 +1,62 @@
 package main
 
 import (
+	"../../../../gen"
```
junedev, line 4: The import order was correct in the old version.

In Go, it is idiomatic to have 3 groups of import statements separated by an empty line.
1. standard library
2. external dependencies
3. local packages

Some people switch or combine 2 and 3 but in general stdlib imports always come first and are a group of its own.
If your IDE gets this wrong you might need to adjust the settings. Some tools like goimports have trouble getting the groups right automatically currently (there are open issues about this) but they will usually honor the groups if you set them up manually.

junedev (changes requested):
Just leaving a few comments now, I did not finish the review yet.

exercises/practice/bowling/.meta/gen.go
eklatzer: I have adjusted the settings, I think it is correct now

gen/gen.go
```diff
@@ -28,68 +33,37 @@ import (
...
+	Input       interface{} `json:"input"`
+	Expected    interface{} `json:"expected"`
 }
 
 // Gen generates the exercise cases_test.go file from the relevant canonical-data.json
-func Gen(exercise string, j interface{}, t *template.Template) error {
-	if dirMetadata == "" {
-		return errors.New("unable to determine current path")
-	}
+func Gen(exercise string, tests map[string]interface{}, t *template.Template) error {
```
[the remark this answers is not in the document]
eklatzer: The goal is the updated all `gen.go` files as many are completely outdated. I think if anyone really wants to update the test cases also updating the generator is possible. I want to tackle this task after finishing this PR. The already changed generators were only to test the new structure and to find weaknesses and try out the handling of interfaces in the generator

eklatzer:
> Hey regarding #2056: I have done the following things:
> 
> * updated `tests.toml`
> * created a new test-case generator
> * implemented the tests of the generator
> 
> Here is what I faced: when using the code from the `example.go` to run the tests, all tests of the `problem-specifications` failed and all with the same problem. Here an example: Input:
> 
> ```
> 4 5 4
> 3 5 5
> 1 5 4
> ```
> 
> Expected output in `problem-specifications`:
> 
> ```
>  "expected": [
>         {
>           "row": 1,
>           "column": 2
>         },
>         {
>           "row": 2,
>           "column": 2
>         },
>         {
>           "row": 3,
>           "column": 2
>         }
>       ]
> 
> -->[]Pair{{1, 2}, {2, 2}, {3, 2}
> ```
> 
> As this test was also implemented in the existing test-cases here the expected value there:
> 
> ```
> []Pair{{0, 1}, {1, 1}, {2, 1}
> ```
> 
> As you can see each value is incremented by 1, compared to the `problem-specifications` (index instead of position). This is also the case for all other test cases. Possible solutions:
> 
> * we adjust the example solution to match the expected `problem-specifications`
> * we increment each value of the expected result by 1, by using the following template:
> 
> ```
> {{printf "%d+1" .Row}},
> {{printf "%d+1" .Column}},
> ```
> 
> Please tell me what you expect here

What about this question? This is the reason why CI is currently failing

andrerfcsantos:
> we adjust the example solution to match the expected problem-specifications

I like this better, because it makes the student's solution match the position numbers provided in the example and in the tests. 

It seems that this exercise originally had the table in the example using 0-based indexes, and when the exercise was introduced in the track, both description and tests were written with that assumption. However, with https://github.com/exercism/problem-specifications/pull/1429 the description and canonical data were updated to use 1-based indexes in problem specs. Then with the change to v3, the exercises' descriptions for the Go track were updated to use the latest version by https://github.com/exercism/go/pull/1410, but the tests were not updated.

eklatzer:
Hey, is it OK to handle this add the end by adding 1 to the index-based values, or do we want to handle this somewhere else?

junedev:
> Hey, is it OK to handle this add the end by adding 1 to the index-based values, or do we want to handle this somewhere else?
> 
> I removed the old tests, as the ones from `problem-specifications` have more test cases and cover similar problems.

@eklatzer I'm not sure I understand the question. If you are referring to how to handle this in the example solution then you are free to handle this however you like. The example solution does not have to be a very good solution, it only needs to be a solution that passes the tests for CI purposes.

exercises/practice/forth/.meta/gen.go
```diff
@@ -12,84 +12,72 @@ func main() {
 	if err != nil {
 		log.Fatal(err)
 	}
-	var j js
-	if err := gen.Gen("forth", &j, t); err != nil {
+	var j = map[string]interface{}{
+		"evaluate": &[]testCase{},
+		//TODO: add test with property `evaluateBoth` to forth_test.go and generate cases in cases_test.go
```
junedev, line 17: Should this todo be part of this PR?

exercises/practice/forth/forth_test.go
```diff
@@ -6,23 +6,20 @@ import (
...
 func TestForth(t *testing.T) {
-	for _, tg := range testGroups {
-		for _, tc := range tg.tests {
-			if v, err := Forth(tc.input); err == nil {
-				var _ error = err
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			v, err := Forth(tc.input)
+			if err == nil {
+				var _ = err
```
junedev, line 13: What is this line (13) for?

exercises/practice/saddle-points/.meta/gen.go
```diff
@@ -0,0 +1,61 @@
+package main
+
+import (
+	"../../../../gen"
```
junedev, line 4: Import order needs to be fixed here as well.

junedev (commented):
Finished the review. Left some more minor comments/questions.

Two more general points (no need to fix everywhere here):
* When adding context for an error via `fmt.Errorf` or similar, the `(%v)` formatting you used is not idiomatic to Go. The old version with the colon is preferred. See for example https://go.dev/blog/go1.13-errors (section "Wrapping errors ..."). If you wonder why this is better, think about what happens when something is wrapped multiple times as it bubbles up the stack.
* I can recommend to make error wrapping a habit. At work, I barely ever return an error as is, I usually always wrap before returning it (see https://docs.gitlab.com/ee/development/go_guide/#adding-context for how to do it in way that you get reasonable output). Remember that by default, there are no stack traces attached to the errors so context is very important.

junedev (approved):

exercises/practice/bowling/.meta/gen.go
```diff
@@ -12,70 +12,52 @@ func main() {
...
-	Cases    []OneCase
-}
-
-// template applied to above data structure generates the Go test cases
-
-type OneCase struct {
-	Description string
-	Property    string
+type Case struct {
+	UUID        string `json:"uuid"`
```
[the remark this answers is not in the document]
junedev: (Same for other exercises.)

junedev:
Two more things I forgot to mention:
* THANKS! :slightly_smiling_face: 
* Please create an issue about updating/extending the README regarding generators. With this new system there are some things to be aware of when writing a generator that should be pointed out (e.g. that cases are flattened, that the key is the property).

gen/gen.go
```diff
@@ -98,167 +73,116 @@ func Gen(exercise string, j interface{}, t *template.Template) error {
...
-	if err := classifyByProperty(j); err != nil {
-		return fmt.Errorf("couldn't auto-classify based on property: %v", err)
+	for property, testCases := range casesPerProperty {
+		log.Printf(" > parsing cases for property %s\n", property)
+		marshal, err := json.Marshal(testCases)
+		if err != nil {
+			return fmt.Errorf("[ERROR] failed to marshal test cases with property %s\n%q", property, err)
+		}
+
+		cache, ok := tests[property]
```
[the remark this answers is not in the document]
eklatzer: I have changed the name and added a comment, but I am not sure if it is better now. What is your opinion?

exercises/practice/bowling/.meta/gen.go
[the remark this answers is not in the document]
eklatzer: I have removed all fields that are not needed

exercises/practice/forth/.meta/gen.go
eklatzer: Was not planned to be part of this PR. How to handle the following test case? `"scenarios": ["local-scope"],`:
```
 {
          "uuid": "3c8bfef3-edbb-49c1-9993-21d4030043cb",
          "description": "only defines locally",
          "comments": [
            "This test is meant to determine if instructions that define words are executed in the state of the program or locally.",
            "For example, in OOP-based approaches, the words should be defined in the same scope,",
            "or local to the instance the instructions were evaluated in."
          ],
          "scenarios": ["local-scope"],
          "property": "evaluateBoth",
          "input": {
            "instructionsFirst": [": + - ;", "1 1 +"],
            "instructionsSecond": ["1 1 +"]
          },
          "expected": [[0], [2]]
        }
```

exercises/practice/forth/forth_test.go
eklatzer: To be honest I don't know. I think I left it there because it existed before I touched the file. I removed it know, as it seems not to be needed

eklatzer:
> Two more things I forgot to mention:
> 
> * THANKS! 🙂
> * Please create an issue about updating/extending the README regarding generators. With this new system there are some things to be aware of when writing a generator that should be pointed out (e.g. that cases are flattened, that the key is the property).

Thanks for the feedback🙂

eklatzer:
> * The `Gen` function is getting quite big and it's doing a lot. It might be worth separating it into several smaller functions that are more testable. However, I'm fine with this being a separate PR - that shouldn't block this one. Let me know what you think, and if you agree in making this a separate PR, I'll create the issue for it.

> Ok, then please create a new issue about updating the exercise generator files that describes the 2 reasons there are issues (the one with the changes signature and the other one with the new canonical data format you noticed earlier).

We should maybe wait with updating the exercise generators until we have split up the `func Gen`.

Here a script to get all exercises with a `gen.go` formatted as task (has to be started at root of project):
```
# ! /bin/bash

find ./exercises/ -name 'gen.go'|while read fname; do
  exercise=$(echo "$fname" | cut -d/ -f4)
  echo "- [ ] $exercise"
done
```

andrerfcsantos (approved):

exercises/practice/forth/.meta/gen.go
junedev: Not exactly sure what that test is supposed to be about. Maybe it would be good to ask in the Slack #discussions about this.
This is not blocking the merge though.

PR #2283, opened 2022-07-04
added new test-cases to exercise chessboard

### LINKED ISSUE #2282: Missing test case for exercise chessboard
Hey, when mentoring `chessboard` I have seen the following submission:
```
func CountInFile(cb Chessboard, file int) int {
    var count int
	if file > 8 {
        return 0
    }
    for _, v := range cb {
        if v[file - 1] {
            count ++
        }
    }
	return count
}
```

The instructions: Return a count of zero (0) if the given file is not a valid one (not between 1 and 8, inclusive).

To really cover this, we should add further test cases. At least a test case with 0 and one with a negative value for `file`

eklatzer:
Closes #2282

### DIFF
--- a/exercises/concept/chessboard/chessboard_test.go
+++ b/exercises/concept/chessboard/chessboard_test.go
@@ -70,7 +70,11 @@ func TestCountInFile(t *testing.T) {
 		{6, 2},
 		{7, 0},
 		{8, 3},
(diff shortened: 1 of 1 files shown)


andrerfcsantos (approved):
🚀

PR #2285, opened 2022-07-06
Add extra test cases for IncrementVoteCount in election_day

### LINKED ISSUE #2281: Missing test-cases for election-day
Hey, when mentoring `election-day` I have seen the following code:
```
// IncrementVoteCount increments the value in a vote counter
func IncrementVoteCount(counter *int, increment int) {
	*counter += 2
}
```
It only passes as there is only the following test case:
```
	{
			name:      "Call to IncrementVoteCount with a pointer to an int with a value of 2 and increment of 2",
			counter:   &twoVotes,
			increment: 2,
			expected:  4,
		},
```

We should add further test cases here

DeathVenom54:
Fixes #2281 
This is my contribution to address the issue.
Added two additional test cases to test the given solution more thoroughly.

### DIFF
--- a/exercises/concept/election-day/election_day_test.go
+++ b/exercises/concept/election-day/election_day_test.go
@@ -60,6 +60,8 @@ func TestVoteCount(t *testing.T) {
 
 func TestIncrementVoteCount(t *testing.T) {
 	twoVotes := 2
+	fiveVotes := 5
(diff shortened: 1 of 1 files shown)


andrerfcsantos (approved):
Thanks for this!

PR #2286, opened 2022-07-06
Tweak weather forecast story for flow

kytrinyx:
This change is purely cosmetic, and there is no open issue requesting it.
It's not strictly necessary.

The weather forecast story had a few turns of phrase that felt a little
bit clunky.

I've tweaked the wording to flow a little bit better.

I also took the opportunity to replace a few exclamation marks with
periods to tone down the enthusiasm a wee bit.


andrerfcsantos (approved):
Thanks! Changes like this one are greatly appreciated.

PR #2287, opened 2022-07-06
Add punctuation to function comments in concept instructions

kytrinyx:
This does not fix an open issue, and is not strictly necessary.

I noticed as I was going through the introductory Hello World exercise that the doc comments in the instructions didn't follow the best practices that we're trying to encourage people to follow. This adds punctuation to a couple of example function comments.


andrerfcsantos (approved):
Nice catch! In the Weather Forecast exercise we teach and enforce this format of comments, so it's good we follow the practices we are teaching and act as an example :)

PR #2290, opened 2022-07-08
added struct literal example as a counterpart to `new` buitin

hellt:
Hi all,
since this struct intro aritcle contains "`new` builtin" section that reasons that `new` is not a great way to instantiate a struct, maybe it makes sense to add a struct literal alternative to it?

### DIFF
--- a/concepts/structs/about.md
+++ b/concepts/structs/about.md
@@ -77,6 +77,16 @@ s := Shape{
 // Output: too few values in Shape{...}
 ```
 
+To create an instance of a `struct` and get a pointer to it use a struct literal:
+
+```go
+s := &Shape {
+    name: "Square",
+    size: 25,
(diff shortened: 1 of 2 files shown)


andrerfcsantos:
I'm on the fence about this.

The inclusion of `new` was just to address the potential confusion between the `New` functions some packages have, including this exercise. The goal was to explain briefly the difference and what `new` does, and why it's not what you want. But we don't want to enter into more detail on pointers because:

- Pointers weren't properly introduced in the track at this point.
- They are not needed for the exercise. This exercise doesn't require creating pointers.
- There's already the [pointers concept](https://exercism.org/tracks/go/concepts/pointers) that goes into more detail about this, and it includes an example of how to create a pointer using the struct literal syntax.

PR #2297, opened 2022-07-10
parsing-log-files: added test case

eklatzer:
Hey, when doing the exercise `parsing-log-files` I have seen that the following solution passes all tests:
```
var validLinePrefixes = map[string]struct{}{"[TRC]": {}, "[DBG]": {}, "[INF]": {}, "[WRN]": {}, "[ERR]": {}, "[FTL]": {}}

func IsValidLine(text string) bool {
	_, validPrefix := validLinePrefixes[text[:5]]
	return validPrefix
}
```
Therefore I have added a test case that is covering a line that is shorter than 5 characters

### DIFF
--- a/exercises/concept/parsing-log-files/.meta/config.json
+++ b/exercises/concept/parsing-log-files/.meta/config.json
@@ -3,6 +3,7 @@
      "norbs57" 
   ],
   "contributors": [
+    "eklatzer"
   ],
   "files": {
     "solution": [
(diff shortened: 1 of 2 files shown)


andrerfcsantos (changes requested):
Additionally, let's add another test case for the empty string, just to make really sure that length is not a factor considered in solutions.

andrerfcsantos (approved):

PR #2303, opened 2022-07-13
census: added test case

eklatzer:
Hey, when mentoring I got the following solution:
```
// HasRequiredInfo determines if a given resident has all of the required information.
func (r *Resident) HasRequiredInfo() bool {
	if r.Name == "" {
        return false
    }
    for _, v := range r.Address{
        if v != "" {
            return true
        }
    }
	return false
}
```
Instructions: In order to be counted, a resident must provide a non-zero value for their name and an address that contains a non-zero value for the `street` key.
It only checks if any key in `r.Address` is set, but not if `street` is set and this passes the tests.

Therefore I have added a test case

### DIFF
--- a/exercises/concept/census/.meta/config.json
+++ b/exercises/concept/census/.meta/config.json
@@ -5,7 +5,8 @@
     "sudomateo"
   ],
   "contributors": [
-    "tehsphinx"
+    "tehsphinx",
+    "eklatzer"
   ],
   "files": {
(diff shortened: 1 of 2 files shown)


burakkarasel:
Great Job!

junedev (approved):

junedev:
Interesting case, thanks for the additional test case.

Issue #2302, opened 2022-07-12
Update of exercise generators


eklatzer:
Hey, as the test case generator was updated and refactored in #2279 we also need to update the generators of the exercises that have one. The signature of `func Gen` changed. It now accepts a `map[string]interface{}` instead of `interface{}`. Additionally, it supports recursively searching the `canonical-data.json` for test cases. All cases are flattened to a list and grouped by the value of `property` of each test case.  The exercise generator has to provide a pointer to a slice of test cases for each property. Example: the `canonical-data.json` contains test cases with the properties `roll` and `score`. Therefore two values have to be provided. The struct used for the properties can (of course) be different if the structure of the `input` or `expected` differ.

Example:
```
var j = map[string]interface{}{
		"roll":  &[]Case{},
		"score": &[]Case{},
	}
	if err := gen.Gen("bowling", j, t); err != nil {
		log.Fatal(err)
	}
```
In the template for the `cases_test.go` the key `.score` or `.roll` can be used:
```
var scoreTestCases = []struct {
	description    string
	previousRolls  []int	// bowling rolls to do before the Score() test
	valid          bool     // true => no error, false => error expected
	score          int	// when .valid == true, the expected score value
	explainText    string   // when .valid == false, error explanation text
}{ {{range .J.score}}
{
	description: {{printf "%q"  .Description}},
	previousRolls: {{printf "%#v" .Input.PreviousRolls}},
	valid: {{printf "%v"  .Valid}},
	score: {{printf "%d"  .Score}},
	explainText: {{printf "%q"  .ExplainText}},
},{{end}}
}
```

Therefore all existing generators have to be updated. The steps to update are:
1. Create the structs for each `property` and set `j` to `map[string]interface{}` with the needed values for each property
2. Update the template and use the previously created keys
**NOTE:** in some cases the type for `input`/`expected` are not the same for all test cases for one `property`. Therefore an `interface{}` has to be used and then it has to be handled in the exercise generator (mostly when a value and an error are provided for the same key in `canonical-data.json`). Examples can be found in the exercises `forth` or `bowling`.

eklatzer:
Currently the following exercises need an updated:
- [ ] acronym
- [ ] all-your-base
- [ ] allergies
- [ ] alphametics
- [ ] anagram
- [ ] armstrong-numbers
- [ ] atbash-cipher
- [ ] binary-search
- [ ] bob
- [ ] book-store
- [x] bowling
- [ ] change
- [ ] clock
[...]

PR #2309, opened 2022-07-13
allergies: updated exercise generator to new test generator

eklatzer:
Hey, here I also adjusted the tests a bit as the inputs seem to have changed

One of #2302

### DIFF
--- a/exercises/practice/allergies/allergies_test.go
+++ b/exercises/practice/allergies/allergies_test.go
@@ -9,9 +9,7 @@ func TestAllergies(t *testing.T) {
 	for _, test := range listTests {
 		t.Run(test.description, func(t *testing.T) {
 			if actual := Allergies(test.score); !sameSliceElements(actual, test.expected) {
-				t.Fatalf("FAIL: Allergies(%d): expected %#v, actual %#v", test.score, test.expected, actual)
-			} else {
-				t.Logf("PASS: Allergic to %#v", test.expected)
+				t.Fatalf("Allergies(%d) = %#v, want: %#v", test.score, actual, test.expected)
 			}
 		})
 	}
@@ -22,7 +20,6 @@ func BenchmarkAllergies(b *testing.B) {
 		b.Skip("skipping benchmark in short mode.")
 	}
 	for i := 0; i < b.N; i++ {
-
 		for _, test := range allergicToTests {
 			Allergies(test.input.score)
 		}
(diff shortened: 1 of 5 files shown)


exercises/practice/allergies/allergies_test.go
```diff
@@ -7,11 +7,13 @@ import (
...
-		if actual := Allergies(test.score); !sameSliceElements(actual, test.expected) {
-			t.Fatalf("FAIL: Allergies(%d): expected %#v, actual %#v", test.score, test.expected, actual)
-		} else {
-			t.Logf("PASS: Allergic to %#v", test.expected)
-		}
+		t.Run(test.description, func(t *testing.T) {
+			if actual := Allergies(test.score); !sameSliceElements(actual, test.expected) {
+				t.Fatalf("FAIL: Allergies(%d): expected %#v, actual %#v", test.score, test.expected, actual)
+			} else {
+				t.Logf("PASS: Allergic to %#v", test.expected)
```
andrerfcsantos, line 14: With subtests, a string logging a `PASS` or `FAIL` is redundant and can be removed as running `go test -v` will make Go print explicitly the result for every test.

exercises/practice/allergies/allergies_test.go
```diff
@@ -7,11 +7,13 @@ import (
...
 func TestAllergies(t *testing.T) {
 	for _, test := range listTests {
-		if actual := Allergies(test.score); !sameSliceElements(actual, test.expected) {
-			t.Fatalf("FAIL: Allergies(%d): expected %#v, actual %#v", test.score, test.expected, actual)
-		} else {
-			t.Logf("PASS: Allergic to %#v", test.expected)
-		}
+		t.Run(test.description, func(t *testing.T) {
+			if actual := Allergies(test.score); !sameSliceElements(actual, test.expected) {
+				t.Fatalf("FAIL: Allergies(%d): expected %#v, actual %#v", test.score, test.expected, actual)
```
andrerfcsantos, line 12:
```suggestion
				t.Fatalf("Allergies(%d)=%#v, expected %#v", test.score, actual, test.expected)
```

Removed the `FAIL` string and made the print results be more in line with what we are doing in other exercises.

andrerfcsantos (changes requested):
Made 2 comments that might apply to other exercises.

exercises/practice/allergies/allergies_test.go
eklatzer: Should be removed nearly everywhere, might have forgotten/not seen sometimes

andrerfcsantos (approved):

exercises/practice/allergies/allergies_test.go
andrerfcsantos: I'll make sure to double check that, no worries :)

PR #2311, opened 2022-07-13
anagram: updated exercise generator to new test generator

eklatzer:
One of #2302 

Here I am not completely sure which of the test cases with `reimplements` to include. Especially for the `words are not anagrams of themselves` thing.

### DIFF
--- a/exercises/practice/anagram/.docs/instructions.md
+++ b/exercises/practice/anagram/.docs/instructions.md
@@ -1,8 +1,13 @@
 # Instructions
 
-An anagram is a rearrangement of letters to form a new word.
-Given a word and a list of candidates, select the sublist of anagrams of the given word.
+An anagram is a rearrangement of letters to form a new word: for example `"owns"` is an anagram of `"snow"`.
+A word is not its own anagram: for example, `"stop"` is not an anagram of `"stop"`.
 
-Given `"listen"` and a list of candidates like `"enlists" "google"
-"inlets" "banana"` the program should return a list containing
-`"inlets"`.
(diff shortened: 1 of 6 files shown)


andrerfcsantos:
> Here I am not completely sure which of the test cases with reimplements to include. Especially for the words are not anagrams of themselves thing.

I think we can include all of them. By the looks of it, seems that `85757361-4535-45fd-ac0e-3810d40debc1` was then divided into 3 subtests. I'm a bit confused why `85757361-4535-45fd-ac0e-3810d40debc1` doesn't have an `include=false`. Did you update the toml using configlet and did it not put it in there?

According to the [instructions here](https://github.com/exercism/configlet#tests), I would expect `configlet sync --tests include -u -e anagram` to do this?

eklatzer:
Hey, when running `configlet sync --tests include -u -e anagram` with an empty `tests.toml` all cases are included because of `--tests include` which includes all unseen. I have now removed all `include=false` statements as requested.

andrerfcsantos (changes requested):
Since tests `85757361-4535-45fd-ac0e-3810d40debc1` and `a0705568-628c-4b55-9798-82e4acde51ca` are being reimplemented, I think they should have `include=false`. Looking at the reimplementations they seem to be more specific and more targeted at testing the specific thing they claim to test, so it makes sense to only have the reimplementations.

eklatzer:
Sorry, I misunderstood you. I thought you wanted to have all test cases

andrerfcsantos (approved):

Issue #2182, opened 2022-04-13
Implement New Concept Exercise: Channels and Select


andrerfcsantos:
## Getting Started

If you have not yet contributed to concept exercises before, this task requires some upfront reading to acquire the necessary background knowledge.

Here you can read about what Concept Exercises are and how they are structured:

- [What are Concept Exercises?](https://exercism.org/docs/building/product/concept-exercises)
- [Concept Exercises](https://exercism.org/docs/building/tracks/concept-exercises)
- [Concepts](https://exercism.org/docs/building/tracks/concepts)

Also, be aware of these general guidelines:

- [Exercism Markdown Specification](https://exercism.org/docs/building/markdown/markdown)
- [Exercism Formatting and Style Guide](https://exercism.org/docs/building/markdown/style-guide)

## Goal

The goal here is to create a new concept exercise that teaches about channels and `select`.
That includes writing a concept and creating a new exercise.

## Concepts

The following concept needs to be created:

- `channels`

You can use the `introduction.md` file of the concept also as `introduction.md` file of the exercise. No need to create different content at this point. Additionally, if you want to save some time it is ok to not have an extensive `about.md` at this point. It can also be mainly the `instructions.md` content, maybe with some additions you would like to make.

## Learning Objectives

The student should learn about the following topics and then practice them in the concept exercise:

- understand what channels are
- the difference between buffered and unbuffered channels
- closing channels
- for loop over a channel: `for x := range ch { ... }`
- use `select` to receive from several channels

## Out-of-scope

- using channels as a way to synchronize goroutines
- having the student create goroutines to work with channels

## Prerequisites

- `functions`
- `for-loops`
- `range-iteration`

These prerequisites should be listed in the `config.json` entry for the exercise.
The prerequisites above are a baseline of the prerequisites that might be needed for the exercise.
When implementing the exercise, feel free to add more prerequisites if the exercise needs them.

## Story

Try to think of a story for the exercise that ties all the tasks together.
The story doesn't need to be highly complex.
Even a small story goes a long way to improve the enjoyment of the exercise.

## Resources

Some links that might be helpful as a starting point and/or for the links section of the concept:

- [Tour of Go: Channels](https://go.dev/tour/concurrency/2)
- [Tour of Go: Buffered Channels](https://go.dev/tour/concurrency/3)
- [Tour of Go: Range and Close](https://go.dev/tour/concurrency/4)
- [Tour of Go: Select](https://go.dev/tour/concurrency/5)

## How to proceed

[...]

PR #2324, opened 2022-07-14
Add channels concept

RezaSi:
for this https://github.com/exercism/go/issues/2182


junedev (changes requested):
The content you present here seems to be copied from https://www.golangbootcamp.com/book/concurrency. The book does not have any license file so we cannot know if the author allows copying without attribution. https://github.com/GoBootcamp/book/issues/48 This is too risky for us.

The content needs to written in your own words, copied from places that explicitly allow copying without attribution (e.g. via an MIT license) or the sentences that are copied need to be quoted and the source needs to be added as a footnote (this is legally only allowed for short sections if not explicitly permitted otherwise).

I only looked at the first part so far. I don't know for how much of the total content this applies to. It's easier for you to know. In any case, please rework all the copied content so it follows the rules above. If you have any questions, let me know.

junedev:
Did you invent the exercise from scratch? If yes, I could start reviewing that part.

RezaSi:
> The content you present here seems to be copied from https://www.golangbootcamp.com/book/concurrency. The book does not have any license file so we cannot know if the author allows copying without attribution. [GoBootcamp/book#48](https://github.com/GoBootcamp/book/issues/48) This is too risky for us.
> 
> The content needs to written in your own words, copied from places that explicitly allow copying without attribution (e.g. via an MIT license) or the sentences that are copied need to be quoted and the source needs to be added as a footnote (this is legally only allowed for short sections if not explicitly permitted otherwise).
> 
> I only looked at the first part so far. I don't know for how much of the total content this applies to. It's easier for you to know. In any case, please rework all the copied content so it follows the rules above. If you have any questions, let me know.

Yes, some part of the intro and about is copied from this site and golang tut, I don't know this, but it's ok, I will try to write it again with my words!

RezaSi:
> Did you invent the exercise from scratch? If yes, I could start reviewing that part.

Yes, it's not copied from somewhere or another exercise, tnx for the review.

junedev (changes requested):
I had a look at the about.md file. I will check out the exercise later on.

RezaSi:
@junedev Tnx for all your advice, I think resolved all, and repair some other parts.
Yes, I'm not a native speaker, if you can help with grammar, it's very welcome.

RezaSi:
@junedev Tnx for all your advice, I think resolved all, and repair some other parts.
Yes, I'm not a native speaker, if you can help with grammar, it's very welcome.

RezaSi:
@junedev it's blocked? :-?

junedev:
@RezaSi Sorry for the wait. I had to discuss with the Exercism leadership how to proceed with this PR. The problem is that the concept content still contains some sentences that are copied without attribution and some others were there were only slight changes made to the original content. This happened  after I mentioned the issue twice already. It is too risky for us to merge content like this and it takes too much time for me to check and discuss every sentence. I will still consider the exercise part (without introduction.md) for merging but before I merge, I will remove the concept content from the PR and set the exercise status to "work in progress". Then there will be a follow up issue about writing the concept from scratch and once someone else worked on this, the exercise can go live.

Hopefully, I can find some time to review the exercise on the weekend but I can't promise. I'm sorry this all takes so long but keep in mind that all track maintainers on Exercism are volunteers that do this work in their free time.

exercises/concept/cache/.meta/example.go
```diff
@@ -0,0 +1,36 @@
...
+			ch <- err
+		}(cache)
+	}
+
+	for range cacheLayers {
+		<-ch
+	}
+
+	return nil
+}
```
junedev, line 36: A couple of thoughts about this task (and the solution):
* Currently, the function always returns nil so it could also just not have a return value at all.
* The instructions should clarify what happens in case Set returns an error.
* Overall, the required solution here is very similar to the one of the first task. Usually we try to bring up some new aspect in each task. Have another look at the learning objectives, maybe the task can be adjusted/re-written to cover something different. For example, currently your tasks don't cover ranging over a channel at all.

exercises/concept/cache/cache.go
```diff
@@ -0,0 +1,9 @@
+package cache
+
+func GetValueFromFastestLayer(cacheLayers []Cache, key string) (value string, err error) {
```
junedev, line 3: Please remove the names from the return values. Those should only be used when really needed as they can lead to mistakes easily.

exercises/concept/cache/cache_struct.go
```diff
@@ -0,0 +1,31 @@
+package cache
```
junedev, line 1: If you follow the suggestion mentioned above with the interface, the cache implementation can be part of the standard test file. If you want to keep it as a separate file, you can list it in the config under the "editor" key so it shows up when solving the exercise but cannot be edited on the website.

junedev, line 1: Instead of "cache_struct.go", "types.go" is a more common name for a file like this.

junedev (changes requested):
As mentioned before, I reviewed the exercise. I like the general idea about fetching data concurrently. I did not review the test file yet as it might change as you make some other adjustments. I will take a closer look once the other comments are addressed.

junedev:
Closing this as it seems abandoned.

PR #2328, opened 2022-07-15
gigasecond: updated exercise generator to new test generator

eklatzer:
One of #2302

### DIFF
--- a/exercises/practice/gigasecond/gigasecond_test.go
+++ b/exercises/practice/gigasecond/gigasecond_test.go
@@ -3,7 +3,6 @@ package gigasecond
 // Write a function AddGigasecond that works with time.Time.
 
 import (
-	"os"
 	"testing"
 	"time"
 )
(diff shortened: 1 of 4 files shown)


exercises/practice/gigasecond/gigasecond_test.go
```diff
@@ -36,17 +32,7 @@ func parse(s string, t *testing.T) time.Time {
...
-		// makes it to the solver though, ask for a bug report.
-		_, statErr := os.Stat("example_gen.go")
-		if statErr == nil || os.Getenv("TRAVIS_GO_VERSION") > "" {
-			t.Fatal(err)
-		} else {
-			t.Log(err)
-			t.Skip("(This is not your fault, and is unexpected.  " +
-				"Please file an issue at https://github.com/exercism/go.)")
-		}
+		t.Fatalf("error in test setup: TestAddGigasecond requires datetime in one of the following formats: \nformat 1:%q\nformat 2:%q\ngot: %q", fmtD, fmtDT, s)
```
eklatzer, line 35: Hey, I think this is a similar case as we had it the last time, so I replaced the handling with the one we discussed the last time. Here is what the output looks like:
```
error in test setup: TestAddGigasecond requires datetime in one of the following formats: 
        format 1:"2006-01-02"
        format 2:"2006-01-02T15:04:05"
        got: "2011-04-25asdfasdf"
```

eklatzer, line 35: Hey, I think this is a similar case as we had it the last time, so I replaced the handling with the one we discussed the last time. Here is what the output looks like:
```
error in test setup: TestAddGigasecond requires datetime in one of the following formats: 
        format 1:"2006-01-02"
        format 2:"2006-01-02T15:04:05"
        got: "2011-04-25asdfasdf"
```

andrerfcsantos (approved):

PR #2331, opened 2022-07-15
hamming: updated exercise generator to new test generator

eklatzer:
One of #2302

### DIFF
--- a/exercises/practice/hamming/hamming_test.go
+++ b/exercises/practice/hamming/hamming_test.go
@@ -10,13 +10,10 @@ func TestHamming(t *testing.T) {
 				if err == nil {
 					t.Errorf("Distance(%q, %q) expected error, got: %d", tc.s1, tc.s2, got)
 				}
-			} else {
-				if err != nil {
-					t.Errorf("Distance(%q, %q) returned unexpected error: %v", tc.s1, tc.s2, err)
-				}
-				if got != tc.want {
-					t.Errorf("Distance(%q, %q) = %d, want %d", tc.s1, tc.s2, got, tc.want)
-				}
+			} else if err != nil {
+				t.Errorf("Distance(%q, %q) returned unexpected error: %v", tc.s1, tc.s2, err)
+			} else if got != tc.want{
+				t.Errorf("Distance(%q, %q) = %d, want %d", tc.s1, tc.s2, got, tc.want)
 			}
 		})
 	}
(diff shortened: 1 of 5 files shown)


exercises/practice/hamming/hamming_test.go
```diff
@@ -4,39 +4,30 @@ import "testing"
...
-				t.Fatalf("Distance(%q, %q) returned unexpected error: %v",
-					tc.s1, tc.s2, err)
-			}
-			if got != tc.want {
-				t.Fatalf("Distance(%q, %q) = %d, want %d.",
-					tc.s1, tc.s2, got, tc.want)
+		t.Run(tc.description, func(t *testing.T) {
+			got, err := Distance(tc.s1, tc.s2)
+			if tc.expectError {
+				if err == nil {
```
junedev, line 10: nested if can be avoided by using `&&`

exercises/practice/hamming/hamming_test.go
```diff
@@ -4,39 +4,30 @@ import "testing"
...
+					t.Errorf("Distance(%q, %q) expected error, got: %d", tc.s1, tc.s2, got)
```
junedev, line 11: should say `got nil`, the value of "got" is not relevant in the message

exercises/practice/hamming/hamming_test.go
```diff
@@ -4,39 +4,30 @@ import "testing"
...
+				}
+			} else {
+				if err != nil {
+					t.Errorf("Distance(%q, %q) returned unexpected error: %v", tc.s1, tc.s2, err)
+				}
+				if got != tc.want {
```
junedev, line 17: can we avoid the nested ifs with "else if"?

exercises/practice/hamming/hamming_test.go
eklatzer: With `if tc.expectError && err == nil` we have one problem: If an error is exepected (`tc.expectError` is `true`) and err is not nil, we would go to the `else`, but I don't think we want to, do we?

junedev: yes, you are right, it would only work if you would add `!tc.expectError` to the conditions below

there is another problem now, the linter does not like the `else if`, instead of the suggested switch maybe just if statements (no else) with Fatalf inside would be better

sorry, I didn't know trying to improve this would lead to such a rabbit hole, feel free to just go back to the old nested version and we leave it at that :see_no_evil:

eklatzer: I have replaced it with the `switch` as I have also used this in other exercises and I like it :)

exercises/practice/hamming/hamming_test.go
eklatzer: We leave it like that as discussed in Slack

junedev (approved):

PR #2334, opened 2022-07-15
Clean up doc comments in concept exercises

kytrinyx:
We have an exercise that explicitly teaches people to use proper punctuation for doc comments.

This adds punctuation where it's missing from doc comments in the stub files for other concept exercises.

I've also removed a doc-style comment from one of the concept exercise test files, since we don't want to implicitly encourage people to write doc comments for their tests.

### DIFF
--- a/exercises/concept/animal-magic/.meta/exemplar.go
+++ b/exercises/concept/animal-magic/.meta/exemplar.go
@@ -5,22 +5,22 @@ import (
 	"time"
 )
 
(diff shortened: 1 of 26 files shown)


andrerfcsantos (approved):
Thanks for this! I'll make a note to check our linting rules so we can catch this in CI in the future.

kytrinyx:
Ah getting CI to take care of it is a great idea. Thank you!

PR #2335, opened 2022-07-15
Add missing test cases for Annalyn's Infiltration

kytrinyx:
In this exercise the instructions say that the prisoner MUST be
awake in order to be rescued, as if they are sleeping they'll be
startled by the rescue attempt, which will alert the guards.

Today we saw a solution that would have allowed Annalyn to rescue
a sleeping prisoner if she didn't have her dog along.

This adds a couple of extra cases for that scenario.

The exemplar correctly implements the function, and should therefore
pass the new tests.


andrerfcsantos (approved):

andrerfcsantos:
Thanks! Since this exercise came from the C# track, and it's probably also present in others, I'll also do a quick check if they implement similar tests.

PR #2336, opened 2022-07-15
lasagna: Fix ElapsedTime doc comment

guzalv:
Instead of the total time needed to prepare + bake a lasagna, which only
depends on the amount of layers, ElapsedTime calculates the time that
has passed until a specific moment in the baking process. This matches
the function signature, as it depends on the amount of layers and the
minutes it has been in the oven so far.

### DIFF
--- a/exercises/concept/lasagna/lasagna.go
+++ b/exercises/concept/lasagna/lasagna.go
@@ -12,7 +12,7 @@ func PreparationTime(numberOfLayers int) int {
 	panic("PreparationTime not implemented")
 }
 
-// ElapsedTime calculates the elapsed cooking time (preparation + baking) based on the amount of layers and the `actual` minutes already in the oven.
+// ElapsedTime calculates the time elapsed cooking the lasagna. This time includes the preparation time and the time the lasagna is baking in the oven.
 func ElapsedTime(numberOfLayers, actualMinutesInOven int) int {
 	panic("ElapsedTime not implemented")
 }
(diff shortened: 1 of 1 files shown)


exercises/concept/lasagna/lasagna.go
```diff
@@ -12,7 +12,7 @@ func PreparationTime(numberOfLayers int) int {
 	panic("PreparationTime not implemented")
 }
 
-// ElapsedTime calculates the total time needed to create and bake a lasagna.
+// ElapsedTime calculates the elapsed cooking time (preparation + baking) based on the amount of layers and the `actual` minutes already in the oven.
```
andrerfcsantos, line 15:
```suggestion
// ElapsedTime calculates the time elapsed cooking the lasagna. This time includes the preparation time and the time the lasagna is baking in the oven.
```

I feel maybe this comment explains the function better?

Let me know what you think the current comment doesn't explain well and if this suggestion fixes it.

andrerfcsantos (changes requested):
Thanks for opening a PR!

Left a small suggestion.

exercises/concept/lasagna/lasagna.go
guzalv: Thanks! I think your suggestion makes the comment more clear :)

guzalv:
Thanks @andrerfcsantos! applied your suggestion

andrerfcsantos (approved):
🚀

PR #2338, opened 2022-07-16
largest-series-product: updated exercise generator to new test generator

eklatzer:
One of #2302


eklatzer:
Marked as draft until discussion in #282 is finished

andrerfcsantos (approved):

andrerfcsantos:
Like june said, the most likely reason for this comes from a time where we didn't provide stubs and the student had to carefully read the tests to figure out the return types and so `var _ error = err` acted both as a compile time check and as a signal to the student that the function should return an error.

PR #2341, opened 2022-07-16
markdown: updated exercise generator to new test generator

### LINKED ISSUE #2302: Update of exercise generators

eklatzer:
Hey, for exercise `markdown` there are some new test cases that are not covered by our current implementation. The provided implementation in `markdown.go` that should be refactored does also not pass the tests. So the TODO's here are:

- [x] Update `example.go` to pass the new tests
- [x] Update `markdown.go` to pass the tests (and discuss what we want to provide)

One of #2302
Closes #2302


andrerfcsantos (approved):
Thanks a lot!

PR #2348, opened 2022-07-16
phone-number: updated exercise generator to new test generator

eklatzer:
One of #2302

What do you think of `func runTests`. As all `TestXXX` where nearly the same I thought something like this could be nice


andrerfcsantos (approved):
I like it :)

Just made small tweaks changing the order of the arguments and wrote one argument per line in the function signature.

PR #2355, opened 2022-07-17
rail-fence-cipher: updated exercise generator to new test generator

eklatzer:
One of #2302 

Added benchmark


andrerfcsantos (approved):
Applied the same tweaks as in the phone number exercise.

PR #2358, opened 2022-07-17
rectangles: updated exercise generator to new test generator

eklatzer:
One of #2302


exercises/practice/rectangles/.meta/gen.go
```diff
@@ -44,14 +37,12 @@ var testCases = []struct {
...
+}{
+{{range .J.rectangles}}{
 	description:	{{printf "%q"  .Description}},
-	input:		[]string{
-		{{range .Input.Strings}}{{printf "%q" .}},
-		{{end}}
-	},
-	expected:	{{printf "%d"  .Expected}},
-},{{end}}
+	input:			{{printf "%#v"  .Input.Strings}},
```
junedev, line 43: The old way of writing the input was much more readable in the context of the exercise as you could see the rectangles. I would recommend to go back to that version.

eklatzer: true

PR #2363, opened 2022-07-17
added the edge test case for `YearsBeforeDesiredBalance()`

hellt:
Adding the edge case when target balance has been already reached on the first iteration.
This will make student to learn checking for edge conditions

### DIFF
--- a/exercises/concept/interest-is-interesting/interest_is_interesting_test.go
+++ b/exercises/concept/interest-is-interesting/interest_is_interesting_test.go
@@ -234,7 +234,7 @@ func TestYearsBeforeDesiredBalance(t *testing.T) {
 			want:          85,
 		},
 		{
-			name:          "Years before balance already reached target balance",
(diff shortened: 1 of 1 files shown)


exercises/concept/interest-is-interesting/interest_is_interesting_test.go
```diff
@@ -233,6 +233,12 @@ func TestYearsBeforeDesiredBalance(t *testing.T) {
 			targetBalance: 12345.6789,
 			want:          85,
 		},
+		{
+			name:          "Years before balance already reached target balance",
```
andrerfcsantos, line 237:
```suggestion
			name:          "Balance is already above target",
```

andrerfcsantos (changes requested):
Thanks! Suggested just a test name change.

andrerfcsantos (approved):

PR #2382, opened 2022-07-18
word-count: updated exercise generator to new test generator

eklatzer:
One of #2302


exercises/practice/word-count/word_count.go
```diff
@@ -1,7 +1,29 @@
...
+	"strings"
+)
+
+// Frequency is a map of the frequency of occurrence keyed to the unique word.
 type Frequency map[string]int
 
+// WordCount returns the map of frequency of words based on the input phrase.
 func WordCount(phrase string) Frequency {
-	panic("Please implement the WordCount function")
+	freq := Frequency{}
```
andrerfcsantos, line 13: This is the stub file, shouldn't contain the solution :)

andrerfcsantos (changes requested):

exercises/practice/word-count/word_count.go
eklatzer: Forgot to add it when testing everything in my local environment

exercises/practice/word-count/word_count.go
```diff
@@ -4,4 +4,4 @@ type Frequency map[string]int
 
 func WordCount(phrase string) Frequency {
 	panic("Please implement the WordCount function")
-}
+}
```
andrerfcsantos, line 7:
```suggestion
}

```

andrerfcsantos (changes requested):

Issue #1892, opened 2021-10-23
Allow additional custom tests in the generator


junedev:
For some exercises it makes sense to add some language specific test cases additional to what is listed in the [problem spec](https://github.com/exercism/problem-specifications) test cases.

Here an example where this would be helpful: https://github.com/exercism/go/issues/1451

Let's discuss ideas how to adjust the generator to incorporate additional test cases. The restriction should be that the new cases adhere to the format of the existing ones (e.g. no new properties are introduced).

andrerfcsantos:
What about having a `canonical-data.append.json` in the `.meta` directory for the exercise?

junedev:
@andrerfcsantos Great idea! That shouldn't be hard and the generator is in that directory already anyway.

So the task here is to change the [generator](https://github.com/exercism/go/blob/main/gen/gen.go) so that it checks whether `.meta/canonical-data.append.json` is present for the exercise. If it it is, the content should be combined with the content of the usual canonical-data.json file.

General information about the generator can be found here: https://github.com/exercism/go#generating-test-cases

junedev:
@eklatzer Not sure whether you mean you want to include this in #2279 but if yes, it would be better if this would be separate PR. Otherwise if the added code for this feature has issues that blocks the other PR for now reason. And the more things are in one PR, the harder it is to review.

eklatzer:
I will wait for #2279 to be merged and then I will do it

eklatzer:
Note for future: we could use the feature of this issue in the exercise `reverse-string` as there is currently the following additional test case:
```
// mutiByteCases adds UTF-8 multi-byte case,
// since the canonical-data.json (generator data source for cases_test.go)
// doesn't have any such cases.
var multiByteCases = []reverseTestCase{
	{
		description: "a multi-byte test case",
		input:       "Hello, 世界",
		expected:    "界世 ,olleH",
	},
}
```

eklatzer:
Same for `roman-numerals` (will also need adjustment in the generator):
```
invalidRomanNumeralTests := []romanNumeralTest{
		{description: "0 is out of range", input: 0},
		{description: "-1 is out of range", input: -1},
		{description: "3001 is out of range", input: 3001},
	}
```

eklatzer:
Same for `two-bucket`:
```
var errorTestCases = []bucketTestCase{
	{
		"Invalid first bucket size",
		0, 5, 5, "one", "one", 1, 0, true,
	},
	{
[...]

PR #2384, opened 2022-07-18
wordy: updated exercise generator to new test generator

eklatzer:
One of #2302


exercises/practice/wordy/cases_test.go
```diff
@@ -1,159 +1,152 @@
...
+		expected:    0,
 	},
 	{
-		"reject prefix notation",
-		"What is plus 1 2?",
-		false,
-		0,
-	},
-	{
-		"reject using default 0 minus number",
```
andrerfcsantos, line 154: It seems this test is no longer present, but the `tests.toml` was not updated?

andrerfcsantos (changes requested):

exercises/practice/wordy/cases_test.go
eklatzer: Seems like it was added directly to `cases_test.go`: https://github.com/exercism/go/commit/e1a2b21f85a930e42bd683e8073e233fa6c16a2b

eklatzer: Should I add it to `wordy_test.go` and then mention it in #1892 as additional test?

andrerfcsantos: Let's not include this test then. I'll make a note to investigate this further later, but at first sight it seems that if this test is important, it should have been added directly to the problem specifications repo. It doesn't seem a test related to any go-specific change to the exercise.

PR #2387, opened 2022-07-19
Expand on parsing and formatting date-time in Golang 

zargold:
I personally had trouble understanding that Jan 2 was a special date and only after a thorough read of the example code in `time.Format` figured out how to use this feature in go. Hopefully, the details added hammer in the point about how to represent dates in Go.

This intends to greater emphasizes the point that users have to represent not just any date, but specifically 1-2-3-4-5-6-7 in their desired/custom format. I found the introduction to the `time` package a bit more brief than some of the other exercise sets so this also helps make the platform more consistently detailed.


andrerfcsantos (changes requested):
Thanks for this and sorry for the delay in reviewing this.

This is such a big source of confusion for people learning Go that I feel something like this is really important.

I'd just change the explanation in the following ways:

- In the first paragraph, I think we should talk a little bit less about the `YYYY mm` format of other languages. I think it's good to briefly say that Go is different from other languages, but we should not try to explain how the the  `YYYY mm` format works, as it might make things even more confusing.
- In the second paragraph, instead of relying on a text explanation, what about just give a couple of examples to drive the point home?

Jasstkn:
> Thanks for this and sorry for the delay in reviewing this.
> 
> This is such a big source of confusion for people learning Go that I feel something like this is really important.
> 
> I'd just change the explanation in the following ways:
> 
> * In the first paragraph, I think we should talk a little bit less about the `YYYY mm` format of other languages. I think it's good to briefly say that Go is different from other languages, but we should not try to explain how the the  `YYYY mm` format works, as it might make things even more confusing.
> * In the second paragraph, instead of relying on a text explanation, what about just give a couple of examples to drive the point home?

Hey. I’ve been also struggling with this concept. [This article](https://yourbasic.org/golang/format-parse-string-time-date-example/) helped me a lot so maybe it’s worth to consider adding it as a part of the concept documentation. 

Also Go by Example states the same if you read carefully(I didn’t): https://gobyexample.com/time-formatting-parsing.

andrerfcsantos:
@Jasstkn We already have the yourbasic link in the docs:

https://github.com/exercism/go/blob/ec502cb0bca355905f2fdd9e58c9eb3fe671a518/concepts/time/links.json#L14-L17

Or are you thinking of somewhere else?

concepts/time/introduction.md
```diff
@@ -4,4 +4,14 @@ In Go, functionality for working with times is provided by the `time` package. T
...
 
-Parsing strings works a little differently in Go. Many languages use a format string with various codes like YYYY for four-digit year, MM for two-digit month, etc. Go instead uses an exact date, `Mon Jan 2 15:04:05 -0700 MST 2006`, to parse strings. You can rearrange these components or omit them as necessary, and you can spell out month and day names, or use their number values, but in order for Go to understand the parts of the string you are parsing, the corresponding parts of the format string must be from this exact date.
+Date-time strings are represented a bit differently in Go. In many languages and libraries, developers represent their desired date-time string format with the [ISO-8601 standard](https://en.wikipedia.org/wiki/ISO_8601). In this syntax, developers must express their input/output date-time string like `YYYY mm:ss`. If developers want the library to represent `5:43 PM` as `17:43` they pass the format string `"HH:mm"`. The `time` package in Go, allows developers to instead represent a specific time in their desired format.
```
junedev, line 7: I don't think ISO 8601 says anything about the syntax (letters) you should use when converting a date to a string. In my understanding, the standard only defines that if you parse a date like "2022-12-01" give that it is supposed to follow the standard then 2022 is the year, 12 the month etc. The letters are only used to describe that in a generic way but are not themselves standardized.
Happy to be provided with some source that says otherwise.

junedev (changes requested):

junedev:
Closing this as it seems abandoned and requires some more work to get right. I created an issue so maybe some else will pick this up.

concepts/time/introduction.md
zargold: I'm pretty sure that ISO 8601 defines that these (mm/MM) be used though not sure if the standard coined these: https://www.loc.gov/standards/datetime/iso-tc154-wg5_n0038_iso_wd_8601-1_2016-02-16.pdf

junedev: Thanks a lot for link! I had trouble finding it myself. Still I find the text not super clear tbh.

In 3.3 there is this part first which makes it sound like the letters are just for writing down the standard.

> To define the date and time representations permitted by this International Standard, use is made of
date and time format representations in which specific characters are used to represent digits or other
characters in date and time representations.

And then there is this bit
> By mutual agreement the parties in information interchange may transfer the date and time format
representations. Only the date and time format representations permitted by this International
Standard shall be used.

which makes it sound more like what you are saying that the standard can also be used for standardizing the representation string itself.

PR #2388, opened 2022-07-20
census: add more tests for Count()

### LINKED ISSUE #2386: census: missing tests for the `Count` function
The last task asks to write a function `Count` that counts all the residents with valid information. However, the invalid residents we have also happen to have a missing address. This means that a solution that only tests for the presence of the address passes the tests:

```go
func Count(residents []*Resident) int {
	result := 0
	for _, resident := range residents {
		if len(resident.Address) != 0 {
			result += 1
		}
	}
	return result
}
```

## Tasks

- [ ] Add more tests to the `Count` function in [census_test.go](https://github.com/exercism/go/blob/main/exercises/concept/census/census_test.go) that includes lists of residents with invalid information or other than just the address. The tests must ensure that testing for just 1 or 2 fields is not enough to consider a resident valid/invalid.

## Working on this issue

If you would like to work on this issue, comment below that you are interested in working on this.

Jasstkn:

fixes #2386 

scope of changes:
- added one testCase with empty name and not empty address with street
- added one testCase with empty name and not empty address with other field than street

this code fails 2 cases from TestCount/some_data_collected now (was 0 before)
```go
func Count(residents []*Resident) int {
	counter := 0
	for _, resident := range residents {
		if len(resident.Address) != 0 {
	            counter += 1
		}
	}
	return counter
}
```

If user validates all fields or refer to already written function, it should be ok.

### DIFF
--- a/exercises/concept/census/census_test.go
+++ b/exercises/concept/census/census_test.go
@@ -228,6 +228,20 @@ func TestCount(t *testing.T) {
 					Age:     0,
 					Address: map[string]string{},
 				},
+				{
+					Name: "",
+					Age:  0,
+					Address: map[string]string{
+						"street": "Main St.",
+					},
+				},
+				{
+					Name: "",
+					Age:  0,
+					Address: map[string]string{
+						"city": "London",
+					},
+				},
 			},
 			want: 1,
 		},


andrerfcsantos (approved):
Thanks!

PR #2391, opened 2022-07-24
[Card Tricks] add admonition

### LINKED ISSUE #2323: Card Tricks: clarify GetItem out-of-bounds handling
Card Tricks used to return the indexed value and true/false for it the index was valid. If the index was invalid, then `0, false` could be returned.

That was changed to returning `-1` for an invalid index. But there is automated advice elsewhere that:

> By convention, when returning an `error` value, other values are not used, and it's considered idiomatic to return the [zero value](https://golang.org/ref/spec#The_zero_value) in this situation. The `error` value makes using an invalid "out of band" value to signal an error redundant in Go.

It seems that returning `-1` for an out-of-band value might be leading the student down the wrong path.

So, as per discussion with @andrerfcsantos on Slack, I'm suggesting it get changed back to returning the index and a bool.

Since the functions concept teaches about multiple return values, perhaps it could be made a prerequisite of Card Tricks? Or maybe the basics concept where functions are introduced could also show a function with multiple return values?

bobahop:
Add admonition for returning out-of-band value.
Closes #2323 

I've done only one other admonition before and don't remember if a link can be embedded in one.
Perhaps the link is not a good idea.

### DIFF
--- a/exercises/concept/card-tricks/.docs/instructions.md
+++ b/exercises/concept/card-tricks/.docs/instructions.md
@@ -28,6 +28,12 @@ If the index is out of bounds (ie. if it is negative or after the end of the sta
 ```go
 card := GetItem([]int{1, 2, 4, 1}, 10) // card == -1
 ```
+~~~~exercism/note
+By convention in Go, an error is returned instead of returning an "out-of-band" value.
+Here the "out-of-band" value is `-1` when a positive integer is expected.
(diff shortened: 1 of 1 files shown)


andrerfcsantos (approved):
Thanks a lot for this, and sorry for the delay in reviewing it.

bobahop:
It looks like the link did not work, as is prints `[zero value][zero value]` inside the note.

PR #2392, opened 2022-07-26
proverb: adjusted difficulty to 3

eklatzer:
Hey, I would suggest to change the difficulty of `proverb` to 3, as the exercise seems to be `easy` instead of `medium` for me


eklatzer:
`proverb` is comparable to `beer-song` as it also has loops, formatting and stuff like that. `beer-song` even seems harder and is `easy`

andrerfcsantos (approved):

PR #2398, opened 2022-07-28
bank-account: small fixes to instructions

guzalv:


andrerfcsantos (approved):
Thanks for this.

In Exercism, as much as possible we try to use [one sentence per line](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line), see [Exercism Markdown Specification](https://exercism.org/docs/building/markdown/markdown).

I've committed to your branch this change, keeping the content you had.

PR #2399, opened 2022-07-29
need-for-speed: added test case to cover a battery level below 100% that still allows…

### LINKED ISSUE #2394: Need For Speed  |  Task 3  |  Drive the Car
Instructions for "Task 3 - Drive the Car" are not clear:

Implement the Drive function that updates the number of meters driven based on the car's speed, and reduces the battery according to the battery drainage. If there is not enough battery to drive one more time the car will not move:

```
speed := 5
batteryDrain := 2
car := NewCar(speed, batteryDrain)
car = Drive(car)
// => Car{speed: 5, batteryDrain: 2, battery: 98, distance: 5}
```

Whats the relation between distance, speed and battery drain ?
OR
Whats the formula to calculate the distance if speed and battery drain are given ?

eklatzer:
… to drive

Closes #2394

### DIFF
--- a/exercises/concept/need-for-speed/need_for_speed_test.go
+++ b/exercises/concept/need-for-speed/need_for_speed_test.go
@@ -94,6 +94,21 @@ func TestDrive(t *testing.T) {
 			},
 		},
 		{
+			name: "Drive the car with exactly enough battery.",
+			car: Car{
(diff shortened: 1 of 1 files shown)


andrerfcsantos (approved):
Thanks for this!

PR #2402, opened 2022-07-30
[Card Tricks] Fix link in instructions.md

### LINKED ISSUE #2401: [Card Tricks] Fix link in admonition
Need to change format of link in admonition so that it renders properly.

Need to change `[zero value][zero value]`
`[zero value]: https://www.geeksforgeeks.org/zero-value-in-golang/`

to

`[zero value](https://www.geeksforgeeks.org/zero-value-in-golang/)`

bobahop:
Fix to hopefully render the link properly in the admonition.

Closes #2401


andrerfcsantos (approved):
Let's try it.

andrerfcsantos:
Seems to work, but I'm wondering if we should remove the backticks, just to make it clearer that it is a link:

image

bobahop:
I used backticks because backticks were used for links elsewhere in the instructions. For instance, the first use of `append`
info
.

PR #2403, opened 2022-08-03
Added run-generators to README

### LINKED ISSUE #2400: Update instructions on how to use the the `run-generators` script
The script `bin/run-generators` was recently changed in #2395

We should have the README briefly mention how to use this script.

eklatzer:
Closes #2400


junedev (commented):
Thanks for working on this. Improving documentation is so important, for new contributors and ourselves in the future.

* For the one command, the code block with the command does not include the GitHub token and then there is the note about using the token if this error occurs. For the other command the token is already mentioned in the command but then there is the same note. Is the token optional for the run-generators command? If yes, it should not be included in the example, if not, the note needs to be rephrased to say "the token is important because ...".
* I find the `NOTE` hard to understand. My best guess is you are trying to say something along the lines of "If you see this error, you probably have been rate limited by GitHub. In this case use a token. The token does not need any scope ..." Maybe you can rephrase this a bit.

eklatzer:
Hey, thanks for the feedback. I have included the token for the script, as it most of the times necessary to provide a token as there are so many generators to run. When running only a single generator it is only needed, if the error occurs. 

Regarding `NOTE` this is completely what I wanted to say, I willst adjust it a bit

README.md
```diff
@@ -314,8 +314,22 @@ For further instructions check out [configlet](https://github.com/exercism/confi
 $ GO111MODULE=off go run exercises/practice/<exercise>/.meta/gen.go
 ```
 
+**NOTE**: If you see the error `json: cannot unmarshal object into Go value of type []gen.Commit` when running the generator you probably have been rate limited by Github.
```
junedev, line 317: nitpick: Github should be GitHub

README.md
```diff
@@ -314,8 +314,22 @@ For further instructions check out [configlet](https://github.com/exercism/confi
...
+Try providing a Github access token with the flag `-github_token="<Token>"`.
+The token does not need any specific scopes as it is only used to fetch infos about commits and the rate limit is higher due to the token.
```
junedev, line 319:
```suggestion
Using the token will result in a higher rate limit.
The token does not need any specific scopes as it is only used to fetch infos about commits.
```

junedev: The same change should be applied below if you decide to accept it.

exercises/practice/word-count/.docs/instructions.md
```diff
@@ -12,7 +12,7 @@ When counting words you can assume the following rules:
 
 1. The count is _case insensitive_ (ie "You", "you", and "YOU" are 3 uses of the same word)
 2. The count is _unordered_; the tests will ignore how words and counts are ordered
-3. Other than the apostrophe in a _contraction_ all forms of _punctuation_ are ignored
+3. Other than the apostrophe in a _contraction_ all forms of _punctuation_ are regarded as spaces
```
junedev, line 15: This should be a separate PR. I know that is a bit annoying but it is easier to review that way and a potential discussion about a completely different topic will not hold back the actual content from being merged. (Not saying there is anything to discuss here but in theory, that could be the case.)

Also remember that we squash commits so you don't have a separate history about the different changes in one PR as you would have otherwise.

eklatzer: Thanks, will make another one in the future. Will wait until the last generator is merged (#2341) and then I will run all generators and everything again

README.md
```diff
@@ -317,8 +317,24 @@ For further instructions check out [configlet](https://github.com/exercism/confi
...
+Try providing a Github access token with the flag `-github_token="<Token>"`.
+Using the token will result in a higher rate limit.
+The token does not need any specific scopes as it is only used to fetch infos about commits.
+
 You should see that some/all of the above files have changed. Commit the changes.
 
+### Synchronizing all exercises with generators
+
+```console
+$ ./bin/run-generators <Github Access Token>
```
junedev: I am still confused with this part. Let's say my token is 123ABC. Looking at the command here I would expect that I need to run `./bin/run-generators 123ABC`. So far so good. But then the note below says I am supposed to pass the token via `-github_token="<Token>"`. Am I supposed to run `./bin/run-generators 123ABC `-github_token="123ABC"`? If yes, why do I have to provide the token twice?

(Note that I added a commit to your branch, please pull before continuing the work.)

README.md
```diff
@@ -317,11 +317,27 @@ For further instructions check out [configlet](https://github.com/exercism/confi
...
+Try providing a GitHub access token with the flag `-github_token="<Token>"`.
+Using the token will result in a higher rate limit.
+The token does not need any specific scopes as it is only used to fetch infos about commits.
+
 You should see that some/all of the above files have changed. Commit the changes.
 
+### Synchronizing all exercises with generators
+
+```console
+$ ./bin/run-generators <GitHub Access Token>
```
junedev, line 330: I am still confused with this part. Let's say my token is 123ABC. Looking at the command here I would expect that I need to run `./bin/run-generators 123ABC`. So far so good. But then the note below says I am supposed to pass the token via `-github_token="<Token>"`. Am I supposed to run `./bin/run-generators 123ABC `-github_token="123ABC"`? If yes, why do I have to provide the token twice?

(Note that I added a commit to your branch, please pull before continuing the work.)

eklatzer: Fixed it, the token only has to be provided as arg, not as flag

README.md
```diff
@@ -317,11 +317,27 @@ For further instructions check out [configlet](https://github.com/exercism/confi
...
+```
+
+**NOTE**: If you see the error `json: cannot unmarshal object into Go value of type []gen.Commit` when running the generator you probably have been rate limited by GitHub.
+Try providing a GitHub access as the first argument of the script.
```
junedev, line 334: :thinking: The code above already includes the token. Do you maybe mean "Make sure you provided the GitHub access token as first argument to the script as shown above."?

eklatzer: I agree that your version is clearer, thanks

PR #2404, opened 2022-08-04
Wrong formatted example

michelebanfi:
Following the others examples starting with `=> ` the previuos example was wronlgy suggesting `\n` at the start of the string.

### DIFF
--- a/exercises/concept/welcome-to-tech-palace/.docs/instructions.md
+++ b/exercises/concept/welcome-to-tech-palace/.docs/instructions.md
@@ -27,8 +27,7 @@ It should return a `string` that consists of 3 lines, a line with the desired nu
 
 ```go
 AddBorder("Welcome!", 10)
-// =>
-// **********
+// => **********
 // Welcome!
 // **********
 ```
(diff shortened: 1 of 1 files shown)


junedev (commented):
Thanks for raising this issue. I am a bit unsure whether the proposed change really makes the situation so much better. The new formatting makes it harder to see that there are the same amount of stars at the top and the bottom and it looks less like the actual output in general. Also someone could potentially still wonder whether or not the space is part of the output.

I am contemplating whether we should make an exception from our usual result format here and put some this here:
```
// should return the following:
// **********
// Welcome!
// **********
```

Any thoughts on this?

michelebanfi:
Yes, that's a good solution, but another way is to remove the first line completely from the code area, it can be a viable solution?


-----
Should return the following:
```
// **********
// Welcome!
// **********
```

junedev:
@michelebanfi I like that! Just make sure you keep the other code block with the example function call above (line 29). Can you update the PR accordingly?

junedev (approved):

PR #2405, opened 2022-08-06
Need For Speed | Added Formula for Calculating Race Distance and Battery Drain Rate

saurami:
Current instructions for completing the `CanFinish` function are not defined.

New instructions provide users a hint that they can use as a formula in this function.

### DIFF
--- a/exercises/concept/need-for-speed/.docs/instructions.md
+++ b/exercises/concept/need-for-speed/.docs/instructions.md
@@ -55,7 +55,10 @@ car = Drive(car)
 
 ## 4. Check if a remote controlled car can finish a race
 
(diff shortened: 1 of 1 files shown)


junedev (changes requested):
You are right that thinking about the formula is often an issue with this exercise. But as you said in your pull request description, what you wrote here is a "hint". For hints, we have a separate file: https://github.com/exercism/go/blob/main/exercises/concept/need-for-speed/.docs/hints.md That would be the better place for the new content you added.

As for the instructions, I think it is a good thing that the student needs to think a bit about how to arrive at the result here. While practicing problem solving is not the primary focus of the learning exercises, not making the exercise too trivial keeps the students engaged. If your intentation is to indicate to the student that they don't need a for loop here but the result can be calcuted instead, then maybe a text that says just that would be a better fit than spelling out the answer.

saurami:
Created new PR: #2415

PR #2406, opened 2022-08-06
Lasagna Master | Pointers Are Not being Passed to AddSecretIngredient Function

saurami:
The instruction for "Task 3 - Add the secret ingredient" states:

> Also, since slice is passed into a function as pointers, changes to the two []string arguments passed into AddSecretIngredient will be modified directly.

which is incorrect because:

1. The introduction section clearly states:

> Strictly speaking, all arguments are passed by value in Go, i.e. a copy is made of the value or data provided to the function. 

2. This is confirmed by the code block of this task

```
friendsList := []string{"noodles", "sauce", "mozzarella", "kampot pepper"}
myList := []string{"noodles", "meat", "sauce", "mozzarella","?"}

AddSecretIngredient(friendsList, myList)
// myList => []string{"noodles", "meat", "sauce", "mozzarella", "kampot pepper"}
```

### DIFF
--- a/exercises/concept/lasagna-master/.docs/instructions.md
+++ b/exercises/concept/lasagna-master/.docs/instructions.md
@@ -49,7 +49,7 @@ Now you want to add that secret ingredient to your recipe as well.
 Write a function `AddSecretIngredient` that accepts two slices of ingredients of type `[]string` as parameters.
 The first parameter is the list your friend sent you, the second is the ingredient list of your own recipe.
 The last element in your ingredient list is always `"?"`. The function should replace it with the last item from your friends list.
-**Note:** `AddSecretIngredient` does not return anything - you should modify the list of your ingredients directly. The list with your friend's ingredients should **not** be modified. Also, since `slice` is passed into a function as pointers, changes to the two `[]string` arguments passed into `AddSecretIngredient` will be modified directly.
+**Note:** `AddSecretIngredient` does not return anything - you should modify the list of your ingredients directly. The list with your friend's ingredients should **not** be modified. 
 
 ```go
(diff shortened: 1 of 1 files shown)


junedev:
@saurami Thanks for bringing this up but I have to reject this change. 
The instructions and the introduction are both correct.

> Also, since slice is passed into a function as pointers, changes to the two []string arguments passed into AddSecretIngredient will be modified directly.

The introduction section explains this in detail in the following sections about "Pointers" and "Exceptions".
https://github.com/exercism/go/blob/main/exercises/concept/lasagna-master/.docs/introduction.md#pointers

> Strictly speaking, all arguments are passed by value in Go, i.e. a copy is made of the value or data provided to the function.

To sum up what is happening in Go: The argument is indeed always copied but in case the value is a pointer (a reference to some data), it is the pointer that is copied, not the underlying data. A slice is always a pointer type so when you pass a slice as argument, the pointer is copied and via that pointer you can modify the underlying data in the slice.

Unfortunately, I don't understand how your code example under point 2 above proves your point.

I would agree that the introduction could be a bit more concise on this topic. Feel free to open an issue about improving it so we don't forget about it.

Also a PR to straighten the English in this sentence is very welcome:
> Also, since slice is passed into a function as pointers, changes to the two []string arguments passed into AddSecretIngredient will be modified directly.

From a general perspective: It is too much too teach all the details about functions and pointers in one concept/exercise that is why we have functions first and then pointers later. But that has the negative side effect that we need to hand wave over the pointer topic a bit in the functions concept. We are aware of that but it was the best tradeoff we could come up with so far.

PR #2407, opened 2022-08-06
Check Security of Source Code

saurami:
Scan source code for vulnerabilities.

Source Repository:
https://github.com/securego/gosec

### DIFF
--- a/.github/workflows/exercise-security.yml
+++ b/.github/workflows/exercise-security.yml
@@ -16,7 +16,7 @@ jobs:
       GO111MODULE: on
     steps:
       - name: Checkout Source
-        uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


saurami (commented):
bumped up version

junedev (changes requested):
This project uses [golanglintci](https://golangci-lint.run/). Adding gosec is a good idea but it should be done by adding it to the list of additional linters that golanglintci should apply. You can find the configuration file you need to edit for this here: https://github.com/exercism/go/blob/main/.golangci.yml

Make sure that the PR that changes the linter config also includes all code changes or nolint directives that might be necessary to make the code pass the new linter setup.

saurami:
gosec added to linter in PR #2416

PR #2409, opened 2022-08-06
Added part about additional test cases to README

### LINKED ISSUE #1892: Allow additional custom tests in the generator

eklatzer:
Closes #1892


junedev (changes requested):
The content related to the actual issue here is fine but the PR also contains the content of another PR. It would be easier if we can discuss/merge them separately. Please make sure this PR only contains the changes for the issue it belongs to.

(Create a branch from main for this, not from some other existing branch and put your changes in there.)

eklatzer:
I have removed the other changes directly in this branch, I created the branch from main but seems like I have copied some other stuff into the README, should be fine now

junedev (approved):
All good now, thanks!

PR #2410, opened 2022-08-06
roman-numerals: made test cases more generic

eklatzer:
Hey, when fixing #1892 I have seen that `roman-numerals` uses two tests for testing invalid and valid test cases. To be more generic I have adjusted it to be one (could be helpful if problem-specifications introduces error test cases in the future)


junedev:
@eklatzer From your description, I don't understand why the PR shows "14 files changed". For some it is hard for me to see what even changed. Could you elaborate what the other changes are or remove them from the PR?

eklatzer:
The real changes were in `roman-numerals`, everything else if from running `gofmt -s -w ./..`

junedev:
Ah ok, I'll review it all together then. In general, it is better to make a separate PR about formatting fixes. (Of course, ideally CI would prevent having formatting errors in the first place.)

junedev (commented):
(I activated "hide whitespace changes", that helped finding the important bits.)

I am struggeling a bit with this PR. I know it is a common practice to squash all kinds of different test cases into one test case struct but that does not make it right. Also the official wiki advises against it, see the second paragraph here: https://github.com/golang/go/wiki/TestComments#table-driven-tests-vs-multiple-test-functions

I understand this PR makes things more consistent but if that means making things worse code quality wise, I don't think it is worth it. Imo a lot of other tests should look more like this one before the change (which is not easily possible when stuff comes from the canonical data directly but which just indicates that we might need to make this possible in the future).
.
@eklatzer If you insist on this change, I would ask @andrerfcsantos to make the final decision on this.

eklatzer:
Hey, I don't insist on it, I only wanted to make it more consistent as it is like this in many other exercises.

andrerfcsantos:
I agree that we should separate the valid and invalid test cases into separate functions. The line `for _, tc := range append(validRomanNumeralTests, invalidRomanNumeralTests...) {` is telling to me, because we clearly have two different sets of tests that we then just join together, and a reader of the code might wonder why separating them in the first place.

Having separate functions would make this not be a source of confusion and would also avoid the slight weirdness of having to put and explain why all generated test cases have `expectError` set to false. I get the consistency argument, however I don't think it's worth pursuing it here.

I'd also prefer for the formatting changes to be made in a separate PR, as all the affected files will have in its history a reference to this commit which is apparently for an unrelated exercise and that can be confusing.

Sorry in advance for the extra work this might require - but I truly believe this is the right thing to do here.

eklatzer:
Thanks for writing back, I will close the PR as  what you describe is the state before my changes. I think you already fixed the format of the generators in the markdown PR, so no additional work

PR #2412, opened 2022-08-06
add a cross-reference to a (base) need-for-speed exercise

hellt:
I think it makes sense make a link to the exercise that is a precursor to this exercise. Otherwise students won't be able to easily refresh what they implemented in the previous exerciese.

### DIFF
--- a/exercises/concept/elons-toys/.docs/instructions.md
+++ b/exercises/concept/elons-toys/.docs/instructions.md
@@ -1,6 +1,6 @@
 # Instructions
 
-Note: This exercise is a continuation of the `need-for-speed` exercise.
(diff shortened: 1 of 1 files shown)


junedev (commented):
Thanks for bringing this up. We try to avoid links in the introduction/instructions as much as possible so students don't get distracted by reading the links. Can you elaborate why you think it is necessary to read the other exercise again before solving this one? 
(Maybe instead of a link we should rather repeat those relevant bits of information so the student does not have to jump back.)

hellt:
Hi @junedev 
I see, and this makes sense. This exercise is a little bit different as it builds on top of a specific exerciese and even refers to it , just without a link.

I think my frustration came from the fact that I didn't see that there was another tab with the `Car` struct and `NewCar` function. So I'd say this was my fault, as I somehow missed this tab entirely and though it was meant for a student to remember what Car struct was from the past exercise

PR #2413, opened 2022-08-06
Modify Chessboard exercise to comply with standard chess notation

cruzfarfan:
The current exercise swaps the definition or file and rank of a chessboard. This could be confusing for people who know chess. 

https://en.wikipedia.org/wiki/Chessboard

Also, the expected response of the first two examples do not match the diagram and the test data provided. This PR corrects both issues while respecting the spirit of the exercise


exercises/concept/chessboard/.docs/instructions.md
```diff
@@ -3,42 +3,42 @@
 As a chess enthusiast, you would like to write your own version of the game. Yes, there may be plenty of implementations of chess available online already, but yours will be unique!
 
 Each square of the chessboard is identified by a letter-number pair:
- - The vertical columns of squares, called files, are numbered 1 through 8.
- - The horizontal rows of squares, called ranks, are labelled A through H.
+ - The vertical columns of squares, called ranks, are numbered 1 through 8.
+ - The horizontal rows of squares, called files, are labelled A through H.
```
andrerfcsantos, line 7:
```suggestion
 - The horizontal rows of squares, called ranks, are numbered 1 through 8.
 - The vertical columns of squares, called files, are labeled A through H.
```

To be correct it should be like this, right?

andrerfcsantos (changes requested):
Looks good, thanks for this.

I recently started watching chess tournaments so I'm completely on board with this change :)

Just made a small suggestion.

junedev:
My husband is into chess and he looked over my shoulder and saw the old diagram in the test file. He was also immediately like "That's not right!". So thanks a lot for the catch and fix @cruzfarfan!

exercises/concept/chessboard/.docs/instructions.md
cruzfarfan: Correct. Thanks for reviewing and caching that!

andrerfcsantos (approved):
Thanks for this!

PR #2415, opened 2022-08-08
Need For Speed | Added Formula for Calculating Race Distance and Battery Drain Rate

saurami:
Based on comment in PR #2405

### DIFF
--- a/exercises/concept/need-for-speed/.docs/hints.md
+++ b/exercises/concept/need-for-speed/.docs/hints.md
@@ -23,6 +23,7 @@
 
 - Assume the car is just starting the race
 - You need to calculate the maximum distance a car can drive with the current level of battery
+- The distance is a product of the car's speed and its *drain rate*. The *drain rate* is calculated by `battery / batteryDrain`.
 - Knowing the maximum distance the car can drive, compare it with the distance of the race track
 
 [struct]: https://tour.golang.org/moretypes/2
--- a/exercises/concept/need-for-speed/.docs/hints.md
+++ b/exercises/concept/need-for-speed/.docs/hints.md
@@ -23,7 +23,8 @@
 
 - Assume the car is just starting the race
 - You need to calculate the maximum distance a car can drive with the current level of battery
-- The distance is a product of the car's speed and its remaining battery. The remaining battery can be calculated by `battery / batteryDrain`.
+- The number of times a car can be driven can be calculated by `battery / batteryDrain`.
+- The maximum distance the car can cover is the product of the car's speed and the number of times it can be driven.
 - Knowing the maximum distance the car can drive, compare it with the distance of the race track
 
 [struct]: https://tour.golang.org/moretypes/2
--- a/exercises/concept/need-for-speed/.docs/hints.md
+++ b/exercises/concept/need-for-speed/.docs/hints.md
@@ -23,7 +23,7 @@
 
 - Assume the car is just starting the race
 - You need to calculate the maximum distance a car can drive with the current level of battery
-- The distance is a product of the car's speed and its *drain rate*. The *drain rate* is calculated by `battery / batteryDrain`.
+- The distance is a product of the car's speed and its remaining battery. The remaining battery can be calculated by `battery / batteryDrain`.
 - Knowing the maximum distance the car can drive, compare it with the distance of the race track
 
 [struct]: https://tour.golang.org/moretypes/2


exercises/concept/need-for-speed/.docs/hints.md
```diff
@@ -23,6 +23,7 @@
 
 - Assume the car is just starting the race
 - You need to calculate the maximum distance a car can drive with the current level of battery
+- The distance is a product of the car's speed and its *drain rate*. The *drain rate* is calculated by `battery / batteryDrain`.
```
junedev, line 26: Sorry, I did not check the content in detail before. Now that I read it, I'm struggling to follow the logic. What is provided in the exercise as "drain" is already a "drain rate". It is the amount of battery needed per `speed` meters of moving the car. Furthermore, the "drain rate" should not depend on the current amount of energy in the battery as it would according to your formula. 

I think the correct name for that part of the equation would be "numberOfDrives" or something along those lines.

junedev (changes requested):

exercises/concept/need-for-speed/.docs/hints.md
saurami: Battery and BatteryDrain are based on whats given in the instructions.

I had raised an issue earlier about the instructions not being clear enough in [this PR](https://github.com/exercism/go/issues/2394)

The drain rate is the time a battery will last. This is derived from battery and battery drain. Therefore, the distance a car will be able to travel would be its speed times the drain rate.

I have finished this exercise and the below code passes all tests:

```
package speed

type Car struct {
	battery int
	batteryDrain int
	speed int
	distance int
}

// NewCar creates a new remote controlled car with full battery and given specifications.
func NewCar(speed, batteryDrain int) Car {
	//panic("Please implement the NewCar function")
    return Car {
		battery: 100, // default value
		batteryDrain: batteryDrain,
		speed: speed,
		distance: 0,  // default value
	}
}

type Track struct {
	distance int
}


// NewTrack creates a new track
func NewTrack(distance int) Track {
	return Track {
		distance: distance,
	}
}

// Drive drives the car one time. If there is not enough battery to drive one more time,
// the car will not move.
func Drive(car Car) Car {
    if car.battery >= car.batteryDrain {
		car.battery  = car.battery  - car.batteryDrain
		car.distance = car.distance + car.speed
	}

	return car
}

// CanFinish checks if a car is able to finish a certain track.
	drainRate := car.battery / car.batteryDrain    // time for battery to drain
	carDistance := car.speed * drainRate           // distance = speed * time

	if track.distance <= carDistance {
		return true
	} else {
		return false
	}
}

```

junedev: > The drain rate is the time a battery will last.

The word "rate" has a very specific official meaning. It cannot mean "some amount of time". A rate is always "something _per_ something else" where both of those things have different units. See https://www.cuemath.com/commercial-math/rate-definition/ for a general explanation about "rate".

Please note that the formula you use is entirely correct. My point is only about the fact that the variable name you use in the code (and in the hint) is misleading.

saurami (commented):
changed "drain rate" to "remaining battery" to make the hint clearer.

saurami:
Hi @junedev - I changed the hint based on your suggestion. 

Please let me know if its appropriate.

exercises/concept/need-for-speed/.docs/hints.md
```diff
@@ -23,6 +23,7 @@
 
 - Assume the car is just starting the race
 - You need to calculate the maximum distance a car can drive with the current level of battery
+- The distance is a product of the car's speed and its remaining battery. The remaining battery can be calculated by `battery / batteryDrain`.
```
junedev, line 26: I am sorry but this is still not correct.

The _remaining battery_ before the race is `battery`. The remaining battery after the race is `battery % batteryDrain`. None of those two interpretations of the term _remaining battery_ fit the expression `battery / batteryDrain`.

As mentioned before, `battery / batteryDrain` describes how many times one can call the "Drive" method on the car, that is why I suggested the name `numberOfDrives`. That multiplied with the speed gives the distance the case can drive in total.

junedev (changes requested):

saurami (commented):
clearer hints for number of drives

junedev (approved):
Looks good now!

SR8H:
Hi @junedev - the hint is not displayed in task 4 over here: https://exercism.org/tracks/go/exercises/need-for-speed/edit

Please let me know if I have to make any changes.

PR #2416, opened 2022-08-08
added gosec to linter based on comment in PR #2407

saurami:
Reference: https://github.com/securego/gosec/blob/master/.golangci.yml

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -8,6 +8,7 @@ linters:
     - gocyclo
     - goimports
     - gofmt
+    - gosec
     - gosimple
     - govet
     - staticcheck


junedev (changes requested):
@saurami As I expected, the code does not pass the linter tests anymore now. See https://github.com/exercism/go/runs/7734762492?check_suite_focus=true.

We have quite some exercises that use math/rand and for those exercises it is usually totally correct not to use crypto/rand. So what you need to do is to add nolint directives in the respective places so the CI checks succeed again. See https://golangci-lint.run/usage/false-positives/#nolint-directive for examples of these directives.

saurami:
Hi @junedev I accidentally closed this PR. Please reopen it after reviewing [this PR](https://github.com/exercism/go/pull/2440)

junedev:
@saurami Please include the fixes you have in #2440 here (you can add more commits to your "patch-5" branch and push them and they will show up here. Then close #2440. That makes it easier to see whether all linter errors are addressed.

saurami:
@junedev I'll go ahead and add this PR's change in #2440 .

There's only one change in this PR and five in that one.

I also have a question - earlier you said:

> We have quite some exercises that use math/rand and for those exercises it is usually totally correct not to use crypto/rand. 

Did you mean only the "Animal Magic" exercise, or do you think others in #2440 as well?

In the syllabus section, only the "Animal Magic" exercise shows up in the Randomness section.

junedev:
Yes, the other way around is fine as well of course. 

I meant all those exercises that you already found. I knew there were a couple of randomness related exercises besides animal magic.

saurami:
Closing this PR as  because changes are incorporated in PR #2440

PR #2419, opened 2022-08-09
updated all exercises with generators

eklatzer:
Hey, I used ./bin/run-generators to update all exercises with generators


andrerfcsantos (approved):
Thanks!

Issue #2417, opened 2022-08-09
Improve test for Animal Magic / GenerateWandEnergy


funwithbots:
The current test validates that different values are generated but it doesn't detect typecasting `rand.Intn` to float64. This passes and I don't believe meets the intent of the challenge.

`
func GenerateWandEnergy() float64 {
   return float64(rand.Intn(12))
}
`

junedev:
@norbs57 In case you are still around, would you like to take a look at this as you are the author of the exercise?

norbs57:
We could add a test that divides the range [0..12] into a number of smaller intervals, say 20 intervals each of size 0.6, e.g. [0..0.6), [0.6..1.2), [1.2..1.8), etc. The test program would count the frequency of generated numbers in each interval, and it would flag up student submissions where the frequency distribution was statistically highly improbable.  

In order to detect more sophisticated "non-random" sequences generated by student programs, we would need to code more advanced statistical randomness tests, see the literature. I am not sure that would be worth it.

junedev:
I like the idea with the intervals. It doesn't have to be more sophisticated just as you said. 
The check for "highly improbable" can also just be something simple like "2 buckets with 0 values inside after some large number of tries".

@norbs57 @funwithbots Anyone up for creating the PR with the new test?

funwithbots:
math/rand pretty much only gives back some variation of `int` or `float` values. I think detecting use of `int` for the test is probably sufficient. If someone goes nuts and does something elaborate with floats or intentionally games the test, I agree it's not worth trying to gauge true randomness. There are so many ways to break this short of a statistical analysis. 

`float64` gives enough digits of precision that after even 10 iterations, if all the values are +/- 1e+6 of an integer, it should fail. Flipping a bool as soon as a non-int based value is returned would do this.

I'm happy to open the PR and add a test if that makes sense. Otherwise, I can take the bucket approach.

norbs57:
The bucket-approach would also deal with submissions that leave out part of the expected range, say they never generate numbers less than 1 or larger than 11.  Having said that, you could also write some separate tests to deal with those.  

The approach of checking frequencies could also be used to improve testing of functions `RollADie`  and `ShuffleAnimals` functions in this exercise.

norbs57:
I am taking a break from Exercism (and Slack) during the Summer. Having said that, I would be happy to check the probability of "false positives" in the "Animal Magic" additional tests. Please drop me an email if you would like me to look at that.

[...]

PR #2420, opened 2022-08-09
Animal Magic: Added test case for GenerateWandEnergy()

funwithbots:
Proposed test case to improve testing of GenerateWandEnergy() task of Animal Magic.
The tests currently pass if a student uses randomly generated ints and then typecasts the result to float64 which doesn't meet the requirements.

`tests` was bumped to ensure that even if a random seed is used, the test should pass. For grins, I ran the bucket test 1M times with various values for `tests` and it passed at 150. I bumped to 200 for extra margin.

Assuming no issues, I will give the same treatment to the other two tasks in the exercise.
#2417

### DIFF
--- a/exercises/concept/animal-magic/animal_magic_test.go
+++ b/exercises/concept/animal-magic/animal_magic_test.go
@@ -62,17 +62,19 @@ func TestWandEnergy(t *testing.T) {
 		last = got
 	}
 	if !foundDifferent {
-		t.Errorf("GenerateWandEnergy() always generates the same number: %f", got)
+		t.Fatalf("GenerateWandEnergy() always generates the same number: %f", got)
 	}
 
-	var emptyBuckets int
-	for _, v := range buckets {
+	var low, high float64
+	for i, v := range buckets {
 		if v == 0 {
-			emptyBuckets++
+			low = float64(i) * bucketSize
+			high = float64(i+1) * bucketSize
+			break
 		}
 	}
-	if emptyBuckets > 1 {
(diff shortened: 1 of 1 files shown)


exercises/concept/animal-magic/animal_magic_test.go
```diff
@@ -55,11 +58,22 @@ func TestWandEnergy(t *testing.T) {
...
 	}
+
+	var emptyBuckets int
+	for _, v := range buckets {
+		if v == 0 {
+			emptyBuckets++
+		}
+	}
+	if emptyBuckets > 1 {
+		t.Errorf("GenerateWandEnergy() results are not uniformly distributed. %d of %d buckets are empty.", emptyBuckets, numBuckets)
```
junedev, line 75: For concept exercises, the student cannot see the test file when working on the website. That means it is important that the error messages are understandable without seeing the code.

I would recommend not talk about "buckets", instead either write a general statement that some sections of the range are not covered by the results or and output one example of an empty section (which you would need to "remember" from the loop above).

junedev (commented):
Thanks for working on this. I had just one comment regarding the failure message.

exercises/concept/animal-magic/animal_magic_test.go
funwithbots: Makes sense. Also no reason to run the entire loop once a failure is detected.

junedev (approved):

junedev:
I reduced the precision for low and high because there were quite a lot of zeros at end. Worked great otherwise. Will merge this in a bit, thanks again.

PR #2422, opened 2022-08-10
Allow whitespace in comments

twocs:
Without updating this test, the comments in go fail. For example, this should be a valid package description comment (assuming it was for weather not builtin): 
/*
	Package builtin provides documentation for Go's predeclared identifiers.
	The items documented here are not actually in package builtin
	but their descriptions here allow godoc to present documentation
	for the language's special identifiers.
*/

### DIFF
--- a/exercises/concept/weather-forecast/weather_forecast_test.go
+++ b/exercises/concept/weather-forecast/weather_forecast_test.go
@@ -112,7 +112,7 @@ func testComment(entityKind, entityName, comment, wantedPrefix string) (ok bool,
 	lowerEntity := strings.ToLower(entityKind)
 
 	// Check if comment has wanted prefix
-	if !strings.HasPrefix(comment, wantedPrefix) {
+	if !strings.HasPrefix(trimmedComment, wantedPrefix) {
 		errorString := fmt.Sprintf("%s comment for %s '%s' should start with '// %s ...': got '// %s'",
 			entityKind, lowerEntity, entityName, wantedPrefix, trimmedComment)
 		return false, errorString


twocs:
Updated tests were submitted by @andrerfcsantos. His exercise with block comment passed the tests, but only because they were far to the left without tabs, unlike the core doc go comments. 
See a failed test here on Iteration 2:
https://exercism.org/tracks/go/exercises/weather-forecast/solutions/twocs

Iteration 1 was successful

exercises/concept/weather-forecast/weather_forecast_test.go
```diff
@@ -112,7 +112,7 @@ func testComment(entityKind, entityName, comment, wantedPrefix string) (ok bool,
 	lowerEntity := strings.ToLower(entityKind)
 
 	// Check if comment has wanted prefix
-	if !strings.HasPrefix(comment, wantedPrefix) {
+	if !strings.HasPrefix(trimmedComment, wantedPrefix) {
```
junedev, line 115: The `testComment` function you modified here is also used for other entities than the package level comment. However, to use block comments e.g. for a function comment would not be idiomatic to Go.

I think the best way to handle this would be to create a separate version of this function and use that when checking the package level comment.

junedev (changes requested):

exercises/concept/weather-forecast/weather_forecast_test.go
twocs: I'm not sure I see your point. The tests pass block comments and line comments identically, for package comments and function comments. If you think that that behaviour is incorrect, I suggest that it's a separate issue altogether. The problem that this fix addresses is that the test author trimmed the spaces, which are not consequential to the code comment, but then tested against the raw comment string instead of the trimmed comment.

twocs: Possibly relevant to see how godoc itself parses code comments (see TestSanitizeFunc for examples of it stripping new lines and tabs, basically it uses HTMLEncode, which strip whitespace.)
https://go.googlesource.com/tools/+/refs/heads/master/godoc/godoc_test.go

junedev: Sorry for the confusion. I did not properly check what exactly is used as input for the function. You are right, the PR is ok as is and recommending line comments would be another issue entirely.

junedev (approved):

PR #2423, opened 2022-08-11
Added hint to auto generated test files that file is autogenerated

### LINKED ISSUE #2421: Add comment at the top of generated files with test cases
Every now and then it happens that contributors directly edit the cases file which is produced by the generator. As we might move towards autogenerating issues about updating exercises that contributors can pick up, this problem might happen more often.

To avoid this, it would be good if the main generator script would add a comment section at the top explaining that the file is auto-generated and should not be edited directly and maybe link to the docs of how to run the generator.

eklatzer:
Hey, any wishes how to improve the text?

Closes #2421


gen/gen.go
```diff
@@ -51,7 +50,7 @@ type Header struct {
 
 // String generates the header for cases_test.go file.
 func (h Header) String() string {
-	return fmt.Sprintf("// Source: %s\n// Commit: %s", h.Origin, h.Commit)
+	return fmt.Sprintf("// AUTOGENERATED FILE: do not change manually (how to run generator: https://github.com/exercism/go#synchronizing-tests-and-instructions)\n// Source: %s\n// Commit: %s\n", h.Origin, h.Commit)
```
junedev, line 53: Maybe the text can be made a bit smoother. Also the line length is rather long for the new comment. Here a potential alternative:
```
// This is an auto-generated file. Do not change it manually. Run the generator to update the file.
// See https://github.com/exercism/go#synchronizing-tests-and-instructions.
```

junedev (approved):
Nice!

PR #2426, opened 2022-08-11
Added generator and gomod-sync to linters and tests

### LINKED ISSUE #2418: CI should run tests and linters for non-exercise folders
Currently, we only run the tests and linter for the exercises but we should also check everything is passed for the generators and for the gomod-sync repo. This can be done by modifying some existing CI script or adding a new one.

eklatzer:
Closes #2418


eklatzer:
Fixed the problem with failing CI on windows by using using an entry in `.gitattributes` as recommended [here](https://github.com/golangci/golangci-lint/issues/580#issuecomment-704913317) in https://github.com/golangci/golangci-lint/issues/580

junedev:
@eklatzer I don't understand the fix. We were running the linter on a windows CI runner before (for all the exercises) and did not have that file and it worked fine. Why is the file needed now?

eklatzer:
To be honest I cannot completely answer, but maybe only some of the files have the problem? Was only a problem for some of the exercise generators, not all.

junedev:
@eklatzer Did you check what line endings those files have that you suspect are the cause of the problem? Maybe they really have the wrong line endings and that should be fixed (potentially in addition to preventing someone committing wrong line endings again).

eklatzer:
Have removed the fix and will check the endings

eklatzer:
@junedev 
>  I don't understand the fix. We were running the linter on a windows CI runner before (for all the exercises) and did not have that file and it worked fine. Why is the file needed now?

To be honest I am also not sure, but seems it has fixed itself, or maybe I changed something with line endings but it works now without any changes

eklatzer:
I added the following parts to CI:
* linting for all `gen.go` files
* linting and test execution for `./gomod-sync` and `./gen`
@junedev Is this everything you wanted to have covered?

.gitattributes
```diff
@@ -0,0 +1 @@
+*.go text eol=lf
```
eklatzer, line 1: I could not finish CI without this line, maybe you find a way

.golangci.yml
```diff
@@ -1,5 +1,5 @@
 run:
-  timeout: 5m
+  timeout: 8m
```
eklatzer, line 2: Increased the timeout as I had a problem with a timeout on windows

eklatzer:
To prevent stuff like in #2451 and to detect forgotten working implementations in the stub I added a CI step, what do you think about this? I am a bit confused why this passes in CI, locally it fails because this branch still contains the old gigasecond. I also tried running the tests on multiple OS/arch combintations. Am I missing something?
```
--- FAIL: TestAddGigasecond (0.00s)
    --- PASS: TestAddGigasecond/date_only_specification_of_time (0.00s)
    --- PASS: TestAddGigasecond/second_test_for_date_only_specification_of_time (0.00s)
    --- PASS: TestAddGigasecond/third_test_for_date_only_specification_of_time (0.00s)
    --- PASS: TestAddGigasecond/full_time_specified (0.00s)
    --- PASS: TestAddGigasecond/full_time_with_day_roll-over (0.00s)
FAIL
FAIL    _/home/eklatzer/dev/Golang/exercism-go/tmp/lint/practice/gigasecond     0.004s
FAIL
[ERROR] stub at ./tmp/lint/practice/gigasecond passed the tests, even thought expect stub to fail
```

.github/workflows/exercise-tests.yml
```diff
@@ -10,7 +10,7 @@ on:
```
junedev, line 10: Re the line endings, if I run the "check for windows line endings" command I found on Stackoverflow for this branch, it still shows a file with the wrong line endings:
```
find . -not -type d -exec file "{}" ";" | grep CRLF
=> ./exercises/practice/forth/.meta/tests.toml: ASCII text, with CRLF line terminators
```
Maybe see how it goes after that one is fixed as well.

.github/workflows/stub-tests.yml
```diff
@@ -0,0 +1,45 @@
...
+      - uses: actions/setup-go@268d8c0ca0432bb2cf416faae41297df9d262d7f
+        with:
+          go-version: ${{ matrix.go-version }}
+
+      - name: Run tests on stubs
+        shell: bash
+        env:
+          GOARCH: ${{ matrix.test-arch }}
+        run: |
+          ./bin/test-without-stubs test-stubs
```
junedev, line 45: That the "test-without-stubs" script now also runs tests with stubs is very confusing. The script should be renamed.

Honestly, I would prefer if you could create a separate PR for adding this bit. It makes this PR even larger, it introduces new problems as you wrote and all of that delays that we can get the other good things you added here merged.

If you extract that bit into a new PR, make sure this PR is up to date with master, then we can resolve whatever issues are left regarding to the original goal (lint and run tests for generator and gomod-sync) and then we can merge this. More can always be done later. Please keep this in mind in general, small PRs are always better, always try to avoid increasing the scope.

junedev (changes requested):

.github/workflows/stub-tests.yml
eklatzer: > Honestly, I would prefer if you could create a separate PR for adding this bit
What should be part of the other PR? Renaming `test-without-stubs`?

.github/workflows/exercise-tests.yml
eklatzer: Could you again check if that is still the case? If I run `find . -not -type d -exec file "{}" ";" | grep CRLF` I don't get any files

.github/workflows/stub-tests.yml
junedev: I meant the whole "running the tests against the stubs and check they fail" part. If you can get it working properly, then you can rename the file and keep it in. Otherwise, as mentioned above, it would be better to tackle "running tests against stubs" in a separate PR.

eklatzer: I renamed the script, is the name fine like this?

I have added a code passing the tests for Gigasecond and as you can see, CI fails, so it is working properly

junedev: Cool, what did you change compared to before where you said it is not working?

The name is fine now.

eklatzer: I did not change anything, I only adjusted the stub to be really passing tests (must be reverted before merging). The problem from https://github.com/exercism/go/pull/2451 is not detected, because in CI the tests do not pass:https://github.com/exercism/go/actions/runs/3073998679/jobs/4966490822#step:4:1840

gen/gen.go
```diff
@@ -181,9 +180,8 @@ func Gen(exercise string, tests map[string]interface{}, t *template.Template) er
 	formattedFileContent, err := format.Source(casesFileContent)
 	if err != nil {
 		fmt.Print("[ERROR] failed to format the output with gofmt (the generated source has a syntax error)")
-		debugFileContent := append(casesFileContent, []byte("// !NOTE: Error during source formatting: Line:Column "+fmt.Sprint(err)+"\n")...)
 		// Save the raw unformatted, error-containing source for purposes of debugging the generator.
-		_ = outputSource("ERROR", casesFile, debugFileContent)
+		_ = outputSource("ERROR", casesFile, append(casesFileContent, []byte(fmt.Sprintf("// !NOTE: Error during source formatting: Line:Column %v\n", err))...))
```
junedev, line 184: Why was the code here changed? The old version seemed more readable. 🤔

gen/remote.go
```diff
@@ -43,7 +43,7 @@ func getRemoteCommit(exercise, token string) (Header, error) {
 	url := fmt.Sprintf(commitsURL, exercise)
 	fmt.Printf("[REMOTE] fetching latest commit (source: %s)\n", url)
 
-	req, err := http.NewRequest("GET", url, nil)
+	req, err := http.NewRequest("GET", url, http.NoBody)
```
junedev, line 46: Is there some linter suggesting this change? In real life, nil is very common to be used here and also mentioned in the official documentation.

junedev (commented):
Left some comments/questions. Will try to help debugging the fmt issue soon.

bin/run-tests
```diff
@@ -50,10 +54,19 @@ EOD
...
+        echo "copy exemplar $exemplar to $tmpstub"
+        cp $exemplar "${tmpstub}/${stub}.go"
+      fi
       echo "copy module $mod to $tmpstub"
       cp $mod $tmpstub
+
+      generator="${dir}.meta/gen.go"
+      if [ -f "${generator}" ]; then
+        echo "copy generator ${generator}"
+        mkdir -p "${tmp}/gen/${stub}" && cp "${generator}" "${tmp}/gen/${stub}/gen.go"
```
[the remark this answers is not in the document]
eklatzer: I did it because this directory is automatically linted, so we don't need to run another lint command

bin/run-tests
```diff
@@ -75,7 +88,7 @@ cleanup () {
   rm -rf $tmp
 }
 
-trap cleanup EXIT INT TERM
+#trap cleanup EXIT INT TERM
```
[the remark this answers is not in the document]
eklatzer: Forgot to remove the `#`, deactivated the cleanup for testing purposes, thanks

bin/run-tests
```diff
@@ -84,10 +97,58 @@ trap cleanup EXIT INT TERM
 # module-aware mode only in places where a go.mod is present.
 export GO111MODULE=auto
 
-# This is the last command in the file,
-# so the exit status of this script is the exit status of go test.
-# If this ever changes, remember to preserve the exit status accordingly.
-# Otherwise CI may falsely report a test as passed when it has failed.
+if [ "$mode" = "$TEST_STUBS_MODE" ];then
```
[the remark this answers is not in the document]
eklatzer: Adjusted it, to print `testing ./tmp/lint/practice/<exercise>` for each exercise, but only the complete output if a test fails

gen/remote.go
eklatzer: I think it was a linter suggestion

gen/gen.go
eklatzer: Linter suggestions, because `debugFileContent := append(casesFileContent, ...)` should assign the value to the same variable as used in the first parameter

exercises/practice/custom-set/.meta/example_slice.go
```diff
@@ -1,4 +1,5 @@
-//+build slice
+//go:build slice
+// +build slice
```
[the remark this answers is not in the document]
eklatzer: Have removed the new ones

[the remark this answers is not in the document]
eklatzer: After running `gofmt -s -w ./..` the new version is added again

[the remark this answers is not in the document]
junedev: It's really annoying that the repo is supposed to function with 1.16 but we run it with 1.18 already. Anyway, then let's keep the two directives in and let's hope older Go versions just ignore the newer one.

gen/gen.go
junedev: So but then why don't we use "casesFileContent" instead of "debugFileContent" but keep in on a separate line for readability?

bin/run-tests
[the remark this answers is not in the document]
junedev: Would be good to leave a comment for this.

.github/workflows/exercise-tests.yml
junedev: It's all good now.

For the general issue, I found this post the most "insightful" so far: https://github.com/golangci/golangci-lint/issues/1018#issuecomment-618594466

The gofmt linting complains about 3 files "at random" on Windows, not 3 specific ones. The default in golangci-lint is to only report 3 errors.
I think there is something that goes wrong when the git checkout is done on windows by the CI. Maybe it is possible to find out what setting for autocrlf is applied by the github action and whether we can change it to something else.

exercises/practice/custom-set/.meta/example_slice.go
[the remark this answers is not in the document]
eklatzer: Why is it supposed to function with Go 1.16? Currently supported is 1.18 and 1.19, can't we require at least 1.17 or 1.18?

gen/gen.go
eklatzer: Changed

.github/workflows/exercise-tests.yml
eklatzer: Using `git config --system core.autocrlf false` seems to really fix it, can we keep this? Thanks for providing the link to the issue.

exercises/practice/custom-set/.meta/example_slice.go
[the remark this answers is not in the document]
junedev: All I'm saying is that we currently specifiy Go 1.16 in all the go.mod files. That means we garantee it works with Go 1.16.
We wanted to update to 1.18 once 1.20 is out and then make use of the generics stuff.

.github/workflows/exercise-tests.yml
junedev: Yes, I think that is a good solution. Well done! If we would prescribe such a setting to every contributor there might be issues but since it is only for the CI windows maschine, I think its totally fine.

junedev:
I will merge this now, let's see how it will work in the wild. 😉

PR #2427, opened 2022-08-12
Robot Simulator: Drop "Define Dir" as it is defined already; update who sends the report.

IsaacG:


IsaacG:
@junedev for review, please.

exercises/practice/robot-simulator/robot_simulator.go
```diff
@@ -3,7 +3,6 @@ package robot
 // See defs.go for other definitions
 
 // Step 1
-// Define Dir type here.
```
junedev, line 6: Dir is already defined but the task is to implement the String method for it. Instead of removing the comment completely, maybe it would be better to change it.

junedev (commented):
I agree, the room makes more sense in the context of the "final position report" mentioned here.

Regarding the other change, I left a comment.

Thanks for working on this.

exercises/practice/robot-simulator/robot_simulator.go
IsaacG: There are 7 other functions which are defined but `panic()`. None of the others have a comment telling you to implement them. The `Define Foo` comments are generally tied to defining something not defined. Implementations of functions are typically handled via a `panic("implement foo")`.

Have a comment to implement `Dir.String()` would be at odds with all the other implements.

junedev: I'm sorry, I did not see there is already the method stub below. That's perfectly fine of course. I thought there was nothing at all hinting that this method needs to be implemented. My mistake.

junedev (approved):

Issue #2098, opened 2022-02-11
Migrate data-driven tests to subtests


artamonovkirill:
Following [this Slack thread](https://exercism-team.slack.com/archives/CAS0EK88H/p1644497987712339) and [this PR discussion](https://github.com/exercism/go/pull/1254/files)

In the track, there are still some data-driven tests like:
```go
func TestSomething(t *testing.T) {
    for _, test := range tests {
        ... some test code ...
        t.Logf("PASS: %s", test.description)
    }
}
```

Such tests could benefit from rewriting them with subtests:
```go
func TestSomething(t *testing.T) {
    for _, test := range tests {
        t.Run(test.description, func(t *testing.T) {
            ... some test code ...
        }
    }
}
```

Pros:
* Easier to parse test output of the test runner
* For an IDE user, it's easy to see which test failed without the need to scroll (sometimes) lengthy output.
* A possibility to run exactly 1 subtest

Cons:
* test names/descriptions lowercased with underscores instead of spaces
* no "fail fast" behaviour without checking the output value from `t.Run`

A (possibly incomplete) list of exercises without subtests:
* [List Ops](https://github.com/exercism/go/blob/73e8ce8bd81bf90d3bc0e97927f0f7fe3f27f4d6/exercises/practice/list-ops/list_ops_test.go)
* [Run Length Encoding](https://github.com/exercism/go/blob/24ac0184920bb6bbfb4f28e363339397068fb172/exercises/practice/run-length-encoding/run_length_encoding_test.go)
* [Change](https://github.com/exercism/go/blob/d219b6f83599280476e57f0ae39b9551bacea6e6/exercises/practice/change/change_test.go)
* [Variable Length Quantity](https://github.com/exercism/go/blob/64d5dd0799d4fbbc26dfea6440e0c1029528fa4c/exercises/practice/variable-length-quantity/variable_length_quantity_test.go)
* [Perfect Numbers](https://github.com/exercism/go/blob/d219b6f83599280476e57f0ae39b9551bacea6e6/exercises/practice/perfect-numbers/perfect_numbers_test.go)
* [Allergies](https://github.com/exercism/go/blob/bcd640d31d52a0e66c5a24095d399947c3205715/exercises/practice/allergies/allergies_test.go)
* [Bowling](https://github.com/exercism/go/blob/64d5dd0799d4fbbc26dfea6440e0c1029528fa4c/exercises/practice/bowling/bowling_test.go)
[...]

PR #2428, opened 2022-08-12
added subtests to chessboard

eklatzer:
One of #2098


exercises/concept/chessboard/chessboard_test.go
```diff
@@ -32,74 +33,70 @@ func newChessboard() Chessboard {
...
+		{in: "D", expected: 0},
+		{in: "E", expected: 2},
+		{in: "F", expected: 0},
+		{in: "G", expected: 1},
+		{in: "H", expected: 7},
+		{in: "Z", expected: 0},
+	}
+	for _, test := range testCases {
+		t.Run(test.in, func(t *testing.T) {
+			if got := CountInFile(cb, test.in); got != test.expected {
```
andrerfcsantos, line 52: What do you think of adding a `description` field to each test case and use that as a test name? I think having a test named "A" or "1" can be confusing. With a description we could be more explicit, as in `"Count of pieces in file A"`, or  `"Count of pieces in rank 1"`

exercises/concept/chessboard/.meta/config.json
```diff
@@ -19,4 +21,4 @@
       ".meta/exemplar.go"
     ]
   }
-}
+}
```
andrerfcsantos, line 24:
```suggestion
}

```

andrerfcsantos (approved):

PR #2429, opened 2022-08-12
added subtests for exercise binary

eklatzer:
One of #2098


eklatzer:
Exercise is deprecated

PR #2432, opened 2022-08-12
added usage of subtests to difference-of-squares

eklatzer:
One of #2098


exercises/practice/difference-of-squares/.meta/config.json
```diff
@@ -30,4 +31,4 @@
   },
   "source": "Problem 6 at Project Euler",
   "source_url": "http://projecteuler.net/problem=6"
-}
+}
```
andrerfcsantos, line 34:
```suggestion
}

```

exercises/practice/difference-of-squares/difference_of_squares_test.go
```diff
@@ -1,35 +1,44 @@
...
+	{input: 10, squareOfSum: 3025, sumOfSquares: 385},
+	{input: 100, squareOfSum: 25502500, sumOfSquares: 338350},
 }
 
 func TestSquareOfSum(t *testing.T) {
 	for _, test := range tests {
-		if s := SquareOfSum(test.n); s != test.sqOfSum {
-			t.Fatalf("SquareOfSum(%d) = %d, want %d", test.n, s, test.sqOfSum)
-		}
+		t.Run(strconv.Itoa(test.input), func(t *testing.T) {
```
andrerfcsantos, line 16: Same suggestion as in #2428, I think we could use a test description here and use it as the subtest name.

eklatzer: Added descriptions, what do you think about them?

andrerfcsantos (approved):

PR #2435, opened 2022-08-13
added usage of subtests food-chain

eklatzer:
One of #2098


exercises/practice/food-chain/.meta/config.json
```diff
@@ -29,4 +30,4 @@
   },
   "source": "Wikipedia",
   "source_url": "http://en.wikipedia.org/wiki/There_Was_an_Old_Lady_Who_Swallowed_a_Fly"
-}
+}
```
andrerfcsantos, line 33:
```suggestion
}

```

exercises/practice/food-chain/food_chain_test.go
```diff
@@ -86,21 +87,23 @@ func diff(got, want string) string {
 
 func TestVerse(t *testing.T) {
 	for v := 1; v <= 8; v++ {
-		if ret := Verse(v); ret != ref[v] {
-			t.Fatalf("Verse(%d) =\n%s\n  want:\n%s\n%s", v, ret, ref[v], diff(ret, ref[v]))
-		}
+		t.Run(strconv.Itoa(v), func(t *testing.T) {
```
andrerfcsantos, line 90: I think we can put a description here too. In the simplest form, it can just be an `strconv.Itoa(v) + " verses"`, but it might be worth to convert the test cases to a slice of test cases, where each test case can be something like:

```go
type testCase struct {
    description string
    verses int
    expectedText string
}
```

eklatzer: To be honest I don't think there is a need of adding a struct and testcases, I have simply added `verse` to the description, what do you think?

andrerfcsantos (approved):

PR #2436, opened 2022-08-13
added usage of subtests for hexadecimal

eklatzer:
One of #2098


junedev:
@eklatzer Isn't this exercise deprecated? This should be visible in the root level config.json. Please don't create PRs for deprecated exercises as it causes unnecessary work.

eklatzer:
Sorry, have not seen that it is deprecated

PR #2437, opened 2022-08-13
fixed bonus tests

eklatzer:
Hey, when updating everything for the generators I seem to have destroyed the bonus test case for two-bucket. Should be fixed now

One of #2098


eklatzer:
@junedev We should consider to add stuff like bonus tests to the linting job of CI

andrerfcsantos (approved):
Thanks for this.

We should be careful adding the bonus test cases to CI, since they often take more time than regular tests, even if the solution is optimal.

PR #2438, opened 2022-08-13
added usage of subtests to the house exercise

eklatzer:
One of #2098

### DIFF
--- a/exercises/practice/house/house_test.go
+++ b/exercises/practice/house/house_test.go
@@ -11,7 +11,7 @@
 package house
 
 import (
-	"strconv"
+	"fmt"
 	"strings"
 	"testing"
 )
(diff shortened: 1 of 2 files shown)


exercises/practice/house/house_test.go
```diff
@@ -112,9 +113,11 @@ that lay in the house that Jack built.`
 
 func TestVerse(t *testing.T) {
 	for v := 0; v < len(expectedVerses); v++ {
-		if ret := Verse(v + 1); ret != expectedVerses[v] {
-			t.Fatalf("Verse(%d) =\n%q\n  want:\n%q", v+1, ret, expectedVerses[v])
-		}
+		t.Run(strconv.Itoa(v+1), func(t *testing.T) {
```
andrerfcsantos, line 116: Similar suggestions to #2435

eklatzer: To be honest I don't think there is a need of adding a struct and testcases, I have simply added `verse` to the description, what do you think?

PR #2440, opened 2022-08-13
Ignore Security Warnings for math/rand

saurami:
Updated test files so that `gosec` ignores `math/rand`.  Files updated based on [this CI job](https://github.com/exercism/go/runs/7734762492?check_suite_focus=true) 

For these exercises `math/rand` is better a better option than `crypto/rand` based on suggestion in PR #2416

### DIFF
--- a/exercises/concept/animal-magic/.meta/exemplar.go
+++ b/exercises/concept/animal-magic/.meta/exemplar.go
@@ -1,4 +1,4 @@
-//nolint:gosec // Randomness is part of an exercise where math.rand is more appropriate than crypto.rand
+//nolint:gosec // In the context of this exercise, it is fine to use math.Rand instead of crypto.Rand.
 package chance
 
 import (
--- a/exercises/concept/animal-magic/.meta/exemplar.go
+++ b/exercises/concept/animal-magic/.meta/exemplar.go
@@ -1,3 +1,4 @@
+//nolint:gosec // Randomness is part of an exercise where math.rand is more appropriate than crypto.rand
 package chance
 
 import (
--- a/exercises/practice/binary-search/binary_search_test.go
+++ b/exercises/practice/binary-search/binary_search_test.go
@@ -1,4 +1,4 @@
-//nolint:gosec // Randomness is part of an exercise where math.rand is more appropriate than crypto.rand
+//nolint:gosec // In the context of this exercise, it is fine to use math.Rand instead of crypto.Rand.
 package binarysearch
 
 import (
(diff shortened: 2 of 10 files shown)


saurami (commented):
added explanation for ignoring gosec on math/rand

saurami:
Hi @junedev please review this PR

saurami:
Fixes for PR #2416

saurami:
Add change from PR #2416

saurami:
added Diffie Hellman practice exercise

saurami:
Hi @junedev can you have a look at this ?

I've added a no lint directive to appropriate files, but still getting an error.

I'm unable to clear GitHub runner cache.

Please suggest.

exercises/concept/animal-magic/.meta/exemplar.go
```diff
@@ -1,3 +1,4 @@
+//nolint:gosec // Randomness is part of an exercise where math.rand is more appropriate than crypto.rand
```
junedev, line 1: I found the phrasing a bit hard to follow and the two comment expressions in one line seemed strange. What do you think about the following instead? (If you like it, please apply it to all the places.)

```suggestion
// In the context of the exercise, it is fine to use math.Rand instead of crypto.Rand.
// The following directive will deactivate the respective linter check.
//nolint:gosec
```

exercises/practice/diffie-hellman/diffie_hellman_test.go
```diff
@@ -1,3 +1,4 @@
+ //nolint:gosec // Randomness is part of an exercise where math.rand is more appropriate than crypto.rand
```
junedev, line 1: Why does the test file need this? I don't see an import for math.rand.

exercises/practice/binary-search/binary_search_test.go
```diff
@@ -1,3 +1,4 @@
+//nolint:gosec // Randomness is part of an exercise where math.rand is more appropriate than crypto.rand
```
junedev, line 1: What do you think about this phrasing instead?

```suggestion
//nolint:gosec // In the context of this exercise, it is fine to use math.Rand instead of crypto.Rand.
```

junedev:
@saurami The bug with CI is really strange. I closed and re-opened the PR to start another run. Sometimes that helps.

junedev:
@saurami I found the issue. There is one space character in front of "//nolint" in the file which CI complains about. If you remove the space, everything should be fine. (I found out because the CI report also contained the message that the code was not formatted correctly.)

exercises/practice/diffie-hellman/diffie_hellman_test.go
saurami: this was a mistake, I fixed it

exercises/practice/binary-search/binary_search_test.go
saurami: Thank you! All files have been updated.

saurami:
All changes made based on suggestions

junedev (approved):
Looks all good now.

PR #2441, opened 2022-08-14
Added Additional Hint for Processing Parts of a Date

saurabhmshr:
Hint for `time.Parse` alone is not helpful. Even minor changes can render the time incorrect. This could be frustrating for learners.

### DIFF
--- a/exercises/concept/booking-up-for-beauty/.docs/hints.md
+++ b/exercises/concept/booking-up-for-beauty/.docs/hints.md
@@ -7,7 +7,7 @@
 ## 1. Parse appointment date
 
 - There is a [method][time.parse] for parsing a `string` into a `Time`.
-- Additional information for formatting parts of a date are in the [time format source][time.format.source]
+- Additional information for formatting parts of a date are in the [time package constants][time.package.constants]
 
 ## 2. Check if an appointment has already passed
 
@@ -31,4 +31,4 @@
 [after]: https://golang.org/pkg/time/#Time.After
 [now]: https://golang.org/pkg/time/#Now
 [hour]: https://golang.org/pkg/time/#Time.Hour
-[time.format.source]: https://go.dev/src/time/format.go
+[time.package.constants]: https://pkg.go.dev/time#pkg-constants
--- a/exercises/concept/booking-up-for-beauty/.docs/hints.md
+++ b/exercises/concept/booking-up-for-beauty/.docs/hints.md
@@ -7,6 +7,7 @@
 ## 1. Parse appointment date
 
 - There is a [method][time.parse] for parsing a `string` into a `Time`.
+- Additional information for formatting parts of a date are in the [time format source][time.format.source]
 
 ## 2. Check if an appointment has already passed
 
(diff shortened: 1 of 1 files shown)


exercises/concept/booking-up-for-beauty/.docs/hints.md
```diff
@@ -7,6 +7,7 @@
 ## 1. Parse appointment date
 
 - There is a [method][time.parse] for parsing a `string` into a `Time`.
+- Additional information for formatting parts of a date are in the [time format source][time.format.source]
```
andrerfcsantos, line 10: By source, you mean the comments in the source?

These comments are part of the documentation, as such, they also appear in the official documentation for the package: https://pkg.go.dev/time#pkg-constants

Linking to the docs would be better instead of the source code would be better here.

andrerfcsantos (changes requested):

exercises/concept/booking-up-for-beauty/.docs/hints.md
saurami: Hi @andrerfcsantos , by source I mean the source code of the time format go file.

I've added this link because the comments at the top (lines 55-63) have exact notations for different parts of a date - year, month, day, hour, etc.

This is needed for declaring a layout when a string has to be formatted to a time layout.

andrerfcsantos: Right. What I'm saying is that the same comments are part of the documentation and are in the link I posted above. Linking to the docs is better than linking to source code.

saurami: I agree that linking docs are better than source code, but that documentation lacks enough information to finish the exercise. The source code is slightly better in terms of proving info and examples.

saurami: @andrerfcsantos  I think its better to copy relevant stuff from the comment section and add it to the exercise. Do you think its better to add it to the instruction or hints section ?

junedev (changes requested):

saurabhmshr (commented):
updated link

junedev (approved):

exercises/concept/booking-up-for-beauty/.docs/hints.md
saurabhmshr: Official docs for the time package constants contains a lot more info than whats asked in the exercise. Even minor formatting changes will produce a completely different time output. That's why I think relevant sections from the official documentation should be a part of the instruction. 

I have updated the link, but don't think it is a good idea. 

Learning to read the official docs is a good thing, but it would be appropriate to expect learners to go through the official docs in the next iteration of a learning exercise based on time.

saurabhmshr:
> The link that André provided at the top contains exactly the information you saw in the source code, you only need to scroll down a bit (after the code block). It is generated from the source code that is why the information is the same. Please change the PR to use the link suggested above instead of a link to the source code. Copying some parts is not necessary. Learning to read the official docs is a good thing as part of the learning process on Exercism.

Official docs for the time package constants contains a lot more info than whats asked in the exercise. Even minor formatting changes will produce a completely different time output. That's why I think relevant sections from the official documentation should be a part of the instruction.

I have updated the link, but don't think it is a good idea.

Learning to read the official docs is a good thing, but it would be appropriate to expect learners to go through the official docs in the next iteration of a learning exercise based on time.

PR #2445, opened 2022-08-14
Add test case for non-ascii anagrams

### LINKED ISSUE #1451: Anagram: add test case with UTF8 characters
In Go, strings are made of UTF8 runes. Therefore, it is not appropriate to assume strings as just English letters, unless it is explicitly said so in the description. 

However, in the exercise Anagram, the test cases only cover English letter strings, although "你好" and "好你" should also be treated as anagrams. This encourages some "clever" solutions using `[26]int` to represent a word, which I think is theoretically wrong, as strings are not a collection of English letters.

Perhaps we should add some test cases to this to capture these scenarios. Also some descriptions could be added for corner cases (e.g. how to treat punctuations? etc)

Or we can specify in the description that the input strings only contain English letters. However, in that case, the better type to use would be `[]byte` instead of `string`.

CoderYihaoWang:
Closes #1451.

A test case was added to ensure that the solution works for non-ascii anagrams.

### DIFF
--- a/exercises/practice/anagram/.meta/gen.go
+++ b/exercises/practice/anagram/.meta/gen.go
@@ -34,14 +34,14 @@ var tmpl = `package anagram
 
 {{.Header}}
 
-type AnagramTest struct {
+type anagramTest struct {
 	description string
 	subject     string
 	candidates  []string
 	expected    []string
 }
 
-var testCases = []AnagramTest{ 
+var testCases = []anagramTest{ 
 	{{range .J.findAnagrams}}{
 		description: {{printf "%q"   .Description}},
 		subject: {{printf "%q"       .Input.Subject}},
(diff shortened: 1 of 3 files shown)


junedev:
@eklatzer Could you help out with providing a review here as you are most familiar with the patterns involved?

eklatzer:
I don't know when I find the time for reviewing it, but is possible

eklatzer:
Just checked it on the phone while in the car. 
I don't think `AnagramTest`should be exported. Personally, I would not declare the variable `nonAsciiTestCases` in the `func init`. Everything else looks good on the phone :)

junedev (changes requested):
@CoderYihaoWang Can you make the adjustment that @eklatzer suggested? That means making AnagramTest unexported (lower case) and moving the code that extends the testCases inside the actual test (above line 32).

CoderYihaoWang:
Sorry I got distracted by something else over the last couple of weeks. Here we go, I've made the requested changes :)

junedev (approved):
I reordered things a bit but otherwise all good now!

PR #2447, opened 2022-08-15
added usage of subtests for exercise matrix

eklatzer:
One of #2098

How to we want to handle the unused testcases?
```
	// undefined
	// {"\n\n", // valid?, 3 rows, 0 columns
	// {"",     // valid?, 0 rows, 0 columns
```
For me the first one would be invalid, as these cases are also invalid:
```
	{description: "first row empty", in: "\n3 4\n5 6", ok: false, rows: nil, cols: nil},
	{description: "middle row empty", in: "1 2\n\n5 6", ok: false, rows: nil, cols: nil},
	{description: "last row empty", in: "1 2\n3 4\n", ok: false, rows: nil, cols: nil},
```
The second one should be valid and a empty matrix

What do you think about this one?

### DIFF
--- a/exercises/practice/matrix/.meta/config.json
+++ b/exercises/practice/matrix/.meta/config.json
@@ -17,7 +17,8 @@
     "robphoenix",
     "sebito91",
     "suzaku",
-    "tleen"
+    "tleen",
+    "eklatzer"
   ],
   "files": {
     "solution": [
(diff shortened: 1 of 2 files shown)


andrerfcsantos (approved):
Since the unused tests are also not part of canonical data, I think we can leave them out for now.

PR #2451, opened 2022-08-17
fixed failure of tests

### LINKED ISSUE #2450: gigasecond: tests pass from the start
When starting the gigasecond exercise, tests pass without modifying the code.

```go
// This is a "stub" file.  It's a little start on your solution.
// It's not a complete solution though; you have to write some code.

// Package gigasecond should have a package comment that summarizes what it's about.
// https://golang.org/doc/effective_go.html#commentary
package gigasecond

// import path for the time package from the standard library
import "time"

// AddGigasecond should have a comment documenting it.
func AddGigasecond(t time.Time) time.Time {
	// Write some code here to pass the test suite.
	// Then remove all the stock comments.
	// They're here to help you get started but they only clutter a finished solution.
	// If you leave them in, reviewers may protest!
	return t
}
```

image

eklatzer:
Hey, I think I have found the problem for #2450

Closes #2450

### DIFF
--- a/exercises/practice/armstrong-numbers/armstrong_numbers_test.go
+++ b/exercises/practice/armstrong-numbers/armstrong_numbers_test.go
@@ -6,7 +6,7 @@ import (
 
 func TestArmstrong(t *testing.T) {
 	for _, tc := range testCases {
-		t.Run(tc.description, func(*testing.T) {
+		t.Run(tc.description, func(t *testing.T) {
 			if actual := IsNumber(tc.input); actual != tc.expected {
 				t.Fatalf("IsNumber(%d) = %t, want: %t", tc.input, actual, tc.expected)
 			}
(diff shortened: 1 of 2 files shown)


junedev (approved):
Thanks!
(That was a very strange effect caused by this missing parameter name here imo.)

iHiD:
This is a good example of where we should really have some CI that runs the test-runner against both the stubs and examplars for all exercises and checks the results are sane btw. @ErikSchierboom I wonder if we could add a global file to all tracks that does this and have it as a CI test on all new commits to `main` (so maybe not-blocking? Or maybe a blocking check on PRs if maintainers don't object too heavily!) We have all the dockerfiles in docker hub so it should be relatively straight-forward?

ErikSchierboom:
@iHiD Yes it would be quite straightforward.

PR #2452, opened 2022-08-18
Exercise grep : fix test case which incorrectly fails on nil slice, but expect empty slice

W8CYE:
This commit addresses an incorrectly failing test case.
	grep_test.go:68: Search("may",[-x],[midsummer-night.txt])
	   got: []
	   want: []

Although both slices are zero length, one is initialized, and one is not.
	expected: []string len: 0, cap: 0, []
	actual:   []string len: 0, cap: 0, nil

Added logic to test for zero length for actual and expected if DeepEqual is false.

### DIFF
--- a/exercises/practice/grep/grep_test.go
+++ b/exercises/practice/grep/grep_test.go
@@ -64,7 +64,13 @@ func TestSearch(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.description, func(t *testing.T) {
 			actual := Search(tc.pattern, tc.flags, tc.files)
-			if !reflect.DeepEqual(actual, tc.expected) && !(len(actual) == 0 && len(tc.expected) == 0) {
+
+			// We do not care whether the result is nil or an empty slice.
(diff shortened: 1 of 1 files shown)


exercises/practice/grep/grep_test.go
```diff
@@ -64,7 +64,7 @@ func TestSearch(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.description, func(t *testing.T) {
 			actual := Search(tc.pattern, tc.flags, tc.files)
-			if !reflect.DeepEqual(actual, tc.expected) {
+			if !reflect.DeepEqual(actual, tc.expected) && !(len(actual) == 0 && len(tc.expected) == 0) {
```
IsaacG, line 67: You may want to be more explicit here, eg (assuming the `==` works):

```go
if want == []string && want == actual {
    continue
}

junedev (approved):
Thanks! As suggested by Isaac, I added a commit, to make things a bit more readable.

PR #2460, opened 2022-08-24
added usage of subtest for protein-translation

eklatzer:
One of #2098 

What is `protein_translation_detailed_test.go` for? I have not found anything meaningful in the history. Can we remove it?


exercises/practice/protein-translation/protein_translation_test.go
```diff
@@ -16,152 +16,149 @@ func TestErrorsNotNil(t *testing.T) {
...
-				t.Fatalf("FAIL: Protein translation test: %s\nExpected error: %q\nActual error: %q",
-					test.input, test.errorExpected, err)
+	for _, tc := range codonTestCases {
+		t.Run(tc.input, func(t *testing.T) {
+			got, err := FromCodon(tc.input)
+			switch {
+			case tc.expectedError != nil:
+				if err == nil {
+					t.Fatalf("FromCodon(%q) expected error: %v, got nil", tc.input, tc.expectedError)
+				}
```
junedev, line 103: The new version seems to be be missing the check that if there was an error and we expected an error, that is is actually the right one as we have ErrStop and ErrInvalidBase above (line 100 in the old code).

exercises/practice/protein-translation/protein_translation_test.go
```diff
@@ -16,152 +16,149 @@ func TestErrorsNotNil(t *testing.T) {
...
-				t.Fatalf("FAIL: RNA translation test: %s\nExpected error: %q\nActual error: %q",
-					test.input, test.errorExpected, err)
+	for _, tc := range rnaTestCases {
+		t.Run(tc.input, func(t *testing.T) {
+			got, err := FromRNA(tc.input)
+			switch {
+			case tc.expectedError != nil:
+				if err == nil {
+					t.Fatalf("FromRNA(%q) expected error: %v, got nil", tc.input, tc.expectedError)
+				}
```
junedev, line 155: same comment as above

junedev (changes requested):
I think we lost some logic here that was present in the old tests, see comment.

junedev:
I agree that protein_translation_detailed_test.go should be removed. The normal benchmark is available in the test file as usual and there is no instructions.append.md or anything like that to explain to the student what the detailed benchmarks are about.

junedev (approved):

PR #2463, opened 2022-08-26
added usage of subtests for queen-attack

eklatzer:
One of #2098

### DIFF
--- a/exercises/practice/queen-attack/queen_attack_test.go
+++ b/exercises/practice/queen-attack/queen_attack_test.go
@@ -42,12 +42,12 @@ func TestCanQueenAttackValid(t *testing.T) {
 }
 
 var invalidTestCases = []testCase{
-	{description: "same square", pos1: "b4", pos2: "b4", expected: false},
-	{description: "position off board", pos1: "a8", pos2: "b9", expected: false},
-	{description: "position off board", pos1: "a0", pos2: "b1", expected: false},
-	{description: "position off board", pos1: "g3", pos2: "i5", expected: false},
-	{description: "invalid position", pos1: "here", pos2: "there", expected: false},
-	{description: "empty position", pos1: "", pos2: "", expected: false},
+	{description: "same square", pos1: "b4", pos2: "b4"},
+	{description: "position off board", pos1: "a8", pos2: "b9"},
+	{description: "position off board", pos1: "a0", pos2: "b1"},
+	{description: "position off board", pos1: "g3", pos2: "i5"},
+	{description: "invalid position", pos1: "here", pos2: "there"},
+	{description: "empty position", pos1: "", pos2: ""},
 }
 
 func TestCanQueenAttackInvalid(t *testing.T) {
--- a/exercises/practice/queen-attack/queen_attack_test.go
+++ b/exercises/practice/queen-attack/queen_attack_test.go
@@ -1,6 +1,7 @@
 package queenattack
 
 import (
+	"fmt"
 	"testing"
 )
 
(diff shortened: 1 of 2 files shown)


exercises/practice/queen-attack/queen_attack_test.go
```diff
@@ -1,51 +1,62 @@
...
+	{description: "common diagonals", pos1: "a6", pos2: "b7", expected: true},
+	{description: "common diagonals", pos1: "d1", pos2: "f3", expected: true},
+	{description: "common diagonals", pos1: "f1", pos2: "a6", expected: true},
+	{description: "common diagonals", pos1: "a1", pos2: "h8", expected: true},
+	{description: "common diagonals", pos1: "a8", pos2: "h1", expected: true},
+}
+
+func TestCanQueenAttackValid(t *testing.T) {
+	for _, tc := range validTestCases {
+		t.Run(tc.description, func(t *testing.T) {
```
junedev, line 31: I would recommend to add the position to the description so it is unique, otherwise Go does this which might be confusing for students:
```
    --- PASS: TestCanQueenAttackValid/common_diagonals (0.00s)
    --- PASS: TestCanQueenAttackValid/common_diagonals#01 (0.00s)
    --- PASS: TestCanQueenAttackValid/common_diagonals#02 (0.00s)
    --- PASS: TestCanQueenAttackValid/common_diagonals#03 (0.00s)
    --- PASS: TestCanQueenAttackValid/common_diagonals#04 (0.00s)
    --- PASS: TestCanQueenAttackValid/common_diagonals#05 (0.00s)
```

If our test cases match the problem-spec test cases, we could use the names from there as well:
https://github.com/exercism/problem-specifications/blob/main/exercises/queen-attack/canonical-data.json

exercises/practice/queen-attack/queen_attack_test.go
```diff
@@ -1,51 +1,62 @@
...
+	}
+}
+
+var invalidTestCases = []testCase{
+	{description: "same square", pos1: "b4", pos2: "b4", expected: false},
+	{description: "position off board", pos1: "a8", pos2: "b9", expected: false},
+	{description: "position off board", pos1: "a0", pos2: "b1", expected: false},
+	{description: "position off board", pos1: "g3", pos2: "i5", expected: false},
+	{description: "invalid position", pos1: "here", pos2: "there", expected: false},
+	{description: "empty position", pos1: "", pos2: "", expected: false},
```
junedev, line 49: Since "expected" is always false for invalid cases (and we correctly do not even check it), I don't think it has to be listed in the test table.

exercises/practice/queen-attack/queen_attack_test.go
eklatzer: Added the position to the name

exercises/practice/queen-attack/queen_attack_test.go
```diff
@@ -1,51 +1,63 @@
...
+	{description: "common diagonals", pos1: "a6", pos2: "b7", expected: true},
+	{description: "common diagonals", pos1: "d1", pos2: "f3", expected: true},
+	{description: "common diagonals", pos1: "f1", pos2: "a6", expected: true},
+	{description: "common diagonals", pos1: "a1", pos2: "h8", expected: true},
+	{description: "common diagonals", pos1: "a8", pos2: "h1", expected: true},
+}
+
+func TestCanQueenAttackValid(t *testing.T) {
+	for _, tc := range validTestCases {
+		t.Run(fmt.Sprintf("%s_%s_%s", tc.description, tc.pos1, tc.pos2), func(t *testing.T) {
```
junedev, line 32: Why the underscores? That is usually something only Go adds instead of the spaces, not something you would write into the test name yourself. 🤔

junedev (commented):
Left one more comment.

exercises/practice/queen-attack/queen_attack_test.go
eklatzer: Better like this?

junedev (approved):

PR #2470, opened 2022-08-28
added usage of subtests for twelve_days

eklatzer:
One of #2098


junedev (approved):
Changing the order here was a good idea!

PR #2471, opened 2022-08-28
added usage of subtests for two-fer

eklatzer:
One of #2098 

What is `example_two_fer_test.go` for? Can we delete it?


junedev (approved):
Finally, I have a proper laptop again and I can start reviewing things. This looks all good, also looks like it was overdue that those tests were cleaned up given that this exercise usually has a lot of students.

PR #2472, opened 2022-08-29
changing output sign to =>

katarzynakawala:
Making output messages consistent


katarzynakawala:
Connected to the issue #2202

junedev (commented):
I'm sorry but the issue states that if `fmt.Println` is used, the format with `Output` is correct.
We do this so that a console output is distinct from a REPL result which we mark with `=>`.

PR #2474, opened 2022-08-30
update code block for direct execution

saurabhmshr:
Updated code block with main function so that code executes locally

### DIFF
--- a/exercises/concept/elons-toys/.docs/introduction.md
+++ b/exercises/concept/elons-toys/.docs/introduction.md
@@ -26,7 +26,6 @@ The method on the struct can be called by:
 ```go
 p := Person{Name: "Bronson"}
 fmt.Println(p.Greetings())
-
 // Output: Welcome Bronson !
 ```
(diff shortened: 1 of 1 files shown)


junedev (commented):
You raised a valid issue here. However, adding the main package and main function makes it harder to identify the code that the concept is actually about. Imagine all examples would include this. That would be a lot of extra boilerplate code to read for people.

As an alternative, I would suggest you split the code example into two separate code blocks. One shows the definition part. Then you have a sentence that says something like "Now the method can be called on the struct." and then the second part (line 26 and 27). WDYT?

SR8H (commented):
split single code block into two

SR8H:
Additionally, I've include the package name in the first code block because the line on top of it says:

> You can only define a method with a receiver whose type is defined in the same package as the method.

SR8H (commented):
removed newline before output

junedev (approved):

Issue #2477, opened 2022-09-02
Insufficent Information in Learning Exercise "The Farm"


saurabhmshr:
There's a lot of gap between the "Introduction" and "Instruction" sections.

In the "Introduction" section, only the first paragraph has information about interfaces. Everything else is about errors.
In the code block, `DivideFood` function receives an interface object and there is no information on this. This makes instructions tough to follow.

Unlike other exercises, multiple tasks are associated with the same function `DivideFood`. For this exercise, there should be two tasks at most - one for dividing food and the other for the silly nephew custom error.

junedev:
I am also unhappy with the learning experience that the exercise provides, especially because the resulting code does not look very idiomatic.

But instead of trying even more small improvements, I am planning a more general restructering. Please see https://github.com/exercism/go/issues/2492. For now, I think the exercise is ok-ish to leave it like it is for a bit longer.

Since a new interfaces exercise will also mean an improved/ more detailed introduction for interfaces, I will close this issue in favor of https://github.com/exercism/go/issues/2492.

PR #2480, opened 2022-09-06
Improved Tests for Learning Exercise "The Farm"

saurabhmshr:
Improved case structure in test function `TestDivideFood`

Changes expected error message type from `string` to `error` in test function `TestDivideFoodSillyNephewError`

### DIFF
--- a/exercises/concept/the-farm/the_farm_test.go
+++ b/exercises/concept/the-farm/the_farm_test.go
@@ -18,95 +18,95 @@ var NonScaleError = errors.New("non-scale error")
 
 func TestDivideFood(t *testing.T) {
 	tests := []struct {
-		description             string
-		fodderAmount            WeightFodder
-		cows                    int
-		wantFodderAmount        float64
-		wantFodderErr           error
+		description      string
+		fodderAmount     WeightFodder
+		cows             int
+		wantFodderAmount float64
+		wantFodderErr    error
 	}{
 		{
-			description:             "100 fodder for 10 cows with no error",
-			fodderAmount:            testWeightFodder{fodder: 100, err: nil},
-			cows:                    10,
-			wantFodderAmount:        10,
-			wantFodderErr:           nil,
+			description:      "100 fodder for 10 cows with no error",
+			fodderAmount:     testWeightFodder{fodder: 100, err: nil},
+			cows:             10,
+			wantFodderAmount: 10,
+			wantFodderErr:    nil,
 		},
 		{
-			description:             "10 fodder for 10 cows with no error",
-			fodderAmount:            testWeightFodder{fodder: 10, err: nil},
-			cows:                    10,
-			wantFodderAmount:        1,
-			wantFodderErr:           nil,
+			description:      "10 fodder for 10 cows with no error",
+			fodderAmount:     testWeightFodder{fodder: 10, err: nil},
+			cows:             10,
+			wantFodderAmount: 1,
+			wantFodderErr:    nil,
 		},
 		{
-			description:             "10.5 fodder for 2 cows with no error",
-			fodderAmount:            testWeightFodder{fodder: 10.5, err: nil},
-			cows:                    2,
-			wantFodderAmount:        5.25,
-			wantFodderErr:           nil,
+			description:      "10.5 fodder for 2 cows with no error",
+			fodderAmount:     testWeightFodder{fodder: 10.5, err: nil},
+			cows:             2,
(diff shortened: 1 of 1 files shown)


SR8H (commented):
removed spaces for error variable

SR8H (commented):
applied go fmt to file

exercises/concept/the-farm/the_farm_test.go
```diff
@@ -14,192 +14,141 @@ func (wf testWeightFodder) FodderAmount() (float64, error) {
...
-					test.wantErrMsg,
-				)
-			}
-		})
+		t.Logf("Testing for %s", test.description)
+
+		fodderWeight := testWeightFodder{fodder: 100, err: nil}
+		gotFodderAmount, gotErrMsg := DivideFood(fodderWeight, test.cows)
+
+		if gotFodderAmount != 0 && gotErrMsg != test.wantErrMsg {
```
junedev, line 149: Errors cannot be compared in this way, see https://go.dev/play/p/rRmtE4fDdQ9
Also, we don't expect a normal sentinel error here, we expect a special type of error that the student should write.

exercises/concept/the-farm/the_farm_test.go
```diff
@@ -14,192 +14,141 @@ func (wf testWeightFodder) FodderAmount() (float64, error) {
...
+			description:      "100 fodder for zero cows with no error",
+			fodderAmount:     testWeightFodder{fodder: 100, err: nil},
+			cows:             0,
+			wantFodderAmount: 0,
+			wantFodderErr:    errors.New("division by zero"),
 		},
 	}
+
 	for _, test := range tests {
-		t.Run(test.description, func(t *testing.T) {
```
junedev, line 125: Why should t.Run (a good practice in Go) be removed in favor of a self-written t.Logf?

exercises/concept/the-farm/the_farm_test.go
```diff
@@ -14,192 +14,141 @@ func (wf testWeightFodder) FodderAmount() (float64, error) {
...
 	}{
 		{
-			description:             "100 fodder for 10 cows",
-			weightFodder:            testWeightFodder{fodder: 100, err: nil},
-			weightFodderDescription: "100 fodder, no error",
-			cows:                    10,
-			wantAmount:              10,
-			wantErr:                 nil,
+			description:      "100 fodder for 10 cows with no error",
+			fodderAmount:     testWeightFodder{fodder: 100, err: nil},
```
junedev, line 29: This field contains a fodder, not an amount.

exercises/concept/the-farm/the_farm_test.go
```diff
@@ -14,192 +14,141 @@ func (wf testWeightFodder) FodderAmount() (float64, error) {
 	return wf.fodder, wf.err
 }
 
+var NonScaleError = errors.New("non-scale error")
```
junedev, line 17: This error was intenionally local to the test case as it is not relevant for other tests.

junedev (commented):
I am sorry but this PR does not improve the exercise for the reasons mentioned in the comments. I will close it for now. Please open an issue describing the problem and the improvement you suggest and we can take it from there.

Keep the comment I made here in mind: https://github.com/exercism/go/issues/2477
If there are very important, concrete fixes that you think should be done, feel free to raise the issue. Otherwise, let's tackle this more generally as explained in the other issue.

exercises/concept/the-farm/the_farm_test.go
saurabh-mish: Running subtests with `t.Run` is a good practice. Its also a good practice to log tests and subtests. 

I can change the tests to have subtests only (`t.Run`)

exercises/concept/the-farm/the_farm_test.go
saurabh-mish: The special type of error that the student should write is already a part of `DivideFood`. This method returns the amount of fodder and error message. A combination of these return values are being compared with expected values from test data.

exercises/concept/the-farm/the_farm_test.go
saurabh-mish: Do you think its a good idea to update `wantAmount` and `wantErr` to `fodderAmount` and `fodderError` ?

exercises/concept/the-farm/the_farm_test.go
saurabh-mish: Agree

PR #2481, opened 2022-09-08
Optional Field Information in Task 1  |  Learning Exercise "Census"

SR8H:
The fact that age is an optional field is stated in task 2. This information should be specified in task 1 for clarity.

### DIFF
--- a/exercises/concept/census/.docs/instructions.md
+++ b/exercises/concept/census/.docs/instructions.md
@@ -6,7 +6,7 @@ The program must be able to create a new resident in the system when given a res
 
 ## 1. Create a new resident
 
-When a census worker collects a resident's information, they need to register that resident by entering their name, age, and address into the system.
+When a census worker collects a resident's information, they need to register that resident by entering their name, age, and address into the system. Age is an optional field.
 
 Implement the function `NewResident` that accepts three arguments:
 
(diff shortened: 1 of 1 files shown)


junedev:
@SR8H Could you explain a bit more why you think the age being optional or not is relevant for the task?

junedev (commented):
@SR8H Could you explain a bit more why you think the age being optional or not is relevant for the task?

SR8H:
I think it's part of general information and not tied to task 1. Instructions for task 1 begin after this specifying that age is optional.

Specifying required and optional fields as early as possible is a better idea. Currently this information is in the middle of task 2, which I think could be clearer.

junedev:
The general introduction is in line 1 to 6. The instructions for task 1 start with the headline for task 1 in line 7 and you added the new text below that. The problem I see is that now, students might think the information that the age is optional is relevant for solving task 1 which is not the case. They might try to add some check whether age is there, they might think the function is sometimes less arguments can be supplied in the constructor. It opens the doors for all kinds of confusion.

The tasks are designed so that only the information that is needed for the current section is given. From experience, that works a lot better then trying to cover all the information in the introduction part at the top as we did/do for some exercises in other tracks. This PR removes the information from the place where it is needed (task 2) and moves it to a place where it is not relevant.

In summary, I'm sorry but I don't think this PR improves the instructions. I will close it.

junedev (changes requested):
The general introduction is in line 1 to 6. The instructions for task 1 start with the headline for task 1 in line 7 and you added the new text below that. The problem I see is that now, students might think the information that the age is optional is relevant for solving task 1 which is not the case. They might try to add some check whether age is there, they might think the function is sometimes less arguments can be supplied in the constructor. It opens the doors for all kinds of confusion.

The tasks are designed so that only the information that is needed for the current section is given. From experience, that works a lot better then trying to cover all the information in the introduction part at the top as we did/do for some exercises in other tracks. This PR removes the information from the place where it is needed (task 2) and moves it to a place where it is not relevant.

In summary, I'm sorry but I don't think this PR improves the instructions. I will close it.

Issue #2484, opened 2022-09-10
Investigate why the test for animal magic sometimes fails on windows


junedev:
The test for the "SeedWithTime" function in animal magic exercise sometimes fails on windows.
See https://github.com/exercism/go/runs/8195983378?check_suite_focus=true for an example.

We already tried to increase the timeout but that didn't solve the problem.

Ravalnath:
@junedev Quick question, why does the time.duration in sleep have to be a random number. Can't we just keep it as time.Duration(10) instead?
I think this has some coincidence when the random number generated is 0. I am able to reproduce this here:
[Go Playground](https://go.dev/play/p/GZHeBWe4vms)

junedev:
For reference, the question was already answered in the PR linked above.

junedev:
@Ravalnath With your hint, I think I got it (might be what you meant above). We select a number between 0 and 10 for the waiting time and the seed also applies for that selection. When the waiting time was 0, the seed might be the same as before, leading to another waiting time of 0 and so on. The test fails. I read that the waiting time is particularly important on Windows as the seeding is slow there, that might be the reason we only see it for Windows. And since it only happens for 0 and not the other numbers (1,2,3, ...) explains why it only happens sometimes (1 out of 10).
So the fix would be to add 1 to the waiting time so it is never 0. Would you like to do a new PR?

junedev:
I will close the issue for now and re-open it in case the problem re-appears.

PR #2486, opened 2022-09-14
Issue# 2484 - Fix for animal magic exercise test failure

Ravalnath:
Fix for Issue [#2484](https://github.com/exercism/go/issues/2484)

If time.Now().UnixNano() is called multiple times within the for loop, chances are the returned time will be the same.
This will produce the same random number as the previous iteration resulting in the failure of test.
The fix will call SeedWithTime() only once before the loop and not for each iteration.

### DIFF
--- a/exercises/concept/animal-magic/animal_magic_test.go
+++ b/exercises/concept/animal-magic/animal_magic_test.go
@@ -11,14 +11,13 @@ import (
 func TestSeedWithTime(t *testing.T) {
 	const tests = 100
 	var last int64
+	SeedWithTime()
 	for i := 0; i < tests; i++ {
-		SeedWithTime()
 		got := rand.Int63()
 		if i > 0 && got != last {
 			return
 		}
 		last = got
-		time.Sleep((time.Duration(rand.Intn(10))) * time.Millisecond)
 	}
 	t.Errorf("SeedWithTime always sets the same seed")
 }


junedev (changes requested):
@Ravalnath Thanks for working on this but your PR unfortunately does not help in the context of the exercise. The test case you modified exists to test whether the student actually used a variable seed and not a fixed number or forgot seeding at all. To check this, we need to check the number we get right after the seeding and see whether that number varies.
Additionally, I don't think running this 100 times would all happen within the same nanosecond (and then that only on windows).

If you have another theory why this is not working on windows, feel free to modify the PR. Otherwise, please close it.

Ravalnath:
@junedev Thanks for providing the context. I will close this PR. Will raise another one if I'm able to figure out the underlying issue specific to windows.

Ravalnath:
@junedev Quick question, why does the time.duration in sleep have to be a random number. Can't just keep it as time.Duration(10) instead?

junedev:
I think the random number was choosen to avoid effects from "too regular seed values". I don't know how it works in reality but just imagine the compiler would not use the full seed but calculate `% 100` and then work with that. If we would advance in time in steps of 100 as well, we would always end up with the same seeding.

The finding above is interesting (although I still don't know why it doesn't work). Would be good if we continue the discussion in the issue though. Can you copy the comment with the link there please?

PR #2487, opened 2022-09-14
Issue# 2484 - Added 1 to the waiting time

Ravalnath:
Issue [#2484](https://github.com/exercism/go/issues/2484)

Add 1 to the wait time so that it never becomes 0 thereby eliminating the scenario where the seed is same as before.


junedev (approved):
🤞 I hope this will help. 🙂

PR #2488, opened 2022-09-15
Flatten Array - Improve error message for nil slice

### LINKED ISSUE #2483: Flatten Array - Allow tests to PASS for both nil and empty slices.
Update flatten_array_test.go so that test cases will pass for both nil and empty slices. Some students have had problems where the test FAILS reporting [] != []:

```
=== RUN   TestFlatten/empty
    flatten_array_test.go:12: Flatten([]) = [], want: []

=== RUN   TestFlatten/all_values_in_nested_list_are_null
    flatten_array_test.go:12: Flatten([<nil> [[[<nil>]]] <nil> <nil> [[<nil> <nil>] <nil>] <nil>]) = [], want: []

--- FAIL: TestFlatten (0.00s)
    --- FAIL: TestFlatten/flattens_a_nested_array (0.00s)
    --- FAIL: TestFlatten/all_values_in_nested_list_are_null (0.00s) 

```

I propose that we implement the same fix used in [Exercise grep : fix test case which incorrectly fails on nil slice, but expect empty slice (PR #2452)](https://github.com/exercism/go/pull/2452)

Original 
```go
func TestFlatten(t *testing.T) {
    for _, tc := range testCases {
        t.Run(tc.description, func(t *testing.T) {
            if actual := Flatten(tc.input); !reflect.DeepEqual(actual, tc.expected) {
                t.Errorf("Flatten(%v) = %v, want: %v", tc.input, actual, tc.expected)
            }
        })
    }
}
```

Proposed
```go
func TestFlatten(t *testing.T) {
    for _, tc := range testCases {
        t.Run(tc.description, func(t *testing.T) {
            actual := Flatten(tc.input)

            // We do not care whether the result is nil or an empty slice.
            if len(tc.expected) == 0 && len(actual) == 0 {
             return
            }
            
            if !reflect.DeepEqual(actual, tc.expected) {
                t.Errorf("Flatten(%v) = %v, want: %v", tc.input, actual, tc.expected)
            }
        })
    }
}
```

Alternatively we could change the error message to explain that zero length and nil are similar, but not the same.  

I am happy to submit the PR based on the feedback to this issue.

W8CYE:
For the Flatten Array Exercise - Update `TestFlatten` to allow initialized zero-length slices or un-initialized(nil) slices to represent the empty slice. 

Fixes #2483

### DIFF
--- a/exercises/practice/flatten-array/flatten_array_test.go
+++ b/exercises/practice/flatten-array/flatten_array_test.go
@@ -8,7 +8,14 @@ import (
 func TestFlatten(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.description, func(t *testing.T) {
-			if actual := Flatten(tc.input); !reflect.DeepEqual(actual, tc.expected) {
+			actual := Flatten(tc.input)
+
+			// We do not care whether the result is nil or an empty slice.
+			if len(tc.expected) == 0 && len(actual) == 0 {
+				return
+			}
+
+			if !reflect.DeepEqual(actual, tc.expected) {
 				t.Errorf("Flatten(%v) = %v, want: %v", tc.input, actual, tc.expected)
 			}
 		})
--- a/exercises/practice/flatten-array/flatten_array_test.go
+++ b/exercises/practice/flatten-array/flatten_array_test.go
@@ -8,15 +8,15 @@ import (
 func TestFlatten(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.description, func(t *testing.T) {
-			actual := Flatten(tc.input)
+			if actual := Flatten(tc.input); !reflect.DeepEqual(actual, tc.expected) {
 
-			// We do not care whether the result is nil or an empty slice.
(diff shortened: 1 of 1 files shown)


junedev (commented):
Please see https://github.com/exercism/go/issues/2483#issuecomment-1250215113

W8CYE:
This PR is ready for review. The changes suggested in issue #2483 have been implemented and will be shown in the style below. 

```
=== RUN   TestFlatten/empty
    flatten_array_test.go:13: Flatten([]) = [], want: []
```

junedev:
@W8CYE Thanks for updating the PR. I added a commit to clean up the nesting a bit. Could you take a look at the latest version (https://github.com/exercism/go/pull/2488/files) whether this looks ok to you?

I also switched to using Fatalf because it removes the need for the return and we want students to work there way TDD style, working on getting one failing test to pass at a time. So it makes sense to stop at the first failure.

W8CYE:
@junedev That is much cleaner. Looks great!

PR #2495, opened 2022-09-23
Anagram: Simplify non-ASCII test cases & add go specific instructions - Fixes #2494

### LINKED ISSUE #2494: Anagram: 'non-ascii' character doesn't match with problem description
It seems that recent pull request (https://github.com/exercism/go/pull/2445, https://github.com/exercism/go/issues/1451) add non-ascii test case.

However, problem description says `The target and candidates are words of one or more ASCII alphabetic characters (A-Z and a-z).`

I'm not sure whether 'word may be contain non-ascii characters' is enough. For example, should we handle non-ascii punctations, i.e., `、` or `。`?

W8CYE:
Updated test cases to only test for single-word anagrams to avoid the complexities of phrases. 

Provided go track-specific suplemental instructions: "The Go exercise includes non-ASCII characters in the test cases, but with Go's first-class support for UTF-8, that should not bother you too much."

Wordsmithing is very welcome. 

I would like to see better non-ASCII test data, but that is probably better done by a native speaker of a language with non-ASCII characters. 

Fixes #2494


exercises/practice/anagram/.docs/instructions.append.md
```diff
@@ -0,0 +1,3 @@
+# Implementation
+
+The Go exercise includes non-ASCII characters in the test cases, but with Go's first-class support for UTF-8, that should not bother you too much.
```
andrerfcsantos, line 3:
```suggestion
Unlike other tracks, the Go version of the exercise includes test cases that use UTF-8 (non-ASCII) characters in the strings.
However, with Go's first-class support for UTF-8 given by the `rune` type, that should not bother you too much.
```

Tried to make things clearer by playing with the words a bit. Also mentioned the `rune` type as a soft hint that if the solution is leveraging runes (instead of bytes for instance), it should be ok.

exercises/practice/anagram/anagram_test.go
```diff
@@ -8,10 +8,16 @@ import (
...
-		subject:     "你好，世界",
-		candidates:  []string{"世界，你好", "hello, 世界", "世界, 你好"},
-		expected:    []string{"世界，你好"},
+		description: "detects unicode anagrams",
+		subject:     "ΑΒΓ",
+		candidates:  []string{"ΒΓΑ", "ΒΓΔ", "γβα"},
+		expected:    []string{"ΒΓΑ", "γβα"},
+	},
+	{
+		description: "eliminates misleading unicode anagrams",
```
andrerfcsantos, line 17: Is this test needed? What is it trying to test?

I get that `Α` and `A` are different runes, despite looking similar. But I can't think of a solution that would pass the other tests and fail this one, but maybe I'm overlooking something.

andrerfcsantos (changes requested):
Thanks for working on this, left some suggestions.

exercises/practice/anagram/anagram_test.go
W8CYE: @andrerfcsantos I took these two test cases from old Exercism test cases when Anagram wasn't limited to ASCII. I will remove this one as it provides little value to the exercise.

andrerfcsantos (approved):
Thanks!

PR #2498, opened 2022-09-24
Sieve: Fix empty slice test

### LINKED ISSUE #2497: Sieve: Returning an empty slice passes all tests
The tests for Sieve of Eratosthenes all pass with a function returning an empty slice. To reproduce, run the tests with the code below which will pass:
```go
func Sieve(limit int) []int {
   return []int{}
}
```
The likely cause appears to be the conditional in the TestSieve function:
```go
// use len() to allow either nil or empty list, because they are not equal by DeepEqual
if len(actual) == 0 || len(tc.expected) == 0 {
   return
}
```

mikejoh12:
Previously any code for the Go sieve exercise which returned an empty slice would pass all the tests. This PR modifies the logic so that a test only passes with an empty slice when such an empty slice is expected by the test.

Fixes #2497


junedev (approved):
Great that you found this!

andrerfcsantos (approved):
Great catch!

PR #2503, opened 2022-09-26
Transpose: Fix empty slice test

### LINKED ISSUE #2502: Transpose: An empty slice passes all the tests
The tests for Transpose all pass with a function returning an empty slice. To reproduce, run the tests with the code below which will pass:

```go
func Transpose(input []string) []string {
   return []string{}
}
```
The likely cause appears to be the || conditional in the TestTranspose function:
```go
if len(actual) == 0 || len(tc.expected) == 0 {
   return
}
```

mikejoh12:
For the Go Transpose problem returning an empty slice in this exercise would pass all the tests. This modifies the TestTranspose function so it only passes an empty slice when such a slice is expected by the test.

Fixes #2502


andrerfcsantos (approved):
Well spotted, thanks!

PR #2515, opened 2022-10-13
Deprecate beer-song and add bottle-song

### LINKED ISSUE #2505: Beer song: deprecate exercise and implement Bottle Song
As per https://github.com/exercism/problem-specifications/pull/2097, we should deprecate the `beer-song` exercise, and implement its replacement, [bottle-song](https://github.com/exercism/problem-specifications/tree/main/exercises/bottle-song).

## Contributing to this issue

Anyone is welcome to contribute to this issue. If you are interested in working on this, just post a comment below saying you want to work on this. Keep in mind that this task requires some exercism knowledge and might not be a good first issue if you are new to contributing.

Here are some docs related to this task:

- [Exercism Docs: Deprecating an exercise](https://exercism.org/docs/building/tracks/deprecated-exercises)
- [Exercism Docs: Practice Exercises](https://exercism.org/docs/building/tracks/practice-exercises)

PakkuDon:
Closes #2505 

As discussed in https://github.com/exercism/problem-specifications/issues/2066 we would like to replace the beer-song exercise with [bottle-song](https://github.com/exercism/problem-specifications/tree/main/exercises/bottle-song). This PR deprecates the beer-song exercise and adds a new bottle-song exercise

### DIFF
--- a/exercises/practice/bottle-song/.meta/config.json
+++ b/exercises/practice/bottle-song/.meta/config.json
@@ -1,5 +1,5 @@
 {
-  "authors": [],
+  "authors": ["PakkuDon"],
   "files": {
     "solution": [
       "bottle_song.go"
--- a/exercises/practice/bottle-song/.meta/config.json
+++ b/exercises/practice/bottle-song/.meta/config.json
@@ -1,0 +1,17 @@
+{
+  "authors": [],
+  "files": {
+    "solution": [
+      "bottle_song.go"
+    ],
+    "test": [
+      "bottle_song_test.go"
+    ],
+    "example": [
+      ".meta/example.go"
+    ]
+  },
+  "blurb": "Produce the lyrics to the popular children's repetitive song: Ten Green Bottles.",
+  "source": "Wikipedia",
+  "source_url": "https://en.wikipedia.org/wiki/Ten_Green_Bottles"
+}
--- a/exercises/practice/bottle-song/.meta/config.json
+++ b/exercises/practice/bottle-song/.meta/config.json
@@ -1,5 +1,7 @@
 {
-  "authors": ["PakkuDon"],
+  "authors": [
+    "PakkuDon"
+  ],
   "files": {
     "solution": [
       "bottle_song.go"
--- a/exercises/practice/bottle-song/.meta/example.go
+++ b/exercises/practice/bottle-song/.meta/example.go
@@ -48,10 +48,14 @@ func verse(n int) []string {
 		}
 	default:
 		return []string{
-			fmt.Sprintf("%s green bottles hanging on the wall,", strings.Title(numberToWord[n])),
-			fmt.Sprintf("%s green bottles hanging on the wall,", strings.Title(numberToWord[n])),
+			fmt.Sprintf("%s green bottles hanging on the wall,", toTitleCase(numberToWord[n])),
+			fmt.Sprintf("%s green bottles hanging on the wall,", toTitleCase(numberToWord[n])),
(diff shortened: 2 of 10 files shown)


exercises/practice/bottle-song/.meta/config.json
```diff
@@ -0,0 +1,17 @@
+{
+  "authors": [],
```
andrerfcsantos, line 2: Feel free to add yourself as the author of the exercise!

exercises/practice/bottle-song/bottle_song_test.go
```diff
@@ -0,0 +1,28 @@
...
+		})
+	}
+}
+
+func equal(a, b []string) bool {
+	if len(b) != len(a) {
+		return false
+	}
+
+	sort.Strings(a)
```
andrerfcsantos, line 25: In this exercise the order of the verses matter, so we shouldn't order the verses to see that they are equal. Checking if the slices are equal as is should be enough. Also, in general we want to consider a nil slice equal to an empty non-nil slice, so if both slices have length 0, consider them equal. We have something similar in other exercises.

andrerfcsantos (changes requested):
Thanks for this and for even making a generator!
Let some comments for now. In general the PR looks good, but I'll do a more through review later.

exercises/practice/bottle-song/bottle_song_test.go
PakkuDon: Good catch! I blindly copy-pasted this from another test 😅 
I've removed the sorts and added another if statement in

exercises/practice/bottle-song/.meta/example.go
```diff
@@ -0,0 +1,61 @@
...
+			fmt.Sprintf("%s green bottles hanging on the wall,", toTitleCase(numberToWord[n])),
+			"And if one green bottle should accidentally fall,",
+			fmt.Sprintf("There'll be %s green bottles hanging on the wall.", numberToWord[n-1]),
+		}
+	}
+}
+
+func toTitleCase(s string) string {
+	return strings.ToUpper(s[0:1]) + strings.ToLower(s[1:])
+}
```
PakkuDon, line 61: I wasn't able to use `strings.Title` as it has been deprecated in Go 1.18 ([Build for reference](https://github.com/exercism/go/actions/runs/3250339595/jobs/5333890548))

Is there a better way to do this? The build suggests using `golang.org/x/text/cases` instead but that would involve installing an external dependency

andrerfcsantos: Temporarily, feel free to copy the `strings.Title` and `strings.isSeparator` function directly: https://github.com/exercism/go/blob/43950ad7c2f1fc054f7487943f6f47a1228ac8af/exercises/practice/accumulate/accumulate_test.go#L10-L55

In the meantime, we are adding support to external packages to the track, and `golang.org/x/text` is one of the dependencies we'll support right out of the bat: https://github.com/exercism/go-test-runner/pull/84

When we have this, we'll replace all the copies of `strings.Title` we have.

PakkuDon: Thanks! I've added that to `bottle_song_test.go` and updated the example solution

andrerfcsantos (approved):
Looks good!

Made some small adjustments to the configs. The difficulty of the exercise seems to not have changed, this was mostly a renaming, so I kept the difficulty for the old exercise. I also left the topics of the old exercise in the config for the new one. This is so that it helps our future work in converting the topics to the `practices` and `prerequisites` fields.

PR #2516, opened 2022-10-15
Update test skip message (and comments) in robot_name_test.go

sbromberger:
Clarify that "sub-optimal" is with respect to time (as opposed to memory).

The original text implies that solutions that do not preallocate 676k names are somehow less optimal than solutions that trade off memory for collision avoidance. This clarifies that this optimality does not apply to the solutions themselves but to their running time.


exercises/practice/robot-name/robot_name_test.go
```diff
@@ -63,8 +63,8 @@ func TestResetName(t *testing.T) {
 	}
 }
 
-// Test 1000 names are unique - this should run reasonably quickly even with a sub-optimal solution
-// (e.g. pick a random name, then pick a new name if it's been seen before)
+// Test 1000 names are unique - this should run reasonably quickly even with a solution that is not optimized
+// with respect to running time. (e.g. pick a random name, then pick a new name if it's been seen before)
```
junedev, line 67: Thanks for the improvement. I am not a native speaker but according to the dictonary it is called "run time" not "running time". Same applies below.

sbromberger: In my field, "runtime" refers to the general state of the system during a program's execution: "runtime libraries", "runtime memory allocation". I deliberately used "running time" to emphasize the fact that we are talking about the classic CPU vs memory tradeoff here since "running time"'s meaning is unambiguously tied to the time it takes for a program to execute (whether that time is measured in wall time or CPU time is immaterial here).

iHiD: I'd personally suggest changing:
> that is not optimised with respect to running time.

to
> that is not particularly fast.

and getting rid of the stuff in brackets altogether. I don't think the example is needed and just gives a clue to the student about what is optimised and what is not.

glennj:
"execution time" was my first thought.

exercises/practice/robot-name/robot_name_test.go
kotp: I have been curious about "runtime" and how different fields use this, or change this.  What is your field?

kotp: Or, rather, which field?  As I see you are involved in more than one.

edgerunner: AFAIK "runtime" refers to the base environment/library/interpreter your code runs on, while "run time" is the actual time **when** your code is running, not its duration.

I'd probably stick to speed-related terms on this one.

junedev (approved):
I applied Jeremys suggestion.

PR #2520, opened 2022-10-20
removed unnecesary line of example code in Meteorology exercise

matusbielik:
I don't really understand why the line of code is there, maybe because of historical reasons?
To me, it makes no sense, the variable is not used anywhere. I've found it confusing when doing the exercise, so I'm proposing to delete it.


andrerfcsantos (approved):
Thanks for this!

> I don't really understand why the line of code is there, maybe because of historical reasons?

Yeah, I'm pretty sure this is what happened. If I recall correctly, the original example only included one temperature unit, then it was changed to include both units and we forgot to remove this.

PR #2523, opened 2022-10-24
Update function signatures provided in matrix exercise

garrettwgg:
A minor change to the test file allows all methods to use a value receiver. This should allow for simpler solutions since students don't need to deal with pointers within their functions.

It is probably possible to change BenchmarkNew to use either a pointer or value (as all other tests are capable of this) but I didn't want to screw anything up so I went with the simplest change.

### DIFF
--- a/exercises/practice/matrix/.meta/example.go
+++ b/exercises/practice/matrix/.meta/example.go
@@ -11,7 +11,7 @@ import (
 
 type Matrix [][]int
 
-func New(s string) (*Matrix, error) {
+func New(s string) (Matrix, error) {
 	var err error
 	lines := strings.Split(s, "\n")
 	m := make(Matrix, len(lines))
@@ -28,39 +28,37 @@ func New(s string) (*Matrix, error) {
 		}
 		m[i] = row
 	}
-	return &m, nil
+	return m, nil
 }
 
-func (m *Matrix) Set(row, col, val int) (ok bool) {
-	mm := *m
-	if row < 0 || row >= len(mm) || col < 0 {
+func (m Matrix) Set(row, col, val int) (ok bool) {
+	if row < 0 || row >= len(m) || col < 0 {
 		return false
 	}
-	if cols := len(mm[0]); col >= cols {
+	if cols := len(m[0]); col >= cols {
 		return false
 	}
-	mm[row][col] = val
+	m[row][col] = val
 	return true
 }
 
-func (m *Matrix) Rows() [][]int {
-	r := make([][]int, len(*m))
-	for i, mr := range *m {
+func (m Matrix) Rows() [][]int {
+	r := make([][]int, len(m))
+	for i, mr := range m {
 		r[i] = append([]int{}, mr...)
 	}
 	return r
 }
 
-func (m *Matrix) Cols() [][]int {
-	mm := *m
-	if len(mm) == 0 {
+func (m Matrix) Cols() [][]int {
+	if len(m) == 0 {
 		return nil
 	}
-	c := make([][]int, len(mm[0]))
+	c := make([][]int, len(m[0]))
 	for i := range c {
-		col := make([]int, len(mm))
+		col := make([]int, len(m))
 		for j := range col {
-			col[j] = mm[j][i]
+			col[j] = m[j][i]
 		}
 		c[i] = col
 	}
--- a/exercises/practice/matrix/matrix.go
+++ b/exercises/practice/matrix/matrix.go
@@ -2,19 +2,19 @@ package matrix
 
 // Define the Matrix type here.
 
-func New(s string) (*Matrix, error) {
+func New(s string) (Matrix, error) {
 	panic("Please implement the New function")
 }
 
 // Cols and Rows must return the results without affecting the matrix.
-func (m *Matrix) Cols() [][]int {
+func (m Matrix) Cols() [][]int {
 	panic("Please implement the Cols function")
 }
 
-func (m *Matrix) Rows() [][]int {
+func (m Matrix) Rows() [][]int {
 	panic("Please implement the Rows function")
 }
 
-func (m *Matrix) Set(row, col, val int) bool {
+func (m Matrix) Set(row, col, val int) bool {
 	panic("Please implement the Set function")
 }
--- a/exercises/practice/matrix/matrix_test.go
+++ b/exercises/practice/matrix/matrix_test.go
@@ -274,7 +274,7 @@ func BenchmarkNew(b *testing.B) {
 	if testing.Short() {
 		b.Skip("skipping benchmark in short mode.")
 	}
-	var matrix *Matrix
+	var matrix Matrix
 	for i := 0; i < b.N; i++ {
 		var err error
 		matrix, err = New("1 2 3 10 11\n4 5 6 11 12\n7 8 9 12 13\n 8 7 6 13 14")


garrettwgg:
OK I think I fixed it.

Getting everything set up is a bit overwhelming. Do I need to be on Linux for things to work? The development setup instructions say:

> Test your clone by cding to the go directory and typing `bin/fetch-golangci-lint` and then `bin/run-tests`. You should see tests pass for all exercises.

Those appear to be bash scripts. I tried running them on WSL Ubuntu but it didn't seem to work.

andrerfcsantos:
> Getting everything set up is a bit overwhelming.

Is there anything in particular you think we can improve on the instructions?

> Do I need to be on Linux for things to work?

Ubuntu or Mac, yes, although WSL should also work.

> Those appear to be bash scripts. I tried running them on WSL Ubuntu but it didn't seem to work.

What failed?

> The development setup instructions say

For the purposes of passing the tests, since we are changing just one exercise, probably the easier thing to do would be to temporarily copy the exemplar solution to the stub file and run the tests with `go test` inside the exercise directory. Those instructions replicate what the CI does, which is run all the tests and linters for all exercises, which might be overkill here.

Since this is your first PR in the repo, we have to approve CI runs, but in future PRs you'll have access directly to the results of CI runs so you won't need to run the development setup locally if you don't want.

andrerfcsantos (approved):

garrettwgg:
I don't think it's the fault of the docs, its just my first time contributing to a project. There are a lot of walls of text to read and some new things to figure out, but that's just the nature of learning something new. Also its a little nerve wracking to submit a pull request for the first time. 😄 

I did indeed copy the contents of the example file to my matrix directory and run the test before submitting.

Here is what I get when trying to set things up. I'm probably just doing something wrong.

```
garrett@GG-RYZEN:/mnt/d/onedrive/programming/exercism/go$ bin/fetch-golangci-lint
-bash: bin/fetch-golangci-lint: /bin/bash^M: bad interpreter: No such file or directory

garrett@GG-RYZEN:/mnt/d/onedrive/programming/exercism/go$ bash bin/fetch-golangci-lint
bin/fetch-golangci-lint: line 2: $'\r': command not found
'olangci/golangci-lint info checking GitHub for tag 'v1.49.0
' - use 'latest' or see https://github.com/golangci/golangci-lint/releases for details

garrett@GG-RYZEN:/mnt/d/onedrive/programming/exercism/go$ bin/run-tests
-bash: bin/run-tests: /bin/bash^M: bad interpreter: No such file or directory

garrett@GG-RYZEN:/mnt/d/onedrive/programming/exercism/go$ bash bin/run-tests
bin/run-tests: line 2: $'\r': command not found
bin/run-tests: line 9: $'\r': command not found
bin/run-tests: line 12: $'\r': command not found
bin/run-tests: line 18: $'\r': command not found
bin/run-tests: line 23: $'\r': command not found
bin/run-tests: line 26: $'\r': command not found
bin/run-tests: line 27: syntax error near unexpected token `$'do\r''
'in/run-tests: line 27: `for parentdir in concept practice; do
garrett@GG-RYZEN:/mnt/d/onedrive/programming/exercism/go$
```

andrerfcsantos:
Ah, the problem there is that in Windows newlines are represented by `\r\n`(aka CRLF - Carriage return, line feed), but in Linux and Mac, they are just represented by `\n` (aka LF - Line feed). Since you probably cloned the repo in Windows, git on windows converted the newlines on the files to use `\r\n`. But when you give such files to a linux system like wsl, bash doesn't understand the `\r` character, so you are seeing that error. To fix that error there are some alternatives:

- Have a version of the repo cloned inside WSL. Clone it using git inside WSL - this will make git clone all files with `\n` newlines, just like bash wants them.
- If you still just want to have 1 copy of the repo in Windows, you can manually remove all the `\r` characters in all repositories in Windows. This can be done in several ways, but might mess with the git diff of the files in Windows.
  - You editor probably lets you choose if newlines should be `\r\n` or `\n`. Choose `\n` and save the file.
  - In the editor, you can do a "Find and Replace" to remove all the `\r`. You might need to enable regex search for this to work.
  - If instead of doing it in the editor you prefer the command line, you can use `sed` in wsl to remove all `\r` instances of all files. Something like `sed -i 's/\r$//g' *` in the scripts directory should work.

mblayman:
Hi! I thought I'd try my hand at updating my matrix problem because I had never tried an updated exercise before. I could be doing something wrong as I'm still pretty new to Go, but it seems like the switch from a pointer to a `Matrix` value as the return type broke the test code.

When I try to run the tests now, I get a compile failure. While I understand that could be expected, I'm uncertain that my solution should be any less acceptable.

```
➜  matrix go test
# matrix [matrix.test]
./matrix_test.go:165:16: invalid operation: got == nil (mismatched types Matrix and untyped nil)
./matrix_test.go:285:15: invalid operation: matrix == nil (mismatched types Matrix and untyped nil)
```

I may be hitting this problem because I defined my `Matrix` as:

```go
type Matrix struct {
	rows [][]int
}
```
while the example code uses a type alias approach instead. Perhaps because I have a struct, the nil value is not acceptable while an array would allow for nil.

Since the `Matrix` value is now the return type, I think it's not possible for `got` or `matrix` to be compared to `nil` for a struct-based approach. The compiler appears to be trying to indicate that the comparison is not valid.

Am I missing something or is this truly an error with the tests now?

Considering that the problem asks the user to define the Matrix type, this change seems overly restricting to me for the solution. Is it wrong to use a struct-based solution?

Also, I apologize if I'm using GitHub issues for this project incorrectly. I hope I'm on track in discussing this from the meta perspective of what is an acceptable solution for defining `Matrix` instead of focusing too narrowly on how I tried to solve the problem.

andrerfcsantos:
@mblayman Thanks for the report. I'll take a closer look later.

Using Github issues for this kind of issues is fine, but alternatively I'd encourage you to use our [new forum](https://forum.exercism.org/) in the future. The advantages of using the forum is that more people can see see issue and comment on it.

andrerfcsantos:
@mblayman Had a chance to take a look at this. Here's my proposal:

- Revert `New` to return a `*Matrix` and adjust the `BenchmarkNew` benchmark so `matrix` is a `*Matrix`
- Keep the method stubs taking a value receiver, but add a note in a comment stating that a pointer receiver can also be used for all the methods.
- (Optional) Change `BenchmarkNew` so `var *matrix Matrix` is not declared explicitly but implicitly with the `:=` like happens in the other tests/benchmarks. The variable is only declared like this because it is needed after the benchmark loop. But I think we can check if the matrix is `nil` at the beginning before the benchmarking loop, in which case both the code outside the loop and the code inside it can have `matrix, err := New(...)`.

Would you like to create a PR for this?

mblayman:
I decided to switch my solution to use `type Matrix [][]int`. The solution is a bit simpler and cleaner, so I think it might be worth leaving the exercise as is now. Thanks for looking into the issue! I'm ok having the piece of mind that wasn't actually doing anything wrong.

tapank:
I started with declaring a struct (just like [mblayman](https://github.com/mblayman)'s first comment above) and could not get it to compile due to the `nil` assignment in the test. Ended up in this thread after looking for solutions elsewhere. Of course, after changing Matrix type declaration to `type Matrix [][]int` as suggested in the previous comment, it all works fine.

The reason I comment here is: many beginner programmers will not be able to figure this out, and are likely to abandon the exercise. I think this exercise needs to be tweaked to support both approaches, i.e. whether Matrix is a `struct` or a `[][]int` type.

andrerfcsantos:
@mblayman @tapank There was some discussion about this recently on the forum. You can see the discussion [here](http://forum.exercism.org/t/go-track-matrix-problem/3655). Double check if [my post](http://forum.exercism.org/t/go-track-matrix-problem/3655/5) makes sense to you and if so feel free to create a PR.

If you have anything more to add, please continue the discussion on the forum for better visibility.

Issue #1335, opened 2020-03-27
sublist : add missing type in downloaded file


Anmol-Sharma:
There is a missing definition for the data type **Relation** of the field `expected` in file cases_test.go. 

This issue is almost a clone of the same problem raised in #1301

PR #2526, opened 2022-10-28
sublist: Fixes sublist exercise by adding missing Relation type.

sebast26:
There is **still** a missing definition for the data type **Relation** of the field expected in file cases_test.go. The exercise in the current state is not solvable.

The issue was not solved by #1335 neither by #1893.

### DIFF
--- a/exercises/practice/sublist/.meta/config.json
+++ b/exercises/practice/sublist/.meta/config.json
@@ -23,7 +23,8 @@
       ".meta/example.go"
     ],
     "editor": [
-      "cases_test.go"
+      "cases_test.go",
+      "relations.go"
     ]
   }
 }
--- a/exercises/practice/sublist/.meta/example.go
+++ b/exercises/practice/sublist/.meta/example.go
@@ -2,17 +2,6 @@ package sublist
 
 import "reflect"
 
-// Relation is the comparison between lists
-type Relation string
-
-// Possible relations
-const (
-	RelationEqual     Relation = "equal"
-	RelationSublist   Relation = "sublist"
-	RelationSuperlist Relation = "superlist"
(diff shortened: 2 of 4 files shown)


andrerfcsantos (changes requested):
Well spotted. However, since the tests implicitly expect specific string values for the `Relation` constants, I think we also should provide:

```go
// Relation is the comparison between lists
type Relation string

// Possible relations
const (
	RelationEqual     Relation = "equal"
	RelationSublist   Relation = "sublist"
	RelationSuperlist Relation = "superlist"
	RelationUnequal   Relation = "unequal"
)
```

How I suggest we do this:

- Create a new file `relations.go` with these contents and make it belong to the same package of the exercise.
- Add `relations.go` to the `files.editor` key of [.meta/config](https://github.com/exercism/go/blob/main/exercises/practice/sublist/.meta/config.json). This will make this file appear as read-only in the online editor.
- Remove the `Relations` definition from [.meta/example.go](https://github.com/exercism/go/blob/main/exercises/practice/sublist/.meta/example.go). Since this relation will be defined in `relations.go`, it doesn't need to be part of the solution file itself.
- Add a quick comment in `sublist.go` saying that the `Relation` type is defined in the `relations.go` file.

Ideally, we would remove the need for `Relation` to be `type Relation string` and we could leave up to the people solving the exercise to choose the type they want and the values they want for the constants. However, this would require changing the test generator for this exercise, which I think can be done in a separate issue/PR.

sebast26:
Hey @andrerfcsantos thanks for tips! I followed them in my new commit. Please review :-)

andrerfcsantos (approved):
Thanks!

PR #2527, opened 2022-10-30
fix: add type info for item parameter

wgwz:


andrerfcsantos:
Thanks for the interest in contributing.

What's the reason behind this change? All variables in the current stub are typed - Go allows to group together consecutive variables with the same type and it's what's happening here.

wgwz:
@andrerfcsantos i didn't know that about go. perhaps i missed it somewhere in the track, but this would also be confusing for other people new to the language, if it wasn't already explicitly introduced.

junedev:
@wgwz The exercise gross store has the concept "functions" as prequisite (exercise "lasagna master"). The introduction document of "functions" explains the syntax used here. A test case can't really cover this so there is no explicit task to apply this. For some things, we need to rely on people reading the text.

PR #2529, opened 2022-10-31
Add missing tests for interest_is_interesting

mintyknight:
Since the wrong logic could clear the tests, I figured that some tests are missing. 
By adding two test scenarios, they covers 

> 1. balance is already the exact same as the target
> 2. result balance would be the exact same as the target

(Following code should have failed the tests, but currently passing. The correct logic should be using < instead of <=)
image


andrerfcsantos (approved):

andrerfcsantos:
Thanks for this!

PR #2531, opened 2022-10-31
the-farm: check "negative fodder" error precedence

### LINKED ISSUE #2530: The Farm: Tests do not cover a condition of the exercise
The exercise reads _"Note that if both a "negative fodder" error and a `SillyNephewError` could be returned, the "negative fodder" error takes precedence."_ However, this is not being tested as shown by solutions, which pass the current version whilst checking for the `SillyNephewError` condition before they check for the negative fodder condition.

luxifr:
fixes #2530

### DIFF
--- a/exercises/concept/the-farm/the_farm_test.go
+++ b/exercises/concept/the-farm/the_farm_test.go
@@ -203,3 +203,40 @@ func TestDivideFoodSillyNephewError(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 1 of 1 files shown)


andrerfcsantos (approved):
Thanks!

CI was failing, but I did a commit fixing it.

luxifr:
> Thanks!
> 
> CI was failing, but I did a commit fixing it.

Thanks for fixing that - I was just about to do the same but you beat me to it :D

PR #2532, opened 2022-11-01
Update tree_building.go

easyyoungg:


andrerfcsantos:
This file contains the stub provided to the student when they start an exercise. Since it is the starter file, it shouldn't obviously contain the solution already. Maybe you meant to commit this somewhere else?

PR #2544, opened 2022-11-12
Remove whitespace in front of exclamation point

UnclassedPenguin:


junedev (commented):
Did you double check that the same text does not appear in other files as well (as mentioned in the auto-generated comment?

PR #2549, opened 2022-11-22
elons-toys: add test for elons_toys to cover track with possible remainder

### LINKED ISSUE #2548: Go track exercise elons-toys passes tests in cases where it should not
When implementing a solution for the mentioned exercise, I stumbled upon the comments of the most starred solution.

There, it was pointed out by the author that some community solutions pass even though they should fail. This happens when the result of a division is assigned to an `int` variable even though it has a remainder.

see https://exercism.org/tracks/go/exercises/elons-toys/solutions/mu8086 for the comments and a link to a playground provided by the author.

do-nat:
The test covers the case when the implementation of `CanFinish` is done in multiple steps where in one of them the result which should be a `float` is assigned to an `int`.

This PR fixes  #2548

example of incorrect solution:

```
func (car Car) CanFinish(trackDistance int) bool {
	neededCycles := trackDistance / car.speed
	return car.battery-(car.batteryDrain*neededCycles) >= 0
}
```
Example of correct solution:
```
func (car *Car) CanFinish(trackDistance int) bool {
	return car.battery-(car.batteryDrain*trackDistance/car.speed) >= 0
}
```

### DIFF
--- a/exercises/concept/elons-toys/elons_toys_test.go
+++ b/exercises/concept/elons-toys/elons_toys_test.go
@@ -140,6 +140,16 @@ func TestCanFinish(t *testing.T) {
 			expected:      true,
 		},
 		{
+			name: "Car has 50% battery. Car cannot finish the race",
+			car: Car{
+				speed:        3,
+				batteryDrain: 3,
+				battery:      50,
+			},
+			trackDistance: 51,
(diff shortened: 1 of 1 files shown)


junedev (approved):
Good test case!

PR #2550, opened 2022-11-22
remove re-initialization

Salamandastron1:


junedev:
Sorry but there is no re-initialization as there are two different variable names.

PR #2551, opened 2022-11-25
Fix grammar

Nick-Wittenberg:
The usage of "it" de-personifies the president, and we can't use "he" or "she" as we don't know the president's gender. The correct grammar is to use "they" in this instance.


junedev (approved):
Thanks for the fix.

PR #2563, opened 2022-12-11
elons-toys: fix test to cover the edge case it should cover

do-nat:
In my previous PR #2549 by accident I chose a trackdistance that is divisible by 3, which makes this test case useless.
With those fixes, the test case does actually fix the issue #2548.


junedev (approved):
🙈 When I checked this PR I mistakenly used the 50 as track distance and thought it works.

PR #2570, opened 2022-12-20
Pangram approaches

bobahop:
Approaches for Pangram


exercises/practice/pangram/.approaches/introduction.md
```diff
@@ -0,0 +1,67 @@
...
+	return alphamask&phrasemask == alphamask
+}
+```
+
+For more information, check the [Bit field approach][approach-bitfield].
+
+## Which approach to use?
+
+The bit field approach benchmarked the fastest.
+To compare performance of the approaches, check the [Performance article][article-performance].
```
junedev, line 63:
```suggestion
To compare performance of the approaches, check the [Performance article][article-performance].

The `strings.ContainsRune()` runes approach is much more readable so if the code is not extremely performance critical, then opting for a readable, easy-to-change solution is recommended.
```

junedev (approved):

PR #2572, opened 2022-12-20
Hamming approaches

bobahop:
Approaches for Hamming.

### DIFF
--- a/exercises/practice/hamming/.approaches/introduction.md
+++ b/exercises/practice/hamming/.approaches/introduction.md
@@ -19,16 +19,15 @@ import "errors"
 
 // Distance accepts two strands and returns their hamming distance.
 func Distance(strand1, strand2 string) (distance int, e error) {
-
 	if len(strand1) != len(strand2) {
-		return distance, errors.New("strands must be of equal length")
+		return 0, errors.New("strands must be of equal length")
 	}
 	for i := 0; i < len(strand1); i++ {
 		if strand1[i] != strand2[i] {
 			distance++
 		}
 	}
-	return distance, e
+	return distance, nil
 }
 ```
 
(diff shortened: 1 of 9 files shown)


exercises/practice/hamming/.approaches/introduction.md
```diff
@@ -0,0 +1,79 @@
...
+	if len(strand1) != len(strand2) {
+		return distance, errors.New("strands must be of equal length")
+	}
+	for i := 0; i < len(strand1); i++ {
+		if strand1[i] != strand2[i] {
+			distance++
+		}
+	}
+	return distance, e
+}
```
junedev, line 32: Hm, this does not look like idiomatic Go code to me ...
* Named returns should only be used if necessary, here they are not. Named returned have the potential that you accidentially assign values to then (especially when working with defer). Also `return 0, errors.New...` makes it clearer that there is no actual distance value returned in the error case. Finally, starting with `distance := 0` before the for-loop makes it much easier to read then making the reader go "Wait, what? Where does distance come from? Ah, there it is. Wait, but what was the zero value again? Ah yes, Go always itializes with a proper value so it is 0, ok, all good."
* I have never seen `e` being used for the error variable, its usually always `err` in Go.
* There should be no empty line below the function signature. Even if gofmt does not enfore this, it is common practice in Go code. Also gofumpt enforces this which is quite popular.
* I am missing an empty line between the if check and the for loop. That would make it clearer where the error check ends and where the actual code begins.

(Potentially, those issues are also present in other solutions you authored and I only noticed here. Not sure.)

junedev (changes requested):
Could you have another look at the solutions here? See comment.

exercises/practice/hamming/.approaches/introduction.md
bobahop: - I don't know that named returns should only be used if necessary. I quite like them. More [here](Ihttps://yourbasic.org/golang/named-return-values-parameters/).

- Yes, I like `err` better`.

- I have no problem getting rid of the empty line below the function signature.

- I also prefer having the empty line between the if and for. Not sure how that slipped by.

Will edit those formatting things. Thanks for all of your attention today!

junedev (approved):

PR #2574, opened 2022-12-20
RNA Transcription approaches

bobahop:
Approaches for RNA Transcription

### DIFF
--- a/exercises/practice/rna-transcription/.approaches/introduction.md
+++ b/exercises/practice/rna-transcription/.approaches/introduction.md
@@ -1,0 +1,109 @@
+# Introduction
+
+There are at least three idomatic approaches to solving RNA Transcription in Go.
+You can use the `strings.Map()` function.
+Or you can use the `map` object to look up the letters.
+Or a `switch` statement can be used to translate the letters.
+
+## General guidance
+
+The key to solving RNA Transcription is TODO
+
+## Approach: `strings.Map()` function
+
+```go
+// Package strand is a small library for returning the RNA complement of a DNA strand.
+package strand
+
+import "strings"
+
+// toRNA returns the RNA complement of the input DNA strand.
+func ToRNA(dna string) string {
+	return strings.Map(func(r rune) rune {
+		switch r {
+		case 'G':
+			return 'C'
+		case 'C':
+			return 'G'
+		case 'T':
+			return 'A'
+		case 'A':
+			return 'U'
+		}
+		return r
+	}, dna)
+}
+```
+
+For more information, check the [`strings.Map()` approach][approach-strings-map].
+
+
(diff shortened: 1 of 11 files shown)


exercises/practice/rna-transcription/.approaches/introduction.md
```diff
@@ -0,0 +1,108 @@
...
+		case 'C':
+			return 'G'
+		case 'T':
+			return 'A'
+		case 'A':
+			return 'U'
+		}
+		return r
+	}, dna)
+}
```
junedev, line 35: Shouldn't the mapping function be defined separately to make this more readable?

exercises/practice/rna-transcription/.approaches/introduction.md
```diff
@@ -0,0 +1,108 @@
...
+For more information, check the [`map` object approach][approach-map-object].
+
+## Approach: `switch` statement
+
+```go
+// Package strand is a small library for returning the RNA complement of a DNA strand.
+package strand
+
+func nucComp(nuc byte) byte {
+
```
junedev, line 72:
```suggestion
func nucComp(nuc byte) byte {
```

exercises/practice/rna-transcription/.approaches/introduction.md
bobahop: I kept it pretty much to what is the currently  [second-most starred solution](https://exercism.org/tracks/go/exercises/rna-transcription/solutions/kgibilterra), so if people saw that and liked it, they could see exactly how it performs. And it is one function call within `ToRNA`, so I don't know that it would buy much to move it.

exercises/practice/rna-transcription/.approaches/introduction.md
bobahop: Not sure what change is intended here.

junedev (approved):

exercises/practice/rna-transcription/.approaches/introduction.md
junedev: It was removing some whitespace that you already removed in a commit you added I think.

exercises/practice/rna-transcription/.approaches/introduction.md
junedev: I find something like `return strings.Map(toComplement, dna)` easier to grasp as it gives the high level story first and then you can dive into the function implementation. In the JS MDN docs, I saw more and more examples in this this style lately https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter#filtering_out_all_small_values.

Stylistic choice I guess.

junedev (approved):

PR #2579, opened 2022-12-23
Strain approaches

bobahop:
Explains the generic approach for Strain.


junedev (commented):
I am unsure about merging this one. This is very much a "workaround" solution for the fact that the exercise tests are not written in a way that you can pass them with just the two generic functions (discard and keep). @andrerfcsantos suggested once we updated to require 1.18 as minimum, we change the exercise accordingly.

andrerfcsantos:
> @andrerfcsantos suggested once we updated to require 1.18 as minimum, we change the exercise accordingly.

Should we just do it?

Go is at 1.20 now. It seems reasonable to set all exercises `go.mod` to 1.18. I believe most people using Go run either the most recent version of Go or the second most recent, so this should be fine.

To be clear, my suggestion here is:

- Upgrade docker images to Go 1.20
- Set the `go.mod` file of exercises to `1.18`
- Regenerate the `go.sum` for external dependencies using a `go.mod` set to 1.18
  - Update [golang.org/x/text/language](https://pkg.go.dev/golang.org/x/text/language) to `0.7.0` to address [GO-2022-1059](https://pkg.go.dev/vuln/GO-2022-1059)

junedev:
@andrerfcsantos Yes, feel free to go ahead with this if you have time.

andrerfcsantos:
@junedev Will do!

junedev:
As discussed, Strain was updated to use generics: https://github.com/exercism/go/pull/2701
I am closing this PR.

@bobahop I'm happy to merge an approach for the new version if you want to create one in the future.

bobahop:
@junedev I think I can just edit this one by deleting what is now irrelevant.

junedev:
Sure, whatever works best for you. Let me know when it's ready to be merged.

bobahop:
@junedev  Okay, I think I made all needed edits correctly, but it's okay with me if you want to double-check.

PR #2589, opened 2023-01-25
Updated the parameter in the the fucntion.

gyan-js:
The func didn't had the parameter name , So I added that for convenience.


andrerfcsantos (approved):
@gyan-js Thanks for the fix.

Despite community contributions being closed, I reopened it and merged since this PR is ready to go without any further changes and adds value.

For future issues you find in the Go track or in exercism in general, feel free to create new topics on [our forum](https://forum.exercism.org/) before opening a PR.

PR #2590, opened 2023-01-25
Added extra example of NeedsLicense function.

gyan-js:
Gave example of NeedsLicense function when implemented for vehicle type truck.


andrerfcsantos (approved):
Thanks!

PR #2594, opened 2023-01-29
Fix typos in introduction.md

denisbr:
Two minor typo fixes.

### DIFF
--- a/concepts/interfaces/about.md
+++ b/concepts/interfaces/about.md
@@ -3,7 +3,7 @@
 ## Interface as a set of methods
 
 In its simplest form, an **interface type** is a set of method signatures.
-Here is an example for an interface defition that includes two methods `Add` and `Value`:
+Here is an example of an interface definition that includes two methods `Add` and `Value`:
 
 ```go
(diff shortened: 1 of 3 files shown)


junedev (approved):
Thanks for this. My spellchecker was not active when I was writing this. 🙈 
I fixed the mistakes you found in the other files as well and added some more fixes.

concepts/interfaces/about.md
```diff
@@ -43,15 +43,15 @@ func (s Stats) SomeOtherMethod() {
 }
 ```
 
-For implementing the interface, it does not matter whether the method has a value or poiter receiver.
+For implementing the interface, it does not matter whether the method has a value or pointer receiver.
 (Revisit the [methods concepts][concept-methods] if you are unsure about those.)
```
Jasstkn, line 47: JFYI: the link seems to be not workin in the UI. Is this intentional?

junedev: Thanks for mentioning this. I will take a look independent of this PR and make another one if needed.

PR #2598, opened 2023-02-05
Fix links in interface concept

junedev:


Jasstkn:
@junedev hey. do you think that it might be useful to add one more link to this learning chapter? 
https://jordanorelli.com/post/32665860244/how-to-use-interfaces-in-go

It helped me a lot to figure out interfaces in a better way.

junedev:
@Jasstkn I added this one now: https://golangbot.com/interfaces-part-1/ It was suggested by Bethany recently and the type of examples is very similar to what I saw in your link. The one from Bethany looked somewhat nicer and rather complete.

PR #2600, opened 2023-02-07
[typo] Add missing "s"

eik3:
Thanks for the Exercism Go track, I'm having great fun.

I think I found a typo in the word "work" and it should be "works".


andrerfcsantos:
@eik3 Since it seems your fork was removed or set private, I couldn't reuse this PR to make further changes.

I made a new PR with the changes: #2608

PR #2612, opened 2023-02-27
Sieve Approach

bobahop:
Approach for Sieve

### DIFF
--- a/exercises/practice/sieve/.approaches/introduction.md
+++ b/exercises/practice/sieve/.approaches/introduction.md
@@ -1,0 +1,39 @@
+# Introduction
+
+There are several variants of a general approach for solving Sieve.
+The general approach is to keep track of composite numbers with a slice of `bool` values.
+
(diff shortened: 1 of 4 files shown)


exercises/practice/sieve/.approaches/introduction.md
```diff
@@ -0,0 +1,37 @@
...
+
+```go
+// Package sieve is a small library for finding prime numbers
+package sieve
+
+// Sieve calculates and returns prime numbers up to the limit
+func Sieve(limit int) []int {
+	composite := make([]bool, limit+1)
+	primes := make([]int, limit/2)
+	prime_index := 0
```
junedev, line 20: Should be `primeIndex` (Official docs: https://go.dev/doc/effective_go#mixed-caps)

junedev: Thanks!

PR #2615, opened 2023-02-28
Secret Handshake approach

bobahop:
Approach for Secret Handshake

### DIFF
--- a/exercises/practice/secret-handshake/.approaches/slice-indexes/content.md
+++ b/exercises/practice/secret-handshake/.approaches/slice-indexes/content.md
@@ -1,0 +1,95 @@
+# Slice indexes
+
+```go
+// Package secret is a small library for interpreting signals for a secret greeting
+package secret
+
+// Handshake interprets what sequence of signals are contained within the number passed in
+func Handshake(code uint) []string {
+	signs := []string{"wink", "double blink", "close your eyes", "jump"}
+	const reverseSigns = 16
+	action, actionIncr, end := 0, 1, len(signs)
+
+	if (code & reverseSigns) != 0 {
+		action = 3
+		actionIncr = -1
+		end = -1
+	}
+	output := []string{}
+
+	for ; action != end; action += actionIncr {
+		if (code & (1 << action)) != 0 {
+			output = append(output, signs[action])
+		}
+	}
+	return output
+}
+```
+
+This approach starts by defining a [slice][slice] of the signals in their normal order.
+A [const] is defined to represent the signal for reversing the order of the other signals.
+It is given a meaningful name so that `16` won't be used as a [magic number][magic-number].
+
+Some variables are defined with values for iterating the slice of signals in their normal order.
+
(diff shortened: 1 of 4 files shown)


exercises/practice/secret-handshake/.approaches/slice-indexes/content.md
```diff
@@ -0,0 +1,95 @@
...
+		if (code & (1 << action)) != 0 {
+			output = append(output, signs[action])
+		}
+	}
+	return output
+}
+```
+
+This approach starts by defining a [slice][slice] of the signals in their normal order.
+A [const] is defined to represent the signal for reversing the order of the other signals.
```
andrerfcsantos, line 30:
```suggestion
A [const][const] is defined to represent the signal for reversing the order of the other signals.
```

exercises/practice/secret-handshake/.approaches/slice-indexes/content.md
```diff
@@ -0,0 +1,95 @@
...
+```go
+if (code & (1 << action)) != 0 {
+  output = append(output, signs[action])
+}
+```
+
+For example, if the number being iterated is `0`, then `1` is shifted left `0` times (so not shifted at all), and the number passed in is ANDed with `00001`.
+If the number passed in is `3`, which is `00011` in binary, then it is ANDed with `00001`.
+`00011` ANDed with `00001` is not equal to `0`, so the signal at the index of the slice of signals is added to the output slice.
+The index used is the number being iterated, which is `0`, so the element at index `0` ("wink`) would be added to the output slice
```
andrerfcsantos, line 78:
```suggestion
The index used is the number being iterated, which is `0`, so the element at index `0` (`"wink"`) would be added to the output slice
```

I guess you meant this?

exercises/practice/secret-handshake/.approaches/slice-indexes/content.md
```diff
@@ -0,0 +1,95 @@
...
+using the [append][append] function.
+
+If the number being iterated is `1`, then `1` is shifted left `1` time, and the number passed in is ANDed with `00010`.
+If the number passed in is `3`, which is `00011` in binary, then it is ANDed with `00010`.
+`00011` ANDed with `00010` is not equal to `0`, so the signal at the index of the slice of signals is added to the output slice.
+The index used is the number being iterated, which is `1`, so the element at index `1` ("double blink`) would be added to the output slice.
```
andrerfcsantos, line 84:
```suggestion
The index used is the number being iterated, which is `1`, so the element at index `1` (`"double blink"`) would be added to the output slice.
```

andrerfcsantos (commented):
A couple of typos, but looks great otherwise 👍

bobahop:
@andrerfcsantos Thanks for the good catches! I need to give the Quality Control department in my brain a stern talking to. Should be ready to merge now.

andrerfcsantos (approved):

PR #2631, opened 2023-03-08
Add hints to the exercise stub to guide the student.

glennj:
This is for the two concept exercises where the stub does not contain the functions already: the student needs to create them before the code can compile.


andrerfcsantos (approved):
Thanks for this!

PR #2644, opened 2023-03-18
Reorder exercises by difficulty

andrerfcsantos:
Following the discussion in: https://forum.exercism.org/t/are-go-practice-exercises-ordered/4464

This makes the order of the exercises make a little more sense.
Also tweaked the difficulty of some exercises according to my experience in the track, hopefully making the difficulty rating more consistent.

### DIFF
--- a/config.json
+++ b/config.json
@@ -482,19 +482,6 @@
         ]
       },
       {
-        "slug": "hamming",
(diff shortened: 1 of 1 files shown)


junedev:
Thanks a lot for this! 💙

PR #2653, opened 2023-03-28
Fix build constraints for robot simulator in regard of Go 1.18 stated…

sebastian0x62:
… in go.mod.


junedev:
@sebastian0x62 Can you clarify? Was it broken before or did you only want to make it nicer?

junedev (approved):

PR #2661, opened 2023-04-12
Fix for wrong formular didn't respect the correct variable type (elons-toys)

sebastian0x62:
See feedback: http://forum.exercism.org/t/bug-wrong-formula-in-elons-toys-exercise/5174/4


junedev (approved):
🙈

PR #2675, opened 2023-05-20
Sync docs

IsaacG:
Sync exercise docs using `configlet sync -u -y --docs`. Revert documents which have diverged significantly from the problem specs.

### DIFF
--- a/exercises/practice/all-your-base/.docs/instructions.md
+++ b/exercises/practice/all-your-base/.docs/instructions.md
@@ -2,31 +2,32 @@
 
 Convert a number, represented as a sequence of digits in one base, to any other base.
 
-Implement general base conversion. Given a number in base **a**,
-represented as a sequence of digits, convert it to base **b**.
+Implement general base conversion.
+Given a number in base **a**, represented as a sequence of digits, convert it to base **b**.
 
 ## Note
 
 - Try to implement the conversion yourself.
   Do not use something else to perform the conversion for you.
 
-## About [Positional Notation](https://en.wikipedia.org/wiki/Positional_notation)
+## About [Positional Notation][positional-notation]
 
-In positional notation, a number in base **b** can be understood as a linear
-combination of powers of **b**.
+In positional notation, a number in base **b** can be understood as a linear combination of powers of **b**.
 
 The number 42, *in base 10*, means:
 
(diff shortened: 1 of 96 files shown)


junedev:
@IsaacG This PR doesn't have a description but it looks like you were syncing all exercise descriptions. Did you check them one by one that the exercise still makes sense (instructions fit to tests etc)? 
André and I noticed a while back that some Go practice exercises were not in line with their problem spec equivalent, append files were not used in all places for changes, different test cases were used etc. Also some exercises in problem spec changed quite a bit (like grade school) so that only updating the docs would lead to an inconsistent state.

junedev (commented):
@IsaacG This PR doesn't have a description but it looks like you were syncing all exercise descriptions. Did you check them one by one that the exercise still makes sense (instructions fit to tests etc)? 
André and I noticed a while back that some Go practice exercises were not in line with their problem spec equivalent, append files were not used in all places for changes, different test cases were used etc. Also some exercises in problem spec changed quite a bit (like grade school) so that only updating the docs would lead to an inconsistent state.

IsaacG:
@junedev I have not gone through them individually, but I can!

Would it make sense to maintain a list of exercises that are (or are not) meant to be in sync with the problem specs? That would all for writing a pretty simple shell script which can be used to sync only the exercises meant to be in sync.

IsaacG:
The following exercises have _some_ degree of diffs. I will revert changes to them. Most of those diffs may be eligible for removal (they don't change the exercises) or just need an append file.

```
 error-handling grade-school grains grep isbn-verifier meetup say wordy
```

junedev:
@IsaacG Thanks a lot for the update. CI did not run for some reason, I will close and re-open the PR, sometimes that helps and I can merge.

IsaacG:
@junedev Thanks! Looks like the checks are all green.

junedev (approved):
Thanks for the update!

Chidinma1709:
Please i really want to learn but i dont understand anything here

IsaacG:
@Chidinma1709 this PR and GitHub in general is not the appropriate place to ask for help. 

See https://exercism.org/docs/using for docs. If you have questions, please post on the forum: https://forum.exercism.org/

PR #2677, opened 2023-05-21
Update instructions for complex-number exercise

ignazioc:
* Introducing `z` in the definition of a complex number
* Formatting to have the code as separate blocks instead of inline with the rest of the text that made reading difficult.


kotp:
Here is [the link to the community forum post.]

[the link to the community forum post.]: https://forum.exercism.org/t/update-instructions-for-complex-number-exercise/5766

junedev:
@ignazioc @kotp This exercise originates from https://github.com/exercism/problem-specifications/tree/main so fixes should be done there and then they are synced down.

PR #2679, opened 2023-05-28
Update Go slices introduction.md

eggled:
Clarify the purpose of the ellipsis in the example code.


junedev:
The ellipsis in Go does not actually "expand" the argument as it would in other languages so I can't merge this as it would be confusing. See https://go.dev/ref/spec#Passing_arguments_to_..._parameters for more details.

PR #2680, opened 2023-05-29
doc: fix example data description

marciodojr:
Hi everyone!

I'm going through the go track and found that the instructions in the exercise [saddle points](https://exercism.org/tracks/go/exercises/saddle-points) have a little error in the markdown. 

**Before**:

image

**After**

image

Please let me know if there is a better way to correct the markdown.


junedev:
Was fixed via another PR.

PR #2690, opened 2023-06-20
Sync docs for the grep exercise

IsaacG:
Sync the grep docs from the problem specs. This reformats much of the instructions but the actual content/meaning doesn't change much.

This *does* result in much fewer example outputs.


junedev (approved):
I agree we should update this even if we loose the examples. We can always further improve problem spec if needed.

PR #2698, opened 2023-08-22
Fix godoc comment variable name

MarvinJWendt:
Single word change, but now it fits with godoc :)


junedev:
Thanks, I updated the other occurrence of this as well.

junedev (approved):

PR #2699, opened 2023-08-24
Update gross_store_test.go

carlosalg:
one of the tests is not in order according to what the instructions indicate,it can be confusing


junedev:
Thanks for this but there is another PR https://github.com/exercism/go/pull/2700 that got the whitespace correct so I will merge that one instead.

PR #2700, opened 2023-08-30
Update test order within gross_store_test.go to match order of steps

hamlim:
Re-arrange the order of tests to match the order of the exercises within the UI.

I was a bit confused when I went through this exercise, I usually approach these in a TDD kind of style, where I do one step, run tests, and verify that the step is working as expected. However this exercise the tests seemed to be out of order compared to the steps.

Of note - I have no idea if this is the "right" fix, I'm only assuming that the order of the tests should also match the order of the steps for them to match within the exercism UI.


junedev (approved):
Thanks for the fix!

PR #2701, opened 2023-09-03
Change strain to use generics

junedev:
@andrerfcsantos suggested that "Strain" would be a good exercise for practicing generics. This PR does that.  The old version of the exercise was very clunky, with generics it makes much more sense.

I added a note on generics in the stub file so people know what they are up for. I also increased the difficultly to 4 to avoid people complaining that an easy exercise needs such an "advanced" concept.

### DIFF
--- a/config.json
+++ b/config.json
@@ -710,7 +710,7 @@
         "uuid": "7fbb0d9d-71d1-4654-8ca0-bb07ddbe6cb5",
         "practices": [],
         "prerequisites": [],
-        "difficulty": 2,
+        "difficulty": 4,
         "topics": [
           "arrays",
           "filtering",
--- a/exercises/practice/strain/.meta/config.json
+++ b/exercises/practice/strain/.meta/config.json
@@ -12,7 +12,8 @@
     "robphoenix",
     "sebito91",
     "soniakeys",
-    "eklatzer"
+    "eklatzer",
+    "junedev"
   ],
   "files": {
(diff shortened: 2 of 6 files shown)


andrerfcsantos:
Looks good, thanks for this!

CI seems to be failing because of `golang.org/x/exp/slices` and some kind of conflict with the linter. Maybe upgrading the linter will fix this?

Also, this package is part of the standard library as of Go 1.21, so when upgrading the track to Go 1.21 we can just replace the import to use the standard library and we won't have this problem.

@junedev Do we think we should try to fix the linter issue for now or focus on upgrading to 1.21?

junedev:
I don't urgently need the slices package for this but it would have been nice.
The real error is this one (which shows when I do "go test" before running the linter):

```
Error: tmp/lint/practice/strain/strain_test.go:8:2: cannot find package "golang.org/x/exp/slices" in any of:
	/opt/hostedtoolcache/go/1.20.7/x64/src/golang.org/x/exp/slices (from $GOROOT)
	/home/runner/go/src/golang.org/x/exp/slices (from $GOPATH)
ok  	_/home/runner/work/go/go/tmp/lint/concept/airport_robot	0.004s
ok  	_/home/runner/work/go/go/tmp/lint/concept/animal_magic	0.004s
```

Not sure why the go test command is not downloading things in a way that it can access it.

PR #2702, opened 2023-09-09
Update introduction.md

dminca:
fix arithmetic typo


junedev:
FYI Line 32 below the table explains why it is not a typo but an intentional example.

dminca:
> FYI Line 32 below the table explains why it is not a typo but an intentional example.

Oh, I've missed that, thank you!

PR #2713, opened 2023-09-30
Tweak code example in binary-search approach

kytrinyx:
The switch statement in the loop-with-switch approach had two explicit cases and a default.

The three cases that exist are:
- equal
- less than
- greater than

If we want to have two explicit cases and a default, then it would feel more symmetrical to have the less-than and greater-than cases be defined explicitly with the equal case being the default.

However, it's hard to argue that any of the three are actually a default here.

Instead I've changed it to have three explicit cases.

Since there is an explicit return if we fall out of the `for` loop, we don't technically need the default.

Thoughts?

### DIFF
--- a/exercises/practice/binary-search/.approaches/loop-with-switch/content.md
+++ b/exercises/practice/binary-search/.approaches/loop-with-switch/content.md
@@ -8,13 +8,13 @@ package binarysearch
 func SearchInts(list []int, key int) int {
 	for left, right := 0, len(list); left != right; {
 		mid := (left + right) / 2
-    
+
 		switch {
 		case list[mid] == key:
 			return mid
 		case key < list[mid]:
 			right = mid
-		default:
+		case key > list[mid]:
 			left = mid + 1
 		}
 	}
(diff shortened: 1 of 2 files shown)


exercises/practice/binary-search/.approaches/loop-with-switch/content.md
```diff
@@ -8,13 +8,13 @@ package binarysearch
 func SearchInts(list []int, key int) int {
 	for left, right := 0, len(list); left != right; {
 		mid := (left + right) / 2
-    
+
```
kytrinyx, line 11: Sorry about stripping this—I didn't realize I did that until after I submitted. Let me know if I should put it back in to avoid extra noise here.

andrerfcsantos (approved):
In general, I think the reason most programmers include a default case is to have peace of mind that all cases are covered. However, in this specific case, I think the 3 case statements make it clear that all cases are covered, even if there's no default block. It's also nicer to read.

exercises/practice/binary-search/.approaches/loop-with-switch/content.md
andrerfcsantos: I don't mind the extra line here.

ErikSchierboom (approved):
I am personally fine with either approach (I've just both).

junedev (approved):

PR #2726, opened 2023-11-03
remove double percent signs

lotharschulz:
fix `TestCalculateResellPrice` test names


junedev (approved):
Thanks! Probably someone wrongly thought those needed to be escaped just like in a formatting string. (Maybe that someone even was myself.)

PR #2735, opened 2023-12-21
clock: update clock exercise instructions to point to correct receiver type documentation

mr-sigma:
The link went to a page which had been relocated. Explicitly linked to the correct section on the new page.


andrerfcsantos (approved):
Thanks for this! 👍

PR #2746, opened 2024-01-15
Add testcase for TestCalculateResellPrice

jesse-kroon:
I've described the issue here.
https://forum.exercism.org/t/expanding-test-cases-for-vehicle-purchase-exercise-in-the-go-track/9066

(Agreed by @andrerfcsantos on [Exercism Community](https://forum.exercism.org/t/expanding-test-cases-for-vehicle-purchase-exercise-in-the-go-track/9066/7))


iHiD:
@andrerfcsantos can we merge this with `[no important files changed]` pls (or whatever the correct magic string is. I don't think we need to rerun all the existing solutions for this change.

andrerfcsantos:
Thanks for this!

PR #2750, opened 2024-01-25
Nucleotide Count Excercise: Mistype fastest method according to "Performance deep dive" page

newbie-k:
"Performance deep dive" page: [here](https://exercism.org/tracks/go/exercises/nucleotide-count/articles/performance)


andrerfcsantos (approved):
Thanks for this!

PR #2753, opened 2024-01-27
Add "D&D Character" practice exercise

jesse-kroon:
As per the #48in24 challenge [implementation status page](https://exercism.org/challenges/48in24/implementation_status).


exercises/practice/dnd-character/dnd_character_test.go
```diff
@@ -0,0 +1,70 @@
...
+	for i := 0; i < b.N; i++ {
+		Modifier(i)
+	}
+}
+
+func inAcceptedRange(score int) bool {
+	return score >= 3 && score <= 18
+}
+
+func assertAbilityScoreInRange(t testing.TB, ability string, score int) {
```
andrerfcsantos, line 64:
```suggestion
func assertCharacterAbilityScoreInRange(t *testing.T, ability string, score int) {
```

Since we are no longer calling this function in `TestAbility`, we can make the name reflect that this function is specific to character abilities.

Using the interface `testing.TB` is unnecessary here, since we are only using it in the context of testing, so receiving the `*testing.T` is more intuitive to readers.

exercises/practice/dnd-character/dnd_character_test.go
```diff
@@ -0,0 +1,70 @@
...
+	t.Helper()
+
+	if !inAcceptedRange(score) {
+		t.Fatalf("%s score is not within accepted range. Got %d", ability, score)
```
andrerfcsantos, line 68: We could change the message to something like "GenerateCharacter() created a character with a score of %d of <ability>, but the score for an ability is expected to be between 3 and 18 inclusive"

exercises/practice/dnd-character/dnd_character_test.go
```diff
@@ -0,0 +1,70 @@
...
+		assertAbilityScoreInRange(t, "Charisma", character.Charisma)
+		assertAbilityScoreInRange(t, "Strength", character.Strength)
+		assertAbilityScoreInRange(t, "Dexterity", character.Dexterity)
+		assertAbilityScoreInRange(t, "Wisdom", character.Wisdom)
+		assertAbilityScoreInRange(t, "Intelligence", character.Intelligence)
+		assertAbilityScoreInRange(t, "Constitution", character.Constitution)
+
+		if character.Hitpoints < 10 {
+			t.Fatalf("Character's base hitpoints are incorrect. Got %d", character.Hitpoints)
+		}
```
andrerfcsantos, line 38: The check here should be if `character.Hitpoints == 10 + Modifier(character.Constitution)`.

The error message should be something like "Got %d hitpoints for a character of %d constitution, expected %d hitpoints".

exercises/practice/dnd-character/.meta/example.go
```diff
@@ -0,0 +1,65 @@
...
+	Charisma     int
+	Hitpoints    int
+}
+
+// Modifier should calculate the correct modifier for a given score
+func Modifier(score int) int {
+	return int(math.Floor(float64(score-10) / 2.0))
+}
+
+// Ability should generate the score for a random ability
```
andrerfcsantos, line 24:
```suggestion
// Ability uses randomness to generate the score for an ability
```

The comment should also be changed in the stub given to students

exercises/practice/dnd-character/.meta/example.go
```diff
@@ -0,0 +1,65 @@
...
+
+	for i := 0; i < 4; i++ {
+		roll := RollDice()
+		scores = append(scores, roll)
+	}
+
+	return CalculateAbilityScore(scores)
+}
+
+// GenerateCharacter should return a Character with valid ability scores
```
andrerfcsantos, line 36:
```suggestion
// GenerateCharacter creates a new Character with random scores for abilities 
```

The comment should also be changed in the stub given to students

andrerfcsantos (approved):
Let's get this live! 🎉

Did some small changes to the tests since your last modifications. After all the work you did I didn't want to bore you with more work. The changes I've made:

- Made code that would generate scores for abilities slightly off the ranges, and noticed that just 100 iterations were not enough to reliably catch those cases. Ran the tests several times and it was somewhat easy to get tests passing when they shouldn't. Changed to 1000 iterations and it was a lot harder to get tests passing with the same wrong code.
- In `TestModifier`, changed the failure message to print ` tc.input.Score` instead of  `tc.input`. In case of failures, it was printing the message as `Modifier({3}) = -3, want -4`, now prints `Modifier(3) = -3, want -4`.
- Changed the order of benchmarks of each function to match the order by which the students have to implement the functions being benchmarked. Go runs the benchmarks by the order they are declared, and it was weird to see `BenchmarkModifier`  be the last benchmark when `Modifier` is the first function students should implement.

---

@jesse-kroon Thanks again for doing this and for pushing it through despite all the hurdles.

jesse-kroon:
Thanks for the final edits @andrerfcsantos ! Good stuff.

PR #2763, opened 2024-02-24
card-tricks: Check if index is >= len(slice)

CodeBleu:
Adjusting so the test will now fail if the `index` value is greater than or equal to the size of the `slice`.

### DIFF
--- a/exercises/concept/card-tricks/card_tricks_test.go
+++ b/exercises/concept/card-tricks/card_tricks_test.go
@@ -51,7 +51,7 @@ func TestGetItem(t *testing.T) {
 			name: "Index out of bounds",
 			args: args{
 				slice: []int{5, 2, 10, 6, 8, 7, 0, 9},
-				index: 8,
+				index: 9,
 			},
 			want: -1,
 		},


andrerfcsantos:
Hey @CodeBleu.

Sorry for the delay in reviewing this. Still interested in working on this?

In this case, I'd suggest to add a new test where the index is strictly bigger than `len(slice)`. That way we have a test for when the index is equal to `len(slice)` and another one for when it is bigger.

CodeBleu:
@andrerfcsantos Does there really need to be 2 tests? Not sure what the benefit is, if an = or > value is used.  There would not be any use case to just check for =, right?  Both will fail.  Maybe I'm missing something,  but I don't see it any different than writing a test that checks for a specific number greater than either...hence just test for >

andrerfcsantos (approved):
After doing some testing, I think I see your point now.

I suggested having two tests to try to be more exhaustive and intentional in the testing. The idea is to have the tests fail even if solutions were just doing `>` or `==` and only pass when they have `>=`. Your change does this, so it's all good.

The benefit of this change is that solutions that just check for `==` will now fail, where previsously they would pass.

I'll merge, thanks for this! And again, sorry for taking so long.

PR #2768, opened 2024-03-16
Single sentence per line for Runes

kotp:
Preparing the patch as discussed in [the forums] I realized that the
"single sentence per line" is not being adhered to as set forth in the
[building exercism] guide.

[the forums]: http://forum.exercism.org/t/potential-misleading-information-on-the-golang-runes-chapter/10082/14
[building exercism]: https://exercism.org/docs/building/markdown/markdown


kotp:
Even though this is closed, the patch is still available. (I deleted the repository on Github since it did not reflect information from there that there was an open and pending thing attached.)  Still that does not preclude bringing in work that is done, nor having to redo it, if this is wanted, it can still be brought in.  (no wasted work)

PR #2769, opened 2024-03-16
Describing a string as a sequence of runes is very confusing

kotp:
As discussed in [the forums], this patch makes the changes desired to
clarify how Strings are related to Runes and hopefully clears up some
confusing and potentially misleading statements.

Ref: https://github.com/exercism/go/pull/2768

[the forums]: https://forum.exercism.org/t/potential-misleading-information-on-the-golang-runes-chapter/10082/1


kotp:
#2768 should be accepted and integrated before this one.

kotp:
Same here, the work is available if you still want it, as a patch.  I deleted my clone of the repository on the platform, but that does not make the work unavailable for integration.

PR #2771, opened 2024-03-22
prime-factors: use slices package to sort and compare test cases

zamai:
> I think this PR would be automatically closed, but wanted to leave it anyway, maybe'll figure out the forum later, but my tiny code reduction PR will be here already. 

In this PR I moved slice sorting and comparison to use standard libs `slices` package, removed helpers that we're necessary before generics were introduced. 

Value of this PR - removes code, enhances readability, showcases useful tools of stdlib :)

### DIFF
--- a/exercises/practice/prime-factors/prime_factors_test.go
+++ b/exercises/practice/prime-factors/prime_factors_test.go
@@ -1,7 +1,7 @@
 package prime
 
 import (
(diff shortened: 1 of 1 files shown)


andrerfcsantos (approved):
Thanks for this. This is something I've been actually think of doing for a while now.

PR #2780, opened 2024-04-16
bird-watcher: Add test case for a week in the end of the log

mugane-dj:
The test cases provided in the exercise for the `BirdsInWeek` function only cover the first week and a week in the middle of the log. An edge case for a week in the end of the log is missing.


andrerfcsantos (approved):
Thanks for this!

PR #2794, opened 2024-05-05
Fix typos in Parsing Log Files instructions

toryc:
This PR fixes a couple typos in the Parsing Log Files instructions.md where "expressions" is plural, but the context shows that it should be singular.


andrerfcsantos:
Already fixed by https://github.com/exercism/go/pull/2797

PR #2797, opened 2024-05-06
Fix typos in regular-expressions concept instructions

toryc:
This PR fixes a couple typos in the Parsing Log Files instructions.md where "expressions" is plural, but the context shows that it should be singular. As discussed in the [forum](https://forum.exercism.org/t/fix-typos-in-parsing-log-files-instructions/11059), I've changed both files where this content is located.


andrerfcsantos (approved):
Thanks for this!

PR #2805, opened 2024-05-21
Update simple_linked_list_test.go with an idempotency test

curiousz-peel:
If the submitter changes the actual list while iterating through its elements, the existing tests won't catch that. Example of a String() method that changes the underlying list: 
func (l *List) String() string {
	var listString string
	for l.head != nil {
		listString += fmt.Sprintf("%d ", l.head.value)
		l.head = l.head.next
	}
	return listString
}


andrerfcsantos:
This is being discussed in the forum: https://forum.exercism.org/t/update-simple-linked-list-test-go-with-an-idempotency-test/11322

PR #2809, opened 2024-06-01
Fix: Type error in lasagna_master_test.go by converting portions to float64

CEOZMaze:
This pull request fixes a type error in lasagna_master_test.go. The test was failing because an integer variable `tt.portions` was being used where a `float64` was expected. This change converts `tt.portions` to `float64` to resolve the error.

Error message:
./lasagna_master_test.go:205:34: cannot use tt.portions (variable of type int) as float64 value in argument to ScaleRecipe

After this change, the tests should pass without any type errors.

image


andrerfcsantos:
Discussion about this in this forum post: https://forum.exercism.org/t/fix-type-error-in-lasagna-master-test-go-by-converting-portions-to-float64/11517

PR #2821, opened 2024-07-19
Update introduction.md

bryansoto:
Minor typo and inconsistent use of spelling of behavior/behaviour.

The use of behavior is on line 25 so luckily it will actually show up in the diff. I just changed line 26 to match it.

If behaviour is preferred, I'm happy to change it.


andrerfcsantos (approved):
Thanks for this!

There were other places where these changes needed to be made. In the interest of not bothering you with more work, I updated the PR to also include the changes in the other files they were needed.

PR #2824, opened 2024-07-28
Describing a string as a sequence of runes is very confusing

kotp:
[Conversation](https://forum.exercism.org/t/potential-misleading-information-on-the-golang-runes-chapter/10082/11).


andrerfcsantos (approved):
Thanks for this!

Rebased your branch into main and force-pushed the changes. I did this because git was complaining about a weird conflict and wouldn't let merge without it being fixed. The conflict was weird, and merging main into the branch didn't work, because when accepting all the changes in the current branch it claimed there was nothing to commit. A simple rebase did the trick.

PR #2826, opened 2024-07-29
Minor correction to the Anumal magic instructions markdown file

Kristianwhittick:
This PR is to clarify the range can also equal zero.

#### Change the text from
Implement a GenerateWandEnergy function. The wand energy should be a random floating point number **between 0.0 and 12.0.**

#### Change the text to
Implement a GenerateWandEnergy function. The wand energy should be a random floating point number **equal or greater than 0.0 and less than 12.0.**


andrerfcsantos (approved):
Thanks for this!

PR #2853, opened 2025-04-10
Remove random.Seed() call from robot-name's approach file

HerrSubset:
From the Go documentation on random.Seed():

Deprecated: As of Go 1.20 there is no reason to call Seed with a random value. Programs that call Seed with a known value to get a specific sequence of results should use New(NewSource(seed)) to obtain a local random generator.

### DIFF
--- a/exercises/practice/robot-name/.approaches/introduction.md
+++ b/exercises/practice/robot-name/.approaches/introduction.md
@@ -20,7 +20,6 @@ package robotname
 import (
 	"fmt"
 	"math/rand"
-	"time"
 )
 
 // Robot is a struct with a string field.
@@ -48,7 +47,6 @@ func generateRobotNames() []string {
 		}
 	}
 
-	rand.Seed(time.Now().UnixNano())
 	rand.Shuffle(len(names), func(i, j int) { names[i], names[j] = names[j], names[i] })
 	return names
 }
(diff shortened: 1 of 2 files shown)


HerrSubset:
Just created [a forum post](https://forum.exercism.org/t/remove-random-seed-call-from-robot-names-approach-file/16824) for this issue as well.

siebenschlaefer:
I'm not involved in the Go track in any way, but I just noticed:
You might want to make that same change in [.approaches/shuffle/content.md](https://github.com/exercism/go/blob/main/exercises/practice/robot-name/.approaches/shuffle/content.md), too, to keep the files in sync.

HerrSubset:
Thanks for checking that @siebenschlaefer, I updated the PR.

I also tested both examples and saw that the `time` import was no longer needed, so removed that too from both files. Both examples now compile and pass the tests!

SleeplessByte (approved):
Looks good. The [sample solution](https://github.com/exercism/go/blob/04d86c950c0c15d0646f4cb9dc1235965de25c8f/exercises/practice/robot-name/.meta/example.go) didn't seed a random number generator.

Can you confirm these approaches actually pass the tests if ran? If yes, I don't see a reason why this cannot be merged...

SleeplessByte:
cc @exercism/go

HerrSubset:
> Can you confirm these approaches actually pass the tests if ran? If yes, I don't see a reason why this cannot be merged...

Yes, tested that last week after siebenschlaefer's comment :)

IsaacG (approved):

PR #2874, opened 2025-07-04
Add flower-field & deprecate minesweeper

kahgoh:
See https://forum.exercism.org/t/suggestion-deprecate-minesweeper-for-flower-field/17967/23

### DIFF
--- a/config.json
+++ b/config.json
@@ -1580,9 +1580,9 @@
         ]
       },
       {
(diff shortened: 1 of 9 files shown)


IsaacG (approved):

IsaacG:
@kahgoh Thank you!

PR #2893, opened 2026-01-05
Fix error message in NewTrack test case

kahghi:
noticed that the error message did not make sense when doing a local `go test`:
image
my output value and the expected value output as the same, and it took awhile for me to realise i hardcoded a value in the NewTrack struct that was causing the error.
the correct test case should compare against the instantiated struc got

### DIFF
--- a/exercises/concept/need-for-speed/need_for_speed_test.go
+++ b/exercises/concept/need-for-speed/need_for_speed_test.go
@@ -65,7 +65,7 @@ func TestNewTrack(t *testing.T) {
 			got := NewTrack(tt.track.distance)
 
 			if got != tt.expected {
-				t.Errorf("NewTrack(%+v) = %+v; expected %+v", tt.track.distance, tt.track, tt.expected)
+				t.Errorf("NewTrack(%+v) = %+v; expected %+v", tt.track.distance, got, tt.expected)
 			}
 		})
 	}


kahghi:
more details below

error message:
```
--- FAIL: TestNewTrack (0.00s)
    --- FAIL: TestNewTrack/Create_a_new_track.#01 (0.00s)
        need_for_speed_test.go:68: NewTrack(360) = {distance:360}; expected {distance:360}
FAIL
exit status 1
FAIL	speed	0.405s
```

need_for_speed.go
```
package speed

// TODO: define the 'Car' type struct
type Car struct {
	battery int
	batteryDrain int
	speed int
	distance int
}

// NewCar creates a new remote controlled car with full battery and given specifications.
func NewCar(speed, batteryDrain int) Car {
	return Car{
		battery: 100,
		batteryDrain: batteryDrain,
		speed: speed,
		distance: 0,
	}
}

// TODO: define the 'Track' type struct
type Track struct {
	distance int
}

// NewTrack creates a new track
func NewTrack(distance int) Track {
	return Track{
		distance: 800,
	}
}

// Drive drives the car one time. If there is not enough battery to drive one more time,
// the car will not move.
func Drive(car Car) Car {
	if car.battery >= car.batteryDrain {
		car.battery -= car.batteryDrain
		car.distance += car.speed
	}
	return car
}

// CanFinish checks if a car is able to finish a certain track.
	numberOfTurns := track.distance / car.speed
	overallBatteryRequired := car.batteryDrain * numberOfTurns
	if overallBatteryRequired <= car.battery {
		return true
	} else {
		return false
	}
}
```

expected behaviour -> test should fail and return that distance was assigned as 800, which is different than the expected 360. 
reality -> test failed with detail that the assigned distance was 360, same as the expected value -> confusion as i could not understand why the tests were failing when the two values were the same.

went to inspect the test cases in need_for_speed_test.go and realised that `TestNewTrack` was referencing its own track value in the error message. realised that the comparison should be done against `got`, similar to `TestNewCar`

<pre><code>
func TestNewTrack(t *testing.T) {
	tests := []struct {
		name     string
		track    Track
		expected Track
	}{
		{
			name: "Create a new track.",
			track: Track{
				distance: 800,
			},
			expected: Track{
				distance: 800,
			},
		},
		{
			name: "Create a new track.",
			track: Track{
				distance: 360,
			},
			expected: Track{
				distance: 360,
			},
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			got := NewTrack(tt.track.distance)

			if got != tt.expected {
				t.Errorf("NewTrack(%+v) = %+v; expected %+v", tt.track.distance, <b>tt.track</b>, tt.expected)
			}
		})
	}
}
</code></pre>


<pre><code>
func TestNewCar(t *testing.T) {
	tests := []struct {
		name     string
		car      Car
		expected Car
	}{
		{
			name: "Create a new car.",
			car: Car{
				speed:        5,
				batteryDrain: 2,
				battery:      100,
			},
			expected: Car{
				speed:        5,
				batteryDrain: 2,
				battery:      100,
			},
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			got := NewCar(tt.car.speed, tt.car.batteryDrain)

			if got != tt.expected {
				t.Errorf("NewCar(%+v,%+v) = %+v; expected %+v", tt.car.speed, tt.car.batteryDrain, <b>got</b>, tt.expected)
			}
		})
	}
}
</code></pre>

andrerfcsantos:
Thanks for this!

PR #2903, opened 2026-02-24
Sync problem specs metadata and docs

BNAndras:
The `error-handling` instructions had a good bit of Go-specific information that I moved over to the existing instructions append. Let me know if you want to pull that into a separate PR so we can take a closer look at the append.

### DIFF
--- a/exercises/practice/zebra-puzzle/.docs/instructions.md
+++ b/exercises/practice/zebra-puzzle/.docs/instructions.md
@@ -12,20 +12,20 @@ The following 15 statements are all known to be true:
 1. There are five houses.
 2. The Englishman lives in the red house.
 3. The Spaniard owns the dog.
-4. Coffee is drunk in the green house.
+4. The person in the green house drinks coffee.
 5. The Ukrainian drinks tea.
 6. The green house is immediately to the right of the ivory house.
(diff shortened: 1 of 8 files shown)


exercises/practice/zebra-puzzle/.docs/instructions.md
```diff
@@ -12,20 +12,20 @@ The following 15 statements are all known to be true:
 1. There are five houses.
 2. The Englishman lives in the red house.
 3. The Spaniard owns the dog.
-4. Coffee is drunk in the green house.
```
IsaacG, line 15: This file needs to go with test changes

BNAndras: Sounds good.

BNAndras: Upon review, there's no unsynced test to add here. It's an instructions-only change, but I did noticed the instructions append isn't a single sentence per line so I can fix the formatting there.

IsaacG: The reformatting might be better in a separate PR to make the move simpler to view in this PR.

PR #2904, opened 2026-02-24
Sync `anagram` tests

BNAndras:
Mark as tiny.

### DIFF
--- a/exercises/practice/anagram/anagram_test.go
+++ b/exercises/practice/anagram/anagram_test.go
@@ -6,18 +6,8 @@ import (
 	"testing"
 )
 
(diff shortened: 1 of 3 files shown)


exercises/practice/anagram/anagram_test.go
```diff
@@ -6,18 +6,8 @@ import (
 	"testing"
 )
 
-var nonAsciiTestCases = []anagramTest{
```
IsaacG, line 9: Why is this being dropped?

BNAndras: There is a canonical Greek test now.

PR #2905, opened 2026-02-24
Sync `armstrong-numbers` tests

BNAndras:
`./bin/run-generators` gives me this error for the two new armstrong tests so I figure these two should be skipped if the numbers can't fit into a int. Mark as tiny.
 
```
2026/02/23 21:15:03 json: cannot unmarshal number 1.8670996100153877e+32 into Go struct field .input.number of type int
exit status 1
```


IsaacG:
Have you tried with a `uint64`?

Issue #2714, opened 2023-10-07
Remove check for short test in benchmarks


junedev:
Benchmarks are not included in normal test runs, so we don't need to do anything to make them excluded.

BNAndras:
@exercism/go, I believe this is the bit that can be potentially removed.

```go
if testing.Short() {
    b.Skip("skipping benchmark in short mode")
}
```

That affects 107 test suites. I tested it on the anagram test suite and currently, `go test -bench . -short` would just run the regular tests. It's not clear why you wouldn't just do `go test .` at that point.

PR #2941, opened 2026-02-28
Implement practice exercise Swift Scheduling

IsaacG:


exercises/practice/swift-scheduling/stub.txt
kahgoh, line 1: What is this file used for? Was this accidentally added?

exercises/practice/swift-scheduling/swift_scheduling_test.go
```diff
@@ -0,0 +1,21 @@
...
+	for _, tc := range testCases {
+		t.Run(tc.description, func(t *testing.T) {
+			if actual := DeliveryDate(tc.start, tc.delivery); actual != tc.expected {
+				t.Fatalf("DeliveryDate(%q, %q) = %q, want: %q", tc.start, tc.delivery, actual, tc.expected)
+			}
+		})
+	}
+}
+
+func BenchmarkDeliveryDate(b *testing.B) {
```
kahgoh, line 15: I noticed this is missing the "skip if short mode" statement that is in other benchmark tests. Should it be here or is there some thinking behind it?  

In case it helps, I dug a bit through the history about adding the check and found issue #1607.

IsaacG: My understanding is that `go test` does not run benchmarks (anymore?). The Go track docs recommend using `go test --bench . --benchmem` to run benchmarks. I believe the short-check is no longer needed.

See #2714 and #2932

exercises/practice/swift-scheduling/stub.txt
IsaacG: Accidental add. Thanks for catching it! Removed.

exercises/practice/swift-scheduling/.meta/gen.go
```diff
@@ -0,0 +1,48 @@
...
+	Input       struct {
+		MeetingStart string `json:"meetingStart"`
+		Description string `json:"description"`
+	} `json:"input"`
+	Expected string `json:"expected"`
+}
+
+var tmpl = `package swiftscheduling
+
+{{.Header}}
```
IsaacG:
```suggestion
var tmpl = `{{.Header}}
```

exercises/practice/swift-scheduling/swift_scheduling_test.go
kahgoh: Ah, that makes sense! Thanks!

exercises/practice/swift-scheduling/.meta/gen.go
```diff
@@ -0,0 +1,46 @@
...
+type testCase struct {
+	Description string `json:"description"`
+	Input       struct {
+		MeetingStart string `json:"meetingStart"`
+		Description string `json:"description"`
+	} `json:"input"`
+	Expected string `json:"expected"`
+}
+
+var tmpl = `{{.Header}}
```
kahgoh, line 31: I notice the package declaration was removed from the template. Doesn't this mean the generated file is going to be missing the package declaration?

IsaacG: It's been moved into the Header and is now auto generated from the slug in another PR.

IsaacG: See #2945 and/or the [updated `Header.String()` code](https://github.com/exercism/go/blob/59434191fb9c9410717ca77c4d28751d310882d7/gen/gen.go#L54).

exercises/practice/swift-scheduling/go.mod
```diff
@@ -0,0 +1,3 @@
+module swiftschedulint
```
kahgoh: Spelling

```suggestion
module swiftscheduling
```

PR #2942, opened 2026-02-28
card-tricks: Fix test case error messages in card-tricks exercise

rahulrao0209:


IsaacG:
I believe this is included in/duplicated by #2943

PR #2961, opened 2026-03-02
Use `%q` to quote strings

### LINKED ISSUE #1956: Use `%q` in test failure messages that contain strings
Currently, we often use `%s` or `%v` as formatting verbs when creating the test failure messages. That causes problems for students because due to the missing quotes around the strings, e.g. trailing whitespace is not visible. In some occasions this was fixed by manually adding quotes around `%s` because we were not aware that this can be done automatically by using `%q`. So we want to make use of `%q` instead.

This tasks needs a bit more investigation regarding how to fix this in case of generated tests before it can be worked on.

IsaacG:
Closes #1956

### DIFF
--- a/exercises/practice/grep/grep_test.go
+++ b/exercises/practice/grep/grep_test.go
@@ -71,7 +71,7 @@ func TestSearch(t *testing.T) {
 			}
 
 			if !reflect.DeepEqual(actual, tc.expected) {
-				t.Errorf("Search(%q, %v, %v)\ngot: %v\nwant: %v", tc.pattern, tc.flags, tc.files, actual, tc.expected)
+				t.Errorf("Search(%q, %q, %q)\ngot: %q\nwant: %q", tc.pattern, tc.flags, tc.files, actual, tc.expected)
 			}
 		})
 	}
--- a/exercises/practice/resistor-color/resistor_color_test.go
+++ b/exercises/practice/resistor-color/resistor_color_test.go
@@ -20,14 +20,14 @@ func TestColors(t *testing.T) {
 			actual := Colors()
 
 			if len(actual) != len(tc.expected) {
-				t.Fatalf("Colors() = %+v, want %+v", actual, tc.expected)
+				t.Fatalf("Colors() = %q, want %q", actual, tc.expected)
 			}
 
 			expectedMap := makeMap(tc.expected)
 			actualMap := makeMap(actual)
 
 			if !mapsEqual(expectedMap, actualMap) {
-				t.Fatalf("Colors() = %+v, want %+v", actual, tc.expected)
+				t.Fatalf("Colors() = %q, want %q", actual, tc.expected)
 			}
 		})
 	}


exercises/practice/grep/grep_test.go
```diff
@@ -71,7 +71,7 @@ func TestSearch(t *testing.T) {
 			}
 
 			if !reflect.DeepEqual(actual, tc.expected) {
-				t.Errorf("Search(%q,%v,%v)\ngot: %v\nwant: %v", tc.pattern, tc.flags, tc.files, actual, tc.expected)
+				t.Errorf("Search(%q, %v, %v)\ngot: %v\nwant: %v", tc.pattern, tc.flags, tc.files, actual, tc.expected)
```
kahgoh: You can also use `%q` even on `[]string`. Quote from  [fmt](https://pkg.go.dev/fmt):

> For compound operands such as slices and structs, the format applies to the elements of each operand, recursively, not to the operand as a whole. Thus %q will quote each element of a slice of strings,

```suggestion
				t.Errorf("Search(%q, %q, %q)\ngot: %q\nwant: %q", tc.pattern, tc.flags, tc.files, actual, tc.expected)
```

exercises/practice/resistor-color/resistor_color_test.go
```diff
@@ -20,14 +20,14 @@ func TestColors(t *testing.T) {
 			actual := Colors()
 
 			if len(actual) != len(tc.expected) {
-				t.Fatalf("Colors(): expected %+v, actual %+v", tc.expected, actual)
+				t.Fatalf("Colors() = %+v, want %+v", actual, tc.expected)
```
kahgoh: I think the`+` is a mistake (the values being formatted are strings, not numbers).

```suggestion
				t.Fatalf("Colors() = %q, want %q", actual, tc.expected)
```

exercises/practice/resistor-color/resistor_color_test.go
```diff
@@ -20,14 +20,14 @@ func TestColors(t *testing.T) {
...
 			}
 
 			expectedMap := makeMap(tc.expected)
 			actualMap := makeMap(actual)
 
 			if !mapsEqual(expectedMap, actualMap) {
-				t.Fatalf("Colors(): expected %+v, actual %+v", tc.expected, actual)
+				t.Fatalf("Colors() = %+v, want %+v", actual, tc.expected)
```
kahgoh:
```suggestion
				t.Fatalf("Colors() = %q, want %q", actual, tc.expected)
```

exercises/practice/state-of-tic-tac-toe/state_of_tic_tac_toe_test.go
```diff
@@ -13,7 +13,7 @@ func TestStateOfTicTacToe(t *testing.T) {
 				t.Fatalf("\n Board: %#v \n Expected error but got nil", c.board)
 
 			case !c.wantErr && err != nil:
-				t.Fatalf("\n Board: %#v \n Expected no errors but got error: %v", c.board, err)
+				t.Fatalf("\n Board: %#v \n Expected no errors but got error: %q", c.board, err)
```
kahgoh, line 16:
```suggestion
				t.Fatalf("\n Board: %#q \n Expected no errors but got error: %q", c.board, err)
```

IsaacG: I was applying `%q` on inputs/outputs values.
Errors are typically not quoted in log entries; they are text messages and not input/output values.

IsaacG: Thank you for reviewing!!

PR #2972, opened 2026-03-06
[docs concepts/regular-expressions] replace "routine" with "method"

chengjilai:
1. The term "routine" is potentially ambiguous in documentation.
2. The opening sentence already refers to them as "16 methods of Regexp", so this change ensures terminology consistency throught out the introduction.
3. This improves readability for new developers.

### DIFF
--- a/concepts/regular-expressions/introduction.md
+++ b/concepts/regular-expressions/introduction.md
@@ -2,11 +2,11 @@
 
 Package [regexp][package-regexp] offers support for regular expressions in Go.
 
-## Syntax 
+## Syntax
 
-The [syntax][regexp-syntax] of the regular expressions accepted is the same general syntax used by Perl, Python, and other languages. 
+The [syntax][regexp-syntax] of the regular expressions accepted is the same general syntax used by Perl, Python, and other languages.
 
-Both the search patterns and the input texts are interpreted as UTF-8.  
+Both the search patterns and the input texts are interpreted as UTF-8.
 
 When using backticks (\`) to make strings, backslashes (`\`)  don't have any special meaning and don't mark the beginning of special characters like tabs `\t` or newlines `\n`:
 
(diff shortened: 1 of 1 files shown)


exercises/concept/parsing-log-files/.docs/introduction.md
```diff
@@ -39,29 +39,29 @@ re, err = regexp.Compile(`a|b)+`)
...
 There are 16 methods of `Regexp` that match a regular expression and identify the matched text.
 Their names are matched by this regular expression:
 
 ```text
 Find(All)?(String)?(Submatch)?(Index)?
 ```
 
-* If `All` is present, the routine matches successive non-overlapping matches of the entire expression.
-* If `String` is present, the argument is a string; otherwise it is a slice of bytes; return values are adjusted as appropriate. 
+* If `All` is present, the method matches successive non-overlapping matches of the entire expression.
```
IsaacG, line 63: The docs also refers to them as "functions" which would be the more appropriate term to use here. Could you change all instances of "method" to "function"?

exercises/concept/parsing-log-files/.docs/introduction.md
IsaacG, line 1: The [concept doc](https://github.com/exercism/go/blob/main/concepts/regular-expressions/introduction.md) also needs to be updated;  they should be kept the same.

exercises/concept/parsing-log-files/.docs/introduction.md
chengjilai: "methods" seems to be more appropriate here. 
[A method is a function with a special receiver argument](https://go.dev/tour/methods/1). Here, a `*regexp.Regexp` is a receiver. 

"function"s in [introduction.md](https://github.com/exercism/go/blob/main/concepts/regular-expressions/introduction.md) are also appropriate. 
They refer to: `regexp.Compile` and `regexp.MustCompile` which are just two functions, not methods, for they do not require receiver arguments

exercises/concept/parsing-log-files/.docs/introduction.md
chengjilai: Thank you! 
The introduction.md is already consistent. Ready for another review. Happy to adjust if I missed something.

exercises/concept/parsing-log-files/.docs/introduction.md
IsaacG: My bad. You're right :)

exercises/concept/parsing-log-files/.docs/introduction.md
IsaacG: The [concept file has the same "routine" on L63](https://github.com/exercism/go/blob/main/concepts/regular-expressions/introduction.md?plain=1#L63) that needs the same change.

chengjilai: Thank you! I missed that file. Just pushed another commit [c0325b3](https://github.com/chengjilai/go-1/commit/c0325b3639f559ebdc04aa3333726d4109928413). Please have a look.

IsaacG: All the other changes should also be synced over :) Those files are currently the same and should remain the same.

chengjilai: Thanks! I overlooked  that. Fixed in [e6902de](https://github.com/chengjilai/go-1/commit/e6902dec0d1e46b5bca8b712da0a5dae0fb0539c)

PR #2990, opened 2026-03-10
[expenses] Update instructions and test case error messages

rahulrao0209:
[no important files changed]

### DIFF
--- a/exercises/concept/expenses/expenses_test.go
+++ b/exercises/concept/expenses/expenses_test.go
@@ -1,8 +1,6 @@
 package expenses
 
-import (
-	"testing"
-)
+import "testing"
 
 var testExpensesRecords = []Record{
 	{


IsaacG:
[Related forum thread](https://forum.exercism.org/t/expenses-proposing-a-few-fixes/34453/)

exercises/concept/expenses/.docs/instructions.md
```diff
@@ -17,7 +21,7 @@ type Record struct {
 }
 ```
 
-A `DaysPeriod` represents a range of days and includes all days from the day `From` up to the day `To`.
+A `DaysPeriod` represents a range of days and includes all days from the day `From` up to the day `To`.
```
IsaacG: It looks like you introduced a trailing space here, as well as below.

exercises/concept/expenses/expenses_test.go
```diff
@@ -255,25 +258,25 @@ func TestCategoryExpenses(t *testing.T) {
...
+				t.Fatalf("CategoryExpenses(%v, %v, %s)=%.2f,%v but a non-nil error was not expected",
+					testExpensesRecords, tC.p, tC.category, got, err)
 			}
 
 			if got != tC.total {
 				var errStr string
 
 				if tC.wantErr {
-					errStr = "unknown category"
+					errStr = fmt.Sprintf("unknown category %s", tC.category)
```
IsaacG:
```suggestion
					errStr = "unknown category " + tC.category
```

exercises/concept/expenses/.docs/instructions.md
rahulrao0209: The trailing space here is so that the next sentence could use a new line.
Below, I've actually removed a trailing space as it served no purpose.

IsaacG: I'm not sure what you mean. The trailing space is not needed here and should not be here. Removing it won't affect the formatted output.

rahulrao0209: If we remove the trailing space the next sentence will not appear on a new line in the markdown.
It will continue along the same line. The other way is to add an explicit &lt;br&gt; tag. Same for all trailing spaces that appear in this document. 
Could you confirm that's not the case for you in your markdown editor? If so, I'll remove the space.

TLDR; to use a new line for each sentence in markdown, we could use 2 spaces or a &lt;br&gt; tag to indicate a line break.

IsaacG: Are you trying to force a linebreak in the formatted output? That could be accomplished with two newlines. But that's not the goal here.

The markdown file is this file, without formatting. 

In this markdown file, each sentence should be on a new line. In the rendered format, maybe sentences are part of a paragraph and do not start on a new line.

rahulrao0209: Ah! Gotcha! Now I understand. I believed the formatted output needed to use new lines too.
Thank you for explaining. I will push an update for this.

exercises/concept/expenses/expenses_test.go
```diff
@@ -1,6 +1,8 @@
 package expenses
 
-import "testing"
+import (
+	"testing"
+)
```
IsaacG:
```suggestion
import "testing"
```

PR #3011, opened 2026-03-14
[card-tricks] Update test failure outputs

IsaacG:


IsaacG:
[Forum thread for reference](https://forum.exercism.org/t/card-tricks-test-expansion/34520/)

PR #3058, opened 2026-04-05
Update `constants` docs

BNAndras:
Related to #3042

### DIFF
--- a/concepts/constants/about.md
+++ b/concepts/constants/about.md
@@ -44,7 +44,7 @@ math.Sqrt(num)  // compile error because a float is expected
 ## What can be a constant
 
 Constants can only hold Booleans, numbers, strings, and runes.
-Because the values are set at compile time, the results of a function call can not be constants:
+Because the values are set at compile time, the results of a function call can not be used for a constant.
 
 ```go
 const s = math.Sqrt(4)  // compile error because the value isn't known at compile time
--- a/concepts/constants/about.md
+++ b/concepts/constants/about.md
@@ -43,7 +43,7 @@ math.Sqrt(num)  // compile error because a float is expected
 
 ## What can be a constant
 
-Constants can only hold booleans, numbers, strings, and runes.
+Constants can only hold Booleans, numbers, strings, and runes.
 Because the values are set at compile time, the results of a function call can not be constants:
 
 ```go
--- a/concepts/constants/about.md
+++ b/concepts/constants/about.md
@@ -23,7 +23,7 @@ Constants are typically declared at package level, though they can also be decla
 ## Typed and untyped constants
 
 Declaring a constant without specifying a type makes it untyped.
-An untyped constant has no set type, but its value determines what types it's compatible with.
+An untyped constant has no set type; its value determines what types it's compatible with.
 As long as the underlying value is compatible, the compiler will convert it to the required type at each use:
 
 ```go
--- a/concepts/constants/about.md
+++ b/concepts/constants/about.md
@@ -1,51 +1,82 @@
 # About
 
-Constants in Go are simple, [unchanging values][const] created with the `const` keyword.
-Constants may be given an explicit type:
+Constants are values set at compile time and cannot be reassigned.
+Use the `const` keyword to declare a single constant:
 
 ```go
-const hi string = "hi" // string
+const greeting = "hello"
 ```
 
-or an implicit, default type:
-
-```go
-const hello = "hello" // string
-```
-
-when the variable needs a type.
-This gives constants more flexibility in Go's type system and allows them to work in a variety of contexts without triggering a compiler error:
-
-```go
-const number2 = 2 // 2 is an untyped numeric constant and does not need to be explicitly given the type float64 as required by the Sqrt method
-sqrt2 := math.Sqrt(number2)
-```
-
-Go does not have enums like some other languages but the `iota` enumerator can be used to create successive untyped integer constants:
+Multiple constants can be declared in a block.
+Within a block, a constant without an explicit value set repeats the previous expression:
 
 ```go
 const (
-    a = 6 * iota // 0
-    b            // 6
-    c            // 12
-    d            // 18
+    greeting    = "hello"
+    buffSize    = 4096
+    chunkSize   // 4096
 )
 ```
 
-Complex types like maps and slices are mutable and cannot be constants; the compiler will throw errors:
+Constants are typically declared at package level, though uncommonly they can also be declared inside a function.
+
+## Typed and untyped constants
+
+Declaring a constant without specifying a type makes it untyped.
+An untyped constant has no set type, but its value determines what types it's compatible with.
+As long as the underlying value is compatible, the compiler will convert it to the required type at each use:
 
 ```go
-func main() {
-    const m = map[int]int{2: 8}
-    const s = []string{"exercism", "v3"}
-}
-// => const initializer map[int]int literal is not a constant
-// => const initializer []string literal is not a constant
+const numA = 2
+
+math.Sqrt(numA)  // numA is treated as float64 here
+var numB int = numA  // numA is treated as int here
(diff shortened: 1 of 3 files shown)


concepts/constants/introduction.md
```diff
@@ -1,31 +1,66 @@
 # Introduction
 
-In Go, a constant is a simple, unchanging value assigned to a name with the `const` keyword:
+Constants are values set at compile time and cannot be reassigned.
```
IsaacG, line 3:
```suggestion
Constants are variables that are set to a value at compile time and cannot be reassigned.
```

concepts/constants/introduction.md
```diff
@@ -1,31 +1,66 @@
...
+const (
+    greeting    = "hello"
+    buffSize    = 4096
+    chunkSize   // 4096
+)
 ```
 
-Go does not have enums like some other languages, but does have a predeclared identifier called `iota` for creating enumerated constants.
-Constants in a block are implicitly repeated:
+Constants are typically declared at package level, though uncommonly they can also be declared inside a function.
```
IsaacG:
```suggestion
Constants are typically declared at package level, though they can also be declared inside a function.
```

concepts/constants/introduction.md
```diff
@@ -1,31 +1,66 @@
...
+
+## Typed and untyped constants
+
+Declaring a constant without specifying a type makes it untyped.
+An untyped constant has no set type, but its value determines what types it's compatible with.
```
IsaacG:
```suggestion
An untyped constant has no set type; its value determines what types it's compatible with.
```

concepts/constants/introduction.md
```diff
@@ -1,31 +1,66 @@
...
+
+```go
+const num int = 2
+
+math.Sqrt(num)  // compile error because a float is expected
+```
+
+## What can be a constant
+
+Constants can only hold booleans, numbers, strings, and runes.
```
IsaacG:
```suggestion
Constants can only hold Booleans, numbers, strings, and runes.
```

concepts/constants/introduction.md
IsaacG, line 1: There's a few differences between the intro and about. I'm not sure if those are intentional. If the two are fairly close, it might be worth making them the same.

concepts/constants/introduction.md
BNAndras: I’d agree since it’s named after a person, but Go uses booleans. We also use it lowercased in our docs (see https://exercism.org/tracks/go/concepts/booleans)

concepts/constants/introduction.md
BNAndras: I’d rather have different content for the two files, but the overlap is fairly significant here. We might as well make them the same.

concepts/constants/introduction.md
IsaacG: I think Bethany set me straight on this. Go (and Python) has a `bool` data type that represents a Boolean value, similar to an `int` type that represents an integer. There is no "Integer" and there is no "boolean".

concepts/constants/introduction.md
IsaacG: I'm all for making them different ... where there's a value to that difference, eg omitting some sections from the intro to make it simpler. But if the content is largely the same, there's no reason not to make them entirely the same.

concepts/constants/introduction.md
BNAndras: Yep. I’m fine with the change. My thought was that this should also be updated across the track where relevant. That can happen before or after this is merged.

concepts/constants/introduction.md
BNAndras: Agreed. I’ll update the docs Monday.

concepts/constants/introduction.md
IsaacG: Definitely. Doing it as a separate PR might be cleaner.

concepts/constants/introduction.md
```diff
@@ -1,31 +1,82 @@
...
+const (
+    greeting    = "hello"
+    buffSize    = 4096
+    chunkSize   // 4096
+)
 ```
 
-Go does not have enums like some other languages, but does have a predeclared identifier called `iota` for creating enumerated constants.
-Constants in a block are implicitly repeated:
+Constants are typically declared at package level, though uncommonly they can also be declared inside a function.
```
IsaacG:
```suggestion
Constants are typically declared at package level, though they can also be declared inside a function.
```

concepts/constants/about.md
```diff
@@ -1,51 +1,82 @@
...
+const (
+    greeting    = "hello"
+    buffSize    = 4096
+    chunkSize   // 4096
+)
 ```
 
-when the variable needs a type.
-This gives constants more flexibility in Go's type system and allows them to work in a variety of contexts without triggering a compiler error:
+Constants are typically declared at package level, though uncommonly they can also be declared inside a function.
```
IsaacG:
```suggestion
Constants are typically declared at package level, though they can also be declared inside a function.
```

concepts/constants/introduction.md
```diff
@@ -1,31 +1,82 @@
...
+
+## Typed and untyped constants
+
+Declaring a constant without specifying a type makes it untyped.
+An untyped constant has no set type, but its value determines what types it's compatible with.
```
IsaacG:
```suggestion
An untyped constant has no set type; its value determines what types it's compatible with.
```

concepts/constants/about.md
```diff
@@ -1,51 +1,82 @@
...
+
+## Typed and untyped constants
+
+Declaring a constant without specifying a type makes it untyped.
+An untyped constant has no set type, but its value determines what types it's compatible with.
```
IsaacG:
```suggestion
An untyped constant has no set type; its value determines what types it's compatible with.
```

concepts/constants/about.md
```diff
@@ -1,51 +1,82 @@
...
+```go
+const num int = 2
+
+math.Sqrt(num)  // compile error because a float is expected
+```
+
+## What can be a constant
+
+Constants can only hold Booleans, numbers, strings, and runes.
+Because the values are set at compile time, the results of a function call can not be constants:
```
IsaacG:
```suggestion
Because the values are set at compile time, the results of a function call can not be used for a constant.
```

concepts/constants/introduction.md
```diff
@@ -1,31 +1,82 @@
...
+```go
+const num int = 2
+
+math.Sqrt(num)  // compile error because a float is expected
+```
+
+## What can be a constant
+
+Constants can only hold Booleans, numbers, strings, and runes.
+Because the values are set at compile time, the results of a function call can not be constants:
```
IsaacG:
```suggestion
Because the values are set at compile time, the results of a function call can not be used for a constant.
```

concepts/constants/introduction.md
```diff
@@ -1,31 +1,82 @@
...
+
+```go
+const s = math.Sqrt(4)  // compile error because the value isn't known at compile time
+```
+
+## Iota
+
+Within a block of constants, `iota` represents each constant's position within the block:
```
IsaacG:
```suggestion
Within a block of constants, `iota` represents a constant's position within the block:
```

concepts/constants/about.md
```diff
@@ -1,51 +1,82 @@
...
+
+```go
+const s = math.Sqrt(4)  // compile error because the value isn't known at compile time
+```
+
+## Iota
+
+Within a block of constants, `iota` represents each constant's position within the block:
```
IsaacG:
```suggestion
Within a block of constants, `iota` represents a constant's position within the block:
```

IsaacG (approved):
Looks good! Thank you. I have a few minor grammatical tweaks I think would help it read a bit smoother before merging.

IsaacG:
Thank you!

PR #3060, opened 2026-04-07
Update `range-iteration` concept docs

BNAndras:
Related to #3042

### DIFF
--- a/concepts/range-iteration/.meta/config.json
+++ b/concepts/range-iteration/.meta/config.json
@@ -1,8 +1,10 @@
 {
-  "blurb": "Go has a special \"for range\" loop to easily iterate over collections of data.",
+  "blurb": "The for range loop iterates over slices, arrays, maps, strings, and channels.",
   "authors": [
     "brugnara",
     "tehsphinx"
   ],
-  "contributors": []
+  "contributors": [
+    "BNAndras"
+  ]
 }
--- a/concepts/range-iteration/about.md
+++ b/concepts/range-iteration/about.md
@@ -80,7 +80,8 @@ for range vals {
 
 ## Range over an Integer
 
-Since Go 1.22, `range` can iterate over an integer directly, yielding values from `0` up to but not including that integer:
+Since Go 1.22, `range` can iterate over an integer directly, yielding values from `0` up to but not including that integer.
+Zero and negative values are valid but produce no iterations.
 
 ```go
 for n := range 3 {
--- a/concepts/range-iteration/about.md
+++ b/concepts/range-iteration/about.md
@@ -55,7 +55,7 @@ for _, v := range vals {
 // 30
 ```
 
-If only the index is needed, assign the value to `_` or omit it completely:
+If only the index is needed, omit it completely:
 
 ```go
 vals := []int{10, 20, 30}
--- a/concepts/range-iteration/about.md
+++ b/concepts/range-iteration/about.md
@@ -1,7 +1,7 @@
 # About
 
-In Go, `range` iterates over slices, arrays, maps, strings, and channels.
-Each iteration yields two values: the index (or key) and a copy of the element at that position.
+In Go, `range` iterates over slices, arrays, maps, and strings.
+Depending on what is being ranged over, each iteration yields one or two values.
 
 ## Iterating over a Slice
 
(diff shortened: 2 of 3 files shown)


concepts/range-iteration/about.md
```diff
@@ -1,99 +1,93 @@
 # About
 
-You now have the knowledge on how to iterate a slice, or a map, in Go.
+In Go, `range` iterates over slices, arrays, maps, strings, and channels.
+Each iteration yields two values: the index (or key) and a copy of the element at that position.
```
IsaacG: This isn't broadly true. It's true for ranging over slices, arrays, maps and strings. It's not true for channels, [iterators](https://pkg.go.dev/iter) and `range int`. It probably shouldn't be stated as generally true.

I think it's generally true that `range` yields _one or two_ values. We probably want to mention iterators. They're used a fair bit in, say, the strings package, eg [`strings.FieldsSeq()`](https://pkg.go.dev/strings#FieldsSeq) or `strings.SplitSeq()`

concepts/range-iteration/about.md
```diff
@@ -1,99 +1,93 @@
...
+    fmt.Println(v)
 }
-// outputs:
 // 10
 // 20
 // 30
 ```
 
-If you want to only print the index, you can replace the `x` with `_`, or simply omit the declaration:
+If only the index is needed, assign the value to `_` or omit it completely:
```
IsaacG:
```suggestion
If only the index is needed, omit it completely:
```

concepts/range-iteration/.meta/config.json
```diff
@@ -1,8 +1,10 @@
 {
-  "blurb": "Go has a special \"for range\" loop to easily iterate over collections of data.",
+  "blurb": "The for range loop iterates over slices, arrays, maps, strings, and channels.",
```
IsaacG, line 2: ...and iterators? Or leave it vauge?

concepts/range-iteration/about.md
```diff
@@ -1,99 +1,118 @@
 # About
 
-You now have the knowledge on how to iterate a slice, or a map, in Go.
+In Go, `range` iterates over slices, arrays, maps, and strings.
```
IsaacG, line 3: And more :)

concepts/range-iteration/about.md
```diff
@@ -1,99 +1,118 @@
...
-To learn more about this see [Go Language Spec: range clause](https://go.dev/ref/spec#RangeClause).
-~~~~
-
-## Iteration omitting key or value
+## Omitting Index or Value
 
-In Go an unused variable will raise an error at build time.
-Sometimes you only need the value, as per the first example:
+Go will not compile if a variable is declared but never used.
+If only the value is needed, assign the index or key to `_` to ignore it.
```
IsaacG, line 38: This line should probably be moved after the codeblock.

concepts/range-iteration/about.md
```diff
@@ -1,99 +1,118 @@
...
-~~~~exercism/note
-It may seem the above output is incorrect, as one would expect the first key/value pair on the declaration of the map `9 10` to be the first one printed and not the last.
-However, maps are unordered by nature - there isn't a first or last key/value pair.
-Because of that, when iterating over the entries of a map, the order by which entries will be visited will be random and not follow any specific pattern.
-This means the above output is possible but might differ from what you get if you try to run this yourself.
-To learn more about this see [Go Language Spec: range clause](https://go.dev/ref/spec#RangeClause).
-~~~~
-
-## Iteration omitting key or value
+## Omitting Index or Value
```
IsaacG, line 35: Should this come after all the range-types? Should we cover ranging over strings? Channels? Maybe mention channels are a later concept?

concepts/range-iteration/.meta/config.json
BNAndras: Probably should make it vague rather than enumerate everything. The previous blurb was incorrectly vague and lost some meaning. What is a collection of data even? I think the elevator pitch here would be something along the lines of "the `for range` syntax can repeat a block of code in a variety of ways." Then in the document, we discuss what that means for the different ways.

IsaacG:
@BNAndras Still planning on working on this?

BNAndras:
Yeah, let me wipe off the cobwebs and I’ll take a look next week.

concepts/range-iteration/about.md
```diff
@@ -1,99 +1,149 @@
 # About
 
-You now have the knowledge on how to iterate a slice, or a map, in Go.
+In Go, `for range` repeats a block of code for values from a range expression.
+That expression determines whether each iteration will yield one value or two.
```
IsaacG:
```suggestion
The range expression determines whether each iteration will yield one value or two.
```

concepts/range-iteration/introduction.md
```diff
@@ -1,100 +1,149 @@
 # Introduction
 
-In Go, you can iterate over a `slice` using `for` and an index, or you can use `range`.
-`range` also allows you to iterate over a `map` or a `channel`.
-This concept will cover iterating over a `map` but iterating over a `channel` is out of the scope for this concept.
+In Go, `for range` repeats a block of code for values from a range expression.
+That expression determines whether each iteration will yield one value or two.
```
IsaacG:
```suggestion
The range expression determines whether each iteration will yield one value or two.
```

IsaacG (approved):

PR #3091, opened 2026-04-17
[react] Refactor API to be less open

YeungOnion:
I think the existing API makes it easy to have a brittle implementation that fails when `react` is used as a package.
The proposed change aims to unexport some of the interface as a minimal change.

---

Frankly, I think it is a bad change; my hope is that this PR will be something else entirely before approval, hoping to have the discussion here, but perhaps an issue is more appropriate.

I think a better change would look like an API that doesn't secretly rely on coupling across private types that expose a public only interface.
I'm not certain that this API is unsatisfiable to extend, but I'm suspicious that it is.
Looking at other solutions, I should be able to take someone's Reactor, \[Input|Compute\]Cell and get the interface to work, but I've not been able to do so.

I think I need someone more experienced to tell me if this is
- a correct analysis
- useful thing to notice for Gophers
- useful to point out in Exercism

I think the core of the issue is that none of the Cell types can express state change through the API, but any implementation must accept an arbitrary Cell as an upstream.
I'd be happy to be wrong, but neither mentor nor I figured it out yet.

Some ideas around different API might be generics:
- only allowing a Reactor to have one type that implements Cell.
- extending to have ComputeCell provide a Recompute

### DIFF
--- a/exercises/practice/react/interfaces.go
+++ b/exercises/practice/react/interfaces.go
@@ -1,6 +1,6 @@
 package react
 
-// A Reactor manages linked cells.
+// A reactor manages linked cells.
 type Reactor interface {
 	// CreateInput creates an input cell linked into the reactor
 	// with the given initial value.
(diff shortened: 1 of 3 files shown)


IsaacG:
Marking the types as non-exported does reduce the scope of what you'd need to worry about (supporting other implementations) but for most students I wouldn't expect them to do anything different based on whether or not the types are exported. This makes me think the value added here is not very large.

 This change would also break all existing solutions, which makes me think the cost of this change is high.

The small value add with a high cost makes this something I don't think is worth doing.

PR #3130, opened 2026-04-27
Update `variables` concept docs

BNAndras:
Related to #3042

### DIFF
--- a/concepts/variables/.meta/config.json
+++ b/concepts/variables/.meta/config.json
@@ -1,5 +1,9 @@
 {
-  "blurb": "TODO: add blurb",
-  "authors": ["ErikSchierboom"],
-  "contributors": []
+  "blurb": "Declare variables, assign values, and understand how Go uses variable types.",
+  "authors": [
+    "ErikSchierboom"
+  ],
+  "contributors": [
+    "BNAndras"
+  ]
 }
--- a/concepts/variables/.meta/config.json
+++ b/concepts/variables/.meta/config.json
@@ -1,5 +1,5 @@
 {
-  "blurb": "Declare variables, assign values, and understand how Go uses variable types.",
+  "blurb": "Declare variables, assign values, and understand variable typing.",
   "authors": [
     "ErikSchierboom"
   ],
--- a/concepts/variables/about.md
+++ b/concepts/variables/about.md
@@ -3,25 +3,34 @@
 Go is statically typed, so every variable will have a type known at compile time.
 Once declared, a variable's type cannot change.
 
-The `var` statement declares a variable and typically its type:
+## Creating a Variable
+
+The `var` statement can declare a new variable's type, set its value, or do both at once:
 
 ```go
-var count int
+var total int     // Type explicitly declared, value implicitly set (see note on zero values below)
+var count int = 1 // Type explicitly declared, value explicitly set
+var sum = 100     // Type implictly declared, value explicitly set
 ```
 
-Go can infer a type from a provided initial value:
+Inside a function, the `:=` short assignment statement can also be used.
 
 ```go
-var count = 1 // count is int
+count := 1 // Type implictly declared, value explicitly set
 ```
 
-Inside a function, the `:=` short assignment statement accomplishes this in a concise format:
+A variable in Go always has a value.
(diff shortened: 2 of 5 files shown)


concepts/variables/introduction.md
```diff
@@ -1,46 +1,45 @@
 # Introduction
 
-Go is a statically-typed language, which means that everything has a type at compile-time.
-Assigning a value to a name is referred to as defining a variable.
-A variable can be defined either by explicitly specifying its type, or by assigning a value to have the Go compiler infer its type based on the assigned value.
+Go is statically typed, so every variable will have a type known at compile time.
+Once declared, a variable's type cannot change.
+
+The `var` statement declares a variable and typically its type:
```
IsaacG:
```suggestion
The `var` statement typically declares both a variable and its type:
```

concepts/variables/introduction.md
```diff
@@ -1,46 +1,45 @@
...
-    return "Hello " + name
-}
+```go
+count := 1
+count = false // compile error: a Boolean is not an int
 ```
 
-Invoking a function is done by specifying the function name and passing arguments for each of the function's parameters.
+A variable in Go always has a value.
+If no initial value is set at declaration, Go provides one, a zero value, based on the variable type.
```
IsaacG:
```suggestion
If no initial value is set at declaration, Go provides one, a zero value, based on the variable type.
(There is a later concept all about zero values.)
```

IsaacG (approved):

concepts/variables/introduction.md
BNAndras: I'll update this tomorrow. I suspect we don't need the second sentence if I just link directly to the zero value concept on Exercism.

concepts/variables/.meta/config.json
```diff
@@ -1,5 +1,9 @@
 {
-  "blurb": "TODO: add blurb",
-  "authors": ["ErikSchierboom"],
-  "contributors": []
+  "blurb": "Declare variables, assign values, and understand how Go uses variable types.",
```
BNAndras: I'm not convinced the concept covers how Go uses variable types so this blurb should be workshopped more. Nothing is coming to mind though.

IsaacG:
```suggestion
  "blurb": "Declare variables, assign values, and understand variable typing.",
```
Something like this maybe?

concepts/variables/about.md
```diff
@@ -1,30 +1,47 @@
...
+count := 1 // Type implictly declared, value explicitly set
+```
 
-// Compiler error when assigning different type
-// count = false
+A variable in Go always has a value.
+If one isn't set at declaration, Go provides one [called a zero value][zero-values], based on the variable's type.
+
+```go
+var total int      // 0
```
IsaacG:
```suggestion
var total int     // 0
```

concepts/variables/introduction.md
```diff
@@ -1,46 +1,47 @@
...
-All parameters must be explicitly typed, there is no type inference for parameters.
-A function can also have multiple return values which must also be explicitly typed.
-Values are returned from functions using the `return` keyword.
-To allow a function to be called by code in other packages, the name of the function must start with a capital letter.
+A variable in Go always has a value.
+If one isn't set at declaration, Go provides one [called a zero value][zero-values], based on the variable's type.
 
 ```go
-package greeting
+var total int      // 0
```
IsaacG:
```suggestion
var total int     // 0
```

IsaacG (approved):

concepts/variables/introduction.md
```diff
@@ -1,46 +1,47 @@
 # Introduction
 
-Go is a statically-typed language, which means that everything has a type at compile-time.
-Assigning a value to a name is referred to as defining a variable.
-A variable can be defined either by explicitly specifying its type, or by assigning a value to have the Go compiler infer its type based on the assigned value.
+Go is statically typed, so every variable will have a type known at compile time.
```
IsaacG, line 3: This feels a bit like a lie. Go has untyped constants. Which aren't variables, but they are placeholders.

IsaacG (approved):

PR #3131, opened 2026-04-27
Update `booleans` concept

BNAndras:
Related to #3042

### DIFF
--- a/concepts/booleans/about.md
+++ b/concepts/booleans/about.md
@@ -13,11 +13,11 @@ knowsCode := false
 
 Logical operators work on one or two Boolean values and produce a Boolean result.
 
-| Operator | Name | Behavior |
-| -------- | ---- | -------- |
+| Operator | Name | Behavior                                       |
+| -------- | ---- | ---------------------------------------------- |
 | `!`      | NOT  | Returns the opposite of a single Boolean value |
 | `&&`     | AND  | Returns `true` only when both sides are `true` |
-| `\|\|`     | OR   | Returns `true` if any side is `true` |
+| `\|\|`   | OR   | Returns `true` if any side is `true`           |
 
 Use `!` (NOT) when you need the opposite of a Boolean value:
 
--- a/concepts/booleans/about.md
+++ b/concepts/booleans/about.md
@@ -1,23 +1,46 @@
 # About
 
-Booleans in Go are represented by the `bool` type, which values can be either `true` or `false`.
-
-Go supports three [boolean operators][logical operators]: `!` (NOT), `&&` (AND), and `||` (OR).
-
-```go
-true || false // => true
-true && false // => false
-!true // => false
-```
-
-The three boolean operators each have a different [_operator precedence_][operators].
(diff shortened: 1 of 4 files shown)


concepts/booleans/about.md
```diff
@@ -1,23 +1,46 @@
 # About
 
-Booleans in Go are represented by the `bool` type, which values can be either `true` or `false`.
+A Boolean value represents whether a condition is `true` or `false`.
+In Go, the `bool` type has only those two values.
```
IsaacG:
```suggestion
In Go, the `bool` type can only be one of those two values.
```

concepts/booleans/about.md
```diff
@@ -1,23 +1,46 @@
...
-Go supports three [boolean operators][logical operators]: `!` (NOT), `&&` (AND), and `||` (OR).
+```go
+isDoorLocked := true
+hasKey := true
+knowsCode := false
+```
+
+## Logical Operators
+
+Logical operators work on one or two `bool` values and produce a Boolean result.
```
IsaacG:
```suggestion
Logical operators work on one or two Boolean values and produce a Boolean result.
```

concepts/booleans/about.md
```diff
@@ -1,23 +1,46 @@
...
+
+| Operator | Name | Behavior |
+| -------- | ---- | -------- |
+| `!`      | NOT  | Returns the opposite of a single Boolean value |
+| `&&`     | AND  | Returns `true` only when both sides are `true` |
+| `||`     | OR   | Returns `true` when at least one side is `true` |
```
IsaacG:
```suggestion
| `\|\|`     | OR   | Returns `true` if any side is `true` |
```

concepts/booleans/about.md
```diff
@@ -1,23 +1,46 @@
...
-[operators]: https://golang.org/ref/spec#Operators
-[logical operators]: https://golang.org/ref/spec#Logical_operators
+Use `||` (OR) when at least one condition must be `true`:
+
+```go
+canOpenDoor := canUnlockDoor || knowsCode // true
+```
+
+Boolean operators are evaluated by default in this order: `!` first, then `&&`, then `||`.
+Use parentheses to override this behavior:
```
IsaacG:
```suggestion
Boolean operators are evaluated in this order: first `!`, then `&&`, then `||`.
Use parentheses to change the order of evaluation:
```

"By default" sounds like you can change the precendence.

IsaacG (approved):

concepts/booleans/about.md
```diff
@@ -1,23 +1,46 @@
 # About
```
IsaacG, line 1: Looks good! Just needs syncing across the `md` files.

concepts/booleans/about.md
BNAndras: The second sentence is a bit tricky, and I'll have to think about it. The parentheses are being used to group expressions, and changing how Go would otherwise group expressions affects the evaluation order. `Use parentheses to change which operations are evaluated first:` might work, but then we should highlight the order with and without parentheses in the code example.

IsaacG (approved):

PR #3136, opened 2026-05-21
[react] Update instructions to reduce focus on API design

YeungOnion:
following discussion in
http://forum.exercism.org/t/react-refactor-api-to-be-less-open/44584

### DIFF
--- a/exercises/practice/react/.docs/instructions.append.md
+++ b/exercises/practice/react/.docs/instructions.append.md
@@ -2,4 +2,4 @@
 
 ## Implementation Notes
 
-Note: For the purpose of this exercise you can assume all Cell objects are created by one of your `Create*()` functions
+Note, for the purpose of this exercise you can assume all `Cell` objects are created by one of your `Create*()` functions.
--- a/exercises/practice/react/.docs/instructions.append.md
+++ b/exercises/practice/react/.docs/instructions.append.md
@@ -1,0 +1,5 @@
+# Instructions append
+
+## Implementation Notes
+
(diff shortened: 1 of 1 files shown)


exercises/practice/react/.docs/instructions.md
```diff
@@ -9,3 +9,5 @@ Implement updates so that when an input value is changed, values propagate to re
 
 In addition, compute cells should allow for registering change notification callbacks.
 Call a cell’s callbacks when the cell’s value in a new stable state has changed from the previous stable state.
+
+Note: For the purpose of this exercise you can assume all Cell objects are created by one of your `Create*()` functions
```
IsaacG: This `instructions.md` file is synced from [the problem specs](https://github.com/exercism/problem-specifications/blob/main/exercises/react/description.md) and should not be manually added.

However, we can append instructions using an `instructions.append.md` file. See [grep for example](https://github.com/exercism/go/blob/main/exercises/practice/pov/.docs/instructions.append.md?plain=1).

IsaacG (changes requested):

YeungOnion:
Thanks for all the help @IsaacG!

exercises/practice/react/.docs/instructions.append.md
```diff
@@ -0,0 +1,5 @@
+# Instructions append
+
+## Implementation Notes
+
+Note: For the purpose of this exercise you can assume all Cell objects are created by one of your `Create*()` functions
```
IsaacG:
```suggestion
Note, for the purpose of this exercise you can assume all `Cell` objects are created by one of your `Create*()` functions.
```

IsaacG (approved):

